PR #41, opened 2019-03-13
Improve documentation for Tangential complex (Fix #16)

### LINKED ISSUE #16: [tangential complex] Documentation improvements
(this is for the python doc, but the C++ one likely has the same issue)
* The user manual presents an example with 4 points and claims that for each point we are going to estimate its tangent subspace. Of course, with 4 points, we aren't going to get good estimates of the tangent subspace at each point... This is super-simplified, normally you have at least dozens of points. Besides, we only claim that we can reconstruct closed curves. Maybe a simple sentence explaining exactly that would help un-confuse users.
* The manual says "For each point p, estimate its tangent subspace Tp (e.g. using PCA)." The current code is hard-coded to use PCA, as far as I can tell from the doc (it would be nice if it allowed to plug other tangent space estimators). The sentence kind of gives a reader the impression that they are responsible for this tangent space estimation and they should look into PCA or alternatives, when (currently) that's an implementation detail that users cannot access.

VincentRouvreau:


src/Tangential_complex/doc/Intro_tangential_complex.h
```diff
@@ -46,9 +46,10 @@ An extensive description of the Tangential complex can be found in \cite tangent
 \subsection whatisthetc What is a Tangential Complex?
 
 Let us start with the description of the Tangential complex of a simple example, with \f$ k=1 \f$ and \f$ d=2 \f$.
-The input data is 4 points \f$ P \f$ located on a curve embedded in 2D.
+Only 4 points will be displayed (more are required for PCA) to simplify the figures. \f$ P \f$ located on a closed
+curve embedded in 2D.
```
mglisse: We end up with a strange sentence without a verb?

src/Tangential_complex/doc/Intro_tangential_complex.h
```diff
@@ -46,17 +48,19 @@ An extensive description of the Tangential complex can be found in \cite tangent
 \subsection whatisthetc What is a Tangential Complex?
 
 Let us start with the description of the Tangential complex of a simple example, with \f$ k=1 \f$ and \f$ d=2 \f$.
-The input data is 4 points \f$ P \f$ located on a curve embedded in 2D.
+The point set \f$ \mathscr P \f$ is located on a closed curve embedded in 2D.
+Only 4 points will be displayed (more are required for PCA) to simplify the figures.
 \image html "tc_example_01.png" "The input"
-For each point \f$ p \f$, estimate its tangent subspace \f$ T_p \f$ (e.g. using PCA).
+For each point \f$ P \f$, estimate its tangent subspace \f$ T_p \f$ using PCA.
```
mglisse: T_p should be T_P I guess.

PR #50, opened 2019-04-19
My modif

siddharthpritam:


VincentRouvreau:
I merged it in my own repo instead of using the main one

PR #52, opened 2019-05-06
Persistence interval shall use numpy array

VincentRouvreau:
* numpy is now a hard dependency.
* persistence_intervals_in_dimension and read_persistence_intervals_in_dimension now returns a numpy array of dimension 2 (list of [birth, death]).


src/cython/cython/cubical_complex.pyx
```diff
@@ -5,13 +5,15 @@ from libcpp.string cimport string
...
+from numpy import array as np_array
+
 """This file is part of the Gudhi Library. The Gudhi library
    (Geometric Understanding in Higher Dimensions) is a generic C++
    library for computational topology.
 
    Author(s):       Vincent Rouvreau
 
-   Copyright (C) 2016 Inria
+   Copyright (C) 2019 Inria
```
mglisse: I often see copyrights of the form 2016-2019 instead of just the latest year. I doubt that matters though.

src/cython/cython/cubical_complex.pyx
```diff
@@ -195,4 +197,4 @@ cdef class CubicalComplex:
         else:
             print("intervals_in_dim function requires persistence function"
                   " to be launched first.")
-        return intervals_result
+        return np_array(intervals_result)
```
mglisse: Does that mean that cython creates a list of pairs, and we convert it to a numpy array afterwards? That's ok as a starting point to define the interface, but eventually we will want to produce a numpy array directly.

src/cython/cython/persistence_graphical_tools.py
```diff
@@ -318,8 +315,8 @@ def plot_persistence_density(persistence=[], persistence_file='',
                                  reverse=True)[:max_intervals]
 
         # Set as numpy array birth and death (remove undefined values - inf and NaN)
-        birth = np.asarray([(interval[1][0]) for interval in persistence_dim if (math.isfinite(interval[1][1]) and math.isfinite(interval[1][0]))])
-        death = np.asarray([(interval[1][1]) for interval in persistence_dim if (math.isfinite(interval[1][1]) and math.isfinite(interval[1][0]))])
+        birth = np.asarray([(interval[1][0]) for interval in persistence_dim if (isfinite(interval[1][1]) and isfinite(interval[1][0]))])
+        death = np.asarray([(interval[1][1]) for interval in persistence_dim if (isfinite(interval[1][1]) and isfinite(interval[1][0]))])
```
mglisse: If you start from a numpy array a, you can keep only the rows with a finite second column using `a[np.isfinite(a[:,1])]`.

src/cython/cython/cubical_complex.pyx
mglisse: (Hmm github seems to have lost my comment, probably I clicked on "resolve" and thought it would also post the text) The intermediate object we convert to a numpy array is not a python list, it is a C++ish vector of pairs (via cython), hopefully converting that to a numpy array is not as slow as building an intermediate list would be. Really returning a numpy array from C++ code seems surprisingly hard from what I have seen (and not just with cython), so I think this will do for now.

src/cython/cython/persistence_graphical_tools.py
VincentRouvreau: In the new version 7e8685458f684f347b2aa6a77a288c249c722f25 I tried to use numpy array as much as possible. It makes a 0.3 out of 4 seconds improvement for plot_persistence_density over 740 intervals pairs.

Issue #53, opened 2019-05-10
Missing example of witness complex from distance matrix


mglisse:
It would be nice to have an example showing how to build a witness complex starting from a distance matrix (select a set of landmarks, build the nearest landmark table, etc). I believe there used to be such an example in C++ a long time ago but it got removed.

Trying to write such code in python, the format of the nearest landmark table is not convenient, it contains both integers and floats and thus cannot be a numpy array. I was starting from a distance matrix in a numpy array, selecting a set of landmarks, keeping only the corresponding columns in the matrix. Then I can either argsort or sort the rows but it is inconvenient to get both results mixed so I can pass the result to the witness complex.

PR #54, opened 2019-05-10
Clarify doc of witness complex

mglisse:
(related to issue #53 but it does not fix it)


mglisse:
> I have tried the examples src/cython/example/witness_complex_from_nearest_landmark_table.py and src/Witness_complex/example/example_nearest_landmark_table.cpp by trying to change the threshold value, but the generated simplex tree seems to be always the same. Is it normal ?

Did you try making the threshold smaller than 0.1?

PR #59, opened 2019-05-22
Strong collapse

VincentRouvreau:


mglisse:
The user manual does not include, or link to, any example on how to use it.

mglisse:
It is a bit strange to pass the number of points to the constructor and then the graph to initialize_variant. Is that useful?

mglisse:
`Flag_complex_sparse_matrix` is documented, although it isn't clear to me that a user is supposed to manipulate that directly.
On the other hand, the documentation of `Flag_complex_strong_collapse` talks repeatedly about `Flag_complex_tower_assembler`, which is not documented.
At some point we will want a documented interface to handle towers, strong collapses are not the only way to produce them.

mglisse:
`get_distance_matrix` returns a distance matrix. Sure, I could guess that from the name. But what does this distance matrix represent? And what kind of type is Distance_matrix, what interface does it have?

VincentRouvreau:
> .

Done on 421ec25e639ce9190c360a3709eacf537ae01d97

VincentRouvreau:
> It is a bit strange to pass the number of points to the constructor and then the graph to initialize_variant. Is that useful?

I removed initialize_exact_version and initialize_approximate_version and made 2 constructors that can build the exact version, or the approximate one. I hope it is clearer.
c9fb8166970f2ddca749c9d6708a5d91fb42536d

VincentRouvreau:
> It is a bit strange to pass the number of points to the constructor and then the graph to initialize_variant. Is that useful?



> The user manual does not include, or link to, any example on how to use it.

I added a simple example in the documentation : ac1be78217508792be4ff44cddb67b7164bdcf4e

VincentRouvreau:
> `get_distance_matrix` returns a distance matrix. Sure, I could guess that from the name. But what does this distance matrix represent? And what kind of type is Distance_matrix, what interface does it have?

I moved Distance_matrix type in `Flag_complex_sparse_matrix` for documentation purpose : 76ef89433ec85daa7ac147d36cf9f33786a58335
I cannot add this type in tower (we want to hide this class), nor in Flag_complex_strong_collapse (circular dependency).
The other solution is to let this type definition as global to the namespace, but the type would require to be added to the doxygen group strong collapse, and the type documentation would land on user manual.

src/Strong_collapse/include/gudhi/Strong_collapse/Flag_complex_tower_assembler.h
```diff
@@ -0,0 +1,175 @@
...
+      }
+    } else {
+      std::cerr << "Unable to open file " << outFile;
+      exit(-1);
+    }
+  }
+
+  typename Flag_complex_sparse_matrix<SimplicialComplex>::Distance_matrix distance_matrix() {
+    std::size_t non_zero_rw = flag_filtration_.num_vertices();
+    Filtration_value inf = std::numeric_limits<Filtration_value>::max();
```
mglisse: That's a misleading name. `max` is the largest **finite** value. For infinity, you can replace `max` with `infinity`. Then it depends which one you want...

src/Strong_collapse/doc/Intro_strong_collapse.h
```diff
@@ -0,0 +1,115 @@
...
+ * (here Rips) complexes connected with inclusions. 
+ * After perfoming the reduction the filtration reduces to a flag-tower, which is a sequence of flag complexes
+ * connected through more general (inclusion and contractions) simplicial maps. 
+ * However to compute the persistent homology of any tower one needs to convert it to an equivalent filtration. Section
+ * 4 of \cite strongcollapsesocg2019 descibes an efficient strategy to convert a flag tower to a flag filtration, 
+ * using only the 1-skeletons of the complexes, which we implement here.
+ * 
+ * \subsection strong_collapse_from_points_example Example from a point cloud and a distance function
+ * 
+ * This example builds the edge graph from the given points, threshold value, and distance function.
```
mglisse: Without context, I have no idea what the threshold value is about. Looking at it, it appears that it is a maximum on the edge lengths that we consider?

src/Strong_collapse/example/strong_collapse_from_points.cpp
```diff
@@ -0,0 +1,56 @@
...
+  points.push_back({2.0, 19.0});
+  points.push_back({9.0, 17.0});
+  
+  // ----------------------------------------------------------------------------
+  // Edge graph construction from the list of points
+  // ----------------------------------------------------------------------------
+  Filtered_edges_vector edge_graph =
+      Gudhi::compute_edge_graph<Gudhi::Filtered_edges_vector, Simplicial_complex>(
+          points,
+          12.,  // means all edges
```
mglisse: Infinity, or maybe 10000, would more clearly indicate "all edges".

mglisse:
From the documentation, I don't see any link to the other examples. I searched both in the doc for this specific package, and in the global "examples" link in the left bar.

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
  */
-template< typename SimplicialComplexForProximityGraph
-          , typename ForwardPointRange
-          , typename Distance >
-Proximity_graph<SimplicialComplexForProximityGraph> compute_proximity_graph(
-    const ForwardPointRange& points,
-    typename SimplicialComplexForProximityGraph::Filtration_value threshold,
+template <template <class> class Edge_graph, class SimplicialComplexForEdgeGraph, typename ForwardPointRange,
+          typename Distance>
+Edge_graph<SimplicialComplexForEdgeGraph> compute_edge_graph(
```
mglisse: I have trouble finding where in the documentation doxygen put this?

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
+      ++ep_iter;
+    }
+    // By default the sort is done on the first element (Filtration_value) in ascending order.
+    std::sort(edges_.begin(), edges_.end());
+  }
 
-/** \brief Computes the proximity graph of the points.
+  Filtered_edge_set edges() const { return edges_; }
+
+  /** \brief Returns the edges vector at a specific index filtration value.
```
mglisse: It doesn't look like this returns a vector... Returns the filtration value of the n-th edge? (the edge of index n?)

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
+    return std::get<0>(*(edges_.begin()));
+  }
+
+  /** \brief Returns the edges vector max filtration value.
+   */
+  typename SimplicialComplexForFilteredEdges::Filtration_value get_filtration_max() const {
+    return std::get<0>(*(edges_.rbegin()));
+  }
+
+  /** \brief Returns the edges vector.
```
mglisse: No?

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
+   */
+  std::size_t size() const { return edges_.size(); }
+
+  /** \brief Returns the sub-filtered edges vector from a `new_threshold` filtration value.
```
mglisse: I think it would be helpful to use simpler English. Returns a vector which contains the edges with filtration value smaller than new_threshold.

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
+   */
+  Filtered_edge_set sub_filter_edges_by_filtration(typename SimplicialComplexForFilteredEdges::Filtration_value new_threshold) const {
```
mglisse: `Filtered_edge_set` is not documented. Looking at what it is, it seems very similar to `Filtered_edges_vector`.
Do you really need to return a heavy vector? If the underlying `Filtered_edges_vector` doesn't disappear, a light `boost::iterator_range` looks like it might suffice.

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
+    auto edge_it = edges_.begin();
+    while ((std::get<0>(*edge_it) < new_threshold) && (edge_it < edges_.end())) {
```
mglisse: Dichotomy? (`std::lower_bound` for instance)

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +31,127 @@ struct vertex_filtration_t {
...
+    }
+#endif  // DEBUG_TRACES
+    return Filtered_edge_set(edges_.begin(), edges_.begin() + new_index + 1);
+  }
+
+private:
+  Filtered_edge_set edges_;
+};
+
+/** \brief Computes the edge graph of the points.
```
mglisse: I am not a fan of the name "edge graph". Having edges is pretty much the definition of a graph... *Rips graph*? *Ball graph*? That is, if you don't like the original name: *proximity graph*.

Issue #12, opened 2019-03-07
[Simplex_tree] Simplex_tree insert functions handles badly duplicated vertices in list


VincentRouvreau:
```python
st = gd.SimplexTree()
st.insert([0,1])
st.insert([1,2])
st.insert([1,1]) # No error, even return True 

print(st.get_filtration()) # => crash
```

Seems to come from insert_simplex_and_subfaces. Check insert function behaviour

PR #60, opened 2019-05-22
Simplex tree insert duplicated vertices fix vincent

VincentRouvreau:
Fix issue #12


VincentRouvreau:
Fix issue #12

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -811,19 +814,30 @@ class Simplex_tree {
...
       } else {
-	// FIXME: this interface makes no sense, and it doesn't seem to be tested.
-	insertion_result.first = null_simplex();
+        // FIXME: this interface makes no sense, and it doesn't seem to be tested.
+        insertion_result.first = null_simplex();
       }
     }
-    if (++first == last) return insertion_result;
+    // Next iteration to avoid consecutive equal values
+    while ((vertex_one == *first) && (first < last)) {
```
mglisse: At a glance, I would test if `first!=last` before dereferencing `first`, assuming that `*last` is bad.

mglisse:
I think I'd rather call `std::unique` right after `std::sort` in `insert_simplex`/`insert_simplex_and_subfaces` and keep the others as raw, low-level functions that take already sanitized input (so we can add a GUDHI_CHECK `std::adjacent_find` next to `std::is_sorted` if you want). That or reject repetitions.

VincentRouvreau:
Should be ok this way.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -775,31 +775,26 @@ class Simplex_tree {
...
   }
 
  private:
-  /// Same as insert_simplex_and_subfaces but assumes that the range of vertices is sorted
-  template<class ForwardVertexRange = std::initializer_list<Vertex_handle>>
-  std::pair<Simplex_handle, bool> insert_simplex_and_subfaces_sorted(const ForwardVertexRange& Nsimplex, Filtration_value filt = 0) {
-    auto first = std::begin(Nsimplex);
-    auto last = std::end(Nsimplex);
-    if (first == last)
-      return { null_simplex(), true }; // FIXME: false would make more sense to me.
```
mglisse: Can we keep the comment somewhere, saying that `true` is strange?

PR #62, opened 2019-05-27
Persistent entropy

aluquecerp:
Three files have been included:
- The file Persistent_entropy.h includes two functions. One of them calculate the persistent entropy of a persistence_diagram. The parameters can be used to ignore or keep the infinity bars and to normalize. The other function calculate the persistent entropy of a list of persistence diagrams. They should be given in a file, as a list of the files directions.
- The file files_list.txt is a list of the directions of three persistence diagrams. It should be modified on each computer to be used. The modification needed is just changing the root folder of the persistence diagrams.
- The file persistent_entropy.cpp is an example of usage, using files_list.txt.
- The CMakeLists.txt file has been modified in order to compile persistent_entropy.cpp

### DIFF
--- a/data/persistence_diagram/files_list.txt
+++ b/data/persistence_diagram/files_list.txt
@@ -1,0 +1,6 @@
+/home/luque/Escritorio/Gudhi/gudhi-devel/data/persistence_diagram/first.pers
+/home/luque/Escritorio/Gudhi/gudhi-devel/data/persistence_diagram/rips_on_tore3D_1307.pers
+/home/luque/Escritorio/Gudhi/gudhi-devel/data/persistence_diagram/second.pers
+
+
+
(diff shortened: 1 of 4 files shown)


mglisse:
It seems that the only type of input you take is files. So if I call for instance `Persistent_cohomology::intervals_in_dimension()` and get a `vector<pair<double,double>>`, my only way to compute its entropy is first to write it to a file and then read it back? That doesn't look very convenient.

mglisse:
To clarify, I believe the main entry point should be something like our current [bottleneck_distance](http://gudhi.gforge.inria.fr/doc/latest/group__bottleneck__distance.html#ga03e716e4ecbd47e14e7218b97b23f176), which takes a range of pairs and returns a number. Functions like `persistent_entropy_from_file` would then call that one.
I don't really like the interface of the existing `read_persistence_intervals_in_one_dimension_from_file` (not your fault) with the magical meaning of the value -1 for the last argument (why not -infinity?) (and also the swapping, and the fact that intervals going further than val_inf are not truncated). Your version that has separate keep_inf and val_inf is a bit clearer, although one might argue about packing them as a `boost::optional<double>`.

VincentRouvreau:
We close this PR as the python version is sufficient enough, even if less optimized than this C++ version

PR #64, opened 2019-06-03
Main documentation pages

VincentRouvreau:
With the move of the introduction part into a dedicated web page and a new classification of the GUDHI modules, main documentation pages had to be rewritten.

I took the opportunity to change the style, with smaller images.

To visualize the results:
* Intro http://pages.saclay.inria.fr/vincent.rouvreau/gudhi/doc_improvement//introduction/
* C++ http://pages.saclay.inria.fr/vincent.rouvreau/gudhi/doc_improvement/doc/latest/
* Python http://pages.saclay.inria.fr/vincent.rouvreau/gudhi/doc_improvement/python/latest/


mglisse:
The way we use "reference manual" to denote a shortcut to the documentation of a random class in the package doesn't really work. I think I'd rather keep only the "user manual" links and remove the "reference manual" ones from the page that lists packages.

VincentRouvreau:
Only for the C++ version, then ?
For Python version, sphinx requires a link to reference a page.


----- Mail original -----

mglisse:
The situation does look a bit different.
In C++, the reference manual link really is to one random class (for the witness complex, that's even a concept...). And the user manual contains links to all the classes. So the "reference manual" link to that one random class seems unnecessary and confusing.
In Python, it seems that most packages have a single page of sphinx doc anyway, and for the one package that has several, we link all of them as reference manuals, so that's ok. An alternative would be to put those links inside the user manual.

> sphinx requires a link to reference a page

I assume you mean that this link is currently the only link to those pages, and if we removed it, the page would disappear? The user manual should have links to those pages in one way or another, so if we removed the links from where they are now, we would need to add them elsewhere.

PR #66, opened 2019-06-05
Small fixes

### LINKED ISSUE #7: [Persistent_cohomology] get_persistent_pairs function is not documented
The C++ and Python function is badly documented. Shall be better explained.

### LINKED ISSUE #9: [Persistent_cohomology - doxygen] Persistent_cohomology receive an object of type Complex_ds and not of type Filtered_complex in the documentation
Why does the constructor of Persistent_cohomology receive an object of type Complex_ds and not of type Filtered_complex? I assume that it is the same?
=> Both are the same problem, missing a comment to make doxygen pick them up.

### LINKED ISSUE #20: [doc] OFF file format
The link to http://www.geomview.org/docs/html/OFF.html documenting the OFF file format we use for point sets is repeated in many packages, but not all that use it (for instance the tangential complex, or the python interface). Maybe we should link to our file_formats page, and explain there what variants we support, with an example for a point set so people are not too confused with having to specify edges/faces (just say there are 0).

### LINKED ISSUE #46: left-over debug printing in python SimplexTree.get_cofaces
```python
import gudhi
st = gudhi.SimplexTree()
st.insert([0,1])
st.insert([0,2])
print(st.get_cofaces([0],1))
```
The call to get_cofaces prints stuff. I don't think it should, cython/include/Simplex_tree_interface.h should not have any `std::cout` in it (except if we add a function print or dump).
By the way, the name `Complex` in that file is misleading, get_cofaces returns a chain (a list of simplices) that does not form a complex (boundaries are not included).

### LINKED ISSUE #55: [Simplex_tree] make_filtration_non_decreasing return doc
> @return The filtration modification information.

(same in python) is not clear at all, should be something like

> true if any filtration value was modified, false if the filtration was already non-decreasing.

### LINKED ISSUE #56: Remove references to gitlab
Since the gitlab thing was closed, those are dead links.
Do we get reports about dead links on the gudhi website?

VincentRouvreau:
Fix #56 
Fix #46 
Fix #55 
Fix #9 
Fix #7 
Fix #20


src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -63,12 +63,20 @@ namespace persistent_cohomology {
...
+
+  /** \brief Data stored for each simplex. */
+  typedef typename FilteredComplex::Simplex_key Simplex_key;
+  /** \brief Handle to specify a simplex. */
+  typedef typename FilteredComplex::Simplex_handle Simplex_handle;
+  /** \brief Type for the value of the filtration function. */
+  typedef typename FilteredComplex::Filtration_value Filtration_value;
+  /** \brief Type of element of the field. */
   typedef typename CoefficientField::Element Arith_element;
+  /** \brief Persistent interval type. The Arith_element field is used for the multi-field framework. */
```
mglisse: It could help to specify that the 2 Simplex_handle are for birth and death.

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -159,7 +166,7 @@ class Persistent_cohomology {
       min_length_ = new_length;
     }
 
-    Complex_ds * cpx_;
+    FilteredComplex * cpx_;
```
mglisse: One drawback of just using the template parameter without providing a public typedef for it is that, given a PCOH type, I cannot write any `PCOH::something` to ask what the type of the filtered complex was. It probably doesn't matter, I can't think of a situation where we would need that anyway.

mglisse:
Ok once the birth+death thing is added.
Next time, please don't put so many independent things together, it is much easier to review separately...

PR #69, opened 2019-06-06
Bottleneck documentation

### LINKED ISSUE #11: [Bottleneck_distance - documentation] Distance computation shall be better documented.
An example of Bottleneck distance would help user comprehension.
It shall be mentioned that the Euclidean distance is not used.

VincentRouvreau:
Fix #11 
Fix doxygen warning


src/Bottleneck_distance/doc/Intro_bottleneck_distance.h
```diff
@@ -44,6 +47,43 @@ namespace persistence_diagram {
...
+ * \endcode
+ *
+ * \code Bottleneck distance = 6.5
+ * \endcode
+ *
+ * \image html bottleneck_distance_example.png The point (0, 13) is at 6.5 distance from the diagonal and more specifically from the point (6.5, 6.5)
+ *
+ * \section bottleneckbasicexample Basic example
+ *
+ * This another example computes the bottleneck distance from 2 persistence diagrams:
```
mglisse:
```suggestion
 * This other example computes the bottleneck distance from 2 persistence diagrams:
```

src/Bottleneck_distance/doc/Intro_bottleneck_distance.h
```diff
@@ -44,6 +47,43 @@ namespace persistence_diagram {
...
+
+  double b = Gudhi::persistence_diagram::bottleneck_distance(diag1, diag2);
+  std::cout << "Bottleneck distance = " << b << std::endl;
+}
+ * \endcode
+ *
+ * \code Bottleneck distance = 6.5
+ * \endcode
+ *
+ * \image html bottleneck_distance_example.png The point (0, 13) is at 6.5 distance from the diagonal and more specifically from the point (6.5, 6.5)
```
mglisse:
```suggestion
 * \image html bottleneck_distance_example.png The point (0, 13) is at distance 6.5 from the diagonal and more specifically from the point (6.5, 6.5)
```

mglisse:
When we define the bottleneck distance, we say "such that any couple of matched points are at distance at most b.". We could add something like "where the distance between points is the sup norm in R^2".

src/Bottleneck_distance/doc/Intro_bottleneck_distance.h
```diff
@@ -44,6 +47,43 @@ namespace persistence_diagram {
  * This implementation is based on ideas from "Geometry Helps in Bottleneck Matching and Related Problems"
  * \cite DBLP:journals/algorithmica/EfratIK01. Another relevant publication, although it was not used is
  * "Geometry Helps to Compare Persistence Diagrams" \cite Kerber:2017:GHC:3047249.3064175.
+ * 
+ * \section bottleneckdistanceprecision Distance computation
+ *
+ * Bottleneck distance does not use Euclidean distance, like explained in the following example:
```
mglisse: Looking at the whole page, this seems too prominent. It is just supposed to be a remark, but it appears as the first sentence of the section on computation...

VincentRouvreau: I tried to fix it on 4b401e9 
The figure was also incorrect ((13,0) instead of (0,13))
Tell me if the figure and the distance computation example is "too much".
I also had the citations in the Python documentation version.

PR #70, opened 2019-06-13
Persistent cohomology documentation

VincentRouvreau:
* Highlight persistence_dim_max parameter in plain_homology example
* Only one pcoh constructor to clarify documentation about persistence_dim_max parameter
* Fix alpha complex 3d version on pcoh doxygen page
* Add links to utils pages on pcoh doxygen page

### DIFF
--- a/src/Persistent_cohomology/example/plain_homology.cpp
+++ b/src/Persistent_cohomology/example/plain_homology.cpp
@@ -75,7 +75,8 @@ int main() {
   st.initialize_filtration();
 
   // Class for homology computation
-  // We want persistent homology to be computed for the maximal dimension in the complex (persistence_dim_max = true)
+  // By default, since the complex has dimension 1, only 0-dimensional homology would be computed.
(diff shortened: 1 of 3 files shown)


src/Persistent_cohomology/example/plain_homology.cpp
```diff
@@ -50,26 +50,33 @@ int main() {
...
   st.insert_simplex(edge13);
+  st.insert_simplex_and_subfaces(edge35);
   st.insert_simplex(vertex4);
 
   // Sort the simplices in the order of the filtration
   st.initialize_filtration();
 
   // Class for homology computation
-  Persistent_cohomology pcoh(st);
+  // We want persistent homology to be computed for the maximal dimension in the complex (persistence_dim_max = true)
```
mglisse:
```suggestion
  // By default, since the complex has dimension 1, only 0-dimensional homology would be computed.
  // Here we also want persistent homology to be computed for the maximal dimension in the complex (persistence_dim_max = true)
```

mglisse:
ok with the suggested change.

PR #71, opened 2019-06-13
sklearn-tda

MathieuCarriere:
Branch for the integration of sklearn_tda into Gudhi.


MathieuCarriere:
I don't know if the current location of the python files is the best (gudhi-devel/src/cython)... Also, I guess it would be nice to eventually compile these files with cython, so that we could call them in python with something like gudhi.sktda.PersistenceImage(), although I don't know how to do that.

mglisse:
I thought the plan was to start with just a couple files dropped in cython/, to make it easier and faster to integrate. So the new plan is to do everything at once?
I see you are putting all the sklearn-related stuff in a submodule gudhi.sktda. That seems like a good idea, although the name sktda may be misleading, it looks like a short version of scikit-tda (a different project), while this is about scikit-**learn**.
Ah, from a look at the examples, it looks like everything still ends up at the top-level, say clustering instead of gudhi.sktda.clustering? That's probably what you were asking about. We should indeed find a way to fix it (it may not require calling cython on those files, although that may be one way to do it).
`clustering` is a rather generic name. When we get tomato, should it be in the same submodule?

MathieuCarriere:
Sorry I probably misunderstood the plan. I can create another pull request with only a couple of files to make things easier if you want. Also, I don't know if it simplifies the reviewing process, but at least I added new files with separate commits.

VincentRouvreau:
What I am proposing to you ( @MathieuCarriere  @mglisse @bertrandmichel ) is to accept the merge request in this state (files are just added, but not really integrated into the GUDHI module). And I will work on the integration of this code (check if sklearn is present to add files in the module, add methods in a gudhi.sktda namespace, ...). Are you ok with that ?

mglisse:
@VincentRouvreau  I'd rather keep master clean at all times. Can't you work in a common branch?

VincentRouvreau:
Ok for a common branch.

mglisse:
> You are right about sktda. I think the only reason I used this name was to comply with the package naming convention (so no underscores),

Ah, I didn't know this convention. There are sklearn.cross_decomposition, discriminant_analysis, feature_extraction, etc, so this rule doesn't seem to be followed very strictly.

> but we should eventually have a package that explicitly mention sklearn_tda.

Would gudhi.sklearn work, or would it cause problems to reuse the name of a package?

> I am ok with changing the name `clustering` into something more explicit. Maybe `topological-clustering`?

That doesn't really help. If it comes inside gudhi.sklearn_tda, "topology" is already there (twice if we consider the topic of gudhi). And tomato is also topological. It can make sense for a submodule clustering to contain both tomato and some nerve/cover approach.
But then I am starting to wonder if we should really put inside a submodule sklearn_tda everything that has a method called `fit`. If we are organizing stuff, I think I'd rather have, in gudhi, one submodule for clustering methods, and one submodule for manipulation of diagrams (preprocessing, vectorizing, kernels). If both happen to use a convention from sklearn, that's ok.

> Also, I have implemented (for my own needs) a version of ToMATo in Python which fits into the sklearn_tda framework (that is, with fit and transform methods + possibility to use any graph or density estimation function from scikit-learn). I know that ToMATo is already being implemented in C++, but do you think it would be interesting to also add the Python version to Gudhi?

Could you send me that code? I was looking into integrating tomato in gudhi, and it seems natural to have the interface in python, so I'd like to compare the approaches. The sklearn interface did not seem like a perfect fit (pun!) for tomato.

bertrandmichel:
ok for accepting the merge request in this state

MathieuCarriere:
`gudhi.sklearn` sounds good to me. I also think that having two separated submodules (one for persistence diagram stuff and another for clustering methods such as Mapper and ToMATo) is a good idea.

@mglisse Ok I will send you my code as soon as I put my hands on it (I used it a long time ago...)

mglisse:
@MathieuCarriere Maybe you could remove the clustering/mapper part, it is independent, we will want to discuss how it interacts with CoverComplex, etc. I believe it would help to handle it in a separate PR. (removing a file doesn't mean closing this PR and opening a new one, you can just update your branch)

@bertrandmichel Do you have an opinion for the name of the module that contains various diagram tools (vectorization, kernel, etc)? Mathieu started with gudhi.sktda (well just sktda, but that's a technical issue), we mentioned gudhi.sklearn, but we could consider many things. Some random ideas: gudhi.diag(ram)manip, gudhi.diag(ram)tools, gudhi.persistencerepresentations, gudhi.diagrams, gudhi.sklearntda, gudhi.diagramsklearn, gudhi.sklearndiagrams... (and the variants with an underscore, it isn't forbidden, the guidelines say it is ok if it helps readability)

bertrandmichel:
I also think that two separate PR would be better.

I suggest gudhi.sklearn_representations (I don't think "persistence" is necessary here)

src/cython/sktda/metrics.py
```diff
@@ -0,0 +1,226 @@
...
+            else:
+                print("Gudhi required---returning null matrix")
+
+        Xfit = matrix
+
+        return Xfit
+
+class PersistenceFisherDistance(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the persistence Fisher distance matrix from a list of persistence diagrams. The persistence Fisher distance is obtained by computing the original Fisher distance between the probability distributions associated to the persistence diagrams given by convolving them with a Gaussian kernel. See papers.nips.cc/paper/8205-persistence-fisher-kernel-a-riemannian-manifold-kernel-for-persistence-diagrams for more details.
```
mglisse: Add `http://` or use a bibtex citation?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+
+class BirthPersistenceTransform(BaseEstimator, TransformerMixin):
+    """
+    This is a class for the affine transformation (x,y) -> (x,y-x) to be applied on persistence diagrams. It is a particular scaler for persistence diagram that can be given as input for the DiagramPreprocessor class.
+    """
+    def __init__(self):
+        """
+        Constructor for BirthPersistenceTransform class.
+        """
+        return None
```
mglisse: I have heard that having a `return` in `__init__` is considered bad style...

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+            y (n x 1 array): persistence diagram labels (unused).
+        """
+        return self
+
+    def transform(self, X):
+        """
+        Apply the BirthPersistenceTransform function on the persistence diagrams.
+
+        Parameters:
+            X (list of n x 2 or n x 1 numpy arrays): input persistence diagrams.
```
mglisse: What do n x 1 numpy arrays mean? Are they interpreted as [birth1,death1,birth2,death2,etc]?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Returns:
+            Xfit (list of n x 3 or n x 2 numpy arrays): padded persistence diagrams.
+        """
+        if self.use:
+            Xfit, num_diag = [], len(X)
+            max_card = max([len(diag) for diag in X])
```
mglisse: Ah, I expected this line to be in `fit`. Is it just that it doesn't matter because transform will only ever be called on the same argument as fit?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def transform(self, X):
+        """
+        Add dummy points to each persistence diagram so that they all have the same cardinality. All points are given an additional coordinate indicating if the point was added after padding (0) or already present before (1).  
+
+        Parameters:
+            X (list of n x 2 or n x 1 numpy arrays): input persistence diagrams.
+
+        Returns:
+            Xfit (list of n x 3 or n x 2 numpy arrays): padded persistence diagrams.
```
mglisse: Doesn't it cause problems that this transform returns something of a different shape than the rest, which doesn't directly look like a persistence diagram?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+        """
+        if self.use:
+            Xfit, num_diag = [], len(X)
+            max_card = max([len(diag) for diag in X])
+            for diag in X:
+                [num_pts, dim] = diag.shape
+                diag_pad = np.zeros([max_card, dim+1])
+                diag_pad[:num_pts,:dim] = diag
+                diag_pad[:num_pts, dim] = np.ones(num_pts)
```
mglisse: numpy.pad might make it one line shorter, but that doesn't matter.

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+        Constructor for the ProminentPoints class.
+     
+        Attributes:
+            use (bool): whether to use the class or not (default False).
+            location (string): either "upper" or "lower" (default "upper"). Whether to keep the points that are far away ("upper") or close ("lower") to the diagonal.
+            num_pts (int): cardinality threshold (default 10). If location == "upper", keep the top **num_pts** points that are the farthest away from the diagonal. If location == "lower", keep the top **num_pts** points that are the closest to the diagonal. 
+            threshold (double): distance-to-diagonal threshold (default -1). If location == "upper", keep the points that are at least at a distance **threshold** from the diagonal. If location == "lower", keep the points that are at most at a distance **threshold** from the diagonal. 
+            point_type (string): either "finite" if persistence diagrams are n x 2 numpy arrays, or "essential" if persistence diagrams are n x 1 numpy arrays (default "finite"). If "finite", points are ordered and thresholded by distance-to-diagonal. If "essential", points are ordered and thresholded by first coordinate.
```
mglisse: Ahhh, the n x 1 arrays were for infinite points... Why keep 2D arrays and not use 1D arrays?
We are going to need a global, introductory page for this whole module, that's one place where we could explain how essential features are handled.

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+        Returns:
+            Xfit (list of n x 2 or n x 1 numpy arrays): thresholded persistence diagrams.
+        """
+        if self.use:
+            Xfit, num_diag = [], len(X)
+            for i in range(num_diag):
+                diag = X[i]
+                if self.point_type == "finite":
+                    if diag.shape[0] > 0:
+                        pers       = np.abs(np.matmul(diag[:,:2], [-1., 1.]))
```
mglisse: A matrix multiplication seems like overkill, isn't it the same as `diag[:,1]-diag[:,0]`?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+                        idx_thresh = pers >= self.threshold
+                        thresh_diag, thresh_pers  = diag[idx_thresh.flatten()], pers[idx_thresh.flatten()]
```
mglisse: why do you need to flatten, isn't the array already 1D?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Parameters:
+            X (list of n x 2 or n x 1 numpy arrays): input persistence diagrams.
+            y (n x 1 array): persistence diagram labels (unused).
+        """
+        return self
+
+    def transform(self, X):
+        """
+        If location == "upper", first select the top **num_pts** points that are the farthest away from the diagonal, then select and return from these points the ones that are at least at distance **threshold** from the diagonal for each persistence diagram individually. If location == "lower", first select the top **num_pts** points that are the closest to the diagonal, then select and return from these points the ones that are at most at distance **threshold** from the diagonal for each persistence diagram individually. If point_type == "essential", do the same with first coordinate instead of distance-to-diagonal.
```
mglisse: I don't see any sorting in the essential case.

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+        Returns:
+            Xfit (list of n x 2 or n x 1 numpy arrays): extracted persistence diagrams.
+        """
+        if self.use:
+            Xfit, num_diag = [], len(X)
+            if self.point_type == "finite":
+                for i in range(num_diag):
+                    diag = X[i]
+                    if diag.shape[0] != 0:
+                        idx_fin = diag[:,1] != self.limit
```
mglisse: I wonder if we should test `<limit` instead of `!=limit`?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+                        Xfit.append(diag[idx_fin,:])
+                    else:
+                        Xfit.append(diag)
```
mglisse: `Xfit=[diag[diag[:,1]!=self.limit] for diag in X if diag.shape(0)!=0]`?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+    This is a class for extracting finite or essential points in persistence diagrams.
+    """
+    def __init__(self, use=False, limit=np.inf, point_type="finite"):
+        """
+        Constructor for the DiagramSelector class.
+
+        Attributes:
+            use (bool): whether to use the class or not (default False).
+            limit (double): second coordinate value that is the criterion for being an essential point (default numpy.inf).
+            point_type (string): either "finite" or "essential". The type of the points that are going to be extracted.
```
mglisse: When there are infinite points, separating them is indeed one natural thing to do. Sometimes, people prefer instead to clamp values, i.e. replace any value larger than 42 (including infinity) with 42. Would that fit better as a Scaler, since it keeps the same number of points?

src/cython/sktda/preprocessing.py
```diff
@@ -0,0 +1,274 @@
...
+            X (list of n x 2 or n x 1 numpy arrays): input persistence diagrams.
+
+        Returns:
+            Xfit (list of n x 2 numpy arrays): transformed persistence diagrams.
+        """
+        Xfit = np.matmul(X, np.array([[1., -1.],[0., 1.]]))
+        return Xfit
+
+
+class DiagramPreprocessor(BaseEstimator, TransformerMixin):
```
mglisse: The name looks awfully generic, preprocessing diagrams is what all the classes in this file do. Would DiagramScaler make sense? Or maybe not since it calls scalers instead of being one itself?
Is it standard with sklearn to have a wrapper for a list of scalers instead of using the scalers directly?

src/cython/sktda/vector_methods.py
```diff
@@ -0,0 +1,404 @@
...
+    
+        Returns:
+            Xfit (numpy array with shape (number of diagrams) x (number of pixels = **resolution[0]** x **resolution[1]**)): output persistence images.
+        """
+        num_diag, Xfit = len(X), []
+        for i in range(num_diag):
+
+            diagram, num_pts_in_diag = X[i], X[i].shape[0]
+
+            w = np.ones(num_pts_in_diag)
```
mglisse: np.empty makes it a bit clearer that it will be overwritten anyway.

src/cython/sktda/vector_methods.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        if np.isnan(np.array(self.im_range)).any():
+            pre = DiagramPreprocessor(use=True, scalers=[([0,1], MinMaxScaler())]).fit(X,y)
+            [mx,my],[Mx,My] = pre.scalers[0][1].data_min_, pre.scalers[0][1].data_max_
+            self.im_range = np.where(np.isnan(np.array(self.im_range)), np.array([mx, Mx, my, My]), np.array(self.im_range))
+        return self
+
+    def transform(self, X):
+        """
+        Compute the persistence image for each persistence diagram individually and concatenate the results.
```
mglisse: "concatenate" scared me. I read it as: you have k vectors of length n², and you merge them into one vector of length kn². While actually you are returning a "list" of vectors, which happens to be stored as a 2D array.

src/cython/sktda/vector_methods.py
```diff
@@ -0,0 +1,404 @@
...
+    """
+    def __init__(self, weight=lambda x: 1, resolution=100, sh_range=[np.nan, np.nan]):
+        """
+        Constructor for the Silhouette class.
+
+        Attributes:
+            weight (function): weight function for the persistence diagram points (default constant function, ie lambda x: 1). This function must be defined on 2D points, ie on lists or numpy arrays of the form [p_x,p_y].
+            resolution (int): number of samples for the weighted average (default 100).
+            sh_range ([double, double]): minimum and maximum for the weighted average domain, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). It is the interval on which samples will be drawn uniformly. If one of the values is numpy.nan, it can be computed from the persistence diagrams with the fit() method.
```
mglisse: Could we give this parameter the same name in landscapes and silhouettes? Maybe just `range`, or something more explicit.

src/cython/sktda/metrics.py
```diff
@@ -0,0 +1,226 @@
...
+import sys
+import numpy as np
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.metrics import pairwise_distances
+try:
+    from gudhi import bottleneck_distance
+    USE_GUDHI = True
+except ImportError:
+    USE_GUDHI = False
+    print("Gudhi not found: BottleneckDistance will return null matrix, and exact SlicedWassersteinDistance not available")
```
mglisse: @VincentRouvreau Is that how we want to handle it internally?

mglisse: You mention an exact version, but it doesn't seem to be available now. IIUC, we will add it later as a wrapper for the C++ function (that is perfectly fine). What will it look like? Is it an extra parameter for `__init__` or a different class? Just checking if this has an impact on the desired interface for the approximate class.

src/cython/sktda/metrics.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def transform(self, X):
+        """
+        Compute all sliced Wasserstein distances between the persistence diagrams that were stored after calling the fit() method, and a given list of (possibly different) persistence diagrams.
+
+        Parameters:
+            X (list of n x 2 numpy arrays): input persistence diagrams.
+
+        Returns:
+            Xfit (numpy array of shape (number of diagrams in **diagrams**) x (number of diagrams in X)): matrix of pairwise sliced Wasserstein distances.
```
mglisse: I had a look at scikit-learn, and in particular at functions that return a matrix of pairwise distances/kernels (sklearn.metrics.pairwise.sigmoid_kernel, sklearn.metrics.pairwise.cosine_similarity, etc) and they do not use fit/transform for those, just plain functions. Should we do the same?

src/cython/sktda/metrics.py
```diff
@@ -214,13 +214,19 @@ def transform(self, X):
                         Z = np.concatenate([approx[i], approx_diagonal[i], self.approx_[j], self.approx_diagonal_[j]], axis=0)
                         U, V = np.sum(np.concatenate([approx[i], self.approx_diagonal_[j]], axis=0), axis=0), np.sum(np.concatenate([self.approx_[j], approx_diagonal[i]], axis=0), axis=0) 
                         vectori, vectorj = np.matmul(Z, U.T), np.matmul(Z, V.T)
-                        vectori, vectorj = vectori/np.sum(vectori), vectorj/np.sum(vectorj)
-                        Xfit[i,j] = np.arccos(np.dot(np.sqrt(vectori), np.sqrt(vectorj)))
+                        if np.sum(vectori) != 0:
+                            vectori = vectori/np.sum(vectori)
```
mglisse: Why not give `np.sum(vectori)` a name, to avoid computing it twice?

src/cython/sktda/vector_methods.py
```diff
@@ -286,7 +292,78 @@ def transform(self, X):
                     bc[k] += 1
 
             Xfit.append(np.reshape(bc,[1,-1]))
-            Xfit = np.concatenate(Xfit, 0)
+
+        Xfit = np.concatenate(Xfit, 0)
+
+        return Xfit
+
+class Entropy(BaseEstimator, TransformerMixin):
```
mglisse: Did you see PR #62 and #72 and the discussions starting at https://lists.gforge.inria.fr/pipermail/gudhi-users/2019-March/000221.html and https://lists.gforge.inria.fr/pipermail/gudhi-users/2019-May/000263.html ? Maybe what you have is what we actually want, but it would be nice to acknowledge them and discuss the advantages.

MathieuCarriere:
Oops sorry Marc, I did not see that you have been commenting on the code since June. I will reply to your comments ASAP.

src/cython/sktda/metrics.py
MathieuCarriere: You are right, I added `http://`

src/cython/sktda/preprocessing.py
MathieuCarriere: Indeed, BirthPersistenceTransform is limited to nx2 diagrams. I corrected the doc.

src/cython/sktda/preprocessing.py
MathieuCarriere: You are right, I moved this line to `fit`

src/cython/sktda/preprocessing.py
MathieuCarriere: I agree that this might be confusing (I created this class originally because it is the input format that PersLay expects, I don't know how useful this is for other applications though). I tried to make all other classes compatible, but some are not (like BirthPersistenceTransform, which requires nx2 arrays, or classes in `kernel_methods`). I guess another possibility could be to output two arrays: the padded persistence diagrams, and a boolean array, that would be the corresponding mask, although if we do that, we won't be able to use this class in a sklearn pipeline.

src/cython/sktda/preprocessing.py
MathieuCarriere: Indeed, I shortened the code with `np.pad`

src/cython/sktda/preprocessing.py
MathieuCarriere: I kept 2D arrays because I think sklearn always expect 2D arrays (not sure though). It was just to be consistent between finite and essential representations of diagrams. You are right that we can read the type with the input shape though, so I fixed this.

src/cython/sktda/preprocessing.py
MathieuCarriere: You are right, I removed `matmul`. And yes, `:2` is for the case where diagrams have been padded.

src/cython/sktda/preprocessing.py
MathieuCarriere: No need indeed. I fixed the code.

src/cython/sktda/preprocessing.py
MathieuCarriere: Yes you are right, sorting is not justified here because it is hard to rank the diagram points (as opposed to the finite case, where the distance to the diagonal has a meaning).

src/cython/sktda/preprocessing.py
MathieuCarriere: Yes, that sounds better. I modified the code

src/cython/sktda/preprocessing.py
MathieuCarriere: Nice! Indeed, that does the same thing. I shortened the code

src/cython/sktda/preprocessing.py
MathieuCarriere: Yes, this makes total sense. I added a `Clamping` class that can be called from within `DiagramScaler`

src/cython/sktda/preprocessing.py
MathieuCarriere: I agree, `DiagramScaler` is better. I do not think this kind of wrapper is very standard in `sklearn`, but I did not want to re-implement all the traditional scalers of `sklearn` for persistence diagrams...

src/cython/sktda/vector_methods.py
MathieuCarriere: You are right, I modified the code

src/cython/sktda/vector_methods.py
MathieuCarriere: Yes, you are right, I changed the doc to clarify what is done

src/cython/sktda/vector_methods.py
MathieuCarriere: I renamed it to `sample_range` for `Landscape`, `Silhouette`, `BettiCurve` and `Entropy` classes

src/cython/sktda/metrics.py
MathieuCarriere: I was thinking of including the possibility to call the class with `num_directions` = -1, which would mean: do exact computation. Although, you are right, it is not available right now. It is even worse than this: the last time I checked, the exact computation code in C++ was not working on an example I had, so I have to debug and fix the exact computation code

src/cython/sktda/metrics.py
MathieuCarriere: I agree, usually `sklearn` uses plain functions for this kind of computation. The reason I sticked to a class with `fit` and `transform` methods is because it allows me to treat k-NN with bottleneck distances in the same way than other vector or kernel methods. See the notebook http://localhost:8888/notebooks/3DSeg/3DSegTDA.ipynb for an example

src/cython/sktda/metrics.py
MathieuCarriere: Yes, I modified the code

src/cython/sktda/vector_methods.py
MathieuCarriere: Sorry, I did not know some people were already integrating it in the library. I will check their code

VincentRouvreau:
I merged last master modifications for Python package.
I moved `sktda` in `src/python/gudhi` (no more cython directory).
I commented `DiagramPreprocessor` as it is not defined (all the other tests works perfectly).

You can now `make python`, then:
```python
from gudhi.sktda import Landscape
l = Landscape()
```

mglisse:
`gudhi/__init__.py.in` has code to import everything from all submodules of gudhi. Does it do what we want here, or should we make it handle this submodule differently? I haven't tested.

tlacombe:
There is no link on the main documentation page (/src/python/doc/index.rst) --- you can check it with $ firefox src/python/sphinx/index.html.

mglisse:
The user manual doesn't seem to exist (a file in src/python/doc), and sphinx doesn't seem to generate any documentation for this code (even if I look directly at the list of files it produces).

src/python/gudhi/sktda/kernel_methods.py
```diff
@@ -0,0 +1,202 @@
+"""
+@author: Mathieu Carriere
+All rights reserved
```
mglisse: According to [wikipedia](https://en.wikipedia.org/wiki/All_rights_reserved), this is useless.

mglisse:
I took the liberty to push a file so sphinx will compile the documentation, this isn't meant to be definitive of course, at least a couple lines of introductory documentation in the user manual are required, the file may need a different name, or be split into 4 files depending on how you want the doc to look like, etc.

src/python/gudhi/sktda/kernel_methods.py
```diff
@@ -0,0 +1,202 @@
+"""
+@author: Mathieu Carriere
```
mglisse: This appears verbatim in the sphinx documentation, probably not what you want. Current source files in gudhi seem to handle it differently. On the other hand, I don't see the author for the RipsComplex reference page. We could set `show_authors=True` in doc/conf.py, and add `.. codeauthor:: Me me me` or `.. sectionauthor:: Myself` in the .rst file. Or don't use automation, specify the author manually in the ref page the same way as the user page, and specify the author in the source file in a way that doesn't show in sphinx, as we do in most files currently.

mglisse:
Should the new files have a license header?

mglisse:
It would be good to have at least 1 simple test. We could just run the example (a simple addition to CMakeLists.txt), but it expects some interactive use (it shows pictures that the user has to close). If I run it with `MPLBACKEND=agg` in my environment (not sure if that works under windows), it seems to succeed, with just a warning
> /usr/lib/python3/dist-packages/matplotlib/figure.py:445: UserWarning: Matplotlib is currently using agg, which is a non-GUI backend, so cannot show the figure.

for every `plt.show()`. I can also do `plt.show=lambda:None`, but that would require a way to know if the example is called from the testsuite or not. It may be easier to duplicate the file...

src/cython/sktda/metrics.py
mglisse: Side note: if you find an issue with gudhi (including your own C++ code) that you cannot fix right away, it can be a good idea to file an issue explaining how to repoduce the error, so the information is not lost (even for yourself). Of course if you can fix it immediately, just a pull request is fine, no need for the issue.

src/python/gudhi/sktda/preprocessing.py
```diff
@@ -0,0 +1,299 @@
...
+        Xfit = np.where(X >= self.limit, self.limit * np.ones(X.shape), X)
+        return Xfit
+
+class DiagramScaler(BaseEstimator, TransformerMixin):
+    """
+    This is a class for preprocessing persistence diagrams with a given list of scalers, such as those included in scikit-learn.
+    """
+    def __init__(self, use=False, scalers=[]):
+        """
+        Constructor for the DiagramPreprocessor class.
```
mglisse:
```suggestion
        Constructor for the DiagramScaler class.
```
and similarly in several places below, since the class was renamed.

src/python/example/ex_diagrams.py
```diff
@@ -0,0 +1,108 @@
...
+
+TV = TopologicalVector(threshold=-1)
+tv = TV.fit_transform(diags)
+print("Topological vector is " + str(tv[0,:]))
+
+#diagsT = DiagramPreprocessor(use=True, scalers=[([0,1], BirthPersistenceTransform())]).fit_transform(diags)
+#PI = PersistenceImage(bandwidth=1., weight=lambda x: x[1], im_range=[0,10,0,10], resolution=[100,100])
+#pi = PI.fit_transform(diagsT)
+#plt.imshow(np.flip(np.reshape(pi[0], [100,100]), 0))
+#plt.show()
```
mglisse: Is there a problem with PersistenceImage, that its example is commented out?

src/python/gudhi/sktda/kernel_methods.py
```diff
@@ -0,0 +1,202 @@
...
+# Kernel methods ############################
+#############################################
+
+class SlicedWassersteinKernel(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the sliced Wasserstein kernel matrix from a list of persistence diagrams. The sliced Wasserstein kernel is computed by exponentiating the corresponding sliced Wasserstein distance with a Gaussian kernel. See http://proceedings.mlr.press/v70/carriere17a.html for more details. 
+    """
+    def __init__(self, num_directions=10, bandwidth=1.0):
+        """
+        Constructor for the SlicedWassersteinDistance class.
```
mglisse:
```suggestion
        Constructor for the SlicedWassersteinKernel class.
```

src/python/gudhi/sktda/metrics.py
```diff
@@ -0,0 +1,239 @@
...
+class SlicedWassersteinDistance(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the sliced Wasserstein distance matrix from a list of persistence diagrams. The Sliced Wasserstein distance is computed by projecting the persistence diagrams onto lines, comparing the projections with the 1-norm, and finally integrating over all possible lines. See http://proceedings.mlr.press/v70/carriere17a.html for more details. 
+    """
+    def __init__(self, num_directions=10):
+        """
+        Constructor for the SlicedWassersteinDistance class.
+
+        Attributes:
+            num_directions (int): number of lines to sample uniformly from [-pi,pi] in order to approximate and speed up the distance computation (default 10).
```
mglisse: "sample uniformly" is a bit ambiguous, it often means i.i.d. random values, whereas here the values are evenly spaced.

src/cython/sktda/metrics.py
mglisse: > `num_directions` = -1, which would mean: do exact computation

That sounds good to me (although I would also interpret 0 directions as exact, maybe anything <=0), but we should remove the documentation of this possibility as long as the functionality isn't there.

src/cython/sktda/preprocessing.py
mglisse: Wow, I shouldn't have said anything, the new code is harder to read. But ok.

src/cython/sktda/preprocessing.py
mglisse: So, should the doc say that for essential points (aka nx1 arrays) we return an arbitrary subset of num_pts points? Actually, no, the code seems to compare the birth time to the threshold?

src/python/gudhi/sktda/preprocessing.py
```diff
@@ -0,0 +1,299 @@
...
+
+        Returns:
+            Xfit (list of n x 2 or n x 1 numpy arrays): extracted persistence diagrams.
+        """
+        if self.use:
+            Xfit, num_diag = [], len(X)
+            if self.point_type == "finite":
+                Xfit = [ diag[diag[:,1] < self.limit] if diag.shape[0] != 0 else diag for diag in X]
+            else:
+                Xfit = [ diag[diag[:,1] == self.limit, 0:1] if diag.shape[0] != 0 else diag for diag in X]
```
mglisse:
```suggestion
                Xfit = [ diag[diag[:,1] >= self.limit, 0:1] if diag.shape[0] != 0 else diag for diag in X]
```
So everything is either finite or infinite.

src/python/gudhi/sktda/preprocessing.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        Clamp list of values.
+
+        Parameters:
+            X (numpy array of size n): input list of values.
+
+        Returns:
+            Xfit (numpy array of size n): output list of values.
+        """
+        Xfit = np.where(X >= self.limit, self.limit * np.ones(X.shape), X)
```
mglisse: There is np.full to get a constant array, but np.fmin or np.minimum seem even simpler here.

src/python/gudhi/sktda/preprocessing.py
```diff
@@ -0,0 +1,299 @@
...
+        Parameters:
+            X (list of n x 2 numpy array): input persistence diagrams.
+
+        Returns:
+            Xfit (list of n x 2 numpy array): transformed persistence diagrams.
+        """
+        Xfit = []
+        for diag in X:
+            new_diag = np.empty(diag.shape)
+            np.copyto(new_diag, diag)
```
mglisse: Just `new_diag = np.copy(diag)`?

src/cython/sktda/metrics.py
mglisse: That seems to make sense. Which confuses me: why doesn't sklearn use its own fit/transform interface, if that would make its use more convenient?

src/python/gudhi/sktda/kernel_methods.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        self.diagrams_ = list(X)
+        self.ws_ = [ np.array([self.weight(self.diagrams_[i][j,:]) for j in range(self.diagrams_[i].shape[0])]) for i in range(len(self.diagrams_)) ]
+        if self.kernel_approx is not None:
+            self.approx_ = np.concatenate([np.sum(np.multiply(self.ws_[i][:,np.newaxis], self.kernel_approx.transform(self.diagrams_[i])), axis=0)[np.newaxis,:] for i in range(len(self.diagrams_))])
+        return self
+
+    def transform(self, X):
+        """
+        Compute all sliced Wasserstein kernel values between the persistence diagrams that were stored after calling the fit() method, and a given list of (possibly different) persistence diagrams.
```
mglisse:
```suggestion
        Compute all persistence weighted Gaussian kernel values between the persistence diagrams that were stored after calling the fit() method, and a given list of (possibly different) persistence diagrams.
```

src/python/gudhi/sktda/kernel_methods.py
```diff
@@ -0,0 +1,202 @@
...
+                    for j in range(len(self.diagrams_)):
+                        W = np.matmul(ws[i][:,np.newaxis], self.ws_[j][np.newaxis,:])
+                        E = (1./(np.sqrt(2*np.pi)*self.bandwidth)) * np.exp(-np.square(pairwise_distances(Xp[i], self.diagrams_[j]))/(2*np.square(self.bandwidth)))
+                        Xfit[i,j] = np.sum(np.multiply(W, E))
+        
+        return Xfit
+
+class PersistenceScaleSpaceKernel(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the persistence scale space kernel matrix from a list of persistence diagrams. The persistence scale space kernel is computed by adding the symmetric to the diagonal of each point in each persistence diagram, and then convolving the points with a Gaussian kernel. See https://www.cv-foundation.org/openaccess/content_cvpr_2015/papers/Reininghaus_A_Stable_Multi-Scale_2015_CVPR_paper.pdf for more details.
```
mglisse:
```suggestion
    This is a class for computing the persistence scale space kernel matrix from a list of persistence diagrams. The persistence scale space kernel is computed by adding the symmetric to the diagonal of each point in each persistence diagram, with negative weight, and then convolving the points with a Gaussian kernel. See https://www.cv-foundation.org/openaccess/content_cvpr_2015/papers/Reininghaus_A_Stable_Multi-Scale_2015_CVPR_paper.pdf for more details. 
```
Or something like that?

src/python/gudhi/sktda/metrics.py
```diff
@@ -0,0 +1,239 @@
...
+        Attributes:
+            num_directions (int): number of lines to sample uniformly from [-pi,pi] in order to approximate and speed up the distance computation (default 10). 
+        """
+        self.num_directions = num_directions
+        thetas = np.linspace(-np.pi/2, np.pi/2, num=self.num_directions+1)[np.newaxis,:-1]
+        self.lines_ = np.concatenate([np.cos(thetas), np.sin(thetas)], axis=0)
+
+    def fit(self, X, y=None):
+        """
+        Fit the SlicedWassersteinDistance class on a list of persistence diagrams: persistence diagrams are projected onto the different lines. The diagrams themselves and their projections are then stored in a numpy array called **diagrams**.
```
mglisse: I don't see any member called `diagrams` in this class.

src/python/gudhi/sktda/preprocessing.py
```diff
@@ -0,0 +1,299 @@
...
+            limit (double): clamping value (default np.inf).
+        """
+        self.limit = limit
+
+    def fit(self, X, y=None):
+        """
+        Fit the Clamping class on a list of values (this function actually does nothing but is useful when Clamping is included in a scikit-learn Pipeline).
+
+        Parameters:
+            X (numpy array of size n): input values.
```
mglisse: As written, it should work for any shape (and in particular n x 2 arrays representing diagrams), not just 1d arrays? Or is `size` used in numpy for the total size, independently of the shape? It might be clearer to remove "of size n" and say that Xfit has the same shape as X.
I first wondered if it should remove intervals that become empty (birth>=limit), but I guess that's not its role, I can compose it with another preprocessor that does this, it's good.
If I have a 1d array representing the birth of essential features, with some that are born after `limit`, currently this makes them appear at `limit` instead. I wonder if it would be more logical to remove them instead, if we interpret this as stopping the filtration at some parameter. Not sure.

src/python/gudhi/sktda/preprocessing.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        self.scalers  = scalers
+        self.use      = use
+
+    def fit(self, X, y=None):
+        """
+        Fit the DiagramPreprocessor class on a list of persistence diagrams: persistence diagrams are concatenated in a big numpy array, and scalers are fit (by calling their fit() method) on their corresponding coordinates in this big array.
+
+        Parameters:
+            X (list of n x 2 or n x 1 numpy arrays): input persistence diagrams.
```
mglisse: I guess that all the arrays need to have the same shape, I cannot mix nx2 and nx1 arrays, or even nx2 and mx2, right?

src/cython/sktda/preprocessing.py
mglisse: Ok, it is just this one class, users (and most of your other classes) can just ignore the last coordinate, it doesn't prevent people who want to from adding a different padding class.
Maybe specify in the doc what the dummy point is (the origin (0,0)?).
If it isn't the case yet, it should be easy to have BirthPersistenceTransform accept such arrays, the question is more if it should return nx2 (drop the extra coordinate) or nx3 (ignore but keep it). Not super important though, one can just use BPT before padding.

src/python/gudhi/sktda/vector_methods.py
```diff
@@ -0,0 +1,481 @@
...
+class TopologicalVector(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing topological vectors from a list of persistence diagrams. The topological vector associated to a persistence diagram is the sorted vector of a slight modification of the pairwise distances between the persistence diagram points. See https://diglib.eg.org/handle/10.1111/cgf12692 for more details.
+    """
+    def __init__(self, threshold=10):
+        """
+        Constructor for the TopologicalVector class.
+
+        Attributes:
+            threshold (int): number of distances to keep (default 10). This is the dimension of the topological vector. If -1, this threshold is computed from the list of persistence diagrams by considering the one with the largest number of points and using the dimension of its corresponding topological vector as threshold.
```
mglisse: I wonder if `threshold=0` (maybe any value <=0) should also mean to autodetect? Or for genericity can it make sense to ask for an empty feature?

src/python/gudhi/sktda/vector_methods.py
```diff
@@ -0,0 +1,481 @@
...
+            dim = min(len(vect), thresh)
+            Xfit[i, :dim] = vect[:dim]
+
+        return Xfit
+
+class ComplexPolynomial(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing complex polynomials from a list of persistence diagrams. The persistence diagram points are seen as the roots of some complex polynomial, whose coefficients are returned in a complex vector. See https://link.springer.com/chapter/10.1007%2F978-3-319-23231-7_27 for more details.
+    """
+    def __init__(self, F="R", threshold=10):
```
mglisse: The name `F` doesn't speak a lot (function/map/transform/kind/variant), but I don't mind that much.

src/python/gudhi/sktda/vector_methods.py
```diff
@@ -0,0 +1,481 @@
...
+        """
+        Constructor for the ComplexPolynomial class.
+
+        Attributes:
+           F (char): either "R", "S" or "T" (default "R"). Type of complex polynomial that is going to be computed (explained in https://link.springer.com/chapter/10.1007%2F978-3-319-23231-7_27).
+           threshold (int): number of coefficients (default 10). This is the dimension of the complex vector of coefficients. If -1, this threshold is computed from the list of persistence diagrams by considering the one with the largest number of points and using the dimension of its corresponding complex vector of coefficients as threshold.
```
mglisse: Do you keep the low coefficients or the high coefficients?

mglisse:
> I suggest gudhi.sklearn_representations

Bertrand's sklearn_representations seems to be the last name suggestion in the thread (after sktda, sklearn, sklearn_tda, etc). Is everyone ok with it, can we do one big rename? Any of the suggested names would work for me, I just want us to settle on one.

src/python/gudhi/sktda/vector_methods.py
```diff
@@ -0,0 +1,481 @@
...
+
+            Xfit.append(np.reshape(bc,[1,-1]))
+
+        Xfit = np.concatenate(Xfit, 0)
+
+        return Xfit
+
+class Entropy(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing persistence entropy. Persistence entropy is a statistic for persistence diagrams inspired from Shannon entropy. This statistic can also be used to compute a feature vector, called the entropy summary function. See https://arxiv.org/pdf/1803.08304.pdf for more details.
```
mglisse: Maybe we can mention here: "A previous implementation was contributed by ...", which makes it clear that they are not the authors of this one, but that they did contribute to the project? There may be a more natural place for this comment in the user manual once it exists (a history section?).

MathieuCarriere: I totally agree. I added a short sentence based on your comment, and of course I would not mind having a more detailed comment on this.

src/python/gudhi/sktda/vector_methods.py
MathieuCarriere: High coefficients. Doc should be more explicit now.

src/python/gudhi/sktda/vector_methods.py
MathieuCarriere: I agree. I used those letters to be consistent with the associated paper, but it is hard to make sense of them.

src/python/gudhi/sktda/vector_methods.py
MathieuCarriere: Yes I am not sure either on what should be the best option here. Another possibility is to replace -1 by a string, like "auto" (I think scikit-learn does that often).

src/python/gudhi/sktda/preprocessing.py
MathieuCarriere: Since the code concatenate the diagrams together wrt 1st coordinate, you cannot use nx2 and nx1, but diagrams with different number of points, like nx2 and mx2, should work fine.

src/python/gudhi/sktda/metrics.py
MathieuCarriere: Indeed, the members are actually called `diagrams_` and `approx_diag_`. It should be fixed now.

src/python/gudhi/sktda/kernel_methods.py
MathieuCarriere: Thanks! I fixed this.

src/python/gudhi/sktda/kernel_methods.py
MathieuCarriere: Thanks! I fixed this.

src/python/gudhi/sktda/preprocessing.py
MathieuCarriere: Thanks! I fixed this.

src/python/gudhi/sktda/preprocessing.py
MathieuCarriere: Thanks I fixed this.

src/python/gudhi/sktda/preprocessing.py
MathieuCarriere: Thanks! I fixed this.

src/python/gudhi/sktda/metrics.py
MathieuCarriere: I agree. I replaced with "evenly sampled".

src/python/gudhi/sktda/kernel_methods.py
MathieuCarriere: Thanks! I fixed this.

src/python/example/ex_diagrams.py
MathieuCarriere: I changed a little bit the examples, and forgot to push the new ones. Nothing should be commented in the file now.

src/python/gudhi/sktda/preprocessing.py
MathieuCarriere: Indeed. Should be fixed.

MathieuCarriere:
I am fine with `gudhi.sklearn_representations`.

src/python/gudhi/sktda/metrics.py
mglisse:
```suggestion
            num_directions (int): number of lines evenly sampled on [-pi/2,pi/2] in order to approximate and speed up the distance computation (default 10). 
```
according to the code below?

src/python/gudhi/sktda/vector_methods.py
```diff
@@ -0,0 +1,481 @@
...
+class TopologicalVector(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing topological vectors from a list of persistence diagrams. The topological vector associated to a persistence diagram is the sorted vector of a slight modification of the pairwise distances between the persistence diagram points. See https://diglib.eg.org/handle/10.1111/cgf12692 for more details.
+    """
+    def __init__(self, threshold=10):
+        """
+        Constructor for the TopologicalVector class.
+
+        Attributes:
+            threshold (int): number of distances to keep (default 10). This is the dimension of the topological vector. If , this threshold is computed from the list of persistence diagrams by considering the one with the largest number of points and using the dimension of its corresponding topological vector as threshold.
```
mglisse: Er, did you mean to replace -1 with something?

src/python/gudhi/sktda/vector_methods.py
mglisse: I wasn't talking about R/S/T, indeed you didn't invent those. I was talking about F.

mglisse:
> I am fine with `gudhi.sklearn_representations`.

Apparently @martinroyer would prefer a name without "sklearn" in it. Maybe `gudhi.representations`? I don't have a strong opinion there.

src/python/gudhi/sktda/metrics.py
mglisse: An alternative would have been not to give the names (I don't think users are expected to go look at them), but this is fine, thanks.

bertrandmichel:
I agree and gudhi.representations is fine for me.

mglisse:
@VincentRouvreau the documentation issue in circleci seems to come from
> WARNING: autodoc: failed to import module 'sktda.preprocessing' from module 'gudhi'; the following exception was raised:
> No module named 'sklearn'

Does that mean we just need to add scikit-learn to the CI configuration?

src/python/gudhi/sktda/metrics.py
MathieuCarriere: You are totally right. Just fixed it.

src/python/gudhi/sktda/vector_methods.py
MathieuCarriere: Oops, I started to replace -1 with "auto" and then I decided to wait for your opinion, but I forgot to put -1 back. Just did it.

MathieuCarriere:
I would slightly prefer having `sklearn` in the name, but I am fine with `gudhi.representations`. However, I think we should strongly emphasize (when we write the doc) that it is meant for `scikit-learn` since this is the whole point of the module.

src/python/gudhi/sktda/metrics.py
mglisse: You fixed the other one, not the one I commented on.

MathieuCarriere: Right. Just corrected both.

src/python/gudhi/sktda/vector_methods.py
MathieuCarriere: I see. I replaced "F" with "polynomial_type"

martinroyer:
> I would slightly prefer having `sklearn` in the name, but I am fine with `gudhi.representations`. However, I think we should strongly emphasize (when we write the doc) that it is meant for `scikit-learn` since this is the whole point of the module.

I guess my point is that the emphasis on `sklearn` is conveying information on code design, that is "the module is sklearn-compatible", whereas I would expect (?) a submodule name to be informative as to what it does in essence i.e. "persistence representations in gudhi".

mglisse:
The documentation of functions is written in some informal way, like
> bandwidth (double): blablabla

Sphinx requires some more specific syntax. Looking at existing files, it seems that it expects something more like
> :param bandwidth: blablabla
> :type bandwidth: float

src/python/gudhi/sktda/metrics.py
mglisse: Do we want to also replace it in other places, say for landscapes, betti curves, etc?

martinroyer:
> The documentation of functions is written in some informal way, like
> 
> > bandwidth (double): blablabla
> 
> Sphinx requires some more specific syntax. Looking at existing files, it seems that it expects something more like
> 
> > :param bandwidth: blablabla
> > :type bandwidth: float


This is a problem though, as it makes it harder to read the doc string than @MathieuCarriere 's proposal (consistent with http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

mglisse:
> This is a problem though, as it makes it harder to read the doc string than @MathieuCarriere 's proposal (consistent with http://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings)

Apparently scikit-learn uses a style like Mathieu's, and they use sphinx to generate their doc, so it shouldn't be contradictory. I have no idea what sklearn does differently so sphinx groks their syntax though, it produces something ugly here.

mglisse:
> I have no idea what sklearn does differently so sphinx groks their syntax though

Adding 'numpydoc' or 'sphinx.ext.napoleon' in `extensions` in conf.py is supposed to help.

martinroyer:
> > I have no idea what sklearn does differently so sphinx groks their syntax though
> 
> Adding 'numpydoc' or 'sphinx.ext.napoleon' in `extensions` in conf.py is supposed to help. numpydoc didn't work for me (I didn't investigate). Napoleon does help, but the syntax still doesn't exactly match Mathieu's. At least it doesn't seem to break the existing doc in Sphinx style, which is good.

Yes, I think using 'sphinx.ext.napoleon' and having @MathieuCarriere comply with usage shown in https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html is the right way to proceed! (it looks like the 'Returns:' syntax is the only problem?)

mglisse:
> (it looks like the 'Returns:' syntax is the only problem?)

`__init__` also looks strange, it looks like it should say "parameters" like regular functions, not "attributes".

src/python/gudhi/sktda/kernel_methods.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        Fit the PersistenceScaleSpaceKernel class on a list of persistence diagrams: symmetric to the diagonal of all points are computed and an instance of the PersistenceWeightedGaussianKernel class is fitted on the diagrams and then stored. 
+
+        Parameters:
+            X (list of n x 2 numpy arrays): input persistence diagrams.
+            y (n x 1 array): persistence diagram labels (unused).
+        """
+        self.diagrams_ = list(X)
+        for i in range(len(self.diagrams_)):
+            op_D = np.matmul(self.diagrams_[i], np.array([[0.,1.], [1.,0.]]))
```
mglisse: Out of curiosity, would something like `op_D = self.diagrams_[i][:,[1,0]]` be equivalent?

src/python/gudhi/sktda/vector_methods.py
mglisse: Coming from types languages, my first idea is obviously a magic value of the same type (0 or -1). It is true that python allows passing `None` or some magic string like "auto". If nobody else comments, you get to pick (just be consistent).

mglisse:
> `gudhi/__init__.py.in` has code to import everything from all submodules of gudhi.

I moved sktda to "extra" modules, like wasserstein, which are not auto-imported. `import gudhi` does not import sktda, and after `import gudhi.sktda`, `gudhi.sktda.vector_methods.Landscape` is also available as `gudhi.sktda.Landscape` and that's it.

src/python/gudhi/sktda/kernel_methods.py
MathieuCarriere: Indeed, and your suggestion is simpler. Just changed the code.

mglisse: Ok, if you like it, there is another instance in PersistenceScaleSpaceKernel.

mglisse:
Let me present what remains as a task list. Before merging:
- [ ] write a user manual (it can be separate from the reference manual, like CoverComplex, or 2-in-1, like bottleneck. Should say that it is meant to work with scikit-learn, that diagrams (for 1 dimension) are handled as n x 2 numpy arrays, or n x 1 for points at infinity. A link to the example would be nice), and a 3 line blurb with an image for the main doc page
- [ ] convert docstrings to a sphinx-friendly syntax (starting with "Returns:" that should not name a variable)
- [ ] rename gudhi.sktda to gudhi.representations (or other)
- [ ] add one test (possibly just run the example as a test, if we can omit the graphic part)
- [ ] add scikit-learn to docker for CI
- [ ] settle the question of magic arguments (-1? 0? None? "auto"? "max"?)
- [ ] remove doc of unimplemented feature (exact sliced wasserstein)

Soon after merging (of course it doesn't hurt to solve them before):
- [ ] see how authors appear in the sphinx doc
- [ ] see how we handle builds without CGAL (bottleneck)
- [ ] make one final pass on this super long discussion to see if anything relevant remains and should be changed or filed as an issue for later (I know I skipped a few that didn't seem important enough to be on the list)

@MathieuCarriere do let us know if you need help with some items. I think you should be the one to write the user manual (at least a minimal one), but we can help for the rest, if we know you aren't working on it at the same time.

mglisse:
:disappointed: having the license as a global string doesn't interact so well with automodule (it prints it at the top of the doc, whereas we want it hidden). I think we should instead write it as a comment (lines starting with `#`). Global strings in the file can instead be used to write the documentation, if some prefer having it with the code instead of in the .rst file.

mglisse:
If you want to look at the generated doc but aren't able to build it yourself, the procedure is:
- at the bottom of this page, there is a list of checks. Possibly click on "show all checks" to unfold the list.
- Click on "Details" next to the check called "build"
- click on "python" (one of 5 tests)
- click on "artifacts". If you don't see any such link, go to circleci.com and log in with your github account, then try again.
- click on tmp/sphinx/index.html

src/python/gudhi/sktda/vector_methods.py
mglisse: Let's just stick to -1 since that's what the code currently does, less work.

mglisse:
> I can also do `plt.show=lambda:None`

That's what I ended up doing, importing the example from a test. It isn't pretty...

mglisse:
I pushed some documentation (very minimalist). As far as I am concerned, the branch can be merged. Although I would **really** like someone to write a more suitable documentation...

MathieuCarriere:
Hi Marc, sorry for the late answer, my schedule is crazy these days. I will try to write a nice doc tomorrow.

src/python/gudhi/representations/preprocessing.py
```diff
@@ -0,0 +1,305 @@
...
+        Constructor for BirthPersistenceTransform class.
+        """
+        return None
+
+    def fit(self, X, y=None):
+        """
+        Fit the BirthPersistenceTransform class on a list of persistence diagrams (this function actually does nothing but is useful when BirthPersistenceTransform is included in a scikit-learn Pipeline).
+
+        Parameters:
+            X (n x 2 numpy array): input persistence diagrams.
```
mglisse: Should the type actually be "**list of** n x 2 numpy arrays", like in transform? I know it is ignored anyway...

Issue #35, opened 2019-03-07
[Global - Copyright] Python Copyrights mention that Gudhi is a generic C++ library, but no mention about Python


VincentRouvreau:
The actual Copyrights says:
This file is part of the Gudhi Library. The Gudhi library (Geometric Understanding in Higher Dimensions) is a generic C++ library for computational topology.
It shall say:
This file is part of the Gudhi Library. The Gudhi library (Geometric Understanding in Higher Dimensions) is a generic C++ library, with a Python interface, for computational topology.
All Copyrights to be modified.

VincentRouvreau:
Or we could skip mentioning any specific language.

PR #76, opened 2019-06-27
Gpl to mit

VincentRouvreau:
From GPL v3 license to MIT.
[Licensing documentation](http://pages.saclay.inria.fr/vincent.rouvreau/gudhi/gpl_to_mit/licensing/) from a branch gpl_to_mit in https://github.com/GUDHI/GUDHI.github.io
[Python documentation](http://pages.saclay.inria.fr/vincent.rouvreau/gudhi/gpl_to_mit/python/latest/)
[C++ documentation](http://pages.saclay.inria.fr/vincent.rouvreau/gudhi/gpl_to_mit/python/latest/)


.travis.yml
```diff
@@ -9,32 +9,32 @@ matrix:
   include:
     # A. Mac OSX
     - os: osx
-      osx_image: xcode9.4
+      osx_image: xcode10.2
```
mglisse: What does that have to do with a license change?

mglisse:
Some parts are on the website, so not in this repos, so I'll comment as discussion:
typos "Some of the GUDHI modules depend-s- on third part/ies/y/ libraries"

mglisse:
Maybe mention in the licensing page that we write "Copyright: MIT (GPL v3)" in the package list for exactly the situation described: our code is MIT, but there is a dependency on GPL code, so for practical purposes for a user it is as if this package was GPL.

data/points/generator/aurelien_alvarez_surfaces_in_R8.py
```diff
@@ -1,28 +1,19 @@
-#   This file is part of the Gudhi Library. The Gudhi library
-#   (Geometric Understanding in Higher Dimensions) is a generic C++
-#   library for computational topology.
-#  
-#   Author(s):       Aurélien Alvarez
```
mglisse: Did Aurélien give his permission for the change to this particular file?

mglisse:
homsimpl: removing the binary is sensible, not just for the license.
GudhUI still tries to call this program. It would be nice to mention homsimpl/CHOMP somewhere. At least make sure that the error message is clear, when it isn't installed, about what one can install to get this functionality.

src/cython/cython/alpha_complex.pyx
```diff
@@ -5,19 +5,19 @@ from libcpp.string cimport string
...
-   Modification(s):
-     - YYYY/MM Author: Description of the modification
+    Modification(s):
+      - YYYY/MM Author: Description of the modification
 """
 
 __author__ = "Vincent Rouvreau"
 __copyright__ = "Copyright (C) 2016 Inria"
-__license__ = "GPL v3"
+__license__ = "MIT"
```
mglisse: I think it might be better to keep GPL there. The license of the file is written above, it is MIT. But the complete package including the dependency on CGAL is GPL. And if someone queries gudhi.AlphaComplex.__license__ (or however that shows up in python), it would be better if they got the information: GPL.

.travis.yml
VincentRouvreau: I roll back the modifications. It appears only in this branch, but it is general now.

VincentRouvreau:
@mglisse : thanks for the review. I fixed all the points, tell me if you are ok with it.

VincentRouvreau:
Fixes #35

src/GudhUI/model/Model.h
```diff
@@ -307,9 +307,15 @@ class Model {
  private:
   void run_chomp() {
     save_complex_in_file_for_chomp();
+    std::cout << "POUET" << std::endl;
```
mglisse: COIN!

VincentRouvreau: oops, sorry : https://github.com/VincentRouvreau/gudhi-devel/commit/0a4ece477416422f9d007bbd7eb4c1322d723fcb

src/GudhUI/model/Model.h
```diff
@@ -307,9 +307,15 @@ class Model {
...
     std::cout << "Call CHOMP library\n";
-    int returnValue = system("utils/homsimpl chomp.sim");
-    std::cout << "CHOMP returns" << returnValue << std::endl;
+    int returnValue = system("which homsimpl");
+    if (returnValue != 0) {
+        std::cout << "CHOMP must be installed and available in PATH for this functionnality." << std::endl;
+        return;
+    }
+    returnValue = system("homsimpl chomp.sim");
+    std::cout << "CHOMP returns " << returnValue << std::endl;
```
mglisse: I think it might be enough to test the returnValue of `system("homsimpl chomp.sim")`, and if it is non-zero, explain that homsimpl (from CHOMP) failed, which could happen in particular if it isn't installed or in the PATH. I am a bit wary of using `which`, but maybe it has advantages?

VincentRouvreau: https://github.com/VincentRouvreau/gudhi-devel/commit/4d8dc4477942d196a304e71db4918ac71862e597

PR #83, opened 2019-08-08
Split gudhi python in modules

VincentRouvreau:


mglisse:
Should we rename the directory cython -> python, while we are moving files around?

mglisse:
It seems that the size of the plugins more than doubled in the process. I guess that was to be expected, and it could have been much worse.

mglisse:
So for now we keep the types documented as `gudhi.AlphaComplex` without mentioning any subpackages, keeping those as implementation details, and we will see later if we want to expose them. Makes sense.

src/cython/gudhi/alpha_complex.pyx
```diff
@@ -106,6 +110,8 @@ cdef class AlphaComplex:
         :returns: A simplex tree created from the Delaunay Triangulation.
         :rtype: SimplexTree
         """
-        simplex_tree = SimplexTree()
-        self.thisptr.create_simplex_tree(simplex_tree.thisptr, max_alpha_square)
-        return simplex_tree
+        stree = SimplexTree()
+        cdef intptr_t stree_int_ptr=stree.thisptr
+        cdef Simplex_tree_interface_full_featured* stree_ptr = <Simplex_tree_interface_full_featured*>stree_int_ptr
```
mglisse: Does cython complain if you try to save a line or two? Cast `stree.thisptr` without creating a variable `stree_int_ptr` for instance? It is distracting to see many lines of code that essentially do nothing (unless cython does require them).

src/cython/gudhi/rips_complex.pyx
```diff
@@ -93,6 +96,7 @@ cdef class RipsComplex:
         :returns: A simplex tree created from the Delaunay Triangulation.
         :rtype: SimplexTree
         """
-        simplex_tree = SimplexTree()
-        self.thisref.create_simplex_tree(simplex_tree.thisptr, max_dimension)
-        return simplex_tree
+        stree = SimplexTree()
+        cdef intptr_t stree_ptr=stree.thisptr
+        self.thisref.create_simplex_tree(<Simplex_tree_interface_full_featured*>stree_ptr, max_dimension)
```
mglisse: Ah, you did use one line fewer here :wink:

src/cython/gudhi/simplex_tree.pyx
```diff
@@ -79,18 +44,20 @@ cdef class SimplexTree:
 
     # The real cython constructor
     def __cinit__(self):
-        self.thisptr = new Simplex_tree_interface_full_featured()
+        cdef Simplex_tree_interface_full_featured* ptr = new Simplex_tree_interface_full_featured()
+        self.thisptr = <intptr_t>ptr
```
mglisse: Again, does cython let you write that as a single line, without naming the variable `ptr`?

src/cython/gudhi/simplex_tree.pyx
```diff
@@ -68,7 +27,13 @@ cdef class SimplexTree:
...
     of the simplex tree.
     """
-    cdef Simplex_tree_interface_full_featured * thisptr
+    # unfortunately 'cdef public Simplex_tree_interface_full_featured* thisptr' is not possible
+    # Use intptr_t instead to cast the pointer
+    cdef public intptr_t thisptr
+
+    # Get the pointer casted as it should be
+    cdef Simplex_tree_interface_full_featured* get_ptr(self):
+        return <Simplex_tree_interface_full_featured*>(self.thisptr)
```
mglisse: I wonder if we are allowed to define a property here. If so, it would hide a bit more the fact that thisptr is not actually a `Simplex_tree_interface_full_featured*`. But `get_ptr` is fine.

src/cython/gudhi/alpha_complex.pyx
VincentRouvreau: I tried this but cython complains.

```
        stree = SimplexTree()
        cdef Simplex_tree_interface_full_featured* stree_ptr = <Simplex_tree_interface_full_featured*>(stree.thisptr)
                                                              ^
------------------------------------------------------------

alpha_complex.pyx:115:63: Python objects cannot be cast to pointers of primitive types
```

src/cython/gudhi/rips_complex.pyx
VincentRouvreau: You are right. Done here: 3c96720c786462827820f4d35e720c8f2084867b

src/cython/gudhi/simplex_tree.pyx
VincentRouvreau: I agree. Done on 4fd93dbc06b3020ddd856aed2d8ff37e1d27e517
I do not see any other occurrence.

VincentRouvreau:
> Should we rename the directory cython -> python, while we are moving files around? 
> And rename the cmake target? At least one plugin is pure python.

Ok. Both are done on 68753b3c28321e28eedd5829c94234da84e25c8d

> Or will we use a different directory/target for pybind11 plugins?

Let's try with your first proposition.

Issue #74, opened 2019-06-20
[Alpha_complex] Building a Delaunay triangulation (without filtration)


mglisse:
I have seen several times people use Alpha_complex to build a Delaunay triangulation, and then ignore the filtration values (overwrite them with other values). IIRC, computing the filtration values is by far the most costly part of Alpha_complex, so this seems wasteful. It could be nice to have an option to say that we don't care about the filtration values.
An alternative way to resolve this would be to have a (simple, copy-pastable) example showing how to build a Simplex_tree representing the Delaunay triangulation of a point cloud, possibly using CGAL's python interface (or anything simple).

VincentRouvreau:
What would be the best:
- Delaunay triangulation as a new module (Alpha complex could inherit from it)
- A new parameter in constructor (the desired filtration value) to trigger or not the computation. The Delaunay triangulation would have to be documented in Alpha complex this way.

VincentRouvreau:
Closed by #165

PR #86, opened 2019-08-22
Delaunay triangulation for alpha complex

VincentRouvreau:
A prototype to fix issue #74 
It is easy for dD Alpha complex as the filtration is computed right after the delaunay triangulation.
It is not that trivial for the 3d case (not done yet).


src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -217,13 +230,16 @@ class Alpha_complex {
...
+   * It also computes the filtration values accordingly to the \ref createcomplexalgorithm if default_filtration_value
+   * is not set.
    *
    * \tparam SimplicialComplexForAlpha must meet `SimplicialComplexForAlpha` concept.
    * 
    * @param[in] complex SimplicialComplexForAlpha to be created.
    * @param[in] max_alpha_square maximum for alpha square value. Default value is +\f$\infty\f$, and there is very
    * little point using anything else since it does not save time.
+   * @param[in] default_filtration_value Default Delaunay triangulation filtration value. Set this value if filtration
+   * values are not needed to be computed. Default value is NaN (which means compute the filtration values).
```
mglisse: Why not a boolean if you only care for NaN / not-NaN?

src/Bottleneck_distance/include/gudhi/Bottleneck.h
```diff
@@ -5,22 +5,30 @@
...
 #include <algorithm>  // for max
 #include <limits>  // for numeric_limits
 
 #include <cmath>
 #include <cfloat>  // FLT_EVAL_METHOD
 
+// Make compilation fail - required for external projects - https://github.com/GUDHI/gudhi-devel/issues/10
+#if CGAL_VERSION_NR < 1041101000
+# error Alpha_complex_3d is only available for CGAL >= 4.11
+#endif
```
mglisse: These changes have nothing to do with this PR, what are they doing here?

PR #93, opened 2019-09-20
C++ 14 is the new standard for the project

VincentRouvreau:


mglisse:
Ok for me. It may not be the cleanest way to specify the standard, but this just bumps the number.
I wonder if there are things that can be cleaned up as a consequence.
Aren't there also places in the doc that need updating? "The library uses c++11" in 2 files.

PR #94, opened 2019-09-23
Wdist theo

tlacombe:
Adding wasserstein computation based on the Python Optimal Transport package


mglisse:
Wrong branch, see #95 instead.

PR #95, opened 2019-09-23
wasserstein distance added on fork

tlacombe:


src/python/test/test_wasserstein_distance.py
```diff
@@ -0,0 +1,22 @@
+import gudhi
+
+""" This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+    Author(s):       Theo Lacombe
+
+    Copyright (C) 2016 Inria
```
mglisse: 2019

src/python/test/test_wasserstein_distance.py
```diff
@@ -0,0 +1,22 @@
...
+__author__ = "Theo Lacombe"
+__copyright__ = "Copyright (C) 2016 Inria"
+__license__ = "MIT"
+
+
+def test_basic_bottleneck():
+    diag1 = np.array([[2.7, 3.7], [9.6, 14.0], [34.2, 34.974]])
+    diag2 = np.array([[2.8, 4.45], [9.5, 14.1]])
+
+    assert gudhi.wasserstein_distance(diag1, diag2) == 1.4453593023967701
```
mglisse: Might be safer to test that `abs(gudhi.wasserstein_distance(diag1, diag2) - 1.4453593023967701)` is smaller than some small constant.
The test needs to be mentioned in src/python/CMakeLists.txt.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
+import numpy as np
+import scipy.spatial.distance as sc
+try:
+    import ot
+except ImportError:
+    print("POT (Python Optimal Transport) package is not installed. Try to run $ pip install POT")
```
mglisse: @VincentRouvreau should we also mention "conda install -c conda-forge pot"?

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Cf = np.vstack((Cf, Cdy[None,:]))
+
+    return Cf
+
+
+def wasserstein_distance(X, Y, p=2., q=2.):
+    '''
+    param X, Y: (n x 2) and (m x 2) numpy array (points of persistence diagrams)
+    param q: Ground metric (i.e. norm l_q); Default value is 2 (euclidean norm).
```
mglisse: I think I would add that it is "in the plane" or "in ℝ²". It is already clear, but it doesn't hurt to be doubly clear.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
...
+    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+    Author(s):       Theo Lacombe
+
+    Copyright (C) 2016 Inria
+
+    Modification(s):
+      - YYYY/MM Author: Description of the modification
+"""
+
+def proj_on_diag(X):
```
mglisse: Should we hide those "private" functions somehow? Give them a name starting with `_`? Add some keyword for sphinx?

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
...
+    param Y: (m x 2) np.array encoding the second diagram.
+    param q: Ground metric (i.e. norm l_q).
+    param p: exponent for the Wasserstein metric.
+    return: (n+1) x (m+1) np.array encoding the cost matrix C. 
+                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) and its orthogonal proj onto the diagonal.
+                note also that C[n+1, m+1] = 0  (it costs nothing to move from the diagonal to the diagonal).
+    '''
+    Xdiag = proj_on_diag(X)
+    Ydiag = proj_on_diag(Y)
+    if np.isinf(p):
```
mglisse: Did you mean to test `q` here? Maybe this is a sign that we need a few more tests?

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
...
+    '''
+    param X, Y: (n x 2) and (m x 2) numpy array (points of persistence diagrams)
+    param q: Ground metric (i.e. norm l_q); Default value is 2 (euclidean norm).
+    param p: exponent for Wasserstein; Default value is 2.
+    return: float, the p-Wasserstein distance (1 <= p < infty) with respect to the q-norm as ground metric.
+    '''
+    M = build_dist_matrix(X, Y, p=p, q=q)
+    n = len(X)
+    m = len(Y)
+    a = 1.0 / (n + m) * np.ones(n)  # weight vector of the input diagram. Uniform here.
```
mglisse: `np.full`?

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
...
+    hat_a = np.append(a, m/(n+m))  # so that we have a probability measure, required by POT
```
mglisse: `m/(n+m)` only work in python 3? I would be fine with declaring that gudhi stops supporting python 2, but it shouldn't come just because of one division here.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,75 @@
...
+    b = 1.0 / (n + m) * np.ones(m)  # weight vector of the input diagram. Uniform here.
+    hat_b = np.append(b, n/(m+n))  # so that we have a probability measure, required by POT
+
+    # Comptuation of the otcost using the ot.emd2 library.
+    # Note: it is the squared Wasserstein distance.
+    ot_cost = (n+m) * ot.emd2(hat_a, hat_b, M)
+
+    return np.power(ot_cost, 1./p)
```
mglisse: Any particular reason why you use `**` in some places and `np.power` in others? Is it `**` for matrices and `np.power` for scalars?

src/python/gudhi/wasserstein.py
tlacombe: Hum you're right. It would be better to add `from __future__ import division` on top of the file, or cast n (or m) as floats. Sorry for this, I only work with Python3 (which is a reasonable idea considering https://pythonclock.org/ ).

src/python/gudhi/wasserstein.py
tlacombe: Done !

src/python/gudhi/wasserstein.py
tlacombe: Indeed. Solved

src/python/gudhi/wasserstein.py
tlacombe: I added a _ in front of them of the moment, hoping for sphinx to do the job.

src/python/gudhi/wasserstein.py
tlacombe: Done !

src/python/gudhi/wasserstein.py
tlacombe: I added this suggestion (and left the pip one, just in case.).

src/python/test/test_wasserstein_distance.py
tlacombe: It seems the tests work this way and are consistent with those done for the `bottleneck_distance` ones. I added them in the aforementioned files.

mglisse:
@VincentRouvreau I suspect that we need to add "pot" to the machine running the tests. Can you handle it?

VincentRouvreau:
> @VincentRouvreau I suspect that we need to add "pot" to the machine running the tests. Can you handle it?

Ok I am looking for it.

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -0,0 +1,14 @@
...
+   |      ../../doc/Bottleneck_distance/perturb_pd.png               | persistence diagrams. It's the minimum value c that can be achieve by|                                                                  |
+   |      :figclass: align-center                                    | a perfect matching between the points of the two diagrams (+ all the | :Introduced in: GUDHI 2.0.0                                      |
+   |                                                                 | diagonal points), where the value of a matching is defined as the    |                                                                  |
+   |      Wasserstein distance is the p-th root of the sum of the    | p-th root of the sum of all edges lengths to the power p. Edges      | :Copyright: MIT (`GPL v3 </licensing/>`_)                        |
+   |      edges lengths to the power p.                              | lengths are measured in norm q, for $1 \leq q \leq \infty$.          |                                                                  |
```
mglisse: From other files, it seems that math stuff does not use $ but :math: and backticks.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,96 @@
...
+    param p: exponent for Wasserstein; Default value is 2.
+    return: float, the total persistence of the diagram (that is, its distance to the empty diagram).    
+    '''
+    Xdiag = _proj_on_diag(X)
+    return (np.sum(np.linalg.norm(X - Xdiag, ord=q, axis=1)**p))**(1/p)
+
+
+def wasserstein_distance(X, Y, p=2., q=2.):
+    '''
+    param X, Y: (n x 2) and (m x 2) numpy array (points of persistence diagrams)
```
mglisse: Those don't appear nicely in the doc. Maybe with an extra `:` before "param"? After `make sphinx`, the html doc should be available in some subdirectory, you can check how it looks.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,98 @@
...
+
+def _perstot(X, p, q):
+    '''
+    :param X: (n x 2) numpy array (points of a given diagram)
+    :param q: Ground metric on the (upper-half) plane (i.e. norm l_q in R^2); Default value is 2 (euclidean norm).
+    :param p: exponent for Wasserstein; Default value is 2.
+    :returns: float, the total persistence of the diagram (that is, its distance to the empty diagram).    
+    '''
+    Xdiag = _proj_on_diag(X)
+    return (np.sum(np.linalg.norm(X - Xdiag, ord=q, axis=1)**p))**(1/p)
```
mglisse: Maybe `1./p` to be sure?

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -0,0 +1,14 @@
...
+   | .. figure::                                                     | The p-Wasserstein distance measures the similarity between two       | :Author: Theo Lacombe                                            |
+   |      ../../doc/Bottleneck_distance/perturb_pd.png               | persistence diagrams. It's the minimum value c that can be achieve by|                                                                  |
+   |      :figclass: align-center                                    | a perfect matching between the points of the two diagrams (+ all the | :Introduced in: GUDHI 2.0.0                                      |
+   |                                                                 | diagonal points), where the value of a matching is defined as the    |                                                                  |
+   |      Wasserstein distance is the p-th root of the sum of the    | p-th root of the sum of all edges lengths to the power p. Edges      | :Copyright: MIT (`GPL v3 </licensing/>`_)                        |
```
mglisse: Where is the GPL license coming from?

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -0,0 +1,14 @@
+.. table::
+   :widths: 30 50 20
+
+   +-----------------------------------------------------------------+----------------------------------------------------------------------+------------------------------------------------------------------+
+   | .. figure::                                                     | The p-Wasserstein distance measures the similarity between two       | :Author: Theo Lacombe                                            |
+   |      ../../doc/Bottleneck_distance/perturb_pd.png               | persistence diagrams. It's the minimum value c that can be achieve by|                                                                  |
+   |      :figclass: align-center                                    | a perfect matching between the points of the two diagrams (+ all the | :Introduced in: GUDHI 2.0.0                                      |
```
mglisse: The version number will need updating (3.1.0?).

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def wasserstein_distance(X, Y, p=2., q=2.):
+    '''
+    :param X: (n x 2) np.array encoding the (points of the) first diagram.
```
mglisse: Spell it `numpy.array` in the doc?

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,98 @@
...
+    :param Y: (m x 2) np.array encoding the second diagram.
+    :param q: Ground metric on the (upper-half) plane (i.e. norm l_q in R^2); Default value is 2 (euclidean norm).
```
mglisse: I believe we are supposed to capitalize `Euclidean` in English.

src/python/gudhi/wasserstein.py
mglisse: Maybe here (and not just in the example section) we could say "finite points", or advertise the fact that we do not want infinity.

mglisse:
@MathieuCarriere Hello, are you happy with the interface and function name here? IIRC you wanted to use Wasserstein distances.

src/python/doc/wasserstein_distance_sum.inc
tlacombe: From copy-pasting the doc of bottleneck distance... I guess I should remove it ? (POT is under MIT).

src/python/gudhi/wasserstein.py
tlacombe: good idea :+1:

src/python/doc/wasserstein_distance_sum.inc
tlacombe: I updated it to 3.1.0 for the moment.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,98 @@
...
+    Copyright (C) 2019 Inria
+
+    Modification(s):
+      - YYYY/MM Author: Description of the modification
+"""
+
+def _proj_on_diag(X):
+    '''
+    :param X: (n x 2) array encoding the points of a persistent diagram.
+    :returns: (n x 2) arary encoding the (respective orthogonal) projections of the points onto the diagonal
```
```suggestion
    :returns: (n x 2) array encoding the (respective orthogonal) projections of the points onto the diagonal
```

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -0,0 +1,14 @@
+.. table::
+   :widths: 30 50 20
+
+   +-----------------------------------------------------------------+----------------------------------------------------------------------+------------------------------------------------------------------+
+   | .. figure::                                                     | The p-Wasserstein distance measures the similarity between two       | :Author: Theo Lacombe                                            |
+   |      ../../doc/Bottleneck_distance/perturb_pd.png               | persistence diagrams. It's the minimum value c that can be achieve by|                                                                  |
```
mglisse:
```suggestion
   |      ../../doc/Bottleneck_distance/perturb_pd.png               | persistence diagrams. It's the minimum value c that can be achieved by|                                                                  |
```

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -0,0 +1,14 @@
...
+   |      :figclass: align-center                                    | a perfect matching between the points of the two diagrams (+ all the | :Introduced in: GUDHI 3.1.0                                      |
+   |                                                                 | diagonal points), where the value of a matching is defined as the    |                                                                  |
+   |      Wasserstein distance is the p-th root of the sum of the    | p-th root of the sum of all edges lengths to the power p. Edges      | :Copyright: MIT                                                  |
```
mglisse: I think it is "edge lengths", no 's' on "edge".

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,98 @@
...
+    return (np.sum(np.linalg.norm(X - Xdiag, ord=q, axis=1)**p))**(1./p)
+
+
+def wasserstein_distance(X, Y, p=2., q=2.):
+    '''
+    :param X: (n x 2) numpy.array encoding the (finite points of the) first diagram. Must not contain essential points (i.e. with infinite coordinate).
+    :param Y: (m x 2) numpy.array encoding the second diagram.
+    :param q: Ground metric on the (upper-half) plane (i.e. norm l_q in R^2); Default value is 2 (euclidean norm).
+    :param p: exponent for Wasserstein; Default value is 2.
+    :returns: the p-Wasserstein distance (1 <= p < infty) with respect to the q-norm as ground metric.
```
```suggestion
    :returns: the p-Wasserstein distance (1 <= p < infinity) with respect to the q-norm as ground metric.
```

mglisse:
With the typo fixes, I think this is good to go.

src/python/doc/wasserstein_distance_sum.inc
tlacombe: For that one, I need to do a separate commit as the doc is very sensitive to the position of the | .

src/python/doc/wasserstein_distance_sum.inc
tlacombe: Indeed it seems so, corrected !

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,98 @@
...
+
+    M = _build_dist_matrix(X, Y, p=p, q=q)
+    a = np.full(n+1, 1. / (n + m) )  # weight vector of the input diagram. Uniform here.
+    a[-1] = a[-1] * m                # normalized so that we have a probability measure, required by POT
+    b = np.full(m+1, 1. / (n + m) )  # weight vector of the input diagram. Uniform here.
+    b[-1] = b[-1] * n                # so that we have a probability measure, required by POT
+
+    # Comptuation of the otcost using the ot.emd2 library.
+    # Note: it is the squared Wasserstein distance.
+    ot_cost = (n+m) * ot.emd2(a, b, M)
```
mglisse: The documentation of ot.emd2 says that M must be "a c-order array with type float64". I think we can rely on cdist, etc to produce a c-order array, but I am more worried about float64. If a user calls wasserstein_distance with numpy arrays of float32, what happens? Unless it just works, we should say something in the doc (or tweak it so it just works).

mglisse:
I looked a bit at the benchmark, playing with wdist.py there.
First, `processes=...` does not seem to have any effect, it never used multiple threads. Is 5000 not large enough for that?
Second, the number of iterations is not sufficient for the example with 5000 points. It stops with a value of 6.589 and a warning, while if I increase numItermax to 1000000 it successfully finds 6.3108, but it takes 3 times as long, a bit longer than hera. The curves seem to indicate that hera may be significantly faster with even more points.
Should we increase the default number of iterations? To how much? Do we need to make it visible to a user? That would suck, but currently what is a user of gudhi supposed to do if they see the warning about increasing numItermax?

mglisse:
We should probably also mention POT in installation.rst.

src/python/gudhi/wasserstein.py
mglisse: I tried, I called wasserstein_distance on float32 numpy arrays, and it just worked, so let's ignore this.

mglisse:
I think for now you should increase the number of iterations (pick whatever big number), then Vincent merges the branch (after the release, unless 3.0 already has its own branch) if he does not have further comments, and we can work on improvements after that.

mglisse:
For later: as is, if I do not have POT installed, I get the message "POT (Python Optimal Transport) package is not installed." as soon as I do `import gudhi`. As our list of dependencies increases (scipy, sklearn, later pytorch and tensorflow), that won't scale very well. Some alternatives I can think of:

- import ot inside wasserstein_distance so it doesn't complain until we try to use it
- don't import everything into gudhi so eagerly, make users import gudhi.wasserstein (that's probably what we will do for sktda, and for things using pytorch or tensorflow). If we go that way, it is better to decide it soon, because it changes what the user should write.
- complain later: still import early, but only print the error message when the function is called

Also, scipy is used without protection, which looks a bit inconsistent.

mglisse:
>     * don't import everything into gudhi so eagerly, make users import gudhi.wasserstein (that's probably what we will do for sktda, and for things using pytorch or tensorflow).

Would it be enough to remove the line that adds wasserstein to GUDHI_PYTHON_MODULES? Or maybe add it to some GUDHI_PYTHON_EXTRA_MODULES instead that would be listed in `__all__` but not used in the automatic import? Of course doc/example would need updating then.

src/python/gudhi/wasserstein.py
```diff
@@ -0,0 +1,99 @@
...
+        Cxd = np.linalg.norm(X - Xdiag, ord=q, axis=1)**p
+        Cdy = np.linalg.norm(Y - Ydiag, ord=q, axis=1)**p
+    else:
+        C = sc.cdist(X,Y, metric='minkowski', p=q)**p
+        Cxd = np.linalg.norm(X - Xdiag, ord=q, axis=1)**p
+        Cdy = np.linalg.norm(Y - Ydiag, ord=q, axis=1)**p
+    Cf = np.hstack((C, Cxd[:,None]))
+    Cdy = np.append(Cdy, 0)
+
+    Cf = np.vstack((Cf, Cdy[None,:]))
```
mglisse: Since I discovered it today, let me mention the existence of np.block, when you want to build a matrix from 4 blocks like here. I am not writing it so the code gets changed, I just wanted to share the knowledge.

VincentRouvreau: > Since I discovered it today, let me mention the existence of np.block, when you want to build a matrix from 4 blocks like here. I am not writing it so the code gets changed, I just wanted to share the knowledge.

Maybe we should trace it in a new issue ?

mglisse: It isn't worth the trouble. There are always going to be 10 different ways of doing things in numpy...

Issue #108, opened 2019-10-15
C++ documentation is not generated with Doxygen 1.8.16


VincentRouvreau:
When building the documentation with Doxygen 1.8.16, no specific warning, nor errors are generated, but the documentation is not generated.
By this, I mean there is an almost empty index.html with just the title.

VincentRouvreau:
@skachano : this bug is now fixed in master

PR #115, opened 2019-10-17
fix documentation not being generated with Doxygen 1.8.16

jmarino:
This fixes issue #108 
Caused by a known [bug](https://github.com/doxygen/doxygen/issues/7190) in Doxygen-1.8.16.


VincentRouvreau:
I tested the fix using conda and installing doxygen 1.8.16 and the proposed fix allow to generate the documentation. The result is not very nice (identation issue in the TOC, image sizes are not really nice, ...) and there are a lot of new warnings, but it works and will allow the dev to continue to generate the documentation.
I also tested the fix with doxygen 1.8.13 and the fix has no impact.

Issue #113, opened 2019-10-17
[python] replace "is" with "=="


mglisse:
`if max_plots is not 1000` looks wrong, I think we want `!=`. I am getting "Deprecated parameter. It has been replaced by max_intervals" for every call to the graphical tools.
```
>>> a=1000
>>> a is 1000
False
>>> a==1000
True
```

mglisse:
Python 3.8 shows
> SyntaxWarning: "is" with a literal. Did you mean "=="?

:+1:

VincentRouvreau:
Fixed by #137 and #119

Issue #109, opened 2019-10-16
[python - persistence_graphical tools] Dimension 1 color in the palette is not visible with video projector


VincentRouvreau:
For persistence diagram and bar code, the dimension 1 color in the palette (a light green) is not distinguishable when projecting the video on a white screen

mglisse:
Aren't we using the default from matplotlib? Has this been reported to them?

VincentRouvreau:
> Aren't we using the default from matplotlib? Has this been reported to them?

No. I defined the palette, but it was not a good idea. Matplotlib has its own palette (even used by bokeh).

VincentRouvreau:
Fixed by #119

Issue #111, opened 2019-10-17
[python - persistence_graphical tools] Graphical tools from a persistence file is broken


VincentRouvreau:
Since gudhi 3.0.0 (from a python module to a package), graphical tool when giving as an input a persistence file is not plotting the correct data. i.e.:
```
import gudhi
gudhi.plot_persistence_diagram(persistence_file='data/persistence_diagram/rips_on_tore3D_1307.pers').show()
gudhi.plot_persistence_diagram(persistence_file='data/persistence_diagram/rips_on_tore3D_1307.pers').show()
```
(have to do it twice, first time is correct)
After investigation, the read function only returns (0., 0.) with an incorrect dimension.
read function hash seems correct.
I also tried `strace -f -o toto.txt python toto.py` but the correct file is read.
I tried in a miniconda env reinstalling python 3.7, cython, ... but the same error appears.

mglisse:
I cannot reproduce this issue (debian testing, python 3.7.5rc1).

VincentRouvreau:
The function to read persistence file is using `sscanf` function. This function is `locale` sensitive.
I tried to reproduce [this matplotlib issue](https://github.com/matplotlib/matplotlib/issues/6706), but this is not exactly my case:
```python
import gudhi                                                                                                                                                                                          
import locale                                                                                                                                                                                         
print(locale.getlocale())                                                                                                                                                                             
# >>> ('en_US', 'UTF-8')

gudhi.read_persistence_intervals_in_dimension('../../../data/persistence_diagram/rips_on_tore3D_1307.pers')   
# The values are correct                                                                                        
# >>> array([[0.138335 ,       inf],
# >>>        [0.104347 ,       inf],
# >>>        [0.0983494,       inf],
# >>>        ...,
# >>>        [0.104771 , 0.104801 ],
# >>>        [0.103878 , 0.103908 ],
# >>>        [0.11169  , 0.111695 ]])

print(locale.getlocale())                                                                                                                                                                             
# >>> ('en_US', 'UTF-8')

import matplotlib.pyplot as plot                                                                                                                                                                      
print(locale.getlocale())                                                                                                                                                                            
# >>> ('en_US', 'UTF-8')

gudhi.read_persistence_intervals_in_dimension('../../../data/persistence_diagram/rips_on_tore3D_1307.pers')                                                                                          
# The values are wrong because of sscanf returned value that is not correct
# >>> array([[2., 0.],
# >>>        [1., 0.],
[...]

PR #119, opened 2019-10-24
Persistence graphical tool improvements

VincentRouvreau:
- Fix issue #113 (replace `is not` with `!=`)
- Fix issue #109  (replace palette with a more visible one)


src/common/include/gudhi/reader_utils.h
```diff
@@ -305,8 +305,12 @@ void read_persistence_intervals_and_dimension(std::string const& filename, Outpu
     std::string line;
     getline(in, line);
     if (line.length() != 0 && line[0] != '#') {
-      double numbers[4];
+      double numbers[4] = {0.};
```
mglisse: Is it because of one of gcc's idiotic warnings?

VincentRouvreau: I add a weird bug and could not find an explanation. I tried to set the double array before filling it, but it did not help.
I did not remove it, but I can do it.

mglisse: If it does not help for anything, then yes, I'd rather not add the initialization.
Is the weird bug still there?

VincentRouvreau: Yes it is still there, even with the master branch. It does not seem to come from the development done in this branch

VincentRouvreau: I removed the initialization on 57c1f0f4b36b934c8fc54c8691f77b9b6d0eff9d

mglisse: Ça ne colle pas bien avec le DEBUG_TRACES en dessous. Il faudrait soit que le code dans DEBUG_TRACES ne lise que les n premiers éléments, soit qu'on vire ce code, soit qu'on initialise (avec 4 qNaN par exemple ?).

mglisse: > Yes it is still there, even with the master branch. It does not seem to come from the development done in this branch

ok, then file an issue and let's ignore it for this PR.

VincentRouvreau: Ok.
I reworked traces on 1e3b4f6c56409dde933341701575dc105ede12db
The issue is tracked by #111

PR #123, opened 2019-10-31
Add a get_point method for Alpha_complex_3d.

VincentRouvreau:
- Add a get_point method for Alpha_complex_3d.
- Change Point_3 type to be either Weighted_point_3 or Bare_point_3 (new type).
- Add some get_point method tests.

### DIFF
--- a/src/Alpha_complex/benchmark/Alpha_complex_3d_benchmark.cpp
+++ b/src/Alpha_complex/benchmark/Alpha_complex_3d_benchmark.cpp
@@ -71,10 +71,10 @@ void benchmark_points_on_torus_3D(const std::string& msg) {
   for (int nb_points = 1000; nb_points <= 125000; nb_points *= 5) {
     std::cout << "  Alpha complex 3d on torus with " << nb_points << " points." << std::endl;
     std::vector<K::Point_d> points_on_torus = Gudhi::generate_points_on_torus_3D<K>(nb_points, 1.0, 0.5);
-    std::vector<typename Alpha_complex_3d::Point_3> points;
+    std::vector<typename Alpha_complex_3d::Bare_point_3> points;
 
     for (auto p : points_on_torus) {
-      points.push_back(typename Alpha_complex_3d::Point_3(p[0], p[1], p[2]));
+      points.push_back(typename Alpha_complex_3d::Bare_point_3(p[0], p[1], p[2]));
     }
 
     Gudhi::Clock ac_create_clock("    benchmark_points_on_torus_3D - Alpha complex 3d creation");
@@ -115,7 +115,7 @@ void benchmark_weighted_points_on_torus_3D(const std::string& msg) {
     std::cout << "  Alpha complex 3d on torus with " << nb_points << " points." << std::endl;
     std::vector<K::Point_d> points_on_torus = Gudhi::generate_points_on_torus_3D<K>(nb_points, 1.0, 0.5);
 
-    using Point = typename Weighted_alpha_complex_3d::Point_3;
+    using Point = typename Weighted_alpha_complex_3d::Bare_point_3;
     using Weighted_point = typename Weighted_alpha_complex_3d::Weighted_point_3;
 
     std::vector<Weighted_point> points;
@@ -158,7 +158,7 @@ void benchmark_periodic_points(const std::string& msg) {
 
   for (double nb_points = 10.; nb_points <= 40.; nb_points += 10.) {
     std::cout << "  Periodic alpha complex 3d with " << nb_points * nb_points * nb_points << " points." << std::endl;
-    using Point = typename Periodic_alpha_complex_3d::Point_3;
+    using Point = typename Periodic_alpha_complex_3d::Bare_point_3;
     std::vector<Point> points;
 
     for (double i = 0; i < nb_points; i++) {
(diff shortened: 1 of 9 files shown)


src/Alpha_complex/benchmark/Alpha_complex_3d_benchmark.cpp
```diff
@@ -71,10 +71,10 @@ void benchmark_points_on_torus_3D(const std::string& msg) {
   for (int nb_points = 1000; nb_points <= 125000; nb_points *= 5) {
     std::cout << "  Alpha complex 3d on torus with " << nb_points << " points." << std::endl;
     std::vector<K::Point_d> points_on_torus = Gudhi::generate_points_on_torus_3D<K>(nb_points, 1.0, 0.5);
-    std::vector<typename Alpha_complex_3d::Point_3> points;
+    std::vector<typename Alpha_complex_3d::Bare_point_3> points;
```
mglisse: Why change it here? I would stick to the nicer Point_3 by default, and only use Bare/Weighted in the weighted case.

src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
```diff
@@ -567,9 +573,32 @@ Weighted_alpha_complex_3d::Weighted_point_3 wp0(Weighted_alpha_complex_3d::Point
     return true;
   }
 
+  /** \brief get_point returns the point corresponding to the vertex given as parameter.
+   *
+   * @param[in] vertex Vertex handle of the point to retrieve.
+   * @return The point found.
+   * @exception std::out_of_range In case vertex is not found (cf. std::vector::at).
+   */
+  const Point_3& get_point(std::size_t vertex) {
```
mglisse: Not making this method `const` (with a mutable `cgal_vertex_iterator_vector`) advertises the fact that this is not thread-safe, ok.

src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
```diff
@@ -567,9 +573,32 @@ Weighted_alpha_complex_3d::Weighted_point_3 wp0(Weighted_alpha_complex_3d::Point
...
+    if (map_cgal_simplex_tree.size() != cgal_vertex_iterator_vector.size()) {
+      cgal_vertex_iterator_vector.resize(map_cgal_simplex_tree.size());
+      for (auto map_iterator = map_cgal_simplex_tree.begin(); map_iterator != map_cgal_simplex_tree.end(); map_iterator++) {
```
mglisse: Still not fond of range-for loops?

src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
```diff
@@ -264,13 +269,13 @@ Weighted_alpha_complex_3d::Weighted_point_3 wp0(Weighted_alpha_complex_3d::Point
  public:
   /** \brief Alpha_complex constructor from a list of points.
    *
-   * @param[in] points Range of points to triangulate. Points must be in `Alpha_complex_3d::Point_3` or
+   * @param[in] points Range of points to triangulate. Points must be in `Alpha_complex_3d::Bare_point_3` or
    * `Alpha_complex_3d::Weighted_point_3`.
```
mglisse: So aren't they always `Alpha_complex_3d::Point_3`?

src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
```diff
@@ -225,30 +225,35 @@ class Alpha_complex_3d {
...
  *
  * Note: This type is defined to void if Alpha complex is not weighted.
  *
  * */
   using Weighted_point_3 = typename Triangulation_3<Kernel, Tds, Weighted, Periodic>::Weighted_point_3;
 
+  /** \brief `Alpha_complex_3d::Point_3` type is either a `Alpha_complex_3d::Bare_point_3` (Weighted = false) or a
+   * `Alpha_complex_3d::Weighted_point_3` (Weighted = true).
+   */
+  using Point_3 = typename std::conditional<Weighted, Weighted_point_3, Bare_point_3>::type;
```
mglisse: Is it just `typename Alpha_shape_3::Point`?

mglisse:
> This is the reason why the vertices may be not contiguous.

I think this was copied from the dD case but doesn't apply here, because we do not use the order of the input points as vertex handle in the complex. I expect it is always contiguous with the current 3d code.

src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
VincentRouvreau: From my point of view, it is better to explain that Point_3 is a Bare_point_3 or a Weighted_point_3, than using a nested Alpha_shape type, no ? I might be wrong...

mglisse: We could do both: for instance keep your doc, and change the code.

mglisse:
> I expect it is always contiguous with the current 3d code.

At some point (not this PR) we could even consider promising this, so people can use a contiguous simplex tree for it.

src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
VincentRouvreau: Done on 6d2e5d1cbdecd86ad4d56fc8902542caca2cf2bd

PR #127, opened 2019-11-05
Alpha complex with cgal epeck d

VincentRouvreau:


src/Alpha_complex/example/Fast_alpha_complex_from_off.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  // WARNING : CGAL::Epick_d is fast but not safe (unlike CGAL::Epeck_d)
+  // (i.e. when the points are on a grid)
+  using Fast_kernel = CGAL::Epick_d<CGAL::Dynamic_dimension_tag>;
+  // ----------------------------------------------------------------------------
+  // Init of an alpha complex from an OFF file
+  // ----------------------------------------------------------------------------
+  Gudhi::alpha_complex::Alpha_complex<Fast_kernel> alpha_complex_from_file(off_file_name);
+
+  std::streambuf* streambufffer;
```
mglisse: Are the 3 'f' on purpose?

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -184,6 +190,11 @@ class Alpha_complex {
  private:
   template<typename InputPointRange >
   void init_from_range(const InputPointRange& points) {
+    #if CGAL_VERSION_NR < 1050000000
+    if (Is_Epick_D<Kernel>::value)
+      std::cerr << "It is strongly advised to use a CGAL version from 5.0 for performance reasons." << std::endl;
```
mglisse: Er, the warning should be for people using CGAL-4.x with **Epeck_d**. For Epick_d, it doesn't matter if they have 4.x or 5.0.

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -248,7 +261,8 @@ class Alpha_complex {
   template <typename SimplicialComplexForAlpha,
             typename Filtration_value = typename SimplicialComplexForAlpha::Filtration_value>
   bool create_complex(SimplicialComplexForAlpha& complex,
-                      Filtration_value max_alpha_square = std::numeric_limits<Filtration_value>::infinity()) {
+                      Filtration_value max_alpha_square = std::numeric_limits<Filtration_value>::infinity(),
+                      bool exact = false) {
```
mglisse: I was wondering if we should try to write this extra argument in a way that will naturally extend if we ever add a safe2 mode. But I guess nobody will use exact anyway, so we can slightly break the interface when the time comes if needed.

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -324,9 +338,18 @@ class Alpha_complex {
...
+              // With CGAL >= 4.11 and < 5.X, CGAL::exact do not work as it is always exact values
+              // This is why it is slow and 5.X is advised
+              alpha_complex_filtration = CGAL::to_double(squared_radius(pointVector.begin(), pointVector.end()));
+#else  // CGAL_VERSION_NR < 1050000000
+              if (exact) {
+                alpha_complex_filtration = CGAL::to_double(CGAL::exact(squared_radius(pointVector.begin(), pointVector.end())));
+              } else {
+                alpha_complex_filtration = CGAL::to_double(squared_radius(pointVector.begin(), pointVector.end()));
+              }
+#endif  //CGAL_VERSION_NR < 1050000000
```
mglisse: Let me suggest:
```c++
  CGAL::NT_converter<typename Geom_traits::FT, Filtration_value> cv;
  auto sqrad = squared_radius(pointVector.begin(), pointVector.end());
#if CGAL_VERSION_NR >= 1050000000
  if(exact) CGAL::exact(sqrad);
#endif
  alpha_complex_filtration = cv(sqrad);
```

mglisse:
Maybe I missed it, but I think we need to document that if you use Epick_d, you get something fast, but potentially badly wrong in some cases. If you use Epeck_d and pass exact=true to create_complex, you get exact (well, converted to Filtration_value) values. And if you use Epeck_d and pass exact=false (the default), you get something safe, i.e. the values can be wrong by at most ... And note that when computing persistence, this means you can still get unwanted intervals of small relative length but large absolute length, like [10^12,1.000001*10^12] which has length 10^6.

Did you get a chance to run one quick benchmark?

src/Alpha_complex/example/Fast_alpha_complex_from_off.cpp
VincentRouvreau: cf. 98469d9e80a881ab4dae6358b7a7e64ce90784b2

src/Alpha_complex/include/gudhi/Alpha_complex.h
VincentRouvreau: cf. ea296b4a93c37b0e19d4605c1be3e950ed8cd3c2

src/Alpha_complex/include/gudhi/Alpha_complex.h
VincentRouvreau: cf. 5a2c3559da581136edb7facf5578bd1443080ef3

mglisse:
For the record, the benchmark shows Epeck_d and Epick_d performance are quite similar, meaning that there is no reason to keep Epick_d as the default :smile:

mglisse:
<pre>Using the default CGAL::Epeck_d makes the construction safe. If you pass exact=true to create_complex, the filtration values are the exact ones converted to the filtration value type of the simplicial complex. This can be very slow. If you pass exact=false (the default), the filtration values are only guaranteed to have a small multiplicative error compared to the exact value, see &lt;a href="https://doc.cgal.org/latest/Number_types/classCGAL_1_1Lazy__exact__nt.html"&gt;`CGAL::Lazy_exact_nt&lt;NT&gt;::set_relative_precision_of_to_double`&lt;/a&gt; for details. A drawback, when computing persistence, is that an empty exact interval [10^12,10^12] may become a non-empty approximate interval [10^12,10^12+10^6]. Using CGAL::Epick_d makes the computations slightly faster, and the combinatorics are still exact, but the computation of filtration values can exceptionally be arbitrarily bad. In all cases, we still guarantee that the output is a valid filtration (faces have a filtration value no larger than their cofaces).</pre>
could be a starting point...

mglisse:
>  Maybe there is a better place.

The place where you put it looks fine to me. Although we could say something in the reference page (file Alpha_complex.h through doxygen) as well.

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -63,17 +67,20 @@ namespace alpha_complex {
  * 
  * Please refer to \ref alpha_complex for examples.
  *
- * The complex is a template class requiring an Epick_d <a target="_blank"
+ * The complex is a template class requiring an <a target="_blank"
+ * href="https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epeck__d.html">CGAL::Epeck_d</a>,
+ * or an <a target="_blank"
+ * href="https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epick__d.html">CGAL::Epick_d</a> <a target="_blank"
```
mglisse: The same sentence is present in the user manual and should be updated there as well.

VincentRouvreau: You are right. I added it here : c8be137d15f40f40c4150f0c0fb9776cd680a36c
I also mentionned the CGAL 5.X performance here.

PR #141, opened 2019-11-14
Make the license a comment instead of a string

mglisse:
I think this interacts better with sphinx automodule.
Looking at some other projects, a string at the beginning of the file is meant to be used as docstring for the module, and hidden stuff is better as a comment.

### DIFF
--- a/src/python/gudhi/__init__.py.in
+++ b/src/python/gudhi/__init__.py.in
@@ -1,14 +1,13 @@
 from importlib import import_module
 
-"""This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
-    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
-    Author(s):       Vincent Rouvreau
-
-   Copyright (C) 2016  Inria
-
-   Modification(s):
-     - YYYY/MM Author: Description of the modification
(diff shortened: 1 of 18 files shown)


VincentRouvreau:
Sorry, I missed this one, but the license comment is ok, thanks to this [PEP8  recommendation](https://www.python.org/dev/peps/pep-0008/#id23).
As you always put your import at the first line of your file, it should work.
I can add this remark to sklearn-tda review too.

mglisse:
> Sorry, I missed this one, but the license comment is ok, thanks to this [PEP8 recommendation](https://www.python.org/dev/peps/pep-0008/#id23).

Does this mean you agree with this PR and we can merge it?

> As you always put your import at the first line of your file, it should work.

Actually, PEP8 seems to say you should put the imports *after* the license comment (and after the docstring if there is one).

VincentRouvreau:
> Does this mean you agree with this PR and we can merge it?

Yes, I just approved.

> Actually, PEP8 seems to say you should put the imports _after_ the license comment (and after the docstring if there is one). But that's a separate issue, I didn't want to mix independent choices.

Yes you are right, sorry for the misunderstanding of the PEP8.
I have checked in some other projects, and your proposal is correct.

PR #151, opened 2019-11-21
Use find_package(Python) instead of PythonInterp

### LINKED ISSUE #139: FindPythonInterp deprecated
```cmake
find_package(PythonInterp)
```
should be replaced with
```cmake
```

mglisse:
Apparently find_package(PythonInterp) is deprecated since 3.12. However, this patch requires cmake >= 3.12, which may be a bit too recent (1 year old next week)...
It is also a bit less convenient to specify a non-default version of python (I have 3.8 and 3.7, but only 3.7 has the various modules installed): I first run cmake (it picks python3.8, the most recent), then edit the python executable with ccmake, and let it configure/generate.

Fix #139.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.12)
 
 project(GUDHIdev)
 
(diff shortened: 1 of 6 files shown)


mglisse:
CGAL-5.3 will require CMake >= 3.14, so I think it would be fine if Gudhi-3.5 did as well.

Issue #116, opened 2019-10-20
[cython] Use `except +`


mglisse:
For instance, in rips_complex.py, one can write:
```diff
-void create_simplex_tree(Simplex_tree_interface_full_featured* simplex_tree, int dim_max)
+void create_simplex_tree(Simplex_tree_interface_full_featured* simplex_tree, int dim_max) except +
```
so that when there isn't enough memory, the C++ bad_alloc exception is turned into a python exception instead of crashing the whole python process.

mglisse:
There are probably a number of other places where we could add `except +`, but I guess we don't need an issue for that, if it doesn't contain any specific pointers to places where it is missing.

VincentRouvreau:
> There are probably a number of other places where we could add `except +`, but I guess we don't need an issue for that, if it doesn't contain any specific pointers to places where it is missing.

Do you want to keep it open ?

mglisse:
Nah, we can reopen it when we find a specific place where we should add `except +`.

PR #152, opened 2019-11-21
Add some `except +` for cython

mglisse:
I just added a few, I haven't tested many yet.
Strangely, doing a parallel build with
> python3 setup.py build_ext --inplace -j20

breaks the exception conversion for get_point in AlphaComplex, and rebuilding just this module makes it work. Serial builds also work.

From issue #116.


mglisse:
The issue with parallel builds seems to have disappeared here, even before I rebased on master. Strange, but I'll take it :smile:

VincentRouvreau:
> The issue with parallel builds seems to have disappeared here, even before I rebased on master. Strange, but I'll take it smile

Just hope there will be no collision with language_level not set in Cython development.

PR #159, opened 2019-11-26
 Fix various links for sphinx

mglisse:
+ some small doc changes along the way
+ yet another sphinx extension (I added it to the dockerfile in the branch, but that's ignored until the merge, so circleci failures are expected)

### DIFF
--- a/src/python/doc/witness_complex_user.rst
+++ b/src/python/doc/witness_complex_user.rst
@@ -47,7 +47,7 @@ which leads to definitions of **weak relaxed witness complex** (or just relaxed 
 In particular case of 0-relaxation, weak complex corresponds to **witness complex** introduced in
 :cite:`de2004topological`, whereas 0-relaxed strong witness complex consists of just vertices and is not very
 interesting. Hence for small relaxation weak version is preferable.
-However, to capture the homotopy type (for example using Gudhi::persistent_cohomology::Persistent_cohomology) it is
+However, to capture the homotopy type (for example using :func:`gudhi.SimplexTree.persistence`) it is
 often necessary to work with higher filtration values. In this case strong relaxed witness complex is faster to compute
 and offers similar results.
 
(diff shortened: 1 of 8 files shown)


mglisse:
@VincentRouvreau do we expect that merging the branch will automatically update docker and make circleci work? Or should I first merge the docker patch so we can see CI results before merging this branch?

VincentRouvreau:
> @VincentRouvreau do we expect that merging the branch will automatically update docker and make circleci work? Or should I first merge the docker patch so we can see CI results before merging this branch?

Docker hub is constructing the image when there are modifications in the master branch.
This is the reason why you have to merge the modification of Dockerfile_for_circleci_image into master first.
Here it is strange that circleCI was not launched on your merge request...

mglisse:
>  This is the reason why you have to merge the modification of Dockerfile_for_circleci_image into master first.

Well, it should work if I merge the whole branch into master, but then we won't have seen the test results.

>  Here it is strange that circleCI was not launched on your merge request...

It always does that on my branches. It is launched on the push to my fork, on which circleci is enabled, and I can see the results when I go to circleci.com. When I make a pull request, it skips rerunning a test it has already run elsewhere (that's sensible), but for some reason it does not add a link to that :disappointed: .

src/python/doc/witness_complex_user.rst
```diff
@@ -69,7 +69,7 @@ The construction of the Euclidean versions of complexes follow the same scheme:
 
 In the non-Euclidean classes, the lists of nearest landmarks are supposed to be given as input.
 
-The constructors take on the steps 1 and 2, while the function 'create_complex' executes the step 3.
+The constructors take on the steps 1 and 2, while the function :func:`!create_complex` executes the step 3.
```
VincentRouvreau: It does not make a real link, but I suppose it is because we cannot link to witness_complex_ref, strong_witness_complex_ref, euclidean_witness_complex_ref and euclidean_strong_witness_complex_ref, no ?

mglisse: Yes, at least that was my goal: give it a consistent look, but since I can't make it link to 4 things (and I don't want to add a list of links after the word), don't make it a link at all (with `!`).

PR #162, opened 2019-12-02
Representations and wasserstein fixes

VincentRouvreau:
pep8 wasserstein and representation modules.
Add a corresponding image to representation.


mglisse:
Were the changes produced by some tool (which one?) or manually?

src/python/example/diagram_vectorizations_distances_kernels.py
```diff
@@ -5,22 +5,25 @@
...
+
+D = np.array([[0., 4.], [1., 2.], [3., 8.], [6., 8.], [0., np.inf],
+              [5., np.inf]])
 diags = [D]
 
 diags = DiagramSelector(use=True, point_type="finite").fit_transform(diags)
-diags = DiagramScaler(use=True, scalers=[([0,1], MinMaxScaler())]).fit_transform(diags)
-diags = DiagramScaler(use=True, scalers=[([1], Clamping(limit=.9))]).fit_transform(diags)
+diags = DiagramScaler(use=True,
+                      scalers=[([0, 1], MinMaxScaler())]).fit_transform(diags)
```
mglisse: I have seen suggestions with
```
diags = (DiagramScaler(use=True, scalers=[([0,1], MinMaxScaler())])
         .fit_transform(diags))
```
I don't know if that's better or worse...

src/python/example/diagram_vectorizations_distances_kernels.py
```diff
@@ -102,7 +118,9 @@ def arctan(C,p):
 Y = PSS.transform(diags2)
 print("PSS kernel is " + str(Y[0][0]))
 
-PSS = PersistenceScaleSpaceKernel(kernel_approx=RBFSampler(gamma=1./2, n_components=100000).fit(np.ones([1,2])))
+PSS = PersistenceScaleSpaceKernel(
+    kernel_approx=RBFSampler(gamma=1. /
+                             2, n_components=100000).fit(np.ones([1, 2])))
```
mglisse: That seems like a strange place to break, or even to add spaces. Also, why replace 1. with 1.0 in Wasserstein but not here?

VincentRouvreau:
> Were the changes produced by some tool (which one?) or manually?

I used `black` for wasserstein, but I had to use `yapf` for representations (some of the code was not nicely idented with `black`):
```
                        if self.location == "lower":
                            new_diag = np.concatenate(
                                [
                                    thresh_diag[
                                        sort_index[
                                            min(self.num_pts, thresh_diag.shape[0]) :
                                        ],
                                        :,
                                    ],
                                    diag[~idx_thresh],
                                ],
                                axis=0,
                            )
```

This explains why there is 1.0 in wasserstein.py (because of `black`) and 1. in the other files (`yapf` does not modify 1. to 1.0).
I can restart from scratch to have a nice identation, as you want.

src/python/example/diagram_vectorizations_distances_kernels.py
mglisse: This should have been `gamma=.5` anyway, and then both black and yapf (all 4 default styles) give something vaguely sensible.

mglisse:
> I used `black` for wasserstein, but I had to use `yapf` for representations

How do you use them? When I try locally, they modify the code, but they keep the long lines in docstrings. Did you use a separate tool for that?

> not nicely idented with `black`

I've seen worse.

src/python/example/diagram_vectorizations_distances_kernels.py
```diff
@@ -49,50 +54,61 @@ def pow(n):
...
 plt.show()
 
-D = np.array([[1.,5.],[3.,6.],[2.,7.]])
+D = np.array([[1., 5.], [3., 6.], [2., 7.]])
 diags2 = [D]
 
-diags2 = DiagramScaler(use=True, scalers=[([0,1], MinMaxScaler())]).fit_transform(diags2)
+diags2 = DiagramScaler(use=True,
+                       scalers=[([0,
+                                  1], MinMaxScaler())]).fit_transform(diags2)
```
mglisse: `yapf3 --style google` avoids the ugly break in the middle of `[0, 1]`, but that appears to be random, making  `scalers` one letter longer makes it break there again.

PR #163, opened 2019-12-03
Fix #134 - Vector_vertex_iterator has become a map

### LINKED ISSUE #134: [Alpha_complex] Vector_vertex_iterator has become a map
In commit fb22bc9ca84f5b3c55a598bf0c903a73c117e783, this happened
```diff
-  typedef typename std::vector< CGAL_vertex_iterator > Vector_vertex_iterator;
+  typedef typename std::map< Vertex_handle, CGAL_vertex_iterator > Vector_vertex_iterator;
```
(and now Vertex_handle is being replaced by size_t). The name `Vector_vertex_iterator` is misleading. And I don't understand the point of using a map instead of a vector (if there are duplicate points, some entries in the vector may be unused, but that shouldn't be a problem). At the very least `unordered_map` would be faster.

VincentRouvreau:
Fix #134

### DIFF
--- a/src/Alpha_complex/include/gudhi/Alpha_complex.h
+++ b/src/Alpha_complex/include/gudhi/Alpha_complex.h
@@ -121,8 +121,8 @@ class Alpha_complex {
   // size_type type from CGAL.
   typedef typename Delaunay_triangulation::size_type size_type;
 
-  // Map type to switch from simplex tree vertex handle to CGAL vertex iterator.
-  typedef typename std::map< std::size_t, CGAL_vertex_iterator > Vector_vertex_iterator;
+  // Structure to switch from simplex tree vertex handle to CGAL vertex iterator.
+  typedef typename std::vector< CGAL_vertex_iterator > Vector_vertex_iterator;
 
  private:
   /** \brief Vertex iterator vector to switch from simplex tree vertex handle to CGAL vertex iterator.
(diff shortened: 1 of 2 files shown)


mglisse:
Is `number_of_vertices()` still valid after this, in case there were duplicate points? Should we ask `triangulation_` instead?

VincentRouvreau:
> Is `number_of_vertices()` still valid after this, in case there were duplicate points? 

In case there are duplicated points (let's say 7 points + 2 points that are already there), when I iterate on finite full cells vertices, I can reach the point where the vertices iterator `data()` method returns 8 instead of 1 (or 7 instead of 0).
For sure I have to use `vertex_handle_to_iterator_.resize(point_cloud.size());`

> Should we ask `triangulation_` instead? I am not sure what this function is for. It says that the number of vertices and points is the same, but that's not true...

I am not sure of the aim of this method anymore. Maybe I thought this method would help when you want to use the `get_point` method. Do we keep this method, but returns `triangulation_->number_of_vertices()` instead ? Maybe it would be helpful to detect if some points are duplicated `(points.size() != alphacomplex.number_of_vertices())`

> By the way, `triangulation_->number_of_vertices() > 0` looks equivalent to `triangulation_->empty()` in create_complex.

That's for sure.

mglisse:
>  For sure I have to use `vertex_handle_to_iterator_.resize(point_cloud.size());`

Yes.

> Maybe it would be helpful to detect if some points are duplicated `(points.size() != alphacomplex.number_of_vertices())`

After create_complex, we can get that information from the simplicial complex. But yes, we can keep it if you like, it doesn't hurt as long as it matches its doc.

VincentRouvreau:
> After create_complex, we can get that information from the simplicial complex. But yes, we can keep it if you like, it doesn't hurt as long as it matches its doc.

I removed the method on 329637d2a0a806955a29faccf38915c3fb7cd2fd and modify the tests accordingly.

PR #165, opened 2019-12-04
Delaunay triangulation for alpha complex in dD

VincentRouvreau:
A prototype to fix issue #74 for dD Alpha complex as the filtration is computed right after the delaunay triangulation.
It is not that trivial for the 3d case (not done on this pull request).


src/python/include/Alpha_complex_interface.h
```diff
@@ -60,7 +60,7 @@ class Alpha_complex_interface {
   }
 
   void create_simplex_tree(Simplex_tree_interface<>* simplex_tree, double max_alpha_square) {
-    alpha_complex_->create_complex(*simplex_tree, max_alpha_square);
+    alpha_complex_->create_complex(*simplex_tree, max_alpha_square, false, false);
```
mglisse: It doesn't hurt, but it wasn't necessary, and could cause conflicts if you are doing the Epeck_d thing at the same time.

mglisse:
Do we end up with NaN as filtration value everywhere? Does it cause problems if we pass a simplex tree with the no-filtration option, in debug mode? (at least assign_filtration has an assertion IIRC)
I need to think a bit if we want to use 0 or some user-specified filtration value.

src/python/include/Alpha_complex_interface.h
VincentRouvreau: Yes you are right. I rolled back this modification on 7cd38c50d3843fd222d4210507bad948afe63912

VincentRouvreau:
> Do we end up with NaN as filtration value everywhere?

Yes.

> Does it cause problems if we pass a simplex tree with the no-filtration option, in debug mode? (at least assign_filtration has an assertion IIRC)

It asserts as `insert_simplex_and_subfaces` is calling `assign_filtration` (in the recursive call).

> I need to think a bit if we want to use 0 or some user-specified filtration value.

The signature is `insert_simplex_and_subfaces(const InputVertexRange& Nsimplex, Filtration_value filtration = 0)`, we could think of 2 implementations with the 2 signatures:
+ `insert_simplex_and_subfaces(const InputVertexRange& Nsimplex)`.
+ `insert_simplex_and_subfaces(const InputVertexRange& Nsimplex, Filtration_value filtration)`.

Do you have any use case ? From my point of view, users will build a simplex tree with `NaN` values and will set filtration values with their own values.

mglisse:
> Do you have any use case ?

Build the complex, set some values for the vertices, and `make_filtration_non_decreasing`. I am not sure what happens to NaN in this case, -infinity would work.

VincentRouvreau:
All is fixed and will work when #172 will be fixed. cf. https://gist.github.com/VincentRouvreau/feb35ac8f3b590068cd64d0745798b2e

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -248,16 +248,20 @@ class Alpha_complex {
...
    * 
    * @param[in] complex SimplicialComplexForAlpha to be created.
    * @param[in] max_alpha_square maximum for alpha square value. Default value is +\f$\infty\f$, and there is very
-   * little point using anything else since it does not save time.
+   * little point using anything else since it does not save time. Useless if `default_filtration_value` is set to
+   * `true`.
    * @param[in] exact Exact filtration values computation. Not exact if `Kernel` is not <a target="_blank"
    * href="https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epeck__d.html">CGAL::Epeck_d</a>.
-   * 
+   * @param[in] default_filtration_value Set this value to `true` if filtration values are not needed to be computed.
```
mglisse: If we want people to be able to use it with `assign_filtration` on the vertices and `make_filtration_non_decreasing`, we need to document what filtration value simplices end up with.

src/Alpha_complex/test/Delaunay_complex_unit_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+  std::vector<Point> points;
+  // 50 points on a 4-sphere
+  points = Gudhi::generate_points_on_sphere_d<TestedKernel>(10, 5, 1.);
+
+  Gudhi::alpha_complex::Alpha_complex<TestedKernel> alpha_complex(points);
+
+  // Alpha complex
+  Gudhi::Simplex_tree<> stree_from_alpha_complex;
+  BOOST_CHECK(alpha_complex.create_complex(stree_from_alpha_complex));
+  stree_from_alpha_complex.initialize_filtration();
```
mglisse: Why? filtration_simplex_range already does it, and you can use complex_simplex_range instead.

src/Alpha_complex/test/Delaunay_complex_unit_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+  // Check all the simplices from alpha complex are in the Delaunay complex
+  for (auto f_simplex : stree_from_alpha_complex.filtration_simplex_range()) {
+    std::vector<Gudhi::Simplex_tree<>::Vertex_handle> simplex;
+    for (Gudhi::Simplex_tree<>::Vertex_handle vertex : stree_from_alpha_complex.simplex_vertex_range(f_simplex)) {
+      std::cout << "(" << vertex << ")";
+      simplex.push_back(vertex);
+    }
+    std::cout << std::endl;
+    Gudhi::Simplex_tree<>::Simplex_handle sh = stree_from_delaunay_complex.find(simplex);
```
mglisse: Can't we pass `simplex_vertex_range(f_simplex)` directly to `find`? It is a range.

mglisse:
I expect the python version will be in a separate PR? (asking in case the issue gets automatically closed)

VincentRouvreau:
> I expect the python version will be in a separate PR? (asking in case the issue gets automatically closed)

Yes, that's it. I first have to think about the interface.

src/Alpha_complex/test/Delaunay_complex_unit_test.cpp
VincentRouvreau: I don't think so because here we have 2 different simplex trees. The first one is constructed as an alpha complex and the second as a delaunay complex. This means the simplex handles (iterator over the dictionnary) won't be the same for the 2 data structures. I just check I have the same triangulations in the 2 simplex trees, and filtration values are `NaN` for the Delaunay complex

mglisse: I didn't say to use the Simple_handle of one complex in the other. All I am saying is that simplex_vertex_range returns a range (of Vertex_handle, i.e. something like int), and that you are copying it to a vector, just to pass it to another function as a range. The intermediate vector seems unnecessary. Am I missing something?

VincentRouvreau: Ok, I got it. I removed the vector copy. It was just here in case it fails, but it shall not.

PR #167, opened 2019-12-05
Wbary theo

tlacombe:
Implementation of barycenters for persistence diagram. 

In its current state, support "Lagrangian" estimation of the barycenter.


mglisse:
This seems to be based on an old version of master. You should git rebase (will require to force push afterwards). Then you can edit CMakeLists.txt to let it know about your new file(s).

tlacombe:
I hope I managed to update master in my fork and merge it with my branch wbary-theo correctly. I also updated CMakeLists (and added a test file) as suggested.

src/python/doc/barycenter_user.rst
```diff
@@ -0,0 +1,51 @@
+:orphan:
+
+.. To get rid of WARNING: document isn't included in any toctree
+
+Wasserstein distance user manual
```
mglisse: copy-paste?

tlacombe: Indeed. Should I keep the header about the toctree btw ?

mglisse:
I wonder if this should be merged with wasserstein.py (or make it wasserstein/distance.py and wasserstein/barycenter.py), they have the same external dependencies, some common helper functions, and the barycenter seems to be a wasserstein barycenter.
Any opinions?

tlacombe:
I agree with the idea of putting the files in a same folder, for which the same could be "wasserstein", or "transport" or something else. Especially considering I might write further functions in the future that will also use POT + similar helper functions.

mglisse:
@martinroyer do you want to comment on the organization?

Putting things together would also apply to the documentation.

src/python/doc/barycenter_user.rst
mglisse: The question about toctree is for @VincentRouvreau , but my guess is that you have to keep some arbitrary text there, so it might as well be text that explains why it is there.

VincentRouvreau: We have no toctree element in our documentation. It seems annoying for sphinx as it outputs some warning for each `*.rst`. You can keep `:orphan:` and the comment to fix the warning.

martinroyer:
> @martinroyer do you want to comment on the organization?

Weirdly enough, I am thinking of a mix-up between what exists and what you suggested:
- it seems like the material from wasserstein.py now has a natural place in gudhi.representation.metrics
- I like @tlacombe 's suggestion of a transport file, probably at gudhi.representation.transport
- and then I would keep barycenter.py where it is (making use of representation.transport.optimal_matching()) and barycenters could be available at gudhi.barycenter

martinroyer:
I find that overall the function names are too short and not explicit enough (_proj_on_diag, _build_dist_matrix, _mean, ... I know I tend to prefer longer names).

mglisse:
> I find that overall the function names are too short and not explicit enough (_proj_on_diag, _build_dist_matrix, _mean, ... I know I tend to prefer longer names).

Is it mostly the fact that they are abbreviated that you don't like, and would you be fine with `_projection_on_the_diagonal` and `_build_distance_matrix`? For private local functions like this, I think the context reduces a bit the need for long names. Could you give an example of the kind of name you would prefer?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,342 @@
...
+    #dg4 = np.array([])
+    #
+    #bary, a = lagrangian_barycenter(pdiagset=[dg1, dg2, dg3, dg4],init=3, verbose=True)
+    #_plot_barycenter([dg1, dg2, dg3, dg4], bary, a)
+    #message = "Wasserstein barycenter estimated:"    
+    #print(message)
+    #print(bary)
+
+if __name__=="__main__":
+    _sanity_check(verbose = True)
```
mglisse: Ah, so you have a lot of testing code (and a lot is commented out) in the source file itself, and running the file runs the tests. It seems to be a somewhat common practice in python, but since we haven't used it in gudhi yet, I'd like to understand if this is useful.
The part of _sanity_check that currently doesn't have a `#` looks like it might as well go in the separate test file. Is this function meant to be run regularly as part of the testsuite, or only when debugging manually? What about _plot_barycenter, is it meant for debugging? For an example or tutorial? As a public interface?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,342 @@
...
+    return np.array([(w[0] + w[1])/2 , (w[0] + w[1])/2])
+
+
+def _proj_on_diag_array(X):
+    '''
+    :param X: (n x 2) array encoding the points of a persistent diagram.
+    :returns: (n x 2) array encoding the (respective orthogonal) projections of the points onto the diagonal
+    '''
+    Z = (X[:,0] + X[:,1]) / 2.
+    return np.array([Z , Z]).T
```
mglisse: `Z = X.mean(axis=1)` and `return Z.reshape(-1,1).repeat(2,axis=1)` would be less readable, ok.

src/python/gudhi/barycenter.py
tlacombe: Hum... I am not sure what follows "__test__": is useful actually. It is mostly an artifact coming from my trials. If this is not a spread practice in gudhi, I should remove it.

Regarding _plot_barycenter, I think this function should be public at some point as it could be convenient, but I didn't know which approach I should take. Currently, it is somewhat cumbersome to use, so I preferred to hide it.

src/python/gudhi/barycenter.py
tlacombe: I guess the Z = X.mean(axis=1) is as readable as what I have done (for the second part, I would think `np.array([Z, Z]).T` is more readable, but I didn't investigate if it has any impact in terms of performances (this is however not the bottleneck of the algorithm anyway).

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def _build_dist_matrix(X, Y, p=2., q=2.):
+    '''
+    :param X: (n x 2) numpy.array encoding the (points of the) first diagram.
+    :param Y: (m x 2) numpy.array encoding the second diagram.
+    :param q: Ground metric (i.e. norm l_q).
+    :param p: exponent for the Wasserstein metric.
```
mglisse: So the function actually returns a cost matrix where the costs are `dist**p`.

src/python/gudhi/barycenter.py
mglisse: If you consider some code useless, then yes, please remove it.
For the plot function, moving it to an example could be a reasonable compromise.

src/python/gudhi/barycenter.py
tlacombe: Indeed, I should change the name accordingly.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,342 @@
...
+        Cdy = np.linalg.norm(Y - Ydiag, ord=q, axis=1)**p
+    Cf = np.hstack((C, Cxd[:,None]))
+    Cdy = np.append(Cdy, 0)
+
+    Cf = np.vstack((Cf, Cdy[None,:]))
+
+    return Cf
+
+
+def _optimal_matching(X, Y):
```
mglisse: Is that more or less the function that Hari needs, and which could become public in the future?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,342 @@
...
+        return (k * w + (m-k) * w_delta) / m
+    else:
+        return np.array([0, 0])
+
+
+def lagrangian_barycenter(pdiagset, init=None, verbose=False):
+    """
+        Compute the estimated barycenter computed with the algorithm provided
+        by Turner et al (2014).
+        It is a local minima of the corresponding Frechet function.
```
mglisse:
```suggestion
        It is a local minimum of the corresponding Frechet function.
```

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,342 @@
...
+                    If None, init is made on a random diagram from the dataset. 
+                    Otherwise, it must be an int (then we init with diagset[init])
+                    or a (n x 2) numpy.array enconding a persistence diagram with n points.
+    :param verbose: if True, returns additional information about the
+                        barycenters (assignment and energy).
+    :returns: If not verbose (default), a numpy.array encoding
+                    the barycenter estimate (local minima of the energy function). 
+                    If verbose, returns a triplet (Y, a, e)
+                    where Y is the barycenter estimate, a is the assignments between the
+                    points of Y and thoses of the diagrams,
```
mglisse: How are those assignments encoded? Essentially, how do I read `a`?

tlacombe:
@martinroyer , @mglisse : 
Concerning the organization of the folder(s), I have no insight on what should be the best.
To me, `wasserstein_distance` must be at the same level as `bottleneck_distance`, as those are (mathematically speaking) the same functions up to a change in the parameter. It seems the latter is accessible for `gudhi.representations.metrics` so should be the former. 

Concerning the names, as @mglisse said, I wouldn't think that complete names are required for hidden functions. But I agree it does not cost much to write a longer name.

src/python/gudhi/barycenter.py
tlacombe: Well, I do not really know what Hari needs.
But yes, I agree that this function, or something similar, should be public in the future (in `representations.` I guess?)

src/python/gudhi/barycenter.py
tlacombe: Good question, I will detail the docs accordingly.

src/python/gudhi/barycenter.py
tlacombe: What do you mean by "moving it to an example"? In the docs?

src/python/gudhi/barycenter.py
tlacombe: Note also that for the current available function `lagrangian_barycenter` one must use `p=2`. I decided to let this parameter `p` as I plan to implement other `barycenter` functions that could work with different values of `p`.

src/python/gudhi/barycenter.py
mglisse: There is a directory `example/`, next to `test/`, where you can add files. Or you can write a tutorial for https://github.com/GUDHI/TDA-tutorial .

mglisse:
>     * it seems like the material from wasserstein.py now has a natural place in gudhi.representation.metrics
> 
>     * I like @tlacombe 's suggestion of a transport file, probably at gudhi.representation.transport
> 
>     * and then I would keep barycenter.py where it is (making use of representation.transport.optimal_matching()) and barycenters could be available at gudhi.barycenter

I forget why we called it gudhi.representations and not gudhi.diagrams :confused:
Alex is also working on a vectorization of diagrams that uses optimal transport, we don't want to confuse the two.
gudhi.representations.metrics.wasserstein.lagrangian_barycenter is getting a bit long (I'd say that metrics is the easiest one to skip).
I am not really convinced by gudhi.barycenter, I think this goes back to our discussion where I find integer.addition more natural, while you prefer addition.integer...
If we start putting more stuff in gudhi.representations, its doc will need some work.

tlacombe:
> > ```
> > * it seems like the material from wasserstein.py now has a natural place in gudhi.representation.metrics
> > 
> > * I like @tlacombe 's suggestion of a transport file, probably at gudhi.representation.transport
> > 
> > * and then I would keep barycenter.py where it is (making use of representation.transport.optimal_matching()) and barycenters could be available at gudhi.barycenter
> > ```
> 
> I forget why we called it gudhi.representations and not gudhi.diagrams confused
> Alex is also working on a vectorization of diagrams that uses optimal transport, we don't want to confuse the two.
> gudhi.representations.metrics.wasserstein.lagrangian_barycenter is getting a bit long (I'd say that metrics is the easiest one to skip).
> I am not really convinced by gudhi.barycenter, I think this goes back to our discussion where I find integer.addition more natural, while you prefer addition.integer...
> If we start putting more stuff in gudhi.representations, its doc will need some work.

I guess `representations` could be justified by the fact that we could introduce (in the long run) some other "topological representations" of our data (like multi-persistence based descriptors).

I think that `wasserstein_distance()` should be called from `something.metrics` as it is a metric. 
As `barycenter` works on a set of persistence diagrams, I guess it should be accessible as `representations.barycenter`.

src/python/gudhi/barycenter.py
tlacombe: @mglisse : I will write a tutorial, it seems to be a good idea as there are some "maths" involved. I was wondering, in a tutorial, should I consider the user have access to the "live" version of gudhi? (i.e. with `lagrangian_barycenter` in this case). I am afraid that it could be somewhat misleading.

tlacombe:
Aside from organizational aspects (where should we put this file), is there something left to be done for a "v1" integration ?

src/python/doc/barycenter_sum.inc
```diff
@@ -0,0 +1,24 @@
...
+   |      ./img/barycenter.png                                       | mean in a non linear space such as the one of persistence diagrams.  |                                                                  |
+   |      :figclass: align-center                                    | Given a set of persistence diagrams :math:`\mu_1 \dots \mu_n`, it is | :Introduced in: GUDHI 3.1.0                                      |
+   |                                                                 | defined as a minimizer of the variance functional, that is of        |                                                                  |
+   |      Illustration of Frechet mean between persistence           | :math:`\mu \mapsto \sum_{i=1}^n d_2(\mu,\mu_i)^2`.                   | :Copyright: MIT                                                  |
+   |      diagrams.                                                  | where :math:`d_2` denotes the Wasserstein-2 distance between persis- |                                                                  |
```
mglisse: Words split with '-' end up looking weird (same for "available" below), let's avoid that.

tlacombe: Ok I will modify that.

mglisse:
Is it obvious from the name that a Lagrangian barycenter is an approximation of the barycenter for the Wasserstein distance? That will of course be less of a problem if we move it to gudhi.wasserstein.*.

tlacombe:
> Is it obvious from the name that a Lagrangian barycenter is an approximation of the barycenter for the Wasserstein distance?*.

Not at all actually. Here, "lagrangian" is to be opposed to "Eulerian", as in physics (and mathematical) modelling. Lagrangian means that you focus on updating "particles locations" (what does the current algorithm), while Eulerian means you focus on updating "the mass in a given position" (encoding positions with a discrete grid in practice). Point is that "Lagrangian approaches" are generally not convex (think of k-means solved by the Lloyd algorithm... or this algorithm), while Eulerian approaches give convex problem (at least in the case of barycenter estimation), at the cost of introducing a grid. This terminology is rather standard in Optimal Transport (basically, transporting point clouds (Lagrangian) vs transporting histograms (Eulerian), among the only things you can do in practice.)

As I plan to do some "Eulerian barycenter" in further implementations, I thought about naming this one `lagrangian_barycenter`.
Now I realize how intricate it is... 

What could be a better naming ? The main properties of this algorithm are:
1) It is not convex (thus get stuck in local minima)
2) It really looks like a Lloyd algorithm, and more generally an EM-algo (we update points locations, then we update matchings, then we update points locations, then... until a fix point is reached.)
3) It is somewhat "combinatorial" (opposed to the "variational" algorithms I plan to add later)
Another option could be to name it after its creators (K.Turner et al.).

mglisse:
`gudhi.wasserstein.lagrangian_barycenter` looks ok to me. It is mostly the "wasserstein" part that I am missing in the current `gudhi.barycenter.lagrangian_barycenter`, since I guess we could look for barycenters for other metrics. Otherwise, `wasserstein_barycenter`? `wasserstein_lagrangian_barycenter`?
(we are really not consistent with snake case vs camel case in gudhi python)

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,227 @@
...
+
+    # Initialisation of barycenter
+    if init is None:
+        i0 = np.random.randint(m)  # Index of first state for the barycenter
+        Y = X[i0].copy() #copy() ensure that we do not modify X[i0]
+    else:
+        if type(init)==int:
+            Y = X[init].copy()
+        else:
+            Y = init.copy()
```
VincentRouvreau: This use case is not tested. Do you have any way to test it ?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,227 @@
...
+        for i in range(m):
+            edges, cost = _optimal_matching(Y, X[i], withcost=True)
+            print(edges)
+            groupings.append([x_i_j for (y_j, x_i_j) in enumerate(edges) if y_j < n_y])
+            energy += cost
+            log["groupings"] = groupings
+        energy = energy/m
+        log["energy"] = energy
+
+        return Y, log
```
VincentRouvreau: This use case is not tested. Do you have any way to test it ?
A test of verbose mode would be useful. Maybe ok if you relaunch `test_lagrangian_barycenter` in verbose mode ?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,227 @@
...
+                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) and its orthogonal proj onto the diagonal.
+                note also that C[n+1, m+1] = 0  (it costs nothing to move from the diagonal to the diagonal).
+                Note that for lagrangian_barycenter, one must use p=q=2. 
+    '''
+    Xdiag = _proj_on_diag_array(X)
+    Ydiag = _proj_on_diag_array(Y)
+    if np.isinf(q):
+        C = sc.cdist(X, Y, metric='chebyshev')**p
+        Cxd = np.linalg.norm(X - Xdiag, ord=q, axis=1)**p
+        Cdy = np.linalg.norm(Y - Ydiag, ord=q, axis=1)**p
```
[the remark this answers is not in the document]
tlacombe: Actually, the current algorithm only works if p=q=2 (btw, I should rename these variables with respect to the name we recently chose for `wasserstein_distance.py` ...!)

Allowing for different `p,q` is just a preliminary step for further algorithms (working differently, and should work with other values for `p,q`). This shouldn't be used for now. In the current implementation, the user shouldn't be able to change it as these are not parameters of `lagrangian_barycenter` (they are only used in intermediate functions, like `_optimal_matching`, with default parameters).
Would it be safer to remove it?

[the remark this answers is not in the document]
VincentRouvreau: If it is some dead code, it is safer to remove it, thanks. If not, please test it.
Ok for the renaming.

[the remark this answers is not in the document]
tlacombe: Well I wouldn't say it is some dead code. It is rather some "not-born-yet" code ; in sense that I will need it in upcoming implementations of barycenter estimation with different approaches. But it is not used yet.

But ok, I will rewrite it for now (restricting to the case p=q=2), and will reintroduce this version when implementing other algorithms.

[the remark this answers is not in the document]
mglisse: From Vincent's comment, it seems that he would also be happy if you kept the code, and just added a very small test that calling _build_dist_matrix returns what is expected (the notion of unit test means testing some private functions as well, not just the final public one). Also, just moving this code to wasserstein.py (or importing stuff from wasserstein.py) would mean that this isn't dead code anymore, since the duplicate function there is used, no?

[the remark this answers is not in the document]
tlacombe: Yes I could do that and that would be the proper way to go. When I started implementing this, `wasserstein.py` wasn't merge to `master`. I think I actually need a bit of re-writing for this code overall.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,227 @@
...
+                    - assigments, a list of list of pairs (i,j),
+                        That is, a[k] = [(i, j) ...], where (i,j) indicates that X[i] is matched to Y[j]
+                        if i > len(X) or j > len(Y), it means they represent the diagonal.
+                    - energy, a float representing the Frechet mean value obtained.
+    """
+    X = pdiagset  # to shorten notations, not a copy
+    m = len(X)  # number of diagrams we are averaging
+    if m == 0:
+        print("Warning: computing barycenter of empty diag set. Returns None")
+        return None
```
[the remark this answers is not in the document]
tlacombe: Tested now

src/python/gudhi/barycenter.py
tlacombe: Tested now

src/python/gudhi/barycenter.py
tlacombe: Tested now

src/python/gudhi/barycenter.py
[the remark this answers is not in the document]
tlacombe: I finally removed _build_dist_matrix from this file and imported the one of `wasserstein.py`.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,227 @@
...
+    M = _build_dist_matrix(X, Y)
+
+    a = np.full(n+1, 1. / (n + m) )  # weight vector of the input diagram. Uniform here.
+    a[-1] = a[-1] * m                # normalized so that we have a probability measure, required by POT
+    b = np.full(m+1, 1. / (n + m) )  # weight vector of the input diagram. Uniform here.
+    b[-1] = b[-1] * n                # so that we have a probability measure, required by POT
+    P = ot.emd(a=a, b=b, M=M)*(n+m)
+    # Note : it seems POT return a permutation matrix in this situation, ie a vertex of the constraint set (generically true).
+    if withcost:
+        cost = np.sqrt(np.sum(np.multiply(P, M)))
```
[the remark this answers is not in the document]
tlacombe: Added a test.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+# This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+# See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+# Author(s):       Theo Lacombe
+#
+# Copyright (C) 2019 Inria
+#
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
```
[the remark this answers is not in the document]
tlacombe: I am currently working on not having lines with more than 80 char (sorry, I misclicked by asking a new review, I pushing to get this done). 

However, it seems that even the docstring is more than 80c long, is it normal?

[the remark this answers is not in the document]
tlacombe: Done in the latest commit

src/python/doc/barycenter_user.rst
```diff
@@ -0,0 +1,51 @@
...
+Function
+--------
+.. autofunction:: gudhi.barycenter.lagrangian_barycenter
+
+
+Basic example
+-------------
+
+This example computes the Frechet mean (aka Wasserstein barycenter) between four persistence diagrams.
+It is initialized on the 4th diagram, which is the empty diagram. It is encoded by np.array([]).
```
mglisse: What is the significance of the initialization diagram? Does it only change the speed? Does it change the result? How can a user decide how to initialize?
I don't think you need to say that np.array([]) is empty.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+import ot
+import numpy as np
+import scipy.spatial.distance as sc
+
+from gudhi.wasserstein import _build_dist_matrix, _perstot
+
+
+def _proj_on_diag(w):
```
mglisse: You already have that one in gudhi.wasserstein.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,233 @@
...
+                                that is the mean of squared distances 
+                                of observations to the output.
+                    - nb_iter, integer representing the number of iterations 
+                               performed before convergence of the algorithm.
+    """
+    X = pdiagset  # to shorten notations, not a copy
+    m = len(X)  # number of diagrams we are averaging
+    if m == 0:
+        print("Warning: computing barycenter of empty diag set. Returns None")
+        return None
```
mglisse: I haven't had time to think, are there obvious advantages returning None rather than an empty diagram for instance?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,233 @@
...
+        cost = _perstot(X, order=2, internal_p=2)**2
+        if withcost:
+            return res, cost
+        else:
+            return res
+
+    # we know X, Y are not empty diags now
+    M = _build_dist_matrix(X, Y, order=2, internal_p=2)
+
+    a = np.full(n+1, 1. / (n + m) )
```
mglisse: Did you clarify what happens with integer weights, without renormalization?

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,233 @@
...
+    a[-1] = a[-1] * m                
+    b = np.full(m+1, 1. / (n + m) )  
+    b[-1] = b[-1] * n                
+    P = ot.emd(a=a, b=b, M=M)*(n+m)
+    # Note : it seems POT returns a permutation matrix in this situation, 
+    # ie a vertex of the constraint set (generically true).
+    if withcost:
+        cost = np.sum(np.multiply(P, M))
+    P[P < 0.5] = 0  # dirty trick to avoid some numerical issues... to improve.
+    res = np.nonzero(P)
```
mglisse: Would `np.argwhere` help?

src/python/gudhi/barycenter.py
tlacombe: Well, for me, the empty diagrams would be the barycenter of several empty diagrams, but averaging "no" diagrams might be a bit weird, but I agree that the empty diagram could be a correct answer (just like in R, when you say a sum over the empty set is 0). 
I just think that averaging an empty set of diagrams is a weird idea, so it would be adequate to have a warning there. If you're running Lloyd algo for K-Means for instance, and one centroid got "unaffected", you may not want to tell at the next iteration that it's "0" (or the emptydgm), right?

src/python/doc/barycenter_user.rst
tlacombe: It actually has a lot of importance: the energy functional is not convex in the PD space so the output (and its quality) depends on the initialization. There is however no heuristic (afaik) to pick a good initial diagram. 

I should change the default initialization (which is currently "one of the dgm at random") as it might be counter intuitive. Maybe we could start with the empty dgm (which I think is a very bad choice actually...), or the "first dgm of the pdiagset", so that the user gets a deterministic output?

I will implement later another estimation procedure ("Eulerian" instead of "Lagrangian"), which will be convex this time (at the cost of optimizing on a regular grid, returning a "persistence surface" instead of an image).

mglisse: My questions meant that this requires more documentation, not particularly that the code needs changing. Offering the user an option to initialize differently is more useful if the user understands what that means.

src/python/gudhi/barycenter.py
tlacombe: Actually they take slightly different objects, the one here returns the projection onto the diagonal of a single point, while the one in `gudhi.wasserstein` does the same for an array of points. 

However, this one is only used once so I just removed it and slightly changed the code in `_mean`.

src/python/gudhi/barycenter.py
tlacombe: Well yes, it seems re-normalization is actually useless. I don't know why I thought it was required... (I think that the error message "check that a, b are in the simplex" misled me).
I think it's better to remove it then, it probably reduces numerical error impact.

src/python/gudhi/barycenter.py
tlacombe: Yes it does !

src/python/doc/barycenter_user.rst
tlacombe: Improved the doc (hopefully).

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,229 @@
...
+    """
+    :param x: a list of 2D-points, off diagonal, x_0... x_{k-1}
+    :param m: total amount of points taken into account, 
+                that is we have (m-k) copies of diagonal
+    :returns: the weighted mean of x with (m-k) copies of the diagonal
+    """
+    k = len(x)
+    if k > 0:
+        w = np.mean(x, axis=0)
+        w_delta = (w[0] + w[1]) / 2 * np.ones(2)
```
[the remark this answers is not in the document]
tlacombe: Hum, I don't know if this is related, but in `wasserstein.py` we used single quote ` ''' ` for block comments, so I will do this change to be consistent (maybe it will improve something).

tlacombe:
> As I said in my review, you can go with 120 characters per line.
> But 80 is good too +1

Yes, but I had some >120 ; so anyway I went for 80 (a bit of overkill). I am trying to get rid of the errors with appveyor and travisCI, but I have no understanding of these (except that it comes from Sphinx).

VincentRouvreau:
> I am trying to get rid of the errors with appveyor and travisCI, but I have no understanding of these (except that it comes from Sphinx).

You have to merge last modifications from upstream/master and more specifically this one 63e4222e528317b7e0385bf5881393ff2f97fa80
"Normally", only doxygen target on travis shall fail. Brew is broken on travis and do not want to install doxygen anymore :cry:

tlacombe:
Is it ready for merge?

mglisse:
> Is it ready for merge?

Sorry, not yet. I'll try to find time...

tlacombe:
> > Is it ready for merge?
> 
> Sorry, not yet. I'll try to find time...

No problem, there is no emergency, I just wanted to know if I had something to do right now.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,224 @@
...
+    if m == 0:
+        print("Warning: computing barycenter of empty diag set. Returns None")
+        return None
+    
+    # store the number of off-diagonal point for each of the X_i
+    nb_off_diag = np.array([len(X_i) for X_i in X])  
+    # Initialisation of barycenter
+    if init is None:
+        i0 = np.random.randint(m)  # Index of first state for the barycenter
+        Y = X[i0].copy() #copy() ensure that we do not modify X[i0]
```
mglisse: Does that mean that you modify pdiagset during the algorithm? It may be worth documenting if this is the case, as the user may be confused if his diagrams change unexpectedly.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,224 @@
...
+        else:
+            Y = init.copy()
+
+    nb_iter = 0
+
+    converged = False  # stoping criterion
+    while not converged:
+        nb_iter += 1
+        K = len(Y)  # current nb of points in Y (some might be on diagonal)
+        G = np.zeros((K, m), dtype=int)-1  # will store for each j, the (index)
```
mglisse: `numpy.full`

src/python/gudhi/barycenter.py
mglisse: Once your other branch is merged, will you be able to remove this function completely?

tlacombe: I think so.

src/python/gudhi/barycenter.py
tlacombe: No, sorry, maybe my line-comment is not clear. 
I do modify `Y`, so that if I wrote `Y = X[i0]`, I would also modify `X|i0]` which is not the expected behavior. Maybe it is not worth documenting on this, it is very basic.

So the answer is no : I do not modify `pdiagset` .

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,224 @@
...
+        return res, cost
+    return res
+
+
+def lagrangian_barycenter(pdiagset, init=None, verbose=False):
+    '''
+    :param pdiagset: a list of size m containing numpy.array of shape (n x 2) 
+                        (n can variate), encoding a set of 
+                        persistence diagrams with only finite coordinates. 
+                        If empty, returns None.
```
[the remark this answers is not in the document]
tlacombe: I moved it to the :returns: section.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,224 @@
...
+                    the barycenter estimate 
+                    (local minima of the energy function). 
+                    If verbose, returns a couple (Y, log)
+                    where Y is the barycenter estimate,
+                    and log is a dict that contains additional informations:
+                    - groupings, a list of list of pairs (i,j),
+                        That is, G[k] = [(i, j) ...], where (i,j) indicates 
+                        that X[i] is matched to Y[j]
+                        if i > len(X) or j > len(Y), it means they 
+                        represent the diagonal.
```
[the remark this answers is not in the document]
tlacombe: Well, it shouldn't matter much, but I guess consistency is better than inconsistency.

src/python/gudhi/barycenter.py
tlacombe: changed

src/python/gudhi/barycenter.py
mglisse: Ah, yes. In retrospect, it is pretty obvious...

mglisse: Ah, yes. In retrospect, it is pretty obvious...

tlacombe: I removed the line comment

src/python/gudhi/barycenter.py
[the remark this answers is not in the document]
tlacombe: Actually the problem comes from the doc, the code returns -1. Doc updated.

src/python/gudhi/barycenter.py
```diff
@@ -0,0 +1,224 @@
...
+    :param verbose: if True, returns additional information about the
+                        barycenter.
+    :returns: If not verbose (default), a numpy.array encoding
+                    the barycenter estimate 
+                    (local minima of the energy function). 
+                    If verbose, returns a couple (Y, log)
+                    where Y is the barycenter estimate,
+                    and log is a dict that contains additional informations:
+                    - groupings, a list of list of pairs (i,j),
+                        That is, G[k] = [(i, j) ...], where (i,j) indicates
```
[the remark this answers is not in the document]
tlacombe: (hopefully) solved.

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -3,11 +3,11 @@
...
-   |                                                                 | diagonal points), where the value of a matching is defined as the    |                                                                  |
-   |      Wasserstein distance is the q-th root of the sum of the    | q-th root of the sum of all edge lengths to the power q. Edge lengths| :License: MIT                                                    |
-   |      edge lengths to the power q.                               | are measured in norm p, for :math:`1 \leq p \leq \infty`.            |                                                                  |
+   |      ../../doc/Bottleneck_distance/perturb_pd.png               | persistence diagrams using the sum of all edges lengths (instead of  |                                                                  |
+   |      :figclass: align-center                                    | the maximum). It allows to define sophisticated objects such as      | :Introduced in: GUDHI 3.1.0                                      |
```
mglisse, line 7: You probably had a conflict while merging, you need to check what the conflict is in that case and edit the file. Here, "Introduced in" has been replaced by "Since" and "Copyright" by "License".

src/python/doc/wasserstein_distance_sum.inc
```diff
@@ -3,11 +3,11 @@
...
+   |                                                                 | barycenters of a family of persistence diagrams.                     |                                                                  |
+   |      Wasserstein distance is the q-th root of the sum of the    |                                                                      | :Copyright: MIT                                                  |
```
mglisse, line 9: The legend of the figure seems redundant, and takes space below the figure. If you want to keep the text, I think it would look better in the middle, with the rest of the text.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -9,10 +9,16 @@ Definition
 
 .. include:: wasserstein_distance_sum.inc
 
-Functions
----------
-This implementation uses the Python Optimal Transport library and is based on
-ideas from "Large Scale Computation of Means and Cluster for Persistence
+The q-Wasserstein distance is defined as the minimal value 
+by a perfect matching between the points of the two diagrams (+ all
```
mglisse, line 13: "by" sounds strange, maybe "of" or "for"? If you think "by" is fine, then ok.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -84,3 +90,80 @@ The output is:
...
+     :figclass: align-center                                   
+                                                               
+     Illustration of Frechet mean between persistence          
+     diagrams.                                                 
+
+
+.. autofunction:: gudhi.barycenter.lagrangian_barycenter
+
+Basic example
+-------------
```
mglisse, line 129: Should we make the "basic example" sections **sub**sections? (technically, that just means using whatever other character than '-' to underline it, say "^" or "=")

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -84,3 +90,80 @@ The output is:
...
+
+This example estimates the Frechet mean (aka Wasserstein barycenter) between 
+four persistence diagrams.
+It is initialized on the 4th diagram.
+As the algorithm is not convex, its output depends on the initialization and 
+is only a local minimum of the objective function.
+Initialization can be either given as an integer (in which case the i-th 
+diagram of the list is used as initial estimate) or as a diagram. 
+If None, it will randomly select one of the diagram of the list
```
mglisse, line 138:
```suggestion
If None, it will randomly select one of the diagrams of the list 
```

src/python/gudhi/wasserstein/barycenter.py
```diff
@@ -0,0 +1,158 @@
...
+#
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
+
+
+import ot
+import numpy as np
+import scipy.spatial.distance as sc
+
+from gudhi.wasserstein import wasserstein_distance
```
mglisse, line 15:
```suggestion
from .wasserstein import wasserstein_distance
```
should work. I don't know if it is really better though...

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -84,3 +90,80 @@ The output is:
...
+
+.. figure::                                                    
+     ./img/barycenter.png                                      
+     :figclass: align-center                                   
+                                                               
+     Illustration of Frechet mean between persistence          
+     diagrams.                                                 
+
+
+.. autofunction:: gudhi.barycenter.lagrangian_barycenter
```
mglisse, line 126: Missing `wasserstein`, it doesn't show up in the doc.

src/python/doc/wasserstein_distance_user.rst
tlacombe: Agreed

src/python/doc/wasserstein_distance_user.rst
tlacombe: I guess I accidentally removed "minimal value *achieved* by"

src/python/doc/wasserstein_distance_sum.inc
tlacombe: Yes I did have ; I guess I merged improperly. I'll try to solve that.

src/python/doc/wasserstein_distance_sum.inc
tlacombe: I removed the text. I guess the doc is explicit enough.

src/python/doc/wasserstein_distance_user.rst
tlacombe: Done

src/python/doc/wasserstein_distance_user.rst
tlacombe: Done

src/python/gudhi/wasserstein/barycenter.py
```diff
@@ -0,0 +1,158 @@
...
+    :param pdiagset: a list of size m containing numpy.array of shape (n x 2) 
+                    (n can variate), encoding a set of 
+                    persistence diagrams with only finite coordinates. 
+    :param init: The initial value for barycenter estimate. 
+                    If None, init is made on a random diagram from the dataset. 
+                    Otherwise, it must be an int 
+                    (then we init with diagset[init])
+                    or a (n x 2) numpy.array enconding 
+                    a persistence diagram with n points.
+    :param verbose: if True, returns additional information about the
```
mglisse, line 45: I think at least for init and verbose you can give the type of the argument, using the usual sphinx syntax.

src/python/gudhi/wasserstein/barycenter.py
```diff
@@ -0,0 +1,158 @@
...
+                    barycenter.
+    :returns: If not verbose (default), a numpy.array encoding
+                    the barycenter estimate of pdiagset
+                    (local minima of the energy function).
```
mglisse, line 49:
```suggestion
                    (local minimum of the energy function). 
```

src/python/gudhi/wasserstein/barycenter.py
```diff
@@ -0,0 +1,158 @@
...
+    :param verbose: if True, returns additional information about the
+                    barycenter.
+    :returns: If not verbose (default), a numpy.array encoding
+                    the barycenter estimate of pdiagset
+                    (local minima of the energy function). 
+                    If pdiagset is empty, returns None.
+                    If verbose, returns a couple (Y, log)
+                    where Y is the barycenter estimate,
+                    and log is a dict that contains additional informations:
+                    - groupings, a list of list of pairs (i,j),
```
mglisse, line 54: The list isn't nicely formatted. Can you experiment a bit with making it look prettier? (a blank line before the list can help, but I don't think it is enough)

src/python/gudhi/wasserstein/barycenter.py
```diff
@@ -0,0 +1,158 @@
...
+                    That is, G[k] = [(i, j) ...], where (i,j) indicates 
+                    that X[i] is matched to Y[j]
+                    if i = -1 or j = -1, it means they
```
mglisse, line 57: Are you missing a `.` and a capital letter? Or `;`?

tlacombe: Done

src/python/gudhi/wasserstein/barycenter.py
tlacombe: Done

src/python/gudhi/wasserstein/barycenter.py
```diff
@@ -0,0 +1,160 @@
...
+        # Step 1 : compute optimal matching (Y, X_i) for each X_i
+        #          and create new points in Y if needed
+        for i in range(m):
+            _, indices = wasserstein_distance(Y, X[i], matching=True, order=2., internal_p=2.)
+            for y_j, x_i_j in indices:
+                if y_j >= 0:  # we matched an off diagonal point to x_i_j...
+                    if x_i_j >= 0:  # ...which is also an off-diagonal point.
+                        G[y_j, i] = x_i_j
+                    else:  # ...which is a diagonal point
+                        G[y_j, i] = -1  # -1 stands for the diagonal (mask)
```
mglisse, line 112: Isn't -1 the only possible negative value? Then this is the same as always writing x_i_j to G[y_j, i].

src/python/test/test_wasserstein_barycenter.py
```diff
@@ -0,0 +1,46 @@
...
+    res = np.array([[0.27916667, 0.55416667], [0.7375,  0.7625], [0.2375, 0.2625]])
+
+    dg7 = np.array([[0.1, 0.15], [0.1, 0.7], [0.2, 0.22], [0.55, 0.84], [0.11, 0.91], [0.61, 0.75], [0.33, 0.46], [0.12, 0.41], [0.32, 0.48]])
+    dg8 = np.array([[0., 4.], [4, 8]])
+   
+    # error crit.
+    eps = 1e-7
+
+
+    assert np.linalg.norm(lagrangian_barycenter(pdiagset=[dg1, dg2, dg3, dg4],init=3, verbose=False) - res) < eps
```
mglisse, line 36: In other files, we have used `lagrangian_barycenter(pdiagset=[dg1, dg2, dg3, dg4],init=3, verbose=False) == pytest.approx(res, rel=eps)`.

PR #169, opened 2019-12-06
Fix #107 - Exact alpha complex python version

### LINKED ISSUE #107: AlphaComplex or SimplexTree numeric issues
I have encountered some trouble with numeric stability in AlphaComplex. I attached the minimal code example below. Briefly what I did: I generate a set of points (sin(x), sin(x + PI)), which is the same as (sin(x), -sin(x)). However, when I build AlphaComplex and then a simplex tree, I get out different filtrations for these two embeddings. In case of sin(x + PI), 3-simplices appear, which should not be there as all the points are on the same line. Is there something I am doing wrong?

Minimal code example:
```import gudhi
from gudhi import AlphaComplex, SimplexTree

import math
import numpy as np
import matplotlib
from matplotlib import pyplot as plt

#generate periodic signal
time = np.arange(0, 10, 1)
signal = [math.sin(x) for x in time]
delta = math.pi
delayed = [math.sin(x + delta) for x in time]

#construct embedding
embedding1 = [[signal[i], -signal[i]] for i in range(len(time))]
embedding2 = [[signal[i], delayed[i]] for i in range(len(time))]

fmt = 'Point: %2d | X1: %.16f | Y1: %.16f | X2: %.16f | Y2: %.16f'
print('Printing points | Embedding 1 | Embedding 2')
for i in range(len(time)):
    print(fmt % (i, embedding1[i][0], embedding1[i][1], embedding2[i][0], embedding2[i][1]))

#build alpha complex and simplex tree
alpha_complex1 = AlphaComplex(points=embedding1)
simplex_tree1 = alpha_complex1.create_simplex_tree()

alpha_complex2 = AlphaComplex(points=embedding2)
simplex_tree2 = alpha_complex2.create_simplex_tree()

fmt = '%s -> %f'
print('Filtration for embedding1:')
for filtered_value in simplex_tree1.get_filtration():
    print(fmt % tuple(filtered_value))

print('Filtration for embedding2:')
for filtered_value in simplex_tree2.get_filtration():
    print(fmt % tuple(filtered_value))

diag1 = simplex_tree1.persistence()
diag2 = simplex_tree2.persistence()

pplot1 = gudhi.plot_persistence_diagram(diag1, max_intervals=10, legend=True)
pplot1.show()
pplot2 = gudhi.plot_persistence_diagram(diag2, max_intervals=10, legend=True)
pplot2.show()
```

VincentRouvreau:
Fix #107 
Use Epeck_d instead of Epick_d by default for Alpha complex.
Add exact_version mechanism for Alpha complex in simplex tree creation.
Add MPFR dependency for python module compilation.


mglisse:
This gives a choice between *safe* and *exact*. I understand that adding the *fast* choice is more complicated. But I would like to know what the interface (no need for the actual implementation yet) for choosing *fast* will look like before I validate the interface to choose *exact*... Otherwise, let's just provide *safe* as the only possibility and delay the other alternatives until we have a nice API for that choice.

mglisse:
Doesn't `find_package(CGAL)` already do `find_package(MPFR)`? Here, with just `find_package(CGAL)` and `target_link_libraries(myexe CGAL::CGAL)`, myexe is linked with mpfr and gmp. And `message(${MPFR_LIBRARIES})` prints something. Since we only use MPFR as a dependency for CGAL, I am not sure we really need to handle it explicitly. Or maybe it is easier for now to be explicit, and as soon as we modernize our cmake scripts it can be removed?
(by the way, when MPFR or GMP is missing, CGAL is supposed to disable both and use some slower alternative, although there are bugs, and it is ok if for now we are stricter and require gmp+mpfr as dependencies for CGAL)

VincentRouvreau:
> Doesn't `find_package(CGAL)` already do `find_package(MPFR)`? Here, with just `find_package(CGAL)` and `target_link_libraries(myexe CGAL::CGAL)`, myexe is linked with mpfr and gmp. And `message(${MPFR_LIBRARIES})` prints something. Since we only use MPFR as a dependency for CGAL, I am not sure we really need to handle it explicitly. Or maybe it is easier for now to be explicit, and as soon as we modernize our cmake scripts it can be removed?
> (by the way, when MPFR or GMP is missing, CGAL is supposed to disable both and use some slower alternative, although there are bugs, and it is ok if for now we are stricter and require gmp+mpfr as dependencies for CGAL)

Oh yes, you are completely right. After testing it, I rolled back FindMPFR mechanism on 0474f6a62622608b96eac3a553a081e148cbcabc

VincentRouvreau:
> This gives a choice between _safe_ and _exact_. I understand that adding the _fast_ choice is more complicated. But I would like to know what the interface (no need for the actual implementation yet) for choosing _fast_ will look like before I validate the interface to choose _exact_... Otherwise, let's just provide _safe_ as the only possibility and delay the other alternatives until we have a nice API for that choice.

Do you want this PR to do only the _safe_ mode and to propose another PR to see what could be the interfaces ? Or do I propose something on this PR ?

mglisse:
> Do you want this PR to do only the _safe_ mode and to propose another PR to see what could be the interfaces ? Or do I propose something on this PR ?

Doing only *safe* is one possibility.
Note that "propose something" doesn't need to be in the form of a patch, it can be a 1-line comment showing the new arguments for some function, or whatever text is sufficient to understand what it will look like for a user. I don't care about the internal details or what the content of functions will be, only how a user will call them.

VincentRouvreau:
> Doing only _safe_ is one possibility.

Ok let's go this way. I rolled back exact version

> Note that "propose something" doesn't need to be in the form of a patch, it can be a 1-line comment showing the new arguments for some function, or whatever text is sufficient to understand what it will look like for a user. I don't care about the internal details or what the content of functions will be, only how a user will call them.

I will do it on a separate PR, it will be clearer and bug fix is the priority.

VincentRouvreau:
@jasnaurbancic I added the code you provide in the issue description to test the proposed fix.

PR #173, opened 2019-12-09
Example in doc + fix on Landscape

### LINKED ISSUE #166: [python representations] Landscapes null entries
When we do not specify the bounds for a landscape, it automatically picks the min and max of the births and deaths. This may be right for some other classes. But for landscapes, the value at min or at max is always 0, so this creates 2*num_landscapes coordinates that are always 0.

Some possibilities would be:
* pretend we asked for resolution+2, and drop the extremes
* use half-values: instead of "min, min+t, min+2*t, ..., max", use "min+t/2, min+3*t/2, ... max-t/2".

MathieuCarriere:
I added an inline example as suggested by Marc in the doc of representations, and I fixed the Landscape by discarding the coordinates that are always 0 in Landscapes
Fix #166


src/python/doc/representations.rst
```diff
@@ -8,9 +8,9 @@ Representations manual
...
+This module, originally available at https://github.com/MathieuCarriere/sklearn-tda and named sklearn_tda, aims at bridging the gap between persistence diagrams and machine learning, by providing implementations of most of the vector representations for persistence diagrams in the literature, in a scikit-learn format. More specifically, it provides tools, using the scikit-learn standard interface, to compute distances and kernels on persistence diagrams, and to convert these diagrams into vectors in Euclidean space.
 
-A diagram is represented as a numpy array of shape (n,2), as can be obtained from :func:`~gudhi.SimplexTree.persistence_intervals_in_dimension` for instance. Points at infinity are represented as a numpy array of shape (n,1), storing only the birth time.
+A diagram is represented as a numpy array of shape (n,2), as can be obtained from `SimplexTree.persistence_intervals_in_dimension` for instance. Points at infinity are represented as a numpy array of shape (n,1), storing only the birth time.
```
mglisse: Can you explain why this change? The previous version showed the text "persistence_intervals_in_dimension" and it was clickable and led to the doc for that function in SimplexTree.

src/python/example/diagram_vectorizations_distances_kernels.py
```diff
@@ -26,9 +26,9 @@
 
 LS = Landscape(resolution=1000)
 L = LS.fit_transform(diags)
-plt.plot(L[0][:1000])
-plt.plot(L[0][1000:2000])
-plt.plot(L[0][2000:3000])
+plt.plot(L[0][:999])
+plt.plot(L[0][999:2*999])
```
mglisse: why not `999:1999`?

src/python/doc/representations.rst
```diff
@@ -46,3 +46,24 @@ Metrics
...
+    D = np.array([[0.,4.],[1.,2.],[3.,8.],[6.,8.]])
+    diags = [D]
+    l=Landscape(num_landscapes=2,resolution=10).fit_transform(diags)
+    print(l) 
+
+The output is:
+
+.. testoutput::
+
+    [[0.         1.25707872 2.51415744 1.88561808 0.7856742  2.04275292 3.29983165 2.51415744 1.25707872 0.         0.         0.         0.31426968 0.         0.62853936 0.         0.         0.31426968 1.25707872 0.        ]]
```
mglisse: Did you write that before removing the always-0 components? (the test fails)

src/python/doc/representations.rst
MathieuCarriere: My bad, I messed up something and did not notice. The text should be clickable again now.

src/python/example/diagram_vectorizations_distances_kernels.py
MathieuCarriere: Actually, it should be `998` since the first **and** last coordinates are removed. Fixed it.

src/python/doc/representations.rst
MathieuCarriere: Indeed, I forgot to update the example. Should be good now.

src/python/doc/representations.rst
```diff
@@ -46,3 +46,24 @@ Metrics
...
+    D = np.array([[0.,4.],[1.,2.],[3.,8.],[6.,8.]])
+    diags = [D]
+    l=Landscape(num_landscapes=2,resolution=10).fit_transform(diags)
+    print(l) 
+
+The output is:
+
+.. testoutput::
+
+    [[1.02851895 2.05703791 2.57129739 1.54277843 0.89995409 1.92847304 2.95699199 3.08555686 0.         0.64282435 0.         0.         0.51425948 0.         0.         0.        ]]
```
VincentRouvreau: On your development branch, the output I have is:
```python
[[0.         1.25707872 2.51415744 1.88561808 0.7856742  2.04275292
  3.29983165 2.51415744 1.25707872 0.         0.         0.
  0.31426968 0.         0.62853936 0.         0.         0.31426968
  1.25707872 0.        ]]
```
This may explain why `sphinx_py_test` fails in the test suite.

VincentRouvreau: On your development branch, `sphinx_py_test` fails in the test suite because sphinx `..testouput::` is identation sensitive.

```suggestion
    [[1.02851895 2.05703791 2.57129739 1.54277843 0.89995409 1.92847304
      2.95699199 3.08555686 0.         0.64282435 0.         0.
      0.51425948 0.         0.         0.        ]]
```

src/python/example/diagram_vectorizations_distances_kernels.py
mglisse: Shouldn't your new `+ self.nan_in_range.sum()` mean that `resolution` always matches the number of bins, even after removing the extremes?

src/python/doc/representations.rst
MathieuCarriere: You are totally right, I forgot to update the landscape resolution in some parts of the code. Hopefully it is good now.

src/python/example/diagram_vectorizations_distances_kernels.py
MathieuCarriere: You are totally right, I forgot to update the landscape resolution in some parts of the code. Hopefully it is good now.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -104,10 +104,11 @@ def fit(self, X, y=None):
             X (list of n x 2 numpy arrays): input persistence diagrams.
             y (n x 1 array): persistence diagram labels (unused).
         """
+        self.nan_in_range = np.isnan(np.array(self.sample_range))
         if np.isnan(np.array(self.sample_range)).any():
```
mglisse: Shouldn't this use the new variable defined on the line just above?

MathieuCarriere: Indeed. Should be fixed now.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -121,26 +122,26 @@ def transform(self, X):
...
                 [px,py] = diagram[j,:2]
-                min_idx = np.clip(np.ceil((px          - self.sample_range[0]) / step_x).astype(int), 0, self.resolution)
-                mid_idx = np.clip(np.ceil((0.5*(py+px) - self.sample_range[0]) / step_x).astype(int), 0, self.resolution)
-                max_idx = np.clip(np.ceil((py          - self.sample_range[0]) / step_x).astype(int), 0, self.resolution)
+                min_idx = np.clip(np.ceil((px          - self.sample_range[0]) / step_x).astype(int), 0, self.resolution + self.nan_in_range.sum())
+                mid_idx = np.clip(np.ceil((0.5*(py+px) - self.sample_range[0]) / step_x).astype(int), 0, self.resolution + self.nan_in_range.sum())
+                max_idx = np.clip(np.ceil((py          - self.sample_range[0]) / step_x).astype(int), 0, self.resolution + self.nan_in_range.sum())
 
-                if min_idx < self.resolution and max_idx > 0:
+                if min_idx < self.resolution + self.nan_in_range.sum() and max_idx > 0:
```
mglisse: If we are going to use `self.resolution + self.nan_in_range.sum()` everywhere, maybe we should store that value in self.resolution instead? Or provide an accessor? Not a prerequisite to the merge though.

MathieuCarriere: agreed. just cleaned the code.

PR #176, opened 2019-12-13
Add a more representative image for python persistence representations

VincentRouvreau:


src/python/doc/representations_sum.inc
```diff
@@ -3,7 +3,7 @@
 
    +------------------------------------------------------------------+----------------------------------------------------------------+-----------------------------------------------+
    | .. figure::                                                      | Vectorizations, distances and kernels that work on persistence | :Author: Mathieu Carrière                     |
-   |      ../../doc/Persistence_representations/average_landscape.png | diagrams, compatible with scikit-learn.                        |                                               |
+   |      ../../doc/Persistence_representations/sklearn-tda.png       | diagrams, compatible with scikit-learn.                        |                                               |
```
mglisse: I see that Wasserstein is also hijacking some C++ package to store its image, so it is ok, but we seem to have a directory python/doc/img that could also be used more in the future.

PR #178, opened 2019-12-16
Fix #158 - Add numpy ndarray to init a cubical and a periodic one

### LINKED ISSUE #158: [python - cubical] numpy array as input
`CubicalComplex.__init__` currently takes a 1d vector of values, and a shape (list of dimensions) explaining how to interpret it. It would be very natural, and less error-prone, to accept a numpy array. It could then internally deduce the shape and 1d vector (flat, ravel, reshape...) that the C++ version expects.

VincentRouvreau:
Fix #158

### DIFF
--- a/src/python/gudhi/cubical_complex.pyx
+++ b/src/python/gudhi/cubical_complex.pyx
@@ -77,8 +77,8 @@ cdef class CubicalComplex:
         elif ((dimensions is None) and (top_dimensional_cells is not None)
             and (perseus_file == '')):
             if isinstance(top_dimensional_cells, np.ndarray):
-                dimensions = list(top_dimensional_cells.shape)
-                top_dimensional_cells = top_dimensional_cells.flatten().tolist()
+                dimensions = top_dimensional_cells.shape
+                top_dimensional_cells = top_dimensional_cells.ravel(order='C')
                 self.thisptr = new Bitmap_cubical_complex_base_interface(dimensions, top_dimensional_cells)
             else:
                 print("top_dimensional_cells is not an instance of ndarray. It is a " + type(top_dimensional_cells))
--- a/src/python/gudhi/cubical_complex.pyx
+++ b/src/python/gudhi/cubical_complex.pyx
@@ -5,7 +5,7 @@ from libcpp.string cimport string
 from libcpp cimport bool
 import os
 
-from numpy import array as np_array
+import numpy as np
 
 # This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
 # See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
@@ -47,7 +47,7 @@ cdef class CubicalComplex:
 
     # Fake constructor that does nothing but documenting the constructor
     def __init__(self, dimensions=None, top_dimensional_cells=None,
-                  perseus_file=''):
+                 numpy_array=None, perseus_file=''):
         """CubicalComplex constructor from dimensions and
         top_dimensional_cells or from a Perseus-style file name.
 
@@ -58,16 +58,31 @@ cdef class CubicalComplex:
 
         Or
 
+        :param numpy_array: Filtration values in a d-array.
+        :type numpy_array: numpy ndarray
+
+        Or
+
         :param perseus_file: A Perseus-style file name.
         :type perseus_file: string
         """
 
     # The real cython constructor
     def __cinit__(self, dimensions=None, top_dimensional_cells=None,
-                  perseus_file=''):
-        if (dimensions is not None) and (top_dimensional_cells is not None) and (perseus_file == ''):
+                  numpy_array=None, perseus_file=''):
+        if ((dimensions is not None) and (top_dimensional_cells is not None)
+            and (numpy_array is None) and (perseus_file == '')):
             self.thisptr = new Bitmap_cubical_complex_base_interface(dimensions, top_dimensional_cells)
-        elif (dimensions is None) and (top_dimensional_cells is None) and (perseus_file != ''):
+        elif ((dimensions is None) and (top_dimensional_cells is None)
+            and (numpy_array is not None) and (perseus_file == '')):
+            if isinstance(numpy_array, np.ndarray):
+                dimensions = list(numpy_array.shape)
+                top_dimensional_cells = numpy_array.flatten().tolist()
+                self.thisptr = new Bitmap_cubical_complex_base_interface(dimensions, top_dimensional_cells)
+            else:
+                print("numpy_array is not an instance of ndarray. It is a " + type(numpy_array))
+        elif ((dimensions is None) and (top_dimensional_cells is None)
+            and (numpy_array is None) and (perseus_file != '')):
             if os.path.isfile(perseus_file):
                 self.thisptr = new Bitmap_cubical_complex_base_interface(str.encode(perseus_file))
             else:
@@ -184,4 +199,4 @@ cdef class CubicalComplex:
         else:
             print("intervals_in_dim function requires persistence function"
                   " to be launched first.")
-        return np_array(intervals_result)
+        return np.array(intervals_result)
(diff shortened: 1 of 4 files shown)


src/python/gudhi/cubical_complex.pyx
```diff
@@ -47,7 +47,7 @@ cdef class CubicalComplex:
 
     # Fake constructor that does nothing but documenting the constructor
     def __init__(self, dimensions=None, top_dimensional_cells=None,
-                  perseus_file=''):
+                 numpy_array=None, perseus_file=''):
```
mglisse: I don't like the name `numpy_array`. But it is true that top_dimensional_cells is taken, and we want something that plays the same role. How about this: we keep the same parameters as before, but if `dimensions` is None, top_dimensional_cells is expected to be a numpy array and we deduce the dimensions from its shape. Does that make sense?

src/python/gudhi/periodic_cubical_complex.pyx
```diff
@@ -60,16 +60,41 @@ cdef class PeriodicCubicalComplex:
...
+            and (perseus_file == '')):
+            self.thisptr = new Periodic_cubical_complex_base_interface(dimensions,
+                                                                       top_dimensional_cells,
+                                                                       periodic_dimensions)
+        elif ((dimensions is None) and (top_dimensional_cells is None)
+            and (numpy_array is not None) and (periodic_dimensions is not None)
+            and (perseus_file == '')):
+            if isinstance(numpy_array, np.ndarray):
+                dimensions = list(numpy_array.shape)
+                top_dimensional_cells = numpy_array.flatten().tolist()
```
mglisse: 1. flatten always makes a copy, while some variants make a copy only if needed (flat, ravel, etc)
2. do we really need to convert to a list? I hoped cython would know how to convert a numpy array to a vector, just like it knows how to convert a list.

src/python/gudhi/cubical_complex.pyx
VincentRouvreau: I am ok with your proposal. I will try to make it clear in the documentation.

VincentRouvreau: I am ok with your proposition.
The fix is done on b63be266effe24646823acc59fe397021bb13a3e

src/python/gudhi/periodic_cubical_complex.pyx
VincentRouvreau: You are right. Fixed on 32a9c5e4df26e59f5f376e54c05c7bec92012cff
I used [numpy.ravel('C')](https://docs.scipy.org/doc/numpy/reference/generated/numpy.ravel.html#numpy.ravel). I hope this is the good candidate.

mglisse:
> gudhi.CubicalComplex(top_dimensional_cells=np.array([[1]]))
> CubicalComplex can be constructed from dimensions and top_dimensional_cells or from a Perseus-style file name.
I thought that's how it was supposed to work?

VincentRouvreau:
> I thought that's how it was supposed to work? Did you only change the periodic version?

What version did you use ? The former version was requiring top_dimensions_cells and dimensions. Otherwise you was having this message.
With the version from this PR:
```
import gudhi
import numpy as np                                                      

gudhi.CubicalComplex(top_dimensional_cells=np.array([[1]]))             
<gudhi.cubical_complex.CubicalComplex at 0x7fe589eda410>

cub = gudhi.CubicalComplex(top_dimensional_cells=np.array([[1]]))       

cub.dimension()                                                         
2

cub.num_simplices()                                                     
9
```

mglisse:
I tried (using the periodic version with False for all dimensions) `a=np.array([[3,3],[2,2],[4,4]])` and it gave me a persistence of
> [(0, (2.0, inf)), (0, (2.0, 3.0))]

But in a, the sublevel sets are always connected. This looks like the persistence of `a.reshape(2,3)`.
CubicalComplex uses an unusual indexing. I think to make it work, we need to invert the order of the dimensions that we read from the shape of the array. Or possibly use 'F' instead of 'C', I didn't check if that really matches what the code expects.

mglisse:
> What version did you use ?

Your current branch. Ah, but I had some outdated .cpp files next to the .pyx ones, which probably prevented cython from regenerating them :disappointed: I removed them and it works now.

VincentRouvreau:
> I tried (using the periodic version with False for all dimensions) `a=np.array([[3,3],[2,2],[4,4]])` and it gave me a persistence of
> 
> > [(0, (2.0, inf)), (0, (2.0, 3.0))]
> 
> But in a, the sublevel sets are always connected. This looks like the persistence of `a.reshape(2,3)`.
> CubicalComplex uses an unusual indexing. I think to make it work, we need to invert the order of the dimensions that we read from the shape of the array. Or possibly use 'F' instead of 'C', I didn't check if that really matches what the code expects.

This is because:
```python
import numpy as np
a = np.arange(12).reshape(4,3); a
  array([[ 0,  1,  2],
         [ 3,  4,  5],
         [ 6,  7,  8],
         [ 9, 10, 11]])
a.shape
  (4, 3)
a.ravel('C')
  array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11])
a.ravel('F')
  array([ 0,  3,  6,  9,  1,  4,  7, 10,  2,  5,  8, 11])
```

For the moment, we have no python interface for the different iterators (top dimensions cells, boundary  iterators...). Is there any need ?
By the way, I am proposing to go with `ravel('F')`. Are you ok with that ?

mglisse:
>  By the way, I am proposing to go with `ravel('F')`. Are you ok with that ?

Yes. We can always go back to 'C' and use `reversed(dimensions)` or `dimensions[::-1]` later if that seems better.

mglisse:
> For the moment, we have no python interface for the different iterators (top dimensions cells, boundary iterators...). Is there any need ?

Possibly at some point, and then indeed it will be important to know how we store the array. Rather that iterators, we'd likely want to be able to access the value at coordinates (x,y,z), maybe make the internal vector of cubical complex appear like a numpy array. But I think it can wait.

VincentRouvreau:
> Yes. We can always go back to 'C' and use `reversed(dimensions)` or `dimensions[::-1]` later if that seems better.

Ok I did `ravel('F'')` on ba406558f707f55b638be92d197ae3039595b29a and added some tests to be sure we won't break it.

src/python/gudhi/cubical_complex.pyx
```diff
@@ -58,16 +58,32 @@ cdef class CubicalComplex:
...
     def __cinit__(self, dimensions=None, top_dimensional_cells=None,
                   perseus_file=''):
-        if (dimensions is not None) and (top_dimensional_cells is not None) and (perseus_file == ''):
+        if ((dimensions is not None) and (top_dimensional_cells is not None)
+            and (perseus_file == '')):
             self.thisptr = new Bitmap_cubical_complex_base_interface(dimensions, top_dimensional_cells)
-        elif (dimensions is None) and (top_dimensional_cells is None) and (perseus_file != ''):
+        elif ((dimensions is None) and (top_dimensional_cells is not None)
+            and (perseus_file == '')):
+            if isinstance(top_dimensional_cells, np.ndarray):
```
mglisse: Why so strict? I think it would be convenient to accept lists in the example in cubical_complex_user.rst. `top_dimensional_cells=numpy.array(top_dimensional_cells,copy=false,order='F')` maybe?

VincentRouvreau: Yes you are right. I did the fix you propose.

src/python/gudhi/cubical_complex.pyx
```diff
@@ -60,7 +60,7 @@ cdef class CubicalComplex:
 
         :param top_dimensional_cells: A multidimensional array of cells
             filtration values.
-        :type top_dimensional_cells: numpy ndarray
+        :type top_dimensional_cells: numpy ndarray or list of list of double
```
mglisse: Aren't they called `float` in python? I don't know what the standard python way is to document "or anything convertible to it". I would have kept just "numpy array", because list of double and list of list of list of double and list of array, etc will all work, but I guess having list of list of float as an example doesn't hurt.

Issue #2, opened 2019-03-07
[Cython - simplex_tree] There is no simplex iterators


VincentRouvreau:
We have to use get_filtration() to find the list of simplices.
There shall be an simplex iterator (cf.itertools)

mglisse:
Let me add some details here. The simplex tree data structure was invented to save space. It has size n (number of simplices), while more common representations have size n*d (d is the dimension). If at any point we create a vector of vectors of vertices to represent all the simplices, we have killed the whole point of the simplex tree and should use a different complex instead (which we sadly do not have currently). However, creating such a vector is not necessary to iterate on the simplices. In C++, we have `Complex_simplex_range`/`Complex_simplex_iterator`, and the idea would be to interface that in python. Similarly, get_filtration could wrap iterators on the internal filtration vector, and only build the tuple (simplex, filtration_value) when the iteration reaches that element.
Then come the details, can this be implemented as a generator or does it need something more manual, etc.

yagu0:
It seems that the "vector of vectors of vertices" is the current situation in the C++ functions get_filtration(), get_skeleton(), get_star() and get_cofaces(). So when Python get the result and build an iterator, it's already too late. I like the idea of directly interfacing a C++ iterator, I'll have a look in this direction.

mglisse:
> It seems that the "vector of vectors of vertices" is the current situation in the C++ functions get_filtration(), get_skeleton(), get_star() and get_cofaces(). So when Python get the result and build an iterator, it's already too late.

Yes. We did it that way in the beginning because it was a quick and easy way to have the functionality, and haven't had time to revisit since then.

VincentRouvreau:
When I wrote this issue, I was thinking resolving it by using `yield` on the simplex iterator. I am not sure if this is the correct way to do it, or even if it works, but it seems promising.

PR #185, opened 2020-01-03
#2: add a simplex iterator in Python

yagu0:
Attempt to fix the issue #2 

NOTE: get_filtration() and the introduced method get_unprocessed_filtration() return the same thing:

> f = st.get_filtration()
> g = st.get_unprocessed_filtration()
> f == g
True

So I think get_filtration() could be replaced by the simpler code of the other - temporary, internal - function. The next() method in iterator class could also be simplified (loop "for vertex in filtered_complex[0] ...").

### DIFF
--- a/src/python/gudhi/simplex_tree.pxd
+++ b/src/python/gudhi/simplex_tree.pxd
@@ -18,6 +18,8 @@ __author__ = "Vincent Rouvreau"
 __copyright__ = "Copyright (C) 2016 Inria"
 __license__ = "MIT"
 
+# TODO: add cdef for SimplexTreeIterator ...
+
 cdef extern from "Simplex_tree_interface.h" namespace "Gudhi":
     cdef cppclass Simplex_tree_options_full_featured:
         pass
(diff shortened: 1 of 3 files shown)


mglisse:
> I think get_filtration() could be replaced by the simpler code

Indeed, it looks like `return self.get_ptr().get_filtration()` is sufficient, cython handles the conversion.

We could also return a wrapper to std::vector which avoids doing a conversion. That would happen automatically if we were using pybind11, while in cython it requires writing the wrapper by hand IIUC. However, that doesn't seem worth the trouble, since that's not the final result we want (see the precisions I am going to add to issue #2).

mglisse:
The global map is a clever way to handle nested loops, though it isn't as good for multithreading. But I really don't think it is necessary. I am sorry I haven't had time to look at it yet, I hope I'll find some time...

VincentRouvreau:
While reviewing your code, I tried to do it with yield on PR #211 .
Tell me what you think about this approach.
It allows to remove the global map as @mglisse was talking about.

VincentRouvreau:
I close this PR as we all agree that we will go with #211

PR #189, opened 2020-01-04
Link the doc for subsampling & point file readers, some reorg

### LINKED ISSUE #157: [python - subsampling] Missing from the sphinx documentation
Functions like `sparsify_point_set` have docstrings, but nothing references them so sphinx skips them and they do not appear in the documentation. I noticed it while trying to fix a link to it in the doc for Rips complex.

mglisse:
Some doc wasn't linked from anywhere. Not everything is linked yet, but I didn't want to change everything at the same time.
I moved a few things around on the front-page because it made more sense to me.
I changed `vector[vector[double]]` to `list(list(float))` because when I try it, that's what I get, cython automatically converts. The spelling `list(list(float))` is questionable though. Looking at PEP484 (python 3.5+, but we can just use this syntax to describe types in strings if that's too recent, which doesn't require anything), we could use type annotations like `List[List[float]]` for the return type (or `Iterable[Iterable[float]]` for parameters), and possibly the extension sphinx-autodoc-typehints.

Fix #157.


mglisse:
This could also be a chance to document them as `gudhi.subsampling.*` instead of just `gudhi.*`, if we want.

mglisse:
Part of the idea is that we can't keep adding one huge block in the index page for each small functionality that we add, it does not scale. For instance, I am thinking (for a separate PR) of splitting representations into a few items (maybe metrics, kernels and vectorizations) and move bottleneck/wasserstein away from the front page, into metrics). I think that will make the doc easier to navigate. However, it doesn't fit so well with the fields "author, introduced in, copyright, requires" that we have currently, since it puts together things by different authors, introduced at different times (already a problem when we add functionality to an existing package), with different requirements and thus different (indirect) copyright.

src/python/doc/index.rst
```diff
@@ -37,16 +45,10 @@ Cover complexes
...
-Data structures
-===============
-
-Simplex tree
-------------
+Tangential complex
+==================
 
-.. include:: simplex_tree_sum.inc
+.. include:: tangential_complex_sum.inc
```
VincentRouvreau: It's ok for me to change the modules list, but we have to move the C++ one accordingly and also the [web site introduction](https://gudhi.inria.fr/introduction/).

src/python/doc/point_cloud.rst
```diff
@@ -0,0 +1,22 @@
...
+
+.. autofunction:: gudhi.read_lower_triangular_matrix_from_csv_file
+
+Subsampling
+-----------
+
+.. automodule:: gudhi.subsampling
+   :members:
+   :special-members:
+   :show-inheritance:
```
VincentRouvreau: It is ok for me. We could also imagine rewording [file formats doc](https://gudhi.inria.fr/python/latest/fileformats.html) to add the reader functions here.

PR #192, opened 2020-01-08
Wdist

tlacombe:
Change in variable names : ground metric parameter, formerly named `q`, is now `internal_p`, and wasserstein exponent, formerly named `p`, is now `q`. This aims to be consistent with Hera and following implementations.

### DIFF
--- a/src/python/gudhi/wasserstein.py
+++ b/src/python/gudhi/wasserstein.py
@@ -23,26 +23,26 @@ def _proj_on_diag(X):
     return np.array([Z , Z]).T
 
 
-def _build_dist_matrix(X, Y, p=2., q=2.):
+def _build_dist_matrix(X, Y, q=2., internal_p=2.):
     '''
     :param X: (n x 2) numpy.array encoding the (points of the) first diagram.
     :param Y: (m x 2) numpy.array encoding the second diagram.
-    :param q: Ground metric (i.e. norm l_q).
-    :param p: exponent for the Wasserstein metric.
+    :param internal_p: Ground metric (i.e. norm l_q).
+    :param q: exponent for the Wasserstein metric.
     :returns: (n+1) x (m+1) np.array encoding the cost matrix C. 
                 For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) and its orthogonal proj onto the diagonal.
                 note also that C[n+1, m+1] = 0  (it costs nothing to move from the diagonal to the diagonal).
     '''
     Xdiag = _proj_on_diag(X)
     Ydiag = _proj_on_diag(Y)
-    if np.isinf(q):
-        C = sc.cdist(X,Y, metric='chebyshev')**p
-        Cxd = np.linalg.norm(X - Xdiag, ord=q, axis=1)**p
-        Cdy = np.linalg.norm(Y - Ydiag, ord=q, axis=1)**p
+    if np.isinf(internal_p):
+        C = sc.cdist(X,Y, metric='chebyshev')**q
+        Cxd = np.linalg.norm(X - Xdiag, ord=internal_p, axis=1)**q
+        Cdy = np.linalg.norm(Y - Ydiag, ord=internal_p, axis=1)**q
     else:
-        C = sc.cdist(X,Y, metric='minkowski', p=q)**p
-        Cxd = np.linalg.norm(X - Xdiag, ord=q, axis=1)**p
-        Cdy = np.linalg.norm(Y - Ydiag, ord=q, axis=1)**p
+        C = sc.cdist(X,Y, metric='minkowski', p=internal_p)**q
+        Cxd = np.linalg.norm(X - Xdiag, ord=internal_p, axis=1)**q
+        Cdy = np.linalg.norm(Y - Ydiag, ord=internal_p, axis=1)**q
     Cf = np.hstack((C, Cxd[:,None]))
     Cdy = np.append(Cdy, 0)
 
@@ -51,24 +51,24 @@ def _build_dist_matrix(X, Y, p=2., q=2.):
     return Cf
 
 
-def _perstot(X, p, q):
+def _perstot(X, q, internal_p):
     '''
     :param X: (n x 2) numpy.array (points of a given diagram).
-    :param q: Ground metric on the (upper-half) plane (i.e. norm l_q in R^2); Default value is 2 (Euclidean norm).
-    :param p: exponent for Wasserstein; Default value is 2.
+    :param internal_p: Ground metric on the (upper-half) plane (i.e. norm l_p in R^2); Default value is 2 (Euclidean norm).
+    :param q: exponent for Wasserstein; Default value is 2.
     :returns: float, the total persistence of the diagram (that is, its distance to the empty diagram).    
     '''
     Xdiag = _proj_on_diag(X)
-    return (np.sum(np.linalg.norm(X - Xdiag, ord=q, axis=1)**p))**(1./p)
+    return (np.sum(np.linalg.norm(X - Xdiag, ord=internal_p, axis=1)**q))**(1./q)
 
 
-def wasserstein_distance(X, Y, p=2., q=2.):
+def wasserstein_distance(X, Y, q=2., internal_p=2.):
     '''
     :param X: (n x 2) numpy.array encoding the (finite points of the) first diagram. Must not contain essential points (i.e. with infinite coordinate).
     :param Y: (m x 2) numpy.array encoding the second diagram.
-    :param q: Ground metric on the (upper-half) plane (i.e. norm l_q in R^2); Default value is 2 (euclidean norm).
-    :param p: exponent for Wasserstein; Default value is 2.
-    :returns: the p-Wasserstein distance (1 <= p < infinity) with respect to the q-norm as ground metric.
+    :param internal_p: Ground metric on the (upper-half) plane (i.e. norm l_p in R^2); Default value is 2 (euclidean norm).
+    :param q: exponent for Wasserstein; Default value is 2.
+    :returns: the q-Wasserstein distance (1 <= q < infinity) with respect to the internal_p-norm as ground metric.
     :rtype: float
     '''
     n = len(X)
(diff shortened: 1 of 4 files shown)


mglisse:
> consistent with Hera

Maybe we should clarify what part we are trying to be consistent with.
* In #182, I used the names of the C++ library: file basic_defs_ws.h, struct AuctionParams. Those names look like wasserstein_power, delta, internal_p, etc.
* Hera also has an example, with command line arguments named degree or q (each one has a long and a short name), error or d, internal-p or p, etc.
* The python binding in Dionysus2 uses q, delta and internal_p.
* I don't remember what is done in scikit-tda and giotto.

tlacombe:
Regarding scikit-tda, if my understanding is correct, `wasserstein(dgm1, dgm2)` implemented in https://github.com/scikit-tda/persim/blob/master/persim/wasserstein.py does not allow to play with exponent and ground metric. It calls for `scipy.metrics.pairwise_distances`, which returns the Euclidean norm, and then return the sum of a linear assignment (obtained with `optimize.linear_sum_assignment`). So it should be the W_1 metric for L_2 ground norm.

In giotto, I am not sure I read the code properly, https://github.com/giotto-ai/giotto-learn/blob/master/giotto/diagrams/distance.py , but it seems they have a parameter `p` which should be the Wasserstein exponent. Thus I guess it uses the Euclidean metric as a ground metric (I cannot find the `giotto_wasserstein.py` file used in `_metrics.py`...).

So I actually don't know what should be the correct names... Perhaps we should follow Giotto or scikit-tda?

mglisse:
> I cannot find the `giotto_wasserstein.py`

It is here: https://github.com/giotto-ai/giotto-learn/blob/master/giotto/externals/bindings/wasserstein_bindings.cpp (see PYBIND11_MODULE).

mglisse:
> Perhaps we should follow Giotto or scikit-tda?

Since they don't provide the parameters we are interested in, we can hardly follow them...

tlacombe:
> > I cannot find the `giotto_wasserstein.py`
> 
> It is here: https://github.com/giotto-ai/giotto-learn/blob/master/giotto/externals/bindings/wasserstein_bindings.cpp (see PYBIND11_MODULE).

Thanks ! 
If I understand it means that the function `wasserstein_distance` used afterwards has as default parameters "L_infty ground metric" and "delta=0.01" (relative error criterion if I remember correctly) ; according to the doc in `distances.py`, these cannot be changed by the user, and the wasserstein exponent is named `p`.

So if I summarize,
- Hera uses `wasserstein_exponent` or `q`, and `internal_p` or `p`  (and  `error` or `d` ; but we don't have this one)
- Dyonisus2 uses `q` and `internal_p`
- Scikit-tda does not allow to change these parameters
- Giotto call the wasserstein exponent `p` (and does not allow to change the ground norm).

I don't know which decision we should take.

mglisse:
> Hera uses `wasserstein_exponent` or `q`, and `internal_p` or `p` (and `error` or `d` ; but we don't have this one)

Fixing this:
Hera uses wasserstein_power or degree or q, and internal_p or p, and delta or error or d.

The names we are choosing will also apply to #182, so the approximation thing is relevant.

Let's see if Vincent, @martinroyer, @bertrandmichel (or others!) have an opinion on the naming. I think @MathieuCarriere was also using hera at some point?

tlacombe:
> The names we are choosing will also apply to #182, so the approximation thing is relevant.

Ok I get it. 

As a side consideration, since Hera's implementation of (estimated) Wasserstein distance will be included, what should be the final product for the user? A single function `wasserstein_distance` that would call Hera if `delta > 0` and POT if `delta=0` ? Or something depending on the problem parameters? (say, heuristically uses Hera or POT depending on the size of dgm?) Or simply two different functions?
I don't know if this thread is the correct place to discuss this though.

mglisse:
> what should be the final product for the user? A single function `wasserstein_distance` that would call Hera if `delta > 0` and POT if `delta=0` ?

If we have only the 2 current codes, that looks like a nice possibility (the doc would have to be super clear). You were also talking about adding some regularized versions, which could complicate things. Although if it is hard to give good guarantees about the error between Wasserstein and the regularized version, maybe the regularized version should have a different name (or a different parameter regularization=gamma or something).

> Or something depending on the problem parameters? (say, heuristically uses Hera or POT depending on the size of dgm?) Or simply two different functions?

Those all look like sensible options... Also use Hera if POT is not installed.

We are going to want a gudhi.representations.metrics.WassersteinDistance, it could be the right place to select "automatically" between implementations.

tlacombe:
> You were also talking about adding some regularized versions, which could complicate things. Although if it is hard to give good guarantees about the error between Wasserstein and the regularized version, maybe the regularized version should have a different name (or a different parameter regularization=gamma or something).

I would somewhat advocate for using a different function for the regularized version (as done in POT for instance) as I feel it is a quantity of interest on its own, and also despite it theoretically converges to the exact distance when gamma-->0 (with known bounds), one can have numerical instabilities, thus I think it is safer to treat that as a different function. 

> Also use Hera if POT is not installed.

Indeed, that's a nice point.

MathieuCarriere:
@mglisse indeed I was using `hera` in the first release of `sklearn_tda`. If it simplifies integration, I can pull out the code.

MathieuCarriere:
I basically implemented a Python wrapper for `hera`'s C++ code.

mglisse:
> @mglisse indeed I was using `hera` in the first release of `sklearn_tda`.

We should soon have 2 implementations of Wasserstein in gudhi (pot+hera), so we will be able to dust off that code. But for now, I was only wondering if you (or anyone else) had an opinion on the names of the parameters for Wasserstein: the `p` in W<sup>p</sup> and the `p` in L<sup>p</sup> (the metric on ℝ²) cannot have the same name, and this PR lists the names used elsewhere.

mglisse:
> I basically implemented a Python wrapper for `hera`'s C++ code.

Maybe I should have looked at it before doing #182...

MathieuCarriere:
Actually I didn't include this code in Gudhi, it was only in one of the first releases of `sklearn_tda` when it was a separate GitHub project.

MathieuCarriere:
And I removed it because I wanted something entirely based on Python

MathieuCarriere:
But personally I like the `hera` terminology with `wasserstein_exponent` and `internal_p`

mglisse:
> But personally I like the `hera` terminology with `wasserstein_exponent` and `internal_p`

There seems to some consensus on `internal_p`, let's say this one is settled.

For the other one, Hera uses `wasserstein_power`, `degree` or `q`, not `wasserstein_exponent` (but we can still invent that name if we want). If noone complains, I'm going to decide that we use `wasserstein_power`. Theo, can I ask you to change `q` one last time?

tlacombe:
> > But personally I like the `hera` terminology with `wasserstein_exponent` and `internal_p`
> 
> There seems to some consensus on `internal_p`, let's say this one is settled.
> 
> For the other one, Hera uses `wasserstein_power`, `degree` or `q`, not `wasserstein_exponent` (but we can still invent that name if we want). If noone complains, I'm going to decide that we use `wasserstein_power`. Theo, can I ask you to change `q` one last time?

I'm ok with "wasserstein_power" (although "exponent" would sound great also imo). Sure I'll change it (probably tomorrow morning). I will also update barycenter.py #167 to be consistent with the notations (although for now, the current algorithm only makes sense for wasserstein_power=internal_p=2).

mglisse:
> I'm ok with "wasserstein_power" (although "exponent" would sound great also imo).

Ok, new plan, let's go with either `exponent` or `order`, whichever you prefer. `wasserstein_power` is now out of favor. I reserve the right to change my mind every 5 minutes or so :wink:

src/python/gudhi/wasserstein.py
```diff
@@ -51,24 +51,24 @@ def _build_dist_matrix(X, Y, p=2., q=2.):
...
+def wasserstein_distance(X, Y, order=2., internal_p=2.):
     '''
     :param X: (n x 2) numpy.array encoding the (finite points of the) first diagram. Must not contain essential points (i.e. with infinite coordinate).
     :param Y: (m x 2) numpy.array encoding the second diagram.
-    :param q: Ground metric on the (upper-half) plane (i.e. norm l_q in R^2); Default value is 2 (euclidean norm).
-    :param p: exponent for Wasserstein; Default value is 2.
-    :returns: the p-Wasserstein distance (1 <= p < infinity) with respect to the q-norm as ground metric.
+    :param internal_p: Ground metric on the (upper-half) plane (i.e. norm l_p in R^2); Default value is 2 (euclidean norm).
+    :param order: exponent for Wasserstein; Default value is 2.
+    :returns: the q-Wasserstein distance (1 <= q < infinity) with respect to the internal_p-norm as ground metric.
```
mglisse: We call it `order` in some places and `q` in others, maybe we could clarify a bit. "the Wasserstein distance of order q (1 <= q < infinity)"?

tlacombe: Hum sorry, it's something I just missed. I will commit in few minutes.

tlacombe: Nice suggestion btw, as `order-Wasserstein distance` sounds sloppy ^^

src/python/gudhi/wasserstein.py
```diff
@@ -23,26 +23,26 @@ def _proj_on_diag(X):
...
 
 
-def _build_dist_matrix(X, Y, p=2., q=2.):
+def _build_dist_matrix(X, Y, order=2., internal_p=2.):
     '''
     :param X: (n x 2) numpy.array encoding the (points of the) first diagram.
     :param Y: (m x 2) numpy.array encoding the second diagram.
-    :param q: Ground metric (i.e. norm l_q).
-    :param p: exponent for the Wasserstein metric.
+    :param internal_p: Ground metric (i.e. norm l_q).
```
```suggestion
    :param internal_p: Ground metric (i.e. norm l_p).
```

src/python/gudhi/wasserstein.py
mglisse: The short version ":returns: the Wasserstein distance" without any precision would have worked as well I guess, since the order and internal_p are already described above. No need to change it though.
There are still many `q` in the _sum file, but as you say, we can't replace them with `order` so this is ok.

tlacombe: Maybe I could go for some `order q Wasserstein distance` ? Although I think for the "mathematical" part of the doc, $q$ shouid be ok.

PR #200, opened 2020-01-14
Print warnings to stderr

### LINKED ISSUE #155: Print warnings to stderr
Usually, warnings should be printed on stderr rather than stdout, so they can't get mixed up with regular output. However, in python, I don't think we ever use stderr, only stdout. An audit of the warnings would be nice.

VincentRouvreau:
Fix #155


mglisse:
I am not sure about the use of std::cout for traces. It looks more natural to me to put them in std::clog (or std::cerr), and keep stdout clean for normal program output. But maybe there are reasons to use cout for them?

VincentRouvreau:
@mglisse I know there are a lot of files to be reviewed, but do you agree with this PR ?

src/Hasse_complex/include/gudhi/Hasse_complex.h
```diff
@@ -173,9 +173,9 @@ class Hasse_complex {
   }
 
   void display_simplex(Simplex_handle sh) {
-    std::cout << dimension(sh) << "  ";
-    for (auto sh_b : boundary_simplex_range(sh)) std::cout << sh_b << " ";
-    std::cout << "  " << filtration(sh) << "         key=" << key(sh);
+    std::clog << dimension(sh) << "  ";
```
mglisse: I think (not sure) this one makes more sense in stdout (well, it shouldn't hardcode where it is writing, but that whole class is undocumented anyway, so I don't care about this specific file). The user explicitly asked the program to print this by calling a function called `display_simplex`, this isn't some unnecessary trace. One goal of moving some output to stderr is so that it can easily be separated from the stdout output, not so that stdout disappears. I don't know if there are more interesting examples elsewhere.

mglisse:
For benchmarks, examples and tests, I think writing to stdout was fine. Although now that you have changed it, that looks ok as well, no need to change it back.

src/python/gudhi/off_reader.pyx
```diff
@@ -32,6 +34,6 @@ def read_points_from_off_file(off_file=''):
         if os.path.isfile(off_file):
             return read_points_from_OFF_file(off_file.encode('utf-8'))
         else:
-            print("file " + off_file + " not found.")
+            print("file " + off_file + " not found.", file=sys.stderr)
```
mglisse: Would it make sense to raise FileNotFoundError (or other) instead of printing a warning? I don't know if that should be done here or in read_points_from_OFF_file or further down the chain. That can be a separate issue/PR.

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -561,7 +561,7 @@ class Persistent_cohomology {
    * p1*...*pr is the product of prime numbers pi such that the homology
    * feature exists in homology with Z/piZ coefficients.
    */
-  void output_diagram(std::ostream& ostream = std::cout) {
+  void output_diagram(std::ostream& ostream = std::clog) {
```
mglisse: Similarly here, I think cout is a good default.

PR #204, opened 2020-01-16
Generators

MathieuCarriere:
I added a wrapper for getting the positive and negative simplices associated to the points in persistence diagrams for both `CubicalComplex` and `SimplexTree`. 

Note that `get_persistence_generators` and `get_persistence_pairs` are very similar. The difference is that `get_persistence_generators` returns the simplices **and** the filtration values, whereas `get_persistence_pairs` only returns the simplices. It might be a good idea to merge them into one single function though.


mglisse:
The name doesn't reflect that particularity. get_persistence_pairs_and_values is ugly (we can probably find better), but at least it is clear what the difference with get_persistence_pairs is...
So the idea is that you could just use get_persistence_pairs, and then get the filtration value of each simplex if/when you need it, but that would be too slow? Or just inconvenient?

src/python/gudhi/cubical_complex.pyx
```diff
@@ -85,7 +86,7 @@ cdef class CubicalComplex:
         elif ((dimensions is None) and (top_dimensional_cells is None)
             and (perseus_file != '')):
             if os.path.isfile(perseus_file):
-                self.thisptr = new Bitmap_cubical_complex_base_interface(perseus_file.encode('utf-8'))
+                self.thisptr = new Bitmap_cubical_complex_base_interface(str.encode(perseus_file))
```
mglisse: This seems to revert a recent patch by Vincent, did you have a problematic rebase?

src/python/gudhi/cubical_complex.pyx
```diff
@@ -145,6 +146,32 @@ cdef class CubicalComplex:
...
+            account (strictly greater than min_persistence). Default value is
+            0.0.
+            Sets min_persistence to -1.0 to see all values.
+        :type min_persistence: float.
+        :param persistence_dim_max: If true, the persistent homology for the
+            maximal dimension in the complex is computed. If false, it is
+            ignored. Default is false.
+        :type persistence_dim_max: bool
+        :returns: The persistence of the simplicial complex, together with the corresponding generators, i.e., the positive and negative top-dimensional cells.
+        :rtype:  list of pairs(dimension, pair(index of positive top-dimensional cell, index of negative top-dimensional cell))
```
mglisse: I can imagine why you might want the top-dimensional cells, but then you aren't exactly returning the cell that creates or kills the cycle. When we build a lower-star filtration in a simplicial complex, we get the simplices, not their vertex with maximal filtration. It may get confusing.

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: Oops, no sorry, there is no particular reason as of why I changed it. Probably a bad copy/paste ;-)

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: Yes, you are right. I am creating this branch mainly to do optimization / backprop on persistence diagrams, so this is why I am only interested in the input data, i.e., the top-dimensional cells. But as you said, these are not quite the positive and negative simplices. I'll modify the doc.

MathieuCarriere:
You are right, I can get the pairs with `get_pairs` and then retrieve their corresponding filtration values to match with the points in the persistence diagram, but it is a little bit faster and more convenient to output everything in a single pass

MathieuCarriere:
Actually it is not clear so I reverted to `persistence_pairs` to avoid repetitions

src/python/gudhi/cubical_complex.pyx
```diff
@@ -145,6 +146,32 @@ cdef class CubicalComplex:
             persistence_result = self.pcohptr.get_persistence(homology_coeff_field, min_persistence)
         return persistence_result
 
+    def cofaces_of_cubical_persistence_pairs(self, homology_coeff_field=11, min_persistence=0, persistence_dim_max = False):
```
mglisse: Currently, when we want persistence under a different format, the interface is that the user has to call `persistence()` (with homology_coeff_field, min_persistence, etc), and then they can call other functions like `persistence_intervals_in_dimension` to get it in the format they like. I am not saying we can't change that, just making sure everyone is aware of it.

src/python/gudhi/cubical_complex.pyx
```diff
@@ -145,6 +146,32 @@ cdef class CubicalComplex:
...
+        """This function returns the persistence of the simplicial complex.
```
mglisse: I think we want more description, even more than you put in the "returns" field. Some vague suggestion (split between the beginning and the returns section)

A persistence interval is described by a pair of cells, one that creates the feature and one that kills it. The filtration values of those 2 cells give coordinates for a point in a persistence diagram, or a bar in a barcode. Structurally, in the cubical complexes provided here, the filtration value of any cell is the minimum of the filtration values of the maximal cells that contain it. Connecting persistence diagram coordinates to the corresponding value in the input (i.e. the filtration values of the top-dimensional cells) is useful for differentiation purposes.
This function returns a list of intervals, where for each interval it gives a 2-tuple with its dimension and a 2-tuple of extremities, and for each extremity it gives a 2-tuple of a filtration value and the (index of the) top-dimensional cell it originates from.

src/python/gudhi/cubical_complex.pyx
```diff
@@ -31,6 +31,7 @@ cdef extern from "Persistent_cohomology_interface.h" namespace "Gudhi":
     cdef cppclass Cubical_complex_persistence_interface "Gudhi::Persistent_cohomology_interface<Gudhi::Cubical_complex::Cubical_complex_interface<>>":
         Cubical_complex_persistence_interface(Bitmap_cubical_complex_base_interface * st, bool persistence_dim_max)
         vector[pair[int, pair[double, double]]] get_persistence(int homology_coeff_field, double min_persistence)
+        vector[pair[int, pair[pair[double, int], pair[double, int]]]] get_cofaces_of_cubical_persistence_pairs(int homology_coeff_field, double min_persistence)
```
mglisse: For compatibility with numpy and the rest of the python ecosystem, we are very slowly trying to avoid complicated mixed structures, and use homogeneous arrays where possible (`intervals_in_dimension`). For instance, that could mean a numpy array (n,2) of float, a numpy array (n,2) of int, and only then group those 2 arrays in a tuple, get one such tuple per dimension and put them in a list or dict. I am not saying this is **the** right interface, I would just like you to look at the place where you use this code with tensorflow or other, and see if an interface like that could help (I am pretty sure it would with pytorch or jax where I think input[intarray] is the differentiable diagram we are looking for).

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +73,60 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
     return persistence;
   }
 
+  void top_dimensional_cofaces(std::vector<int> & cofaces, int splx){
+    if (stptr_->dimension(stptr_->simplex(splx)) == stptr_->dimension()){cofaces.push_back(stptr_->simplex(splx));}
+    else{  for (auto v : stptr_->coboundary_simplex_range(stptr_->simplex(splx))){top_dimensional_cofaces(cofaces, stptr_->key(v));}  }
```
mglisse: There is going to be a lot of redundancy here. For instance, in 3d, starting from a vertex, the 8 incident cubes will appear with multiplicity 6? I think you should take advantage of the filtration value. As soon as you find a coface (of codimension 1) with the same filtration value, use that one as base, it will lead to a top-dimensional cell with the same filtration value.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +73,60 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
     return persistence;
   }
 
+  void top_dimensional_cofaces(std::vector<int> & cofaces, int splx){
```
mglisse: I think the cubical complex uses size_t rather than int.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +73,60 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+    if (stptr_->dimension(stptr_->simplex(splx)) == stptr_->dimension()){cofaces.push_back(stptr_->simplex(splx));}
```
mglisse: Can you explain why you use `key` and `simplex`?

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +73,60 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+    else{  for (auto v : stptr_->coboundary_simplex_range(stptr_->simplex(splx))){top_dimensional_cofaces(cofaces, stptr_->key(v));}  }
+  }
+
+  std::vector<std::pair<int, std::pair<std::pair<double, int>, std::pair<double, int>>>> get_cofaces_of_cubical_persistence_pairs(int homology_coeff_field,
+                                                                                                                                  double min_persistence) {
+
+    // Warning: this function is meant to be used with CubicalComplex only!!
+
+    // Gather all top-dimensional cells and store their simplex handles
```
mglisse: There is `top_dimensional_cells_range()` if that helps.

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: OK. I modified the code accordingly to be more consistent. Now the function is more similar to `persistence_pairs()`

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: I shamelessly copy-pasted your nice doc in the file ;-) I can still it even more if you think it is still too rough

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: OK. Now the output is a numpy array (I removed the float outputs)

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Agreed. Thanks, I modified the code.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: No reason. Reverted to simplex handles.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Cool, thanks!

src/python/gudhi/cubical_complex.pyx
```diff
@@ -145,6 +146,33 @@ cdef class CubicalComplex:
             persistence_result = self.pcohptr.get_persistence(homology_coeff_field, min_persistence)
         return persistence_result
 
+    def cofaces_of_persistence_pairs(self):
```
mglisse: I am wondering if having no way to know the dimension of the feature is going to be a problem.

src/python/gudhi/cubical_complex.pyx
```diff
@@ -145,6 +146,33 @@ cdef class CubicalComplex:
...
+
+           This function returns a list of pairs of top-dimensional cells corresponding to 
+           the persistence birth and death cells of the filtration. The cells are represented by 
+           their indices in the input list of top-dimensional cells (and not their indices in the 
+           internal datastructure that includes non-maximal cells). Note that when two adjacent 
+           top-dimensional cells have the same filtration value, we arbitrarily return one of the two
+           when calling the function on one of their common faces.
+
+        :returns: The top-dimensional cells/cofaces of the positive and negative cells.
+        :rtype:  list of pairs(index of positive top-dimensional cell, index of negative top-dimensional cell)
```
mglisse: It looks like you a returning a numpy array (which is good) instead?

src/python/gudhi/simplex_tree.pyx
```diff
@@ -508,7 +508,7 @@ cdef class SimplexTree:
         """
         if self.pcohptr != NULL:
             if persistence_file != '':
-                self.pcohptr.write_output_diagram(persistence_file.encode('utf-8'))
+                self.pcohptr.write_output_diagram(str.encode(persistence_file))
```
mglisse: That still looks like a bad merge.

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex.h
```diff
@@ -340,6 +340,7 @@ class Bitmap_cubical_complex : public T {
    * that provides ranges for the Boundary_simplex_iterator.
    **/
   Boundary_simplex_range boundary_simplex_range(Simplex_handle sh) { return this->get_boundary_of_a_cell(sh); }
+  Boundary_simplex_range coboundary_simplex_range(Simplex_handle sh) { return this->get_coboundary_of_a_cell(sh); }
```
mglisse: I think you need to add some comment for doxygen to document this function. Maybe add a Coboundary_simplex_range typedef? Or if you want to keep this private, why not call get_coboundary_of_a_cell directly in the file where you need it?

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +72,54 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+    else{  
+      for (auto v : stptr_->coboundary_simplex_range(splx)){
+        if(stptr_->filtration(v) == stptr_->filtration(splx)){
+          return top_dimensional_coface(v);
+        }
+      }  
+    }
+  }
+
+  std::vector<std::vector<int>> cofaces_of_cubical_persistence_pairs() {
```
mglisse: Since the inner vectors all have the same size, why not `std::vector<std::array<XXX>>` or `std::vector<std::pair<XXX>>`?

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +72,54 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+
+    std::vector<std::vector<int>> persistence_pairs;
+    for (auto pair : pairs) {
+      int h = stptr_->dimension(get<0>(pair));
+      // Recursively get the top-dimensional cell / coface associated to the persistence generator
+      int face0 = top_dimensional_coface(get<0>(pair));
+      // Retrieve the index of the corresponding top-dimensional cell in the input data
+      int splx0 = order[face0];
+
+      int splx1 = -1;
```
mglisse: I don't remember if it is documented somewhere that -1 means we have an unmatched simplex.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +72,54 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+      if (isfinite(stptr_->filtration(get<1>(pair)))){
+      // Recursively get the top-dimensional cell / coface associated to the persistence generator
+      int face1 = top_dimensional_coface(get<1>(pair));
+      // Retrieve the index of the corresponding top-dimensional cell in the input data
+      splx1 = order[face1];
+      }
+      std::vector<int> vect{ h, splx0, splx1};
```
mglisse: Wait, so you do put the dimension in there and return a nx3 array. That makes sense, but where is that documented?

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +72,54 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+  int top_dimensional_coface(int splx){
+    if (stptr_->dimension(splx) == stptr_->dimension()){return splx;}
+    else{  
+      for (auto v : stptr_->coboundary_simplex_range(splx)){
+        if(stptr_->filtration(v) == stptr_->filtration(splx)){
+          return top_dimensional_coface(v);
+        }
+      }  
+    }
+  }
```
VincentRouvreau: To remove the warning, in case filtration is not found, you shall return something (splx ?):
```suggestion
  int top_dimensional_coface(int splx){
    if (stptr_->dimension(splx) == stptr_->dimension())
      return splx;
    
    for (auto v : stptr_->coboundary_simplex_range(splx)) {
      if(stptr_->filtration(v) == stptr_->filtration(splx)) {
        return top_dimensional_coface(v);
      }
    }
    return splx;
  }
```

src/python/gudhi/simplex_tree.pyx
VincentRouvreau: I agree with Marc, shall be:
```suggestion
                self.pcohptr.write_output_diagram(persistence_file.encode('utf-8'))
```

src/python/gudhi/cubical_complex.pyx
```diff
@@ -145,6 +146,33 @@ cdef class CubicalComplex:
...
+           filtration values of the maximal cells that contain it. Connecting persistence diagram 
+           coordinates to the corresponding value in the input (i.e. the filtration values of 
+           the top-dimensional cells) is useful for differentiation purposes.
+
+           This function returns a list of pairs of top-dimensional cells corresponding to 
+           the persistence birth and death cells of the filtration. The cells are represented by 
+           their indices in the input list of top-dimensional cells (and not their indices in the 
+           internal datastructure that includes non-maximal cells). Note that when two adjacent 
+           top-dimensional cells have the same filtration value, we arbitrarily return one of the two
+           when calling the function on one of their common faces.
```
VincentRouvreau: Remove the 3 spaces to make a better documentation
```suggestion
        feature and one that kills it. The filtration values of those 2 cells give coordinates 
        for a point in a persistence diagram, or a bar in a barcode. Structurally, in the 
        cubical complexes provided here, the filtration value of any cell is the minimum of the 
        filtration values of the maximal cells that contain it. Connecting persistence diagram 
        coordinates to the corresponding value in the input (i.e. the filtration values of 
        the top-dimensional cells) is useful for differentiation purposes.

        This function returns a list of pairs of top-dimensional cells corresponding to 
        the persistence birth and death cells of the filtration. The cells are represented by 
        their indices in the input list of top-dimensional cells (and not their indices in the 
        internal datastructure that includes non-maximal cells). Note that when two adjacent 
        top-dimensional cells have the same filtration value, we arbitrarily return one of the two
        when calling the function on one of their common faces.
```
```

File: src/python/include/Persistent_cohomology_interface.h (hunk 17 above)
Comment (mglisse) [reply src/python/include/Persistent_cohomology_interface.h]: `BOOST_UNREACHABLE_RETURN(something)`?

File: src/python/include/Persistent_cohomology_interface.h [hunk 19]
```diff
@@ -73,6 +72,54 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
     return persistence;
   }
 
+  int top_dimensional_coface(int splx){
```
mglisse: Could we move this utility function to the main classes, not hidden in the python binding code? With a more explicit name.

MathieuCarriere:
Sorry for the long wait guys, I was very busy preparing my applications

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: As you noted below, I actually include the dimension, I just forgot to update the doc ;-)

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: Yes I am. Doc has been updated.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: Thanks. Fixed it.

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex.h
MathieuCarriere: I followed your suggestion and used `get_coboundary_of_a_cell` instead.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: I am not sure how to use C++ array in Cython. Is it the same than vectors and pairs?

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Indeed. I updated the doc.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: I forgot to update the doc ;-). Fixed it.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: I followed Vincent's suggestion.

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: Fixed it.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: You mean in `Persistent_cohomology.h`? Or somewhere else?

MathieuCarriere:
I also added the unitary test in ``, as well as the feature in `periodic_cubical_complex.pyx`. However, I don't think this code makes sense for Simplex_tree, which already has a `get_pairs` function.

src/python/include/Persistent_cohomology_interface.h
mglisse: In src/Bitmap_cubical_complex/include/gudhi/something. It doesn't depend on persistence.

mglisse:
> I don't think this code makes sense for Simplex_tree

Indeed. For the Simplex_tree, I would appreciate if you could take a look at #243, in particular the discussion there. I think it would make sense to try and be vaguely consistent in what we output.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Sure. Should be done now.

MathieuCarriere:
> For the Simplex_tree, I would appreciate if you could take a look at #243, in particular the discussion there.
Sure I'll have a look at that.

src/python/test/test_cubical_complex.py
```diff
@@ -147,3 +147,8 @@ def test_connected_sublevel_sets():
                                  periodic_dimensions = periodic_dimensions)
     assert cub.persistence() == [(0, (2.0, float("inf")))]
     assert cub.betti_numbers() == [1, 0, 0]
+
+def test_connected_sublevel_sets():
+    cub = CubicalComplex(top_dimensional_cells = [[0, 0, 0], [0, 1, 0], [0, 0, 0]])
+    cub.persistence()
+    assert cub.cofaces_of_persistence_pairs() == np.array([[1, 7, 4], [0, 8, -1]])
```
mglisse: This test fails in the continuous integration. You probably want to use `np.array_equal` for that.

MathieuCarriere: Yes, thanks. Fixed it.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +72,43 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+    auto pairs = persistent_cohomology::Persistent_cohomology<FilteredComplex,
+      persistent_cohomology::Field_Zp>::get_persistent_pairs();
+
+    // Gather all top-dimensional cells and store their simplex handles
+    std::vector<int> max_splx; for (auto splx : stptr_->top_dimensional_cells_range()){
+      max_splx.push_back(splx); 
+    }
+    // Sort these simplex handles and compute the ordering function
+    // This function allows to go directly from the simplex handle to the position of the corresponding top-dimensional cell in the input data 
+    std::map<int, int> order; std::sort(max_splx.begin(), max_splx.end());
```
mglisse, line 88: Don't they come already sorted? I am not completely sure because the doc says in lexicographic order, but I think that should match.

src/python/gudhi/periodic_cubical_complex.pyx
```diff
@@ -155,6 +156,35 @@ cdef class PeriodicCubicalComplex:
...
+        the persistence birth and death cells of the filtration. The cells are represented by 
+        their indices in the input list of top-dimensional cells (and not their indices in the 
+        internal datastructure that includes non-maximal cells). Note that when two adjacent 
+        top-dimensional cells have the same filtration value, we arbitrarily return one of the two
+        when calling the function on one of their common faces.
+
+        :returns: The top-dimensional cells/cofaces of the positive and negative cells, together with the corresponding homological dimension.
+        :rtype:  numpy array of integers of shape [number_of_persistence_points, 3], the integers of eah row being: (homological dimension, 
+            index of positive top-dimensional cell, index of negative top-dimensional cell). If the homological feature is essential, i.e., if 
+            the death time is +infinity, then the index of the corresponding negative top-dimensional cell is -1.
```
mglisse, line 178: I think `rtype` should just be the type, the explanation might fit better in the `returns` part.

src/python/include/Persistent_cohomology_interface.h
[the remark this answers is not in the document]
MathieuCarriere: Thanks! Done.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -73,6 +72,43 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+      persistent_cohomology::Field_Zp>::get_persistent_pairs();
+
+    // Gather all top-dimensional cells and store their simplex handles
+    std::vector<int> max_splx; for (auto splx : stptr_->top_dimensional_cells_range()){
+      max_splx.push_back(splx); 
+    }
+    // Sort these simplex handles and compute the ordering function
+    // This function allows to go directly from the simplex handle to the position of the corresponding top-dimensional cell in the input data 
+    std::map<int, int> order; std::sort(max_splx.begin(), max_splx.end()); 
+    for (unsigned int i = 0; i < max_splx.size(); i++)  order.insert(std::make_pair(max_splx[i], i));
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks! Done.

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
```diff
@@ -109,6 +109,14 @@ class Bitmap_cubical_complex_base {
    **/
   virtual inline std::vector<std::size_t> get_coboundary_of_a_cell(std::size_t cell) const;
 
+  /**
+   * This function computes the index of one of the top-dimensional cubes (chosen arbitrarily) associated
+   * to a given simplex handle. Note that the input parameter is not necessarily a cube, it might also
+   * be an edge or vertex of a cube. On the other hand, the output is always indicating the position of
+   * a cube in the data structure.
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks! Done.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Indeed. I commented the sort.

src/python/gudhi/periodic_cubical_complex.pyx
MathieuCarriere: Agreed. Done.

MathieuCarriere:
I updated the output so that it matches with the one in #243

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
```diff
@@ -109,6 +109,15 @@ class Bitmap_cubical_complex_base {
...
   virtual inline std::vector<std::size_t> get_coboundary_of_a_cell(std::size_t cell) const;
 
+  /**
+   * This function finds a top-dimensional cell that is incident to the input cell and has 
+   * the same filtration value. In case several cells are suitable, an arbitrary one is 
+   * returned. Note that the input parameter is not necessarily a cube, it might also
+   * be an edge or vertex of a cube. On the other hand, the output is always indicating the position of
+   * a cube in the data structure. 
+   **/
+  inline int get_top_dimensional_coface_of_a_cell(int splx);
```
mglisse, line 119: int -> size_t (check for other places where this may be needed)

src/python/include/Persistent_cohomology_interface.h
mglisse: I don't like it, we are generating more code, and if there is a bug and we reach it, we are returning a value that looks possible. If we are going to return a value, at least make it something more likely to explode (`return -1; // Unreachable`?).

src/python/gudhi/cubical_complex.pyx
```diff
@@ -170,6 +171,58 @@ cdef class CubicalComplex:
...
+            The indices of the arrays in the list correspond to the homological dimensions, and the 
+            integers of each row in each array correspond to: (index of positive top-dimensional cell, 
+            index of negative top-dimensional cell). 
+            The second list contains the essential features, grouped by dimension.             
+            It contains numpy arrays of shape [number_of_persistence_points, 1].
+            The indices of the arrays in the list correspond to the homological dimensions, and the 
+            integers of each row in each array correspond to: (index of positive top-dimensional cell).
+        """
+        cdef vector[vector[int]] persistence_result
+        if self.pcohptr != NULL:
```
mglisse, line 203: Sorry, I changed things a bit behind your back, can you use an `assert` like the other functions now do?

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -68,6 +68,44 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
     return persistence;
   }
 
+  std::vector<std::vector<int>> cofaces_of_cubical_persistence_pairs() {
+
+    // Warning: this function is meant to be used with CubicalComplex only!!
+
+    auto pairs = persistent_cohomology::Persistent_cohomology<FilteredComplex,
```
mglisse, line 75:
```suggestion
    auto&& pairs = persistent_cohomology::Persistent_cohomology<FilteredComplex,
```

src/python/include/Persistent_cohomology_interface.h
mglisse: Indeed, cython cannot handle array/tuple... A `vector<pair<int,pair<size_t,size_t>>>` might work, but let's postpone that.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -68,6 +68,44 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
     return persistence;
   }
 
+  std::vector<std::vector<int>> cofaces_of_cubical_persistence_pairs() {
```
mglisse, line 71: Could you add a comment before the function that describes the format of its output ? Each element of the vector is a triple (dimension, birth cell, death cell) with -1 to represent unmatched stuff.

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -68,6 +68,44 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+      int splx0 = order[face0];
+
+      int splx1 = -1;
+      if (isfinite(stptr_->filtration(get<1>(pair)))){
+      // Recursively get the top-dimensional cell / coface associated to the persistence generator
+      int face1 = stptr_->get_top_dimensional_coface_of_a_cell(get<1>(pair));
+      // Retrieve the index of the corresponding top-dimensional cell in the input data
+      splx1 = order[face1];
+      }
+      std::vector<int> vect{ h, splx0, splx1};
```
mglisse, line 103: It is a bit surprising to have a very different-looking output in the C++ part and the python part, but that's an implementation detail, it doesn't matter.

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
MathieuCarriere: Thanks. Done.

src/python/gudhi/cubical_complex.pyx
MathieuCarriere: Sure, no worries. Done.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Thanks! Done.

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Sure. Done

src/python/include/Persistent_cohomology_interface.h
MathieuCarriere: Yes I agree. I did this because it was easier to only modify the code in the `.pyx` file. If necessary, I can change it, or I can open an issue later.

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
mglisse: There were two `int` on that line but you only changed one?

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
```diff
@@ -109,6 +109,15 @@ class Bitmap_cubical_complex_base {
    **/
   virtual inline std::vector<std::size_t> get_coboundary_of_a_cell(std::size_t cell) const;
 
+  /**
+   * This function finds a top-dimensional cell that is incident to the input cell and has 
+   * the same filtration value. In case several cells are suitable, an arbitrary one is 
+   * returned. Note that the input parameter is not necessarily a cube, it might also
+   * be an edge or vertex of a cube. On the other hand, the output is always indicating the position of
+   * a cube in the data structure.
```
mglisse, line 117: If you read https://gudhi.inria.fr/doc/latest/group__cubical__complex.html , all the cells are called cube, a vertex is a cube of dimension 0, an edge is a cube of dimension 1, etc. Could you reformulate the last 2 sentences so they match the terminology?

src/python/include/Persistent_cohomology_interface.h
```diff
@@ -67,6 +68,49 @@ persistent_cohomology::Persistent_cohomology<FilteredComplex, persistent_cohomol
...
+  // then the index of top-dimensional coface of negative simplex is arbitrarily set to -1.
+  std::vector<std::vector<int>> cofaces_of_cubical_persistence_pairs() {
+
+    // Warning: this function is meant to be used with CubicalComplex only!!
+
+    auto&& pairs = persistent_cohomology::Persistent_cohomology<FilteredComplex,
+      persistent_cohomology::Field_Zp>::get_persistent_pairs();
+
+    // Gather all top-dimensional cells and store their simplex handles
+    std::vector<int> max_splx; for (auto splx : stptr_->top_dimensional_cells_range()){
```
mglisse, line 84: Let's try to be consistent and use size_t everywhere for cell indices in the full complex. If you want to keep using int for indexes in the list of top-dimensional cells, that may be ok.

mglisse:
I pushed some changes corresponding to my pending comments.

> Note that since recently, cubical complexes can be built from numpy arrays of any dimension, instead of passing a 1d array (the array is interpreted with fortran order, although IIRC we could instead use C order and reverse the shape). We should clarify what the index refers to exactly...

The test doesn't help, it is on a symmetric matrix (we should ban symmetric (or even square) matrices from tests and examples of cubical complexes). Currently, the function makes sense if you use the constructor with dimensions and a 1d top_dimensional_cells, not if you use the convenient one where top_dimensional_cells takes a multidimensional array. I won't block the branch for that, but it seems important to me to document the meaning there.

Issue #148, opened 2019-11-20
[python representations] Deduplicate code


mglisse:
There is some duplicated code in gudhi.representations. For instance, most classes in vector_methods seem to have the same `fit` method that computes `sample_range`. It should be moved to a single function. @martinroyer suggested a base class, I don't have a strong opinion on that.

VincentRouvreau:
@mglisse It seems this has been fixed on #538 
I do not see any other `fit` to be factorized.

PR #205, opened 2020-01-16
Integration of Wasserstein distances in representations module

MathieuCarriere:
I added a class for Wasserstein distances in the `representations` module, so that users can compute these distances in a `sklearn` fashion.


src/python/gudhi/representations/metrics.py
```diff
@@ -10,6 +10,7 @@
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.metrics import pairwise_distances
+from gudhi.wasserstein import wasserstein_distance
```
mglisse: Just a note to myself: what to do when POT is missing?

mglisse:
Not something that will block this PR, but a remark for later: there is quite a lot of repetition in those classes, in particular the double loop, rectangle or triangle if the 2 lists are the same. At some point we may want to parallelize this loop, and it will be painful to write that code 5 times. Sklearn has a single function sklearn.metrics.pairwise_distances that takes as argument the metric, so the code is written only once. Even without doing the same, it may be possible to put quite a bit of code into a common base class or helper function.

MathieuCarriere:
Agreed. I'm on it.

VincentRouvreau:
> Not something that will block this PR, but a remark for later: there is quite a lot of repetition in those classes, in particular the double loop, rectangle or triangle if the 2 lists are the same. At some point we may want to parallelize this loop, and it will be painful to write that code 5 times. Sklearn has a single function sklearn.metrics.pairwise_distances that takes as argument the metric, so the code is written only once. Even without doing the same, it may be possible to put quite a bit of code into a common base class or helper function.

I think this is what you explained in #148 , no ?
Maybe we could update this issue or make a new issue for the parallelization part.

tlacombe:
> Sklearn has a single function sklearn.metrics.pairwise_distances that takes as argument the metric

I would personally really appreciate such a function :-) .

mglisse:
#182 introduces hera, which gives us a second wasserstein_distance function. How do you suggest to handle it here?

mglisse:
> > Sklearn has a single function sklearn.metrics.pairwise_distances that takes as argument the metric
> 
> I would personally really appreciate such a function :-) .

Note that the only thing that prevents us from using sklearn.metrics.pairwise.pairwise_distances (or scipy.spatial.distance.cdist) with `metric=wasserstein_distance` is that they have an assertion that distances are between vectors (1d), while diagrams are matrices (2d). Flattening the diagrams as vectors of length 2*n, and undoing it in the distance, may be sufficient. Maybe a wrapper that does that would be a good way to benefit from sklearn's parallelization, or maybe it isn't general enough. Or how often do we end up having to flatten diagrams so they can fit in sklearn's view of the world?

MathieuCarriere:
Concerning hera, I guess we can add a second parameter to the `wasserstein_distance` method, something like `mode="POT"` or `mode="hera"`?

VincentRouvreau:
> Concerning hera, I guess we can add another parameter to the `wasserstein_distance` method, something like `mode="POT"` or `mode="hera"`?

The way Marc did it is:
```python
from gudhi.hera import wasserstein_distance
# or
from gudhi.wasserstein import wasserstein_distance
```

mglisse:
> Concerning hera, I guess we can add another parameter to the `wasserstein_distance` method, something like `mode="POT"` or `mode="hera"`?

and forward **kwargs from `__init__` to wasserstein_distance so we can pass `delta=.001`? Or have a parameter `delta` and just ignore it if mode=='POT'?

MathieuCarriere:
Maybe the easiest would be specifying something like `metric=="wasserstein_hera"` or `wasserstein_POT` in the `pairwise_persistence_diagram_distances` function. With one of the two selected as default if `metric=="wasserstein"`

src/python/gudhi/representations/metrics.py
```diff
@@ -22,6 +24,108 @@
...
+        vectori_sum, vectorj_sum = np.sum(vectori), np.sum(vectorj)
+        if vectori_sum != 0:
+            vectori = vectori/vectori_sum
+        if vectorj_sum != 0:
+            vectorj = vectorj/vectorj_sum
+        return np.arccos(  min(np.dot(np.sqrt(vectori), np.sqrt(vectorj)), 1.)  )
+
+def sklearn_wrapper(metric, **kwargs):
+    """
+    This function is a wrapper for any metric between two persistence diagrams that takes two numpy arrays of shapes (nx2) and (mx2) as arguments. It turns the metric into another that takes flattened and padded diagrams as inputs.
```
mglisse: I just read a different suggestion on sklearn's github issue 3737: use `np.arange(len(X)).reshape(-1,1)` as point set, and have the metric wrapper compute dist(X[int(a)],X[int(b)]) as distance between a and b. It avoids reshaping, padding, etc.

mglisse:
sklearn.metrics.pairwise_distances isn't as good as I was hoping. For instance, calling it with 2 processes does not help, because it also disables the optimization of only computing half of the symmetric matrix :-1:

src/python/gudhi/representations/metrics.py
MathieuCarriere: Cool! Done.

MathieuCarriere:
Just added the possibility to use Marc's hera-based Wasserstein distance

tlacombe:
> Maybe the easiest would be specifying something like `metric=="wasserstein_hera"` or `"wasserstein_POT"` in the `pairwise_persistence_diagram_distances` function. With one of the two selected as default if `metric=="wasserstein"`

In this case, what's the good way to set the precision parameter of hera ?

src/python/gudhi/representations/metrics.py
```diff
@@ -116,33 +192,54 @@ def transform(self, X):
         Returns:
             numpy array of shape (number of diagrams in **diagrams**) x (number of diagrams in X): matrix of pairwise bottleneck distances.
         """
-        num_diag1 = len(X)
+        if not USE_GUDHI:
+            print("Gudhi built without CGAL: returning a null matrix")
```
mglisse: (#200 is trying to move those messages to stderr)
I wonder if raising an exception may be preferable. What are the use cases for returning a zero matrix? Is it for a case where you compute distance matrices for several distances and try to learn something from all of them together, and then having one as 0 just means any learning algorithm will discard it as not informative?

src/python/gudhi/representations/metrics.py
```diff
@@ -10,6 +10,10 @@
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.metrics import pairwise_distances
+from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
```
mglisse: This one could fail if POT is not installed, and currently we are trying to consider POT as optional. If it isn't too hard, it could be nice to handle that gracefully (otherwise, it isn't so bad).

mglisse:
> In this case, what's the good way to set the precision parameter of hera ?

Mathieu added a parameter delta that is ignored for POT, which makes sense. The main question when choosing how we pass that parameter is if we want an error when we specify delta and use POT. I think we don't, so this method looks fine.

src/python/gudhi/representations/metrics.py
```diff
@@ -116,33 +192,54 @@ def transform(self, X):
...
+        return Xfit
 
-        #if len(self.diagrams_) == len(X) and np.all([np.array_equal(self.diagrams_[i], X[i]) for i in range(len(X))]):
-        if X is self.diagrams_:
-            matrix = np.zeros((num_diag1, num_diag1))
+class WassersteinDistance(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the Wasserstein distance matrix from a list of persistence diagrams. 
+    """
+    def __init__(self, order=2, internal_p=2, mode="pot", delta=0.0001):
```
mglisse: I am not sure about having different default values of delta everywhere, unless there is some particular reason why a different default value makes sense there. If you think the default value in hera.cc is not good, we can change it, no problem.

src/python/gudhi/representations/metrics.py
MathieuCarriere: I totally agree with you. This 0 matrix has no motivation, I also think an exception is better. Just modified the code.

src/python/gudhi/representations/metrics.py
MathieuCarriere: OK so I raised an exception if POT is not installed. Is this what you had in mind?

src/python/gudhi/representations/metrics.py
MathieuCarriere: I agree. I modified to same default value as in `hera.cc`.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,114 @@
...
+    approx1 = np.matmul(D1, lines)
+    diag_proj1 = (1./2) * np.ones((2,2))
+    approx_diag1 = np.matmul(np.matmul(D1, diag_proj1), lines)
+    approx2 = np.matmul(D2, lines)
+    diag_proj2 = (1./2) * np.ones((2,2))
+    approx_diag2 = np.matmul(np.matmul(D2, diag_proj2), lines)
+    A = np.sort(np.concatenate([approx1, approx_diag2], axis=0), axis=0)
+    B = np.sort(np.concatenate([approx2, approx_diag1], axis=0), axis=0)
+    L1 = np.sum(np.abs(A-B), axis=0)
+    return np.mean(L1)
```
mglisse: If I understand this code correctly: it used to be that we could precompute some things about diagrams in `fit` (their projections along each line, and the projections along each line of their diagonal projections, possibly pre-sorted to speed up further sorting). But now we have to recompute that for each diagram as many times as there are other diagrams. I don't know how important it is for this specific distance, but in general it doesn't look so nice. I think those computations should be done in 2 steps, one where we preprocess diagrams (say replace a diagram by all the projections), and then the step where we compute some quantity on each pair of **preprocessed** diagrams. Does that make sense?

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,114 @@
...
+            from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
+            return pairwise_distances(XX, YY, metric=sklearn_wrapper(pot_wasserstein_distance,  X, Y, **kwargs))
+        except ImportError:
+            print("Gudhi built without POT")
+    elif metric == "wasserstein" or metric == "hera_wasserstein":
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(hera_wasserstein_distance,  X, Y, **kwargs))
+    elif metric == "sliced_wasserstein":
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(sliced_wasserstein_distance, X, Y, **kwargs))
+    elif metric == "persistence_fisher":
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(persistence_fisher_distance, X, Y, **kwargs))
```
mglisse: IIRC scikit-learn has a dict { "metric_name": distance_helper_fun, ... }. It can be slightly nicer (and lets sneaky users modify that dict, since it is a global variable), but that's not required at all, I am just mentioning it.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,114 @@
...
+    XX = np.reshape(np.arange(len(X)), [-1,1])
+    YY = None if Y is None else np.reshape(np.arange(len(Y)), [-1,1]) 
+    if metric == "bottleneck":
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(bottleneck_distance, X, Y, **kwargs))
+    elif metric == "pot_wasserstein":
+        try:
+            from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
+            return pairwise_distances(XX, YY, metric=sklearn_wrapper(pot_wasserstein_distance,  X, Y, **kwargs))
+        except ImportError:
+            print("Gudhi built without POT")
```
mglisse: POT is only a runtime dependency. If we don't just let the error message propagate ("ModuleNotFoundError: No module named 'ot'" is already quite readable), we can suggest that in order to compute wasserstein distances, the user should either install POT or use mode="hera" or metric="hera_wasserstein".

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,114 @@
...
+    :param Y: second list of persistence diagrams (optional). If None, pairwise distances are computed from the first list only.
+    :param metric: distance to use. It can be either a string ("sliced_wasserstein", "wasserstein", "bottleneck", "persistence_fisher") or a function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
+    :returns: distance matrix, i.e., numpy array of shape (num diagrams 1 x num diagrams 2)
+    :rtype: float
+    """
+    XX = np.reshape(np.arange(len(X)), [-1,1])
+    YY = None if Y is None else np.reshape(np.arange(len(Y)), [-1,1]) 
+    if metric == "bottleneck":
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(bottleneck_distance, X, Y, **kwargs))
+    elif metric == "pot_wasserstein":
```
mglisse: This option doesn't appear in the doc for `metric`.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,114 @@
...
+        try:
+            from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
+            return pairwise_distances(XX, YY, metric=sklearn_wrapper(pot_wasserstein_distance,  X, Y, **kwargs))
+        except ImportError:
+            print("Gudhi built without POT")
+    elif metric == "wasserstein" or metric == "hera_wasserstein":
```
mglisse: @tlacombe Mathieu is making wasserstein default to hera, are you ok with that or do you suggest pot instead? I don't have any opinion, I just want to make sure you two agree.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,114 @@
...
+    :param X: first list of persistence diagrams. 
+    :param Y: second list of persistence diagrams (optional). If None, pairwise distances are computed from the first list only.
+    :param metric: distance to use. It can be either a string ("sliced_wasserstein", "wasserstein", "bottleneck", "persistence_fisher") or a function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
+    :returns: distance matrix, i.e., numpy array of shape (num diagrams 1 x num diagrams 2)
+    :rtype: float
+    """
+    XX = np.reshape(np.arange(len(X)), [-1,1])
+    YY = None if Y is None else np.reshape(np.arange(len(Y)), [-1,1]) 
+    if metric == "bottleneck":
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(bottleneck_distance, X, Y, **kwargs))
```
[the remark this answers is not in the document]
MathieuCarriere: Indeed it was a mistake. Fixed it.

src/python/gudhi/representations/metrics.py
```diff
@@ -116,33 +188,55 @@ def transform(self, X):
         Returns:
             numpy array of shape (number of diagrams in **diagrams**) x (number of diagrams in X): matrix of pairwise bottleneck distances.
         """
-        num_diag1 = len(X)
+        try:
+            Xfit = pairwise_persistence_diagram_distances(X, self.diagrams_, metric="bottleneck", e=self.epsilon)
+        except ImportError:
+            print("Gudhi built without CGAL")
+        return Xfit
```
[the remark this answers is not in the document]
MathieuCarriere: Yes I fully agree. I still catch the error, but then I rethrow it now, as you suggested.

src/python/gudhi/representations/metrics.py
MathieuCarriere: Yes that totally makes sense. I kept the `sliced_wasserstein_distance` function, but I also added `sliced_wasserstein_distance_on_projections` and `compute_persistence_diagram_projections` to speed up the computation of all pairwise distances.

src/python/gudhi/representations/metrics.py
MathieuCarriere: Thanks, I did the same. The closer I am to sklearn, the better I feel ;-)

src/python/gudhi/representations/metrics.py
MathieuCarriere: Agreed. Just changed the error message.

src/python/gudhi/representations/metrics.py
MathieuCarriere: Thanks, just fixed the doc.

src/python/gudhi/representations/metrics.py
MathieuCarriere: I did this mostly to avoid having a default option that requires an extra dependency, but I don't mind reverting this.

mglisse: That's a good argument in favor of hera indeed.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,154 @@
...
+            return pairwise_distances(XX, YY, metric=sklearn_wrapper(bottleneck_distance, X, Y, **kwargs))
+        except ImportError:
+            print("Gudhi built without CGAL")
+            raise
+    elif metric == "pot_wasserstein":
+        try:
+            from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
+            return pairwise_distances(XX, YY, metric=sklearn_wrapper(pot_wasserstein_distance,  X, Y, **kwargs))
+        except ImportError:
+            print("Gudhi built without POT. Please install POT or use metric='wasserstein' or metric='hera_wasserstein'")
```
mglisse: People using WassersteinDistance with mode="pot" will also see this message. But I guess it doesn't really matter. Please rephrase "Gudhi built without POT" -> "POT not installed" or something like that. Users do **not** need to rebuild gudhi.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -9,13 +9,73 @@
...
+    """
+    DD1 = np.concatenate([D1, D1[:,[1,0]]], axis=0)
+    DD2 = np.concatenate([D2, D2[:,[1,0]]], axis=0)
+    weight_pss = lambda x: 1 if x[1] >= x[0] else -1
+    return 0.5 * persistence_weighted_gaussian_kernel(DD1, DD2, weight=weight_pss, kernel_approx=kernel_approx, bandwidth=bandwidth)
+
+def pairwise_persistence_diagram_kernels(X, Y=None, metric="sliced_wasserstein", **kwargs):
+    """
+    This function computes the kernel matrix between two lists of persistence diagrams given as numpy arrays of shape (nx2).
+    :param X: first list of persistence diagrams.
```
mglisse: The doc looks horrible this way, it doesn't interpret this line as describing a parameter. Maybe skipping a line before would help.
If you have trouble building the doc on your computer, it is visible online: at the bottom of the pull request: "show all tests", "Details" next to the test called "build", "python", "artifacts" (if you don't see any link called artifacts, go to circleci.com, log in using github, and start again).
I notice you changed from Google-style to sphinx-style. Both work, I am just surprised by the difference.

MathieuCarriere: Yes, you are right. Even I don't know exactly why I changed the doc style ;-) I made everything consistent.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+            return metric(X[int(a[0])], Y[int(b[0])], **kwargs)
+    return flat_metric
+
+PAIRWISE_DISTANCE_FUNCTIONS = {
+    "wasserstein": hera_wasserstein_distance,
+    "hera_wasserstein": hera_wasserstein_distance,
+    "persistence_fisher": persistence_fisher_distance,
+}
+
+def pairwise_persistence_diagram_distances(X, Y=None, metric="bottleneck", **kwargs):
```
mglisse: Replying to a point in the audio the other day: it would have been possible to have a class Pairwise_persistence_diagram_distances with fit() and transform(), if we wanted to use a single point of entry and still be usable in pipelines. Now I am not at all saying we should do that, I just felt like replying to that point.

MathieuCarriere: Indeed, that's possible. Although, such a class is not really necessary (it's my opinion of course) because:

---if you want to use sklearn pipelines, it's actually a little bit easier to cross-validate across multiple estimators than cross-validate across different values of a parameter in a single estimator.

---if you don't want to used sklearn pipelines, it is more intuitive to call `pairwise_persistence_diagram_distances()` than going through the fit and transform methods of estimators.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -9,13 +9,82 @@
...
+    return 0.5 * persistence_weighted_gaussian_kernel(DD1, DD2, weight=weight_pss, kernel_approx=kernel_approx, bandwidth=bandwidth)
+
+def pairwise_persistence_diagram_kernels(X, Y=None, metric="sliced_wasserstein", **kwargs):
+    """
+    This function computes the kernel matrix between two lists of persistence diagrams given as numpy arrays of shape (nx2).
+
+    Parameters:    
+        X (list of n numpy arrays of shape (numx2)): first list of persistence diagrams. 
+        Y (list of m numpy arrays of shape (numx2)): second list of persistence diagrams (optional). If None, pairwise kernel values are computed from the first list only.
+        metric: kernel to use. It can be either a string ("sliced_wasserstein", "persistence_scale_space", "persistence_weighted_gaussian", "persistence_fisher") or a function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
```
mglisse, line 70: I am not sure about the name `metric` for this parameter. `kernel` looks more natural. However, `metric` is consistent with distances, and not really wrong since IIUC kernels are associated with metrics. I am ok with keeping `metric`, I just want to make sure it is here by choice and not a copy-paste accident.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -9,13 +9,82 @@
...
+
+    Returns:
+        float: the persistence weighted Gaussian kernel value between persistence diagrams. 
+    """
+    ws1 = np.array([weight(D1[j,:]) for j in range(len(D1))])
+    ws2 = np.array([weight(D2[j,:]) for j in range(len(D2))])
+    if kernel_approx is not None:
+        approx1 = np.sum(np.multiply(ws1[:,np.newaxis], kernel_approx.transform(D1)), axis=0)
+        approx2 = np.sum(np.multiply(ws2[:,np.newaxis], kernel_approx.transform(D2)), axis=0)
+        return (1./(np.sqrt(2*np.pi)*bandwidth)) * np.matmul(approx1, approx2.T)
```
mglisse, line 39: Should this be split, at some point in the future, into a projection function computing approx and a function computing the distance between 2 approximations, when doing pairwise distances on many diagrams? Or is computing the approximation much faster than the distance, or maybe the approximation takes too much memory to keep many around?

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+        try:
+            from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
+            return pairwise_distances(XX, YY, metric=sklearn_wrapper(pot_wasserstein_distance,  X, Y, **kwargs))
+        except ImportError:
+            print("POT (Python Optimal Transport) is not installed. Please install POT or use metric='wasserstein' or metric='hera_wasserstein'")
+            raise
+    elif metric == "sliced_wasserstein":
+        Xproj = compute_persistence_diagram_projections(X, **kwargs)
+        Yproj = None if Y is None else compute_persistence_diagram_projections(Y, **kwargs)
+        return pairwise_distances(XX, YY, metric=sklearn_wrapper(sliced_wasserstein_distance_on_projections, Xproj, Yproj))
```
mglisse, line 170: I was thinking of a general mechanism, like PAIRWISE_DISTANCE_FUNCTIONS but giving 2 functions (projection, distance_on_projections). But if sliced Wasserstein is the only one that benefits from a projection, then that's completely unnecessary and it is indeed better to do it by hand, as you did.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+
+    Returns: 
+        XX (list of n numpy arrays of shape (2*numx2)): list of projected persistence diagrams.
+    """
+    thetas = np.linspace(-np.pi/2, np.pi/2, num=num_directions+1)[np.newaxis,:-1]
+    lines = np.concatenate([np.cos(thetas), np.sin(thetas)], axis=0)
+    XX = [np.vstack([np.matmul(D, lines), np.matmul(np.matmul(D, .5 * np.ones((2,2))), lines)]) for D in X]
+    return XX
+
+def sliced_wasserstein_distance_on_projections(D1, D2):
```
mglisse, line 61: Do we want this function (and the projection) to appear in the documentation? I expect users will be interested in the combined function, or the one that does pairwise distances, but seldom this low-level one.
Normally I wouldn't mind a few extra functions, but the sphinx documentation is getting hard to read, automodule just concatenates the doc of all functions in the order they appear in the file, and the reader may get a bit lost with so many sliced Wasserstein functions. We should not do a reorganization of the doc as part of this PR, or it will never get merged, but I'd like to have an idea of the plan, either not document those 2, or document them and plan to reorganize the doc after this PR is merged (representations.rst can be split into 4 files/pages, instead of automodule you can use autoclass/autofunction and organize them with sections, subsections, and text to explain when to use which one).

mglisse:
I see that you are using the convention that CamelCase is for a class that handles lists of diagrams with fit/transform, while snake_case is for a function that handles just a pair of diagrams. For time delay embedding, after not getting much feedback, we went with just one class, where fit/transform handles lists, while `__call__` handles just one instance. I think that's a bit easier to document, as it groups everything under one class, and it yields a callable that really just takes 2 arguments, which can be convenient. I am not necessarily opposed to your version, I just think it is worth discussing.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+def compute_persistence_diagram_projections(X, num_directions):
+    """
+    This is a function for projecting the points of a list of persistence diagrams (as well as their diagonal projections) onto a fixed number of lines sampled uniformly on [-pi/2, pi/2]. This function can be used as a preprocessing step in order to speed up the running time for computing all pairwise sliced Wasserstein distances / kernel values on a list of persistence diagrams. 
+
+    Parameters:
+        X (list of n numpy arrays of shape (numx2)): list of persistence diagrams. 
+        num_directions (int): number of lines evenly sampled from [-pi/2,pi/2] in order to approximate and speed up the distance computation.
+
+    Returns: 
+        XX (list of n numpy arrays of shape (2*numx2)): list of projected persistence diagrams.
```
mglisse, line 54:
```suggestion
        list of n numpy arrays of shape (2*numx2): list of projected persistence diagrams.
```
or the doc looks strange.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+        vectori = np.sum(np.exp(-np.square(pairwise_distances(Z,U))/(2 * np.square(bandwidth)))/(bandwidth * np.sqrt(2*np.pi)), axis=1)
+        vectorj = np.sum(np.exp(-np.square(pairwise_distances(Z,V))/(2 * np.square(bandwidth)))/(bandwidth * np.sqrt(2*np.pi)), axis=1)
+        vectori_sum, vectorj_sum = np.sum(vectori), np.sum(vectorj)
+        if vectori_sum != 0:
+            vectori = vectori/vectori_sum
+        if vectorj_sum != 0:
+            vectorj = vectorj/vectorj_sum
+        return np.arccos(  min(np.dot(np.sqrt(vectori), np.sqrt(vectorj)), 1.)  )
+
+def sklearn_wrapper(metric, X, Y, **kwargs):
```
mglisse, line 121: I don't think we want that one appearing in the documentation. Different ways I know to get it out of the documentation:
1. rename it `_sklearn_wrapper`
2. don't use automodule in the .rst file, but one autoclass or autofunction for each class or function you do want to appear in the doc
3. still use automodule but list the class/functions that should appear, or list those that should not appear (read sphinx doc for the syntax to do that)

MathieuCarriere: Indeed. I used the first option you suggested :-)

src/python/gudhi/representations/metrics.py
MathieuCarriere: Fixed it.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+    approx_diag1 = np.matmul(np.matmul(D1, diag_proj1), lines)
+    approx2 = np.matmul(D2, lines)
+    diag_proj2 = (1./2) * np.ones((2,2))
+    approx_diag2 = np.matmul(np.matmul(D2, diag_proj2), lines)
+    A = np.sort(np.concatenate([approx1, approx_diag2], axis=0), axis=0)
+    B = np.sort(np.concatenate([approx2, approx_diag1], axis=0), axis=0)
+    L1 = np.sum(np.abs(A-B), axis=0)
+    return np.mean(L1)
+
+def compute_persistence_diagram_projections(X, num_directions):
```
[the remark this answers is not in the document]
MathieuCarriere: It is a private function now.

src/python/gudhi/representations/kernel_methods.py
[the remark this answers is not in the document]
MathieuCarriere: Yes I think it makes sense to ask `metric` to be symmetric. I added it in the doc.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -9,13 +9,82 @@
...
 
+def persistence_weighted_gaussian_kernel(D1, D2, weight=lambda x: 1, kernel_approx=None, bandwidth=1.):
+    """
+    This is a function for computing the persistence weighted Gaussian kernel value from two persistence diagrams. The persistence weighted Gaussian kernel is computed by convolving the persistence diagram points with weighted Gaussian kernels. See http://proceedings.mlr.press/v48/kusano16.html for more details.
+
+    Parameters:
+        D1: (n x 2) numpy.array encoding the (finite points of the) first diagram. Must not contain essential points (i.e. with infinite coordinate).
+        D2: (m x 2) numpy.array encoding the second diagram.
+        bandwidth (double): bandwidth of the Gaussian kernel with which persistence diagrams will be convolved
+        weight: weight function for the persistence diagram points. This function must be defined on 2D points, ie lists or numpy arrays of the form [p_x,p_y].
```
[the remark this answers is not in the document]
MathieuCarriere: Fixed it.

src/python/gudhi/representations/metrics.py
MathieuCarriere: It is a private function now.

src/python/gudhi/representations/metrics.py
MathieuCarriere: sliced Wasserstein is the only one for now so I let it as it is, but I agree it would be interesting to implement more general functions in the future if required.

src/python/gudhi/representations/kernel_methods.py
MathieuCarriere: I actually agree. I did that first to be similar to `sklearn` but after some thoughts I prefer `kernel`, it is much less confusing.

MathieuCarriere:
> I see that you are using the convention that CamelCase is for a class that handles lists of diagrams with fit/transform, while snake_case is for a function that handles just a pair of diagrams. For time delay embedding, after not getting much feedback, we went with just one class, where fit/transform handles lists, while __call__ handles just one instance. I think that's a bit easier to document, as it groups everything under one class, and it yields a callable that really just takes 2 arguments, which can be convenient. I am not necessarily opposed to your version, I just think it is worth discussing.
Or maybe the function taking just 2 diagrams is an implementation detail and it shouldn't appear in the doc at all, that's an other option...

I really like the `__call__` methods, so I added one for almost classes in the `representations` module ;-)

src/python/gudhi/representations/kernel_methods.py
MathieuCarriere: Yes that makes sense. I thought SlicedWasserstein was the only one, but Fisher, PSS and PWG might actually benefit a similar implementation.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -52,7 +121,20 @@ def transform(self, X):
         Returns:
             numpy array of shape (number of diagrams in **diagrams**) x (number of diagrams in X): matrix of pairwise sliced Wasserstein kernel values.
         """
-        return np.exp(-self.sw_.transform(X)/self.bandwidth)
+        return pairwise_persistence_diagram_kernels(X, self.diagrams_, kernel="sliced_wasserstein", bandwidth=self.bandwidth, num_directions=self.num_directions) 
+
+    def __call__(self, diag1, diag2):
```
mglisse, line 126: Actuellement, la doc est triée par ordre alphabétique, donc elle met `__call__` avant `__init__`. Ça surprend, mais ce n'est pas vraiment un problème je pense. `:member-order: bysource` permettrait de dire de respecter l'ordre dans le fichier, mais trier les classes par ordre alphabétique est assez sympa. Il y aurait aussi la possibilité de `autoclass_content = 'both'` (intégrer la doc de `__init__` à celle générale de la classe), mais j'ai l'impression que ça ne peut se faire qu'à l'échelle de tout le projet.

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+    "persistence_fisher": _persistence_fisher_distance,
+}
+
+def pairwise_persistence_diagram_distances(X, Y=None, metric="bottleneck", **kwargs):
+    """
+    This function computes the distance matrix between two lists of persistence diagrams given as numpy arrays of shape (nx2).
+
+    Parameters:
+        X (list of n numpy arrays of shape (numx2)): first list of persistence diagrams. 
+        Y (list of m numpy arrays of shape (numx2)): second list of persistence diagrams (optional). If None, pairwise distances are computed from the first list only.
```
mglisse, line 145: Je ne sais pas s'il y a un moyen simple de dire que `num` n'a pas besoin d'être le même, ni entre X et Y, ni même à l'intérieur de X ou Y. Peut-être une phrase dans representations.rst disant que partout on pourra mélanger des diagrammes avec des nombres de points différents ? (ce n'est pas nouveau avec ce PR)

mglisse:
Let me reduce my ambition a bit so we can merge this soon: if you can clarify the doc about the symmetry for kernels (you choose if it has to be symmetric or not, just say which one), whether distance(x,x)=0, and kwargs, I think the rest can wait (although you are welcome to answer it too).

src/python/example/diagram_vectorizations_distances_kernels.py
```diff
@@ -9,77 +9,63 @@
...
+  SlicedWassersteinKernel, BottleneckDistance, PersistenceFisherKernel, WassersteinDistance
 
-D = np.array([[0.,4.],[1.,2.],[3.,8.],[6.,8.], [0., np.inf], [5., np.inf]])
-diags = [D]
+D1 = np.array([[0.,4.],[1.,2.],[3.,8.],[6.,8.], [0., np.inf], [5., np.inf]])
 
-diags = DiagramSelector(use=True, point_type="finite").fit_transform(diags)
-diags = DiagramScaler(use=True, scalers=[([0,1], MinMaxScaler())]).fit_transform(diags)
-diags = DiagramScaler(use=True, scalers=[([1], Clamping(maximum=.9))]).fit_transform(diags)
+proc1, proc2, proc3 = DiagramSelector(use=True, point_type="finite"), DiagramScaler(use=True, scalers=[([0,1], MinMaxScaler())]), DiagramScaler(use=True, scalers=[([1], Clamping(maximum=.9))])
```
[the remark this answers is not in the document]
MathieuCarriere: Not at all. It is just a result of my strange tastes ;-) I put everything on three lines.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -9,13 +9,82 @@
...
+
+def pairwise_persistence_diagram_kernels(X, Y=None, kernel="sliced_wasserstein", **kwargs):
+    """
+    This function computes the kernel matrix between two lists of persistence diagrams given as numpy arrays of shape (nx2).
+
+    Parameters:    
+        X (list of n numpy arrays of shape (numx2)): first list of persistence diagrams. 
+        Y (list of m numpy arrays of shape (numx2)): second list of persistence diagrams (optional). If None, pairwise kernel values are computed from the first list only.
+        kernel: kernel to use. It can be either a string ("sliced_wasserstein", "persistence_scale_space", "persistence_weighted_gaussian", "persistence_fisher") or a function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
+
```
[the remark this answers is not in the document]
MathieuCarriere: I added a line in the doc

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+    Returns: 
+        float: the sliced Wasserstein distance between persistence diagrams. 
+    """
+    thetas = np.linspace(-np.pi/2, np.pi/2, num=num_directions+1)[np.newaxis,:-1]
+    lines = np.concatenate([np.cos(thetas), np.sin(thetas)], axis=0)
+    approx1 = np.matmul(D1, lines)
+    diag_proj1 = (1./2) * np.ones((2,2))
+    approx_diag1 = np.matmul(np.matmul(D1, diag_proj1), lines)
+    approx2 = np.matmul(D2, lines)
+    diag_proj2 = (1./2) * np.ones((2,2))
```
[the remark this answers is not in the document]
MathieuCarriere: I love your code :-) It would be nice to use Theo's code in order to avoid duplicated code (even though this is probably fine for such a small thing), but I don't want the user to get a warning if he doesn't have POT

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -159,11 +223,20 @@ def transform(self, X):
...
+        Apply PersistenceScaleSpaceKernel on a single pair of persistence diagrams and outputs the result.
+
+        Parameters:
+            diag1 (n x 2 numpy array): first input persistence diagram.
+            diag2 (n x 2 numpy array): second input persistence diagram.
+
+        Returns:
+            float: persistence scale space kernel value.
+        """
+        return _persistence_scale_space_kernel(diag1, diag2, bandwidth=self.bandwidth, kernel_approx=self.kernel_approx)
```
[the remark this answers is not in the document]
MathieuCarriere: Mmmh I don't know because these functions are also called by the `pairwise_` functions

src/python/gudhi/representations/metrics.py
```diff
@@ -10,17 +10,169 @@
...
+}
+
+def pairwise_persistence_diagram_distances(X, Y=None, metric="bottleneck", **kwargs):
+    """
+    This function computes the distance matrix between two lists of persistence diagrams given as numpy arrays of shape (nx2).
+
+    Parameters:
+        X (list of n numpy arrays of shape (numx2)): first list of persistence diagrams. 
+        Y (list of m numpy arrays of shape (numx2)): second list of persistence diagrams (optional). If None, pairwise distances are computed from the first list only.
+        metric: distance to use. It can be either a string ("sliced_wasserstein", "wasserstein", "hera_wasserstein" (Wasserstein distance computed with Hera---note that Hera is also used for the default option "wasserstein"), "pot_wasserstein" (Wasserstein distance computed with POT), "bottleneck", "persistence_fisher") or a symmetric function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
```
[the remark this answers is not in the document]
MathieuCarriere: Yes, you are right. I changed the doc.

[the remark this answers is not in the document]
MathieuCarriere: Yes because some metric functions such as `bottleneck_distance` have extra arguments

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -9,13 +9,82 @@
...
+    return 0.5 * _persistence_weighted_gaussian_kernel(DD1, DD2, weight=weight_pss, kernel_approx=kernel_approx, bandwidth=bandwidth)
+
+def pairwise_persistence_diagram_kernels(X, Y=None, kernel="sliced_wasserstein", **kwargs):
+    """
+    This function computes the kernel matrix between two lists of persistence diagrams given as numpy arrays of shape (nx2).
+
+    Parameters:    
+        X (list of n numpy arrays of shape (numx2)): first list of persistence diagrams. 
+        Y (list of m numpy arrays of shape (numx2)): second list of persistence diagrams (optional). If None, pairwise kernel values are computed from the first list only.
+        kernel: kernel to use. It can be either a string ("sliced_wasserstein", "persistence_scale_space", "persistence_weighted_gaussian", "persistence_fisher") or a function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
```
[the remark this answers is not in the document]
MathieuCarriere: Yep! Thanks, just changed the doc.

src/python/gudhi/representations/metrics.py
MathieuCarriere: I think it is a good idea to mention this, I added a sentence in the `.rst` file

src/python/gudhi/representations/kernel_methods.py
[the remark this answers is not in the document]
mglisse: It doesn't look like it would be harder for the pairwise_ functions to use the classes' `__call__` functions instead.

src/python/gudhi/representations/metrics.py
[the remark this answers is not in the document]
mglisse: I meant it should be added to the doc so people know they can provide a function that takes more than 2 arguments. But you added it in the kwargs description, so I guess that's enough.

src/python/gudhi/representations/metrics.py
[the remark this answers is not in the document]
mglisse: I agree that if we were to reuse that function, it should be in a file that is safer to include, and that reusing a one-liner is not fundamental.

src/python/gudhi/representations/kernel_methods.py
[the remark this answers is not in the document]
mglisse:
```diff
- return pairwise_kernels(XX, YY, metric=_sklearn_wrapper(_persistence_scale_space_kernel, X, Y, **kwargs))
+ return pairwise_kernels(XX, YY, metric=_sklearn_wrapper(PersistenceScaleSpaceKernel(**kwargs), X, Y))
```
(untested, it is just to clarify what I was thinking)

PR #209, opened 2020-01-20
Reorganize the main page of C++ doc

mglisse:
This is similar to what I did in #189, trying to remain consistent.

The position of "contractions" is a bit strange, I wondered about putting it alone, or grouping it with cohomology. I am keeping it close to skeleton-blockers for now, but that doesn't have to be the last word.


mglisse:
If someone has something prettier than this (x<sub>1</sub>,...,x<sub>d</sub>) to offer as the logo for point cloud utilities...

PR #210, opened 2020-01-24
Modern cmake for boost

VincentRouvreau:


VincentRouvreau:
> We may need to check if this changes the minimum version required for boost or cmake. Maybe we are already good.

There is a way to be retro compatible : https://cliutils.gitlab.io/modern-cmake/chapters/packages/Boost.html
And, apparently, it is required if cmake version is newer than boost.

mglisse:
> it is required if cmake version is newer than boost

Unless cmake is even more recent where the workaround is already included? Your choice if you want to add the workaround or not.

PR #211, opened 2020-01-27
Iterator over simplex tree

### LINKED ISSUE #2: [Cython - simplex_tree] There is no simplex iterators

VincentRouvreau:
A proposition for iterators to fix #2 .

The tests won't pass as the interface is broken.
It allows to do:
```python
import gudhi
st = gudhi.SimplexTree()                                                                                                                              
st.insert([1,2,3], 0.5)                                                                                                                               
>>  True

st.insert([4,3,5], 1.)                                                                                                                                
>>  True

for filt in st.get_filtration(): 
    print(filt) 
                                                                                                                                                      
>>  ([1], 0.5)
    ([2], 0.5)
    ([1, 2], 0.5)
    ([3], 0.5)
    ([1, 3], 0.5)
    ([2, 3], 0.5)
    ([1, 2, 3], 0.5)
    ([4], 1.0)
    ([3, 4], 1.0)
    ([5], 1.0)
    ([3, 5], 1.0)
    ([4, 5], 1.0)
    ([3, 4, 5], 1.0)
```


src/python/gudhi/simplex_tree.pyx
```diff
@@ -214,15 +215,14 @@ cdef class SimplexTree:
...
-            ct.append((v, filtered_complex.second))
-        return ct
+        cdef vector[Simplex_tree_simplex_handle].const_iterator it = self.get_ptr().get_filtration_iterator_begin()
+        cdef vector[Simplex_tree_simplex_handle].const_iterator end = self.get_ptr().get_filtration_iterator_end()
+
+        while True:
+            yield(self.get_ptr().get_simplex_filtration(dereference(it)))
+            preincrement(it)
+            if it == end:
+                raise StopIteration
```
mglisse: You don't have to raise the exception yourself, `break` to end the loop is enough (getting to the end of the function is how you finish iterating, with yield generators). So probably `while it != end:` instead?

src/python/include/Simplex_tree_interface.h
```diff
@@ -82,17 +83,12 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
...
-  Filtered_simplices get_filtration() {
-    Base::initialize_filtration();
-    Filtered_simplices filtrations;
-    for (auto f_simplex : Base::filtration_simplex_range()) {
-      Simplex simplex;
-      for (auto vertex : Base::simplex_vertex_range(f_simplex)) {
-        simplex.insert(simplex.begin(), vertex);
-      }
-      filtrations.push_back(std::make_pair(simplex, Base::filtration(f_simplex)));
+  Filtered_simplex get_simplex_filtration(Simplex_handle f_simplex) {
```
mglisse: The name confused me, I thought it would return just the filtration. `get_simplex_and_filtration`? `with_filtration`?

src/python/gudhi/simplex_tree.pyx
```diff
@@ -209,20 +209,18 @@ cdef class SimplexTree:
...
-        :rtype:  list of tuples(simplex, filtration)
+        :rtype:  generator with tuples(simplex, filtration)
         """
         cdef vector[Simplex_tree_simplex_handle].const_iterator it = self.get_ptr().get_filtration_iterator_begin()
         cdef vector[Simplex_tree_simplex_handle].const_iterator end = self.get_ptr().get_filtration_iterator_end()
 
-        while True:
-            yield(self.get_ptr().get_simplex_filtration(dereference(it)))
+        while it != end:
+            yield(self.get_ptr().get_simplex_and_filtration(dereference(it)))
```
mglisse:
```suggestion
            yield self.get_ptr().get_simplex_and_filtration(dereference(it))
```
?
The extra parentheses look strange.

src/python/test/test_alpha_complex.py
```diff
@@ -105,16 +107,18 @@ def test_filtered_alpha():
...
+    assert(next(filtration_generator) == ([1], 0.0))
+    assert(next(filtration_generator) == ([2], 0.0))
+    assert(next(filtration_generator) == ([3], 0.0))
+    assert(next(filtration_generator) == ([0, 1], 0.25))
+    assert(next(filtration_generator) == ([0, 2], 0.25))
+    assert(next(filtration_generator) == ([1, 3], 0.25))
+    assert(next(filtration_generator) == ([2, 3], 0.25))
+    with pytest.raises(StopIteration):
+        next(filtration_generator)
+
```
mglisse: https://stackoverflow.com/q/9983547/1918193 suggests several other ways to compare a generator with something. The simplest is `list(simplex_tree.get_filtration()) == ...`, and there are some fancier ones with izip_longest. What you did works, but it seems a bit painful to have to go through that everywhere.

src/python/include/Simplex_tree_interface.h
```diff
@@ -82,17 +83,12 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
...
-        simplex.insert(simplex.begin(), vertex);
-      }
-      filtrations.push_back(std::make_pair(simplex, Base::filtration(f_simplex)));
+  Simplex_and_filtration get_simplex_and_filtration(Simplex_handle f_simplex) {
+    Simplex simplex;
+    for (auto vertex : Base::simplex_vertex_range(f_simplex)) {
+      simplex.insert(simplex.begin(), vertex);
     }
-    return filtrations;
+    return std::make_pair(simplex, Base::filtration(f_simplex));
```
mglisse:
```suggestion
    return std::make_pair(std::move(simplex), Base::filtration(f_simplex));
```
We are still doing many copies, but that should be one less...

VincentRouvreau:
@mglisse `boundary_simplex_range(SimplexHandle sh)` is still not available. Tell me if you want it. Otherwise, I think I finished to fix your comments.

mglisse:
> @mglisse `boundary_simplex_range(SimplexHandle sh)` is still not available. Tell me if you want it. 

I want it eventually, but not necessarily as a wrapper to the C++ thing (IIRC the issue contains a pure python implementation), and it is better as a separate PR later anyway.

> Otherwise, I think I finished to fix your comments.

Thanks.
Do you want to add a comment near the `..._range().begin()` and `end()` warning that this is ok in this specific case because the range is just a pair of iterators? (we still end up constructing begin and end twice and dropping half of them, but that's not so bad)

src/python/test/test_simplex_tree.py
```diff
@@ -249,3 +249,15 @@ def test_make_filtration_non_decreasing():
     assert st.filtration([3, 4, 5]) == 2.0
     assert st.filtration([3, 4]) == 2.0
     assert st.filtration([4, 5]) == 2.0
+
+def test_simplices_iterator():
+    st = SimplexTree()
+    
+    assert st.insert([0, 1, 2], filtration=4.0) == True
```
mglisse: Normally I am a bit wary of putting code with important side effects in assert: running with `python -O` disables the whole line. But this is just a test, and disabling the inserts just makes the loop below empty, so that should be fine.

PR #214, opened 2020-01-31
About adding timedelay for feature engineering

takeshimeonerespect:
- Additional purpose  
Feature engineering from time series  
- Content  
Convert the input time series to attractor  
- Review point  
  - Source code (Document) `src/python/gudhi/point_cloud/timedelay.py`  
  - Test code `src/python/test/test_point_cloud.py`  
  
Please give me a review.


src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
```
mglisse: Usually we start those source files with a notice that gives the license (MIT), the authors (you? plus someone else?) and the copyright holder (Fujitsu labs?). You can copy and adapt them from some other random file, say subsampling.pyx.

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,56 @@
...
+
+        Returns
+        -------
+        point clouds : list[list[float, float, float]] or list[list[list[float, float, float]]]
+            Makes point cloud every a single time-series data.
+        """
+        ndts = np.array(ts)
+        if ndts.ndim == 1:
+            # for single.
+            return self._transform(ndts).tolist()
```
mglisse: Why convert everything to lists? Why not return the numpy array?

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,56 @@
...
+                np.arange(0, self._dim*self._delay, self._delay))
+        ]
+
+    def transform(self, ts):
+        """Transform method.
+
+        Parameters
+        ----------
+        ts : list[float] or list[list[float]]
+            A single or multiple time-series data.
```
mglisse: Asking for opinions: How about splitting the functionality for 1 vs multiple series into several functions? Say transform would only work on multiple series, because that's what sklearn expects, while `__call__` would only work on a single series? Or some other similar scheme.

src/python/test/test_point_cloud.py
```diff
@@ -0,0 +1,35 @@
+from gudhi.point_cloud.timedelay import TimeDelayEmbedding
+
+def test_normal():
```
mglisse: The test looks good. Maybe we could name the file test_time_delay.py so it is more specific?

mglisse:
Hard question, that doesn't need to be solved here:
we are soon going to need this same functionality to work on torch.tensor. Hari wrote https://github.com/science-enthusiast/pytorch-layer-calling-gudhi/blob/master/tdeLayerOne.py, although I like the indexing you did in `_transform` much better than the explicit double loop. The fact that it is a class that derives from Module and implements a forward function is a detail, it isn't strictly necessary.
I wonder if it will be possible to have a single implementation of time delay embedding, that works just as well whether you give it a numpy.ndarray, a torch.tensor, etc, or if we will need to duplicate the code for each framework (when I say "work for torch.tensor", I mean in a way compatible with autograd, converting to numpy.array doesn't work in that sense). I expect we probably could for a few very simple functions like time delay embedding, tweak the code to make it work for both, but as soon as the function gets larger, it will become harder, so trying hard to share the small functions may be a waste of time. But I may be completely wrong, it is too early for me to have a strong opinion.

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+
+class TimeDelayEmbedding:
+    """Point cloud transformation class.
+
+    Embeds time-series data in the R^d according to Takens' Embedding Theorem
+    and obtains the coordinates of each point.
```
mglisse: It may not be completely obvious for a user what `delay` and `skip` really mean, I believe that a simple example (where delay and skip are different and not 1) here (or in point_cloud.rst), similar to the one in the tests, would help users understand immediately what it does.

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: Thank you for your comment.  
Please correct as follows.  
```
# This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
# See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
# Author(s):       Martin Royer, Yuichi Ike, Masatoshi Takenouchi
#
# Copyright (C) 2019 Inria, Copyright (C) 2019 Fujitsu Laboratories Ltd.
# Modification(s):
#   - YYYY/MM Author: Description of the modification
```

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: When using it alone from the outside, I felt it was better to pass it in an   
orthodox LIST type (Python standard array).  
Until now, it was designed as an inner class so that it could not be used   
from the outside, so we passed data with ndarray.

src/python/gudhi/point_cloud/timedelay.py
mglisse: We could write 2020, although I don't mind 2019. You can continue editing your branch, commit a patch and push to takeshimeonerespect:master. It will automatically appear here if you do that.

takeshimeonerespect: I fixed based on comment.

src/python/test/test_point_cloud.py
takeshimeonerespect: I want to change the file name from test_point_cloud.py to test_time_delay.py.

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: I understood to write such an example.  
is it right?  
```
[For example]
time_series = [1, 2, 3, 4, 5, 6, 7]  
Given skip=2, delay=2, the attractor will be like this.
attractor = [[1, 3, 5], 
             [3, 5, 7]]
```

src/python/test/test_point_cloud.py
mglisse: Sure, you can go ahead, it is just `git mv test_point_cloud.py test_time_delay.py` (then commit and push).

src/python/gudhi/point_cloud/timedelay.py
mglisse: Yes, you can add those few lines to the docstring, just like that, but choosing values skip != delay (maybe 2 and 3) so that users don't confuse them.
If you want to add it in the .rst file instead, it can be presented as code (see for instance wasserstein_distance_user.rst which uses testcode and testoutput), where the testsuite actually executes the 2 lines of code and checks that the output hasn't changed.
I let you choose if you prefer to put it in timedelay.py or point_cloud.rst.

src/python/test/test_point_cloud.py
takeshimeonerespect: I changed a file name.

mglisse: You also need to modify CMakeLists.txt so it tests the new name instead of the old name.

src/python/gudhi/point_cloud/timedelay.py
mglisse: An advantage of separating single vs multiple is that it would let us generalize the function, so it can take as input say a (n,2) array representing a time series with a 2D vector at each time, and it can return a (n-2,6) array where we take 3 consecutive 2D vectors and concatenate them into a 6D vector. This is already what `_transform` does, up to a final call to `reshape`.

takeshimeonerespect: Thank you for valuable comment.  
`__call__`, `_transform` and `transform` have been modified as follows.  
Is code good like this?  
  
- Before  
```
def __call__(self, *args, **kwargs):
    return self.transform(*args, **kwargs)
```
- After  
```
def __call__(self, *args, **kwargs):
    return self._transform(*args, **kwargs)
```  

- Before
```
def _transform(self, ts):
    """Guts of transform method.

    Parameters
    ----------
    ts : ndarray, shape=[n_features]
        A single time-series data.

    Returns
    -------
    attractor : ndarray, shape=[n_points, d]
        `n_points` is the number of points in the R^d.

    """
    return ts[
        np.add.outer(
            np.arange(0, len(ts)-self._delay*(self._dim-1), self._skip),
            np.arange(0, self._dim*self._delay, self._delay))
    ]
```
- After  
```
def _transform(self, ts):
    """Guts of transform method.

    Parameters
    ----------
    ts : ndarray, shape=[n_features]
        A single time-series data.

    Returns
    -------
    attractor : ndarray, shape=[n_points, d]
        `n_points` is the number of points in the R^d.

    """
    ts = np.array(ts)
    return ts[
        np.add.outer(
            np.arange(0, len(ts)-self._delay*(self._dim-1), self._skip),
            np.arange(0, self._dim*self._delay, self._delay))
    ]
```
- Before
```
def transform(self, ts):
    """Transform method."""
    ndts = np.array(ts)
    if ndts.ndim == 1:
        # for single.
        return self._transform(ndts)
    else:
        # for multiple.
        return np.apply_along_axis(self._transform, 1, ndts)
```  
- After
```   
# for multiple
def transform(self, ts):
    """Transform method."""
    ndts = np.array(ts)
    return np.apply_along_axis(self._transform, 1, ndts)
```

src/python/test/test_point_cloud.py
takeshimeonerespect: I succeed test in new name(test_time_delay.py).  
- Test result
new_name_test_2020-02-07

mglisse:
Since you are creating a new subdirectory of gudhi, I believe you need to add a file `__init__.py` in it to satisfy python. An empty file should do. Without `__init__.py`, the subdirectory might be omitted in the conda package, for instance.

src/python/gudhi/point_cloud/timedelay.py
mglisse: I believe numpy arrays (or at least buffer-compatible objects) are the standard when it comes to scientific computing. Scikit-learn uses them as input and output everywhere. Converting back and forth with lists is going to be costly.

src/python/gudhi/point_cloud/timedelay.py
mglisse: Yes, that looks good to me, if you agree with that interface.

mglisse:
For compatibility with sklearn pipelines, I believe we need a `fit(self, X, y=None)` method that does nothing. It also wouldn't hurt to have a fit_transform (that calls fit then transform), which can be added manually or by inheriting from TransformerMixin.

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: The following has been reflected in the code.  
Please comment.
- splitting the functionality for 1 vs multiple series  
- ndts = np.array(ts) means  
- Fix delay and skip example
- add fit (self, X, y = None)  
  
```
class TimeDelayEmbedding:
    """Point cloud transformation class.
    Embeds time-series data in the R^d according to Takens' Embedding Theorem
    and obtains the coordinates of each point.
    Parameters
    ----------
    dim : int, optional (default=3)
        `d` of R^d to be embedded.
    delay : int, optional (default=1)
        Time-Delay embedding.
    skip : int, optional (default=1)
        How often to skip embedded points.
        Given delay=3 and skip=2, an point cloud which is obtained by embedding 
        a single time-series data into R^3 is as follows.
        
    .. code-block:: none
    
    time-series = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    point clouds = [[1, 4, 7], 
                    [3, 6, 9]]
                   
    """
    def __init__(self, dim=3, delay=1, skip=1):
        self._dim = dim
        self._delay = delay
        self._skip = skip

    def __call__(self, ts):
        """Transform method for single time-series data.
        Parameters
        ----------
        ts : list[float]
            A single time-series data.
        Returns
        -------
        point clouds : list[list[float, float, float]]
            Makes point cloud every a single time-series data.
        Raises
        -------
        TypeError
            If the parameter's type does not match the desired type.
        """
        ndts = np.array(ts)
        if ndts.ndim == 1:
            return self._transform(ndts)
        else:
            raise TypeError("Expects 1-dimensional array.")

    def fit(self, ts, y=None):
        return self
    
    def _transform(self, ts):
        """Guts of transform method."""
        return ts[
            np.add.outer(
                np.arange(0, len(ts)-self._delay*(self._dim-1), self._skip),
                np.arange(0, self._dim*self._delay, self._delay))
        ]

    def transform(self, ts):
        """Transform method for multiple time-series data.
        Parameters
        ----------
        ts : list[list[float]]
            Multiple time-series data.
        Attributes
        ----------
        ndts : 
             The ndts means that all time series need to have exactly 
             the same size. 
        Returns
        -------
        point clouds : list[list[list[float, float, float]]]
            Makes point cloud every a single time-series data.
        Raises
        -------
        TypeError
            If the parameter's type does not match the desired type.
        """
        ndts = np.array(ts)
        if ndts.ndim == 2:
            return np.apply_along_axis(self._transform, 1, ndts)
        else:
            raise TypeError("Expects 2-dimensional array.")
```

takeshimeonerespect:
The comment was corrected.
What should I do next?
continuous-integration is pending, what should I do?

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+
+import numpy as np
+
+class TimeDelayEmbedding:
+    """Point cloud transformation class.
+    Embeds time-series data in the R^d according to Takens' Embedding Theorem
+    and obtains the coordinates of each point.
+    Parameters
+    ----------
+    dim : int, optional (default=3)
```
mglisse: Question: should we name this `dim` or something like `repeat` or `copies` (or better, I am not super fond of those 2 names)? For scalar time series, it is the same. But we will probably generalize this function later to accept vector time series, in which case, if we start from a 2D time series, we can embed it in dimension 6 by taking 3 *copies* with delays. In that case, would users rather pass 6 or 3 as argument? I don't know, I think I would more naturally pass 3, but both seem ok, maybe we can stick to `dim`.

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+        ts : list[list[float]]
+            Multiple time-series data.
+        Attributes
+        ----------
+        ndts : 
+             The ndts means that all time series need to have exactly 
+             the same size. 
+        Returns
+        -------
+        point clouds : list[list[list[float, float, float]]]
```
mglisse: It looks like we are not returning a list anymore, but an array?

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: **Name suggestions**
How about in the following order?  
1. replicate
2. recursion
3. iteration
4. rebuild
5. recast

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: Corrected as follows.  
```
points clouds : list of n x 3 numpy arrays
```

src/python/gudhi/point_cloud/timedelay.py
takeshimeonerespect: I suggest as follows.
What do you think?
```python
plane : int, optional (default=3)
        Determination of delayed coordinate system for space.
```

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+# See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+# Author(s):       Martin Royer, Yuichi Ike, Masatoshi Takenouchi
+#
+# Copyright (C) 2020 Inria, Copyright (C) 2020 Fujitsu Laboratories Ltd.
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
+
+import numpy as np
+
+class TimeDelayEmbedding:
```
VincentRouvreau: I have these warnings when generating the documentation:
```
gudhi-devel/build/version/build/python/gudhi/point_cloud/timedelay.py:docstring of gudhi.point_cloud.timedelay.TimeDelayEmbedding:9: WARNING: Unexpected indentation.
gudhi-devel/build/version/build/python/gudhi/point_cloud/timedelay.py:docstring of gudhi.point_cloud.timedelay.TimeDelayEmbedding:11: WARNING: Block quote ends without a blank line; unexpected unindent.
```

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+            If the parameter's type does not match the desired type.
+        """
+        ndts = np.array(ts)
+        if ndts.ndim == 1:
+            return self._transform(ndts)
+        else:
+            raise TypeError("Expects 1-dimensional array.")
+
+    def fit(self, ts, y=None):
+        return self
```
VincentRouvreau: When I perform some coverage of the proposed unitary tests, these lines are not covered:
```bash
cd build/src/python
python3 -m pytest ../../../src/python/test/test_time_delay.py --cov-report term-missing --cov=gudhi
Name                                      Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------
gudhi/point_cloud/timedelay.py               20      6    70%   57, 60, 90-94
```

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+        Raises
+        -------
+        TypeError
+            If the parameter's type does not match the desired type.
+        """
+        ndts = np.array(ts)
+        if ndts.ndim == 2:
+            return np.apply_along_axis(self._transform, 1, ndts)
+        else:
+            raise TypeError("Expects 2-dimensional array.")
```
VincentRouvreau: When I perform some coverage of the proposed unitary tests, these lines are not covered:
```bash
cd build/src/python
python3 -m pytest ../../../src/python/test/test_time_delay.py --cov-report term-missing --cov=gudhi
Name                                      Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------
gudhi/point_cloud/timedelay.py               20      6    70%   57, 60, 90-94
```

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+        Raises
+        -------
+        TypeError
+            If the parameter's type does not match the desired type.
+        """
+        ndts = np.array(ts)
+        if ndts.ndim == 1:
+            return self._transform(ndts)
+        else:
+            raise TypeError("Expects 1-dimensional array.")
```
VincentRouvreau: Please, could you add a test to catch this exception ?
It is quite easy with [pytest to catch exceptions](https://docs.pytest.org/en/3.0.1/assert.html#assertions-about-expected-exceptions).

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,94 @@
...
+#
+# Copyright (C) 2020 Inria, Copyright (C) 2020 Fujitsu Laboratories Ltd.
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
+
+import numpy as np
+
+class TimeDelayEmbedding:
+    """Point cloud transformation class.
+    Embeds time-series data in the R^d according to Takens' Embedding Theorem
```
VincentRouvreau: Maybe add a link to wikipedia:
```suggestion
    Embeds time-series data in the R^d according to [Takens' Embedding Theorem](https://en.wikipedia.org/wiki/Takens%27s_theorem)
```

src/python/gudhi/point_cloud/timedelay.py
mglisse: That doesn't really clarify things. Let's stick to `dim`, as you currently have, it is clear that it is the dimension of the space we are embedding into (and not the number of repetitions).

mglisse:
What do you think of https://github.com/mglisse/gudhi-devel/commit/f2b85dc4ebfed07d71adfc1dfa2b8632d4e17824 ? I generalized a bit (and thus had to remove some checks), fixed some doc issues, used a list instead of apply_along_axis (which is surprisingly slow in my tests) and let `black -l 120` reformat the code (it isn't prettier than the original, but it gives consistency). Of course if we do that, the doc will need some update, at least an example with a vector time series.

takeshimeonerespect:
> What do you think of [mglisse@f6669e4](https://github.com/mglisse/gudhi-devel/commit/f6669e4679d959e1a367d0753afcb873f2d17826) (updated link) ? I generalized a bit (and thus had to remove some checks), fixed some doc issues, used a list instead of apply_along_axis (which is surprisingly slow in my tests) and let `black -l 120` reformat the code (it isn't prettier than the original, but it gives consistency).

Thank you very helpful comments.  
I think this is good.  
takehimeonerespect / gudhi-devel has reflected the pointed out content.

src/python/gudhi/point_cloud/timedelay.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+
+    def __init__(self, dim=3, delay=1, skip=1):
+        self._dim = dim
+        self._delay = delay
+        self._skip = skip
+
+    def __call__(self, ts):
+        """Transform method for single time-series data.
+        Parameters
```
mglisse: You are missing an empty line before Parameters, this seems important for sphinx, otherwise it is interpreted as part of the previous paragraph.

src/python/gudhi/point_cloud/timedelay.py
mglisse: I believe that once you add this link asked by Vincent, we are good to go :+1:

takeshimeonerespect: Thank you very much.  
Now reflected.

mglisse: >  **Go please wait.**

Ok, I'll wait, let me know when I can proceed.

takeshimeonerespect: **Travis CI build is failing, don't you care about this?**
https://travis-ci.org/GUDHI/gudhi-devel/builds/654797009?utm_source=github_status&utm_medium=notification

mglisse: The test that fails is about building the C++ documentation, something that your branch does not touch. It already affects master. It is an issue with installing doxygen using homebrew on macosx (I believe it appeared after travis or homebrew updated something), Vincent is trying to fix it in some other branch. It doesn't need to block your branch.

takeshimeonerespect: Thank you.
I understand well.

mglisse: So, is it ok with you if I merge your branch, or do you still want me to wait?

takeshimeonerespect: I want you to wait.
I'm asking my boss when to merge.

takeshimeonerespect: I'm sorry my reply was late.
Please merge on **March 16 AM**.

PR #215, opened 2020-01-31
Extended persistence

MathieuCarriere:
Addition of extended persistence


MathieuCarriere:
The first commit ("added wasserstein class") is not intentional, I mixed up with another pull request

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,75 @@ class Simplex_tree {
     }
   }
 
+  /** \brief Retrieve good values for extended persistence, and separate the diagrams into the ordinary, relative, extended+ and extended- subdiagrams. Need extend_filtration to be called first!
+   * @param[in] dgm Persistence diagram obtained after calling this->extend_filtration and this->get_persistence.
+   * @return A vector of four persistence diagrams. The first one is Ordinary, the second one is Relative, the third one is Extended+ and the fourth one is Extended-.
+   */
+  std::vector<std::vector<std::pair<int, std::pair<double, double>>>> convert(const std::vector<std::pair<int, std::pair<double, double>>>& dgm){
```
mglisse: The name `convert` seems awfully generic.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,75 @@ class Simplex_tree {
...
+    std::vector<std::vector<std::pair<int, std::pair<double, double>>>> new_dgm(4); double x, y;
+    for(unsigned int i = 0; i < dgm.size(); i++){ int h = dgm[i].first; double px = dgm[i].second.first; double py = dgm[i].second.second;
+      if(std::isinf(py))  continue;
```
mglisse: Can that happen?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,75 @@ class Simplex_tree {
...
+      else{
+        if ((px <= -1) & (py <= -1)){x = minval_ + (maxval_-minval_)*(px + 2); y = minval_ + (maxval_-minval_)*(py + 2); new_dgm[0].push_back(std::make_pair(h, std::make_pair(x,y))); }
+        if ((px >=  1) & (py >=  1)){x = minval_ - (maxval_-minval_)*(px - 2); y = minval_ - (maxval_-minval_)*(py - 2); new_dgm[1].push_back(std::make_pair(h, std::make_pair(x,y))); }
+        if ((px <= -1) & (py >=  1)){x = minval_ + (maxval_-minval_)*(px + 2); y = minval_ - (maxval_-minval_)*(py - 2);
+          if (x <= y)  new_dgm[2].push_back(std::make_pair(h, std::make_pair(x,y)));
+          else  new_dgm[3].push_back(std::make_pair(h, std::make_pair(x,y)));
+        }
+      }
```
mglisse: Maybe add a comment here that you are ignoring the (redundant) points with `px>=1&&py>=1`?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,75 @@ class Simplex_tree {
...
+    }
+    return new_dgm;
+  }
+
+  /** \brief Extend filtration for computing extended persistence.
```
mglisse: Is this something that only works on lower-star filtrations?

mglisse:
> The first commit ("added wasserstein class") is not intentional, I mixed up with another pull request

`git rebase -i` is very convenient to clean up such mistakes, although it requires a forced push afterwards.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Document that not only it adds more simplices in the complex, it also changes the filtration value of existing simplices, so probably a user should not try to use it for anything other than computing persistence. And give a link to the function that converts back filtration values.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Yes, the cone point always induces a point that have infinite persistence (which is thus useless)

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Agreed. I changed to `compute_extended_persistence_subdiagrams`

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Sorry, I don't see what you mean. Am I ignoring points beside the one with infinite persistence?

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Yes. I added this in the doc.

MathieuCarriere: Yes, I modified the doc.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Ah, no, ignore this bad comment.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1487,7 +1487,7 @@ class Simplex_tree {
     return new_dgm;
   }
 
-  /** \brief Extend filtration for computing extended persistence.
+  /** \brief Extend filtration for computing extended persistence. This function only uses the filtration values at the 0-dimensional simplices, and computes the extended persistence diagram induced by the lower-star filtration computed with these values. Note that after calling this function, the filtration values are actually modified. The function compute_extended_persistence_subdiagrams retrieves the original values and separates the extended persistence diagram points w.r.t. their types (Ord, Rel, Ext+, Ext-) and should always be called after computing the persistent homology of the extended simplicial complex.
```
mglisse: I think Vincent may complain that the line is a little bit long :wink:

src/python/gudhi/simplex_tree.pyx
```diff
@@ -386,6 +386,20 @@ cdef class SimplexTree:
         """
         return self.get_ptr().make_filtration_non_decreasing()
 
+    def extend_filtration(self):
+        """ This function extends filtration for computing extended persistence.
+        """
+        return self.get_ptr().extend_filtration()
+
+    def convert(self, dgm):
+        """This function retrieves good values for extended persistence, and separate the diagrams into the ordinary, relative, extended+ and extended- subdiagrams. Need extend_filtration to be called first!
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks! Done.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -386,6 +386,20 @@ cdef class SimplexTree:
         """
         return self.get_ptr().make_filtration_non_decreasing()
 
+    def extend_filtration(self):
+        """ This function extends filtration for computing extended persistence.
+        """
+        return self.get_ptr().extend_filtration()
+
+    def convert(self, dgm):
```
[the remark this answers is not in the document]
MathieuCarriere: Indeed. Should be OK now.

src/python/gudhi/simplex_tree.pxd
```diff
@@ -43,6 +43,8 @@ cdef extern from "Simplex_tree_interface.h" namespace "Gudhi":
         void remove_maximal_simplex(vector[int] simplex)
         bool prune_above_filtration(double filtration)
         bool make_filtration_non_decreasing()
+        void extend_filtration()
+        vector[vector[pair[int, pair[double, double]]]] convert(vector[pair[int, pair[double, double]]])
```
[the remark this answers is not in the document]
MathieuCarriere: Indeed. Should be OK now.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
     }
   }
 
+  /** \brief Retrieve good values for extended persistence, and separate the 
+   * diagrams into the ordinary, relative, extended+ and extended- subdiagrams. 
+   * Need extend_filtration to be called first!
+   * @param[in] dgm Persistence diagram obtained after calling this->extend_filtration 
+   * and this->get_persistence.
+   * @return A vector of four persistence diagrams. The first one is Ordinary, the 
+   * second one is Relative, the third one is Extended+ and the fourth one is Extended-.
```
mglisse: I don't know how obvious it is what Extended + vs - are. Actually, even for the other categories, a notebook showing the categories with pictures will be nice, after this gets in. Is there some paper we could cite somewhere?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
     }
   }
 
+  /** \brief Retrieve good values for extended persistence, and separate the
```
mglisse: "good" values are only approximate with this implementation (numerical errors when going to [-2,-1] and back), I think we want to state that clearly.

mglisse:
Do you intend to replace the private implementation in Nerve_GIC with a call to this one afterwards? (don't do it yet, wait until this is merged, in case the interface changes)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
 
+  /** \brief Retrieve good values for extended persistence, and separate the 
+   * diagrams into the ordinary, relative, extended+ and extended- subdiagrams. 
+   * Need extend_filtration to be called first!
+   * @param[in] dgm Persistence diagram obtained after calling this->extend_filtration 
+   * and this->get_persistence.
+   * @return A vector of four persistence diagrams. The first one is Ordinary, the 
+   * second one is Relative, the third one is Extended+ and the fourth one is Extended-.
+   */
+  std::vector<std::vector<std::pair<int, std::pair<double, double>>>> compute_extended_persistence_subdiagrams(const std::vector<std::pair<int, std::pair<double, double>>>& dgm){
```
mglisse: I was wondering about alternate interfaces, where this function would not be part of the Simplex_tree. Just putting them out there in case they inspire something to someone.

1. We could have something on top of the simplex tree, that presents a different interface to Persistent_cohomology. For instance, we could store minvalue and maxvalue in that object. Or Filtration_value could be something like `pair<bool,double>` and filtration(simplex) would return (true,filt) if simplex ends with maxvert, (false,filt) otherwise. This avoids modifying the filtration values, and the meaning of the normal output of persistence is not too hard to explain. We could even avoid constructing the cone explicitly, and write functions for boundary, filtration, etc that fake its existence, although we would probably need to add a second "key" field to each simplex or use a hashtable on the side to store the key of those phantom simplices.

2. We could have an Extended_persistent_cohomology, that internally uses Persistent_cohomology, and that provides various functions like relative_intervals_in_dimension, etc, already split and already with corrected values.

mglisse:
Mathieu, in #211 Vincent changed get_filtration so that it returns a generator instead of a list, now. This means that `st.get_filtration() == ...` doesn't work directly. The simplest fix for a test is `list(st.get_filtration()) == ...`. Other uses like `for s,f in st.get_filtration():` keep working.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+   * computed with these values. Note that after calling this function, the filtration 
+   * values are actually modified. The function compute_extended_persistence_subdiagrams 
+   * retrieves the original values and separates the extended persistence diagram points 
+   * w.r.t. their types (Ord, Rel, Ext+, Ext-) and should always be called after 
+   * computing the persistent homology of the extended simplicial complex.
+   */
+  void extend_filtration() {
+
+    // Compute maximum and minimum of filtration values
+    int maxvert = -std::numeric_limits<int>::infinity();
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks. Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+    std::vector<double> filt;
+    for (auto sh : this->complex_simplex_range()) {
```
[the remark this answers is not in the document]
MathieuCarriere: Indeed, thanks for the suggestion!

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+      if (this->dimension(sh) == 0){
+        filt.push_back(this->filtration(sh)); 
+        maxvert = std::max(*this->simplex_vertex_range(sh).begin(), maxvert);
+      }
+    }
+    minval_ = *std::min_element(filt.begin(), filt.end());
+    maxval_ = *std::max_element(filt.begin(), filt.end());
```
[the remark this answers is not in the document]
MathieuCarriere: Yes you are right. I removed `filt`.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+    maxvert += 1;
+
+    // Compute vectors of integers corresponding to the Simplex handles
+    std::vector<std::vector<int> > splxs;
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks. Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+    std::vector<double> filt;
+    for (auto sh : this->complex_simplex_range()) {
+      if (this->dimension(sh) == 0){
+        filt.push_back(this->filtration(sh)); 
+        maxvert = std::max(*this->simplex_vertex_range(sh).begin(), maxvert);
+      }
+    }
+    minval_ = *std::min_element(filt.begin(), filt.end());
+    maxval_ = *std::max_element(filt.begin(), filt.end());
+    maxvert += 1;
```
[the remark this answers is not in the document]
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+        vr.push_back(vh);
+      }
+      splxs.push_back(vr);
+    }
+
+    // Add point for coning the simplicial complex
+    int count = this->num_simplices();
+    std::vector<int> cone; 
+    cone.push_back(maxvert); 
+    auto ins = this->insert_simplex(cone, -3);
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks. Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+      }
+      splxs.push_back(vr);
+    }
+
+    // Add point for coning the simplicial complex
+    int count = this->num_simplices();
+    std::vector<int> cone; 
+    cone.push_back(maxvert); 
+    auto ins = this->insert_simplex(cone, -3); 
+    this->assign_key(ins.first, count);
```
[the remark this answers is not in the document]
MathieuCarriere: I forgot the initial reason ;-) I removed the call to these `assign_keys` functions.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
MathieuCarriere: Thanks, this is a great suggestion. It requires myself going a little deeper in the Simplex_tree code though, so for now, I used a quick fix where I just clone the original simplex tree and iterate over its simplices (which takes memory 2*n IIUC).

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+        // Assign value -3 to simplex and cone on simplex
+        this->assign_filtration(sh, -3);
+        auto ins = this->insert_simplex(vr, -3);
+        this->assign_key(ins.first, count);
+      }
+      count++;
+    }
+
+    this->make_filtration_non_decreasing(); 
+    this->initialize_filtration();
```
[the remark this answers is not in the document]
MathieuCarriere: Agreed. Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1467,117 @@ class Simplex_tree {
...
+
```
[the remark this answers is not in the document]
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: I added a link. I have also planned to prepare a notebook illustrating this.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Done.

MathieuCarriere:
> Do you intend to replace the private implementation in Nerve_GIC with a call to this one afterwards? (don't do it yet, wait until this is merged, in case the interface changes)

Yep, that's the plan!

> Mathieu, in #211 Vincent changed get_filtration so that it returns a generator instead of a list, now. This means that st.get_filtration() == ... doesn't work directly. The simplest fix for a test is list(st.get_filtration()) == .... Other uses like for s,f in st.get_filtration(): keep working.

Thanks for the heads-up! I fixed the test file accordingly.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: Since we are actually inside the simplex tree, it would be possible to do:
```c++
for(auto sh=root_.members().begin(); sh!=root_.members().end(); ++sh)
```
(Simplex_handle happens to be a typedef for those iterators)
But I guess the skeleton range is good enough.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1469,6 +1472,124 @@ class Simplex_tree {
...
+    // Compute maximum and minimum of filtration values
+    int maxvert = std::numeric_limits<int>::min();
+    this->minval_ = std::numeric_limits<double>::max();
+    this->maxval_ = std::numeric_limits<double>::min();
+    for (auto sh : this->skeleton_simplex_range(0)) {
+      double f = this->filtration(sh);
+      this->minval_ = std::min(this->minval_, f);
+      this->maxval_ = std::max(this->maxval_, f);
+      maxvert = std::max(*this->simplex_vertex_range(sh).begin(), maxvert);
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks!

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1469,6 +1472,124 @@ class Simplex_tree {
...
+    }
+    
+    assert (maxvert < std::numeric_limits<int>::max());
+    maxvert += 1;
+
+    Simplex_tree* st_copy = new Simplex_tree(*this);
```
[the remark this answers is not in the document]
MathieuCarriere: You are right.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1469,6 +1472,124 @@ class Simplex_tree {
...
+
+    // For each simplex
+    for (auto sh_copy : st_copy->complex_simplex_range()){
+
+      // Locate simplex
+      std::vector<Vertex_handle> vr;
+      for (auto vh : st_copy->simplex_vertex_range(sh_copy)){
+        vr.push_back(vh);
+      }
+      auto sh = this->find(vr);
```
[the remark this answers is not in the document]
MathieuCarriere: That's right, although I still need to get `vr` in order to create the new coned simplex.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1469,6 +1472,124 @@ class Simplex_tree {
...
+    this->minval_ = std::numeric_limits<double>::max();
+    this->maxval_ = std::numeric_limits<double>::min();
+    for (auto sh : this->skeleton_simplex_range(0)) {
+      double f = this->filtration(sh);
+      this->minval_ = std::min(this->minval_, f);
+      this->maxval_ = std::max(this->maxval_, f);
+      maxvert = std::max(*this->simplex_vertex_range(sh).begin(), maxvert);
+    }
+    
+    assert (maxvert < std::numeric_limits<int>::max());
```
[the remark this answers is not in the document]
MathieuCarriere: Sure. FIxed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1469,6 +1472,124 @@ class Simplex_tree {
...
   }
 
+  /** \brief Retrieve good values for extended persistence, and separate the 
+   * diagrams into the ordinary, relative, extended+ and extended- subdiagrams. 
+   * \post This function should be called only if extend_filtration has been called first!
+   * \post The coordinates of the persistence diagram points might be a little different than the
+   * original filtration values due to the internal transformation (scaling to [-2,-1]) that is 
+   * performed on these values during the computation of extended persistence.
+   * @param[in] dgm Persistence diagram obtained after calling this->extend_filtration 
+   * and this->get_persistence.
```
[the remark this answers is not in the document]
MathieuCarriere: Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: Ah, right, I hadn't looked far enough. Could you declare vr before the loop and replace its current declaration with `vr.clear()`, to reduce allocations a bit?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,120 @@ class Simplex_tree {
...
+      auto sh = this->find(vr);
+
+      // Create cone on simplex
+      vr.push_back(maxvert);
+      if (this->dimension(sh) == 0){
+        // Assign ascending value between -2 and -1 to vertex
+        double v = this->filtration(sh);
+        this->assign_filtration(sh, -2 + (v-this->minval_)/(this->maxval_-this->minval_));
+        // Assign descending value between 1 and 2 to cone on vertex
+        this->insert_simplex(vr, 2 - (v-this->minval_)/(this->maxval_-this->minval_));
```
mglisse: I would find it more readable if you stored (v-min)/(max-min) in a variable, so the reader doesn't have to check if this is exactly the same formula of if there is some subtlety (min vs max, sign...).

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,120 @@ class Simplex_tree {
...
+   * computing the persistent homology of the extended simplicial complex.
+   * \post Note that this code creates an extra vertex internally, so you should make sure that
+   * the Simplex tree does not contain a vertex with the largest Vertex_handle. 
+   */
+  void extend_filtration() {
+
+    // Compute maximum and minimum of filtration values
+    int maxvert = std::numeric_limits<int>::min();
+    this->minval_ = std::numeric_limits<double>::max();
+    this->maxval_ = std::numeric_limits<double>::min();
```
mglisse: `lowest`, not `min`. `min` gives you the smallest **positive** double.
Shouldn't all the `double` be `Filtration_value`?

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: Ok for now, but please file an issue once this branch is merged.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,120 @@ class Simplex_tree {
     }
   }
 
+  /** \brief Retrieve good values for extended persistence, and separate the 
+   * diagrams into the ordinary, relative, extended+ and extended- subdiagrams. 
+   * \post This function should be called only if extend_filtration has been called first!
```
mglisse: That looks more like a precondition (`\pre`), something that must be satisfied before the call, than a postcondition, that would be satisfied after the call.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,120 @@ class Simplex_tree {
...
+   * \post The coordinates of the persistence diagram points might be a little different than the
+   * original filtration values due to the internal transformation (scaling to [-2,-1]) that is 
+   * performed on these values during the computation of extended persistence.
+   * @param[in] dgm Persistence diagram obtained after calling this->extend_filtration, 
+   * this->initialize_filtration, and this->compute_persistent_cohomology.
```
mglisse: There is no function compute_persistent_cohomology in Simplex_tree.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,120 @@ class Simplex_tree {
...
+      double px = dgm[i].second.first; 
+      double py = dgm[i].second.second;
+      if(std::isinf(py))  continue;
+      else{
+        if ((px <= -1) & (py <= -1)){
+          x = minval_ + (maxval_-minval_)*(px + 2); 
+          y = minval_ + (maxval_-minval_)*(py + 2); 
+          new_dgm[0].push_back(std::make_pair(h, std::make_pair(x,y))); 
+        }
+        if ((px >=  1) & (py >=  1)){
```
mglisse:
```suggestion
        else if ((px >=  1) & (py >=  1)){
```
Just for the principle.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: Actually, there are many more...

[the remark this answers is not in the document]
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Thanks. Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Agreed.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Sorry, got confused. The doc should be better now.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: One more in numeric_limits?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,119 @@ class Simplex_tree {
...
+    }
+    return new_dgm;
+  }
+
+  /** \brief Extend filtration for computing extended persistence. 
+   * This function only uses the filtration values at the 0-dimensional simplices, 
+   * and computes the extended persistence diagram induced by the lower-star filtration 
+   * computed with these values. 
+   * \post Note that after calling this function, the filtration 
+   * values are actually modified. The function compute_extended_persistence_subdiagrams
```
mglisse: With the proper quotes and/or parentheses, doxygen can generate a hyperlink to the doc of that function.

MathieuCarriere: Cool, thanks!

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
MathieuCarriere: Thanks for spotting it. Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: (thinking aloud)
In a sense, what we would like to say is that this function is meant to take as argument the output of Persistent_cohomology_interface::get_persistence. Which would indicate that it might be a better fit somewhere in the python interface (possibly Simplex_tree_interface, I didn't think carefully).
However, Simplex_tree, as long as it provides extend, should also provide some way to get the original filtration values back. It (`decode_extended_filtration`?) could take as input either a Simplex_handle or a Filtration_value (and Extended_filtration_data if that appears), and return both the corresponding true filtration value, and whether it is UP or DOWN. Then the python wrapper would call that in a loop (it isn't clear if it should use the output of get_persistence, it might want to do its own loop like persistence_pairs or intervals_in_dimension do). And a C++ user could use this more basic function for other things than translating this very specific format of persistence diagram.

What do you think?

mglisse:
Sorry for starting with the details before moving to the higher-level design. I think I can see the end of the tunnel from here :smiley: 
I don't foresee new comments, just discussions on the existing ones.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -126,6 +126,8 @@ class Simplex_tree {
  private:
   typedef typename Dictionary::iterator Dictionary_it;
   typedef typename Dictionary_it::value_type Dit_value_t;
+  Filtration_value minval_;
+  Filtration_value maxval_;
```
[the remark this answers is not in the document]
MathieuCarriere: That's a great suggestion! I created an `Extended_filtration_data` struct (that is created and stored automatically in `Simplex_tree_interface`) which contains these two values.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1482,119 @@ class Simplex_tree {
...
 
+  /** \brief Retrieve good values for extended persistence, and separate the 
+   * diagrams into the ordinary, relative, extended+ and extended- subdiagrams. 
+   * \pre This function should be called only if `extend_filtration()` has been called first!
+   * \post The coordinates of the persistence diagram points might be a little different than the
+   * original filtration values due to the internal transformation (scaling to [-2,-1]) that is 
+   * performed on these values during the computation of extended persistence.
+   * @param[in] dgm Persistence diagram obtained after calling `extend_filtration()`, 
+   * `initialize_filtration()`, and `Gudhi::persistent_cohomology::Persistent_cohomology< FilteredComplex, CoefficientField >::compute_persistent_cohomology()`.
+   * @return A vector of four persistence diagrams. The first one is Ordinary, the
```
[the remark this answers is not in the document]
MathieuCarriere: Yeah, I don't know how to use arrays and tuples in cython...

src/python/test/test_simplex_tree.py
```diff
@@ -250,6 +250,89 @@ def test_make_filtration_non_decreasing():
...
+        ([0, 6], 2.0), 
+        ([0, 2, 6], 2.0), 
+        ([0, 3, 6], 2.0)
+    ]
+
+
+    dgm = st.persistence()
+    L = st.compute_extended_persistence_subdiagrams(dgm)
+    assert L == [
+        [(0, (1.9999999999999998, 2.9999999999999996))],
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks! Done.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,48 @@ cdef class SimplexTree:
...
+            retrieves the original values.
+
+        .. note::
+
+            Note that this code creates an extra vertex internally, so you should make sure that
+            the Simplex_tree does not contain a vertex with the largest Vertex_handle. 
+        """
+        return self.get_ptr().extend_filtration()
+
+    def compute_extended_persistence_subdiagrams(self, dgm):
```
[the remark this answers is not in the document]
MathieuCarriere: That's right. There is no arguments now.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
MathieuCarriere: Actually I modified the API a little bit so that there is no need to call `SimplexTree.persistence()` first. You can do it, but if you don't, it's run internally.

As for a "decoding" function, that would also be nice. Do you mean a function that would loop over all (original, i.e., not coned) simplices, and output the corresponding original filtration values?

MathieuCarriere:
> Sorry for starting with the details before moving to the higher-level design.

No worries, of course! And thanks a lot for all of your feedback!!

src/python/test/test_simplex_tree.py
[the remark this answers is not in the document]
mglisse: Ah, right, the format is too complicated to pass it directly to pytest.approx :disappointed: It happily takes a numpy array, but not nested lists of tuples. I don't have any idea how to make it nicer (except for changing the format of the output), so I guess the test is good like that.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1490,34 +1494,35 @@ class Simplex_tree {
    * performed on these values during the computation of extended persistence.
    * @param[in] dgm Persistence diagram obtained after calling `extend_filtration()`, 
    * `initialize_filtration()`, and `Gudhi::persistent_cohomology::Persistent_cohomology< FilteredComplex, CoefficientField >::compute_persistent_cohomology()`.
+   * @param[in] efd Structure containing the minimum and maximum values of the original filtration
```
mglisse: I think the main defining trait of this parameter is that it is "what extend_filtration returned".

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1535,23 +1540,23 @@ class Simplex_tree {
...
-   * values are actually modified. The function `compute_extended_persistence_subdiagrams()` 
+   * values are actually modified. The function `extended_persistence_subdiagrams()` 
    * retrieves the original values and separates the extended persistence diagram points 
    * w.r.t. their types (Ord, Rel, Ext+, Ext-) and should always be called after 
    * computing the persistent homology of the extended simplicial complex.
    * \pre Note that this code creates an extra vertex internally, so you should make sure that
    * the Simplex tree does not contain a vertex with the largest Vertex_handle. 
    */
-  void extend_filtration() {
+  Extended_filtration_data extend_filtration() {
```
mglisse: In the doc, mention that it returns information needed by the other function.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: "decoding": let me rephrase what I said above, the basic functionality that we expect from Simplex_tree, after calling extend_filtration, is the ability to interpret the current Simplex_tree, and in particular its filtration values. That is, something like:
```c++
std::pair<Filtration_value,bool> decode_extended_filtration(Filtration_value f, Extended_filtration_data const& dat);
```
(or Simplex_handle as first argument, not sure)
where the output has both the original filtration value, and something that says if this was the value in the upward pass or the downward (aka relative) pass. In order to account for the apex, the bool may not be sufficient, it might need some `enum Extended_value_kind { UP, DOWN, EXTRA}` (where we would document that there is always exactly one EXTRA vertex, and that it is never paired in persistence). Or the bool might be enough if we document that in persistence, there is just one infinite interval, defined by an unmatched vertex, that this is not one of the original vertices but an extra one added for technical reasons, and that this function doesn't work for the filtration value of that vertex, so the user is responsible for not calling the function on it.
Ok, that's the basic functionality, and a C++ user should be able to make sense of the output of get_persistent_pairs or intervals_in_dimension (not persistent_betti_numbers, but let's ignore that).

The format `std::vector<std::pair<int, std::pair<Filtration_value, Filtration_value>>>` is something that basically only makes sense in the python interface because get_persistence returns that. In Simplex_tree, if we wanted to provide some function that acts on a whole diagram, it would be more natural if it took as input `vector<Persistent_interval>` for instance (the output of get_persistent_pairs). For that reason, I think we should move extended_persistence_subdiagrams to Simplex_tree_interface (or somewhere else in the python interface). And once it is there, on a similar level to get_persistence, you may not want to use the output of get_persistence and convert it, it may be nicer to work from get_persistent_pairs.

(Note #255 for the future, I hope that we won't have to build this strange list[tuple[int,tuple[float,float]]] forever)

I hope this was clearer than the previous message.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,66 @@ cdef class SimplexTree:
...
+
+        :param homology_coeff_field: The homology coefficient field. Must be a
+            prime number. Default value is 11.
+        :type homology_coeff_field: int.
+        :param min_persistence: The minimum persistence value to take into
+            account (strictly greater than min_persistence). Default value is
+            0.0.
+            Sets min_persistence to -1.0 to see all values.
+        :type min_persistence: float.
+        :param persistence_dim_max: If true, the persistent homology for the
```
mglisse: The default of False is mostly because of Rips, alpha-complex, etc, where homology in the maximal dimension is seldom interesting. Here, we know that we are dealing with a lower-star filtration. Do you expect that we will usually want homology in the maximal dimension or not? If the complex is a graph, after coning, the maximal dimension is 2, but I don't see how we could ever have any H_2 in the cone. So maybe we shouldn't provide the option at all?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1486,124 @@ class Simplex_tree {
...
+   * second one is Relative, the third one is Extended+ and the fourth one is Extended-.
+   * See section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z for a description of these subtypes.
+   */
+  std::vector<std::vector<std::pair<int, std::pair<Filtration_value, Filtration_value>>>> extended_persistence_subdiagrams(const std::vector<std::pair<int, std::pair<Filtration_value, Filtration_value>>>& dgm, const Extended_filtration_data& efd){
+    std::vector<std::vector<std::pair<int, std::pair<Filtration_value, Filtration_value>>>> new_dgm(4); 
+    Filtration_value x, y;
+    Filtration_value minval = efd.minval;
+    Filtration_value maxval = efd.maxval;
+    for(unsigned int i = 0; i < dgm.size(); i++){ 
+      int h = dgm[i].first;
```
mglisse: I wonder if it is more common to keep the dimension, or to correct it, say use h-1 as dimension for features in the relative pass?

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: Hmm, maybe there is one other element that is non-trivial to interpret for the C++ user: the dimension of features.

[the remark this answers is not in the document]
MathieuCarriere: Aaaah, I understand now. I like this idea. I've implemented the basic functionality in C++ and the python wrapper. Tell me if this is what you had in mind.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Agreed. Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Done.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: Yeah now that you mention it, it makes a lot of sense. Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: I would rather keep it like this, and maybe refer to something, for instance the symmetry theorem in https://link.springer.com/article/10.1007/s10208-008-9027-z ?

mglisse: I just wanted to check if the coning did something weird to the dimension. In a graph, a downward branch will appear as dim 0 while an upward branch appears as dim 1, which matches what relative homology says. So this looks good as is.

src/python/gudhi/simplex_tree.pyx
mglisse: Er, shouldn't we pass `False` if there is never any homology in the maximum dimension, to save computation time?

MathieuCarriere: Oops, yes you are right. Done.

mglisse:
Question by @VincentRouvreau during the meeting (rephrased): why is this a member function of Simplex_tree, when it could be done from a free function (which can be put in a separate file or even module)? Nerve_GIC proves that it is possible.
I believe the coning of the complex is an operation that can benefit from a specialized implementation. The current implementation is functional but wastes time (and temporarily some space), and I asked Mathieu to file an issue about rewriting it once the PR is merged (I want to be reasonable and I prefer getting this feature soon with code that is a bit slower than optimal than waiting too long, we can always rewrite it later if needed).
One possibility would be to have Simplex_tree provide a pure coning function, that does the costly part of enlarging the complex (not sure what it would do about filtration values), and then have some simple functions on the side (in a separate file) that call the coning function and rewrite filtration values as expected for extended persistence. That could be a miniature module, although it would likely be clearer to document it in the same module as the Simplex_tree. Unless we expect to build cones for other applications, I am not convinced it is worth the trouble.
Then there is the python part, which is allowed to have a different interface. But for now, the python interface tends to bloat SimplexTree by including everything in it (persistence), so I don't think this makes it much worse.
There, people can debate now...

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,62 @@ cdef class SimplexTree:
...
+        return self.get_ptr().compute_extended_filtration()
+
+    def extended_persistence(self, homology_coeff_field=11, min_persistence=0):
+        """This function retrieves good values for extended persistence, and separate the diagrams 
+        into the Ordinary, Relative, Extended+ and Extended- subdiagrams.
+
+        :param homology_coeff_field: The homology coefficient field. Must be a
+            prime number. Default value is 11.
+        :type homology_coeff_field: int.
+        :param min_persistence: The minimum persistence value to take into
```
mglisse: This seems to apply to modified filtration values. That doesn't matter for 0 or negative values, but it does matter for positive values. I don't know if we should divide what the user passes by max-min, or document this, possibly mention that extended intervals are all kept.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,62 @@ cdef class SimplexTree:
...
+            account (strictly greater than min_persistence). Default value is
+            0.0.
+            Sets min_persistence to -1.0 to see all values.
+        :type min_persistence: float.
+        :returns: A vector of four persistence diagrams. The first one is Ordinary, the second one is Relative, the third one is Extended+ and the fourth one is Extended-. See section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z for a description of these subtypes.
```
mglisse: I think cython turns the vector into a list? You may want to check it once manually with `type(...)` and update the doc if needed.
"of four persistence diagrams" in the format described in `persistence()`.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,62 @@ cdef class SimplexTree:
...
+
+        .. note::
+
+            This function should be called only if :func:`extend_filtration()<gudhi.SimplexTree.extend_filtration>`,
+            :func:`initialize_filtration()<gudhi.SimplexTree.initialize_filtration>`, 
+            and (optionally) :func:`persistence()<gudhi.SimplexTree.persistence>` have been called first!
```
mglisse: I think `initialize_filtration` is optional as well, it is called automatically somewhere from persistence(). I think we might as well remove those 2 lines.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,62 @@ cdef class SimplexTree:
...
+        """
+        cdef vector[pair[int, pair[double, double]]] persistence_result
+        if self.pcohptr == NULL:
+            self.pcohptr = new Simplex_tree_persistence_interface(self.get_ptr(), False)
+            if self.pcohptr != NULL:
+                self.pcohptr.get_persistence(homology_coeff_field, min_persistence)
+        if self.pcohptr != NULL:
+            pairs = self.pcohptr.persistence_pairs()
+            persistence_result = [(len(splx1)-1, [self.filtration(splx1), self.filtration(splx2)]) for [splx1, splx2] in pairs]
+        return self.get_ptr().compute_extended_persistence_subdiagrams(persistence_result)
```
mglisse: This doesn't affect the API so it is ok for now, but in the future I'd rather avoid going through python's persistence_pairs (we create the diagram in C++, convert it to python, convert it back to C++, where we create the extended diagrams, which we then convert to python, that's a bit much), compute_extended_persistence_subdiagrams should be able to read the diagram at pcohptr in C++.

mglisse:
In my view (others may have other ideas), the main remaining question is about the return type. It continues with the inconvenient format of persistence(), and that makes sense, but I'd rather see that format lose its dominance and see more numpy arrays, as in intervals_in_dimension. For several diagrams, we can use a list of such arrays (one per dimension), or a dict (should we index by pairs (extended_type,dimension) or have a hierarchy with a dict of lists or the reverse?), or merge the arrays and store the dimension (and possibly the type) in an extra column (possibly encoded as a float).
I think that's a separate issue though, this PR is following existing practice, can be merged with that interface, and we can handle extended at the same time as regular persistence when we get to updating interfaces.

mglisse:
One thing I thought of while reviewing the PR: a possible interface for python would be that calling extend_filtration triggers an "extended" mode internally, and then later calls to persistence(), intervals_in_dimension(), etc return the extended persistence diagrams. That's much more intrusive though, the current approach seems simpler and more maintainable, with extended persistence code segregated in 2 functions.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -100,6 +102,12 @@ class Simplex_tree {
     void assign_key(Simplex_key);
     Simplex_key key() const;
   };
+  struct Extended_filtration_data {
+    Filtration_value minval;
+    Filtration_value maxval;
+    Extended_filtration_data(){}
+    Extended_filtration_data(Filtration_value vmin, Filtration_value vmax){ minval = vmin; maxval = vmax; }
```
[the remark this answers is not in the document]
MathieuCarriere: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1362,6 +1370,7 @@ class Simplex_tree {
       // Replacing if(f<max) with if(!(f>=max)) would mean that if f is NaN, we replace it with the max of the children.
       // That seems more useful than keeping NaN.
       if (!(simplex.second.filtration() >= max_filt_border_value)) {
+
```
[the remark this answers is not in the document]
MathieuCarriere: Fixed it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1479,6 +1488,108 @@ class Simplex_tree {
...
+    Filtration_value minval = efd.minval;
+    Filtration_value maxval = efd.maxval;
+    if (f >= -2 && f <= -1){
+      p.first = minval + (maxval-minval)*(f + 2); p.second = UP;
+    }
+    else if (f >= 1 && f <= 2){
+      p.first = minval - (maxval-minval)*(f - 2); p.second = DOWN;
+    }
+    else{
+      p.first = -3; p.second = EXTRA;
```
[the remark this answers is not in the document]
MathieuCarriere: Agreed. Moved to quiet_NaN()

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -87,6 +87,8 @@ class Simplex_tree {
   /* \brief Set of nodes sharing a same parent in the simplex tree. */
   typedef Simplex_tree_siblings<Simplex_tree, Dictionary> Siblings;
 
+  enum Extended_simplex_type {UP, DOWN, EXTRA};
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks for the tip. Done.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -396,6 +396,62 @@ cdef class SimplexTree:
...
+
+            Note that after calling this function, the filtration 
+            values are actually modified within the Simplex_tree. 
+            The function :func:`extended_persistence()<gudhi.SimplexTree.extended_persistence>`
+            retrieves the original values.
+
+        .. note::
+
+            Note that this code creates an extra vertex internally, so you should make sure that
+            the Simplex_tree does not contain a vertex with the largest Vertex_handle.
```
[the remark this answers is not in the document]
MathieuCarriere: Sure. Done.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: That's right. I modified the code so that min_persistence filtering is applied to abs(x-y) on the recovered filtration values.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: Indeed. Fixed it.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: Done.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: Mmmh I don't know if there is a way to get the persistence diagram without recomputing persistence. Anyway, I removed the call to persistence_pairs.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: Oh, you moved it outside of the class, that wasn't necessary. Well, I guess it is also ok there, fine.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -42,6 +42,20 @@
...
+/**
+ * \class Extended_simplex_type Simplex_tree.h gudhi/Simplex_tree.h
+ * \brief Extended simplex type data structure for representing the type of simplices in an extended filtration.
+ *
+ * \details The extended simplex type can be either UP (which means
+ * that the simplex was present originally, and is thus part of the ascending extended filtration), DOWN (which means
+ * that the simplex is the cone of an original simplex, and is thus part of the descending extended filtration) or
+ * EXTRA (which means the simplex is the cone point).
+ *
+ * Details may be found in section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z.
```
mglisse: A reminder that we do have access to bibtex in doxygen.

src/python/gudhi/simplex_tree.pyx
mglisse: `get_persistent_pairs()` gives access to the already computed diagram (that's what's used internally by `persistence_pairs()` for instance).

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -42,6 +42,20 @@
...
+ *
+ */
+enum class Extended_simplex_type {UP, DOWN, EXTRA};
```
mglisse: For some reason this does not appear in the documentation :confused:

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Thanks. I just updated the biblio.

src/python/include/Simplex_tree_interface.h
```diff
@@ -124,31 +123,34 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
 
   void compute_extended_filtration() {
     this->efd = this->extend_filtration();
+    this->initialize_filtration();
```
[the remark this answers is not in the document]
MathieuCarriere: For some reason, if `initialize_filtration()` hasn't been called yet, then calling `get_filtration()` in python gives me a segfault (which makes the test fail)... I'm not sure exactly why, so I decided to call `initialize_filtration()` right away.

src/Simplex_tree/include/gudhi/Simplex_tree.h
MathieuCarriere: Are you sure? I see it when I build the doc locally. Although I don't understand why it doesn't show to be part of the Gudhi namespace.

mglisse: Uh, yes, I do see it now. Was it that it didn't appear yesterday? Or I was confused with Extended_filtration_data, which indeed doesn't appear in the doc, but that's on purpose, it is opaque and users just have to know to pass it from one function to the other.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -427,7 +427,7 @@ cdef class SimplexTree:
             0.0.
             Sets min_persistence to -1.0 to see all values.
         :type min_persistence: float.
-        :returns: A list of four persistence diagrams in the format described in :func:`persistence()<gudhi.SimplexTree.persistence>`. The first one is Ordinary, the second one is Relative, the third one is Extended+ and the fourth one is Extended-. See section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z for a description of these subtypes.
+        :returns: A list of four persistence diagrams in the format described in :func:`persistence()<gudhi.SimplexTree.persistence>`. The first one is Ordinary, the second one is Relative, the third one is Extended+ and the fourth one is Extended-. See and https://link.springer.com/article/10.1007/s10208-008-9027-z and section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z for a description of these subtypes.
```
mglisse: The sentence looks strange with a lot of "and". Pasto? Or the first "and" should be "both"?

src/python/include/Simplex_tree_interface.h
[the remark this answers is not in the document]
mglisse: Normally, this kind of issue happens if initialize_filtration was called a first time, then the complex was modified, and then you call get_filtration. I'll give it a look.

src/python/gudhi/simplex_tree.pyx
MathieuCarriere: Oops, yeah bad copy-paste. Fixed it.

src/python/include/Simplex_tree_interface.h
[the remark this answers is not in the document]
mglisse: Yes, that's exactly what happens in your test. Let's say we'll handle that as part of #244 .

mglisse:
I merged it. But if people have comments, they are still welcome!

PR #218, opened 2020-02-04
Fix #186 - Link with TBB for all simplex tree user

### LINKED ISSUE #186: [CMake] Link with TBB in debug mode
If I use `-DCMAKE_BUILD_TYPE=Debug` with cmake, linking the example mini_simplex_tree.cpp fails.
<pre>/usr/bin/c++   -Wall -pedantic -g  -rdynamic CMakeFiles/Simplex_tree_example_mini_simplex_tree.dir/mini_simplex_tree.cpp.o  -o Simplex_tree_example_mini_simplex_tree -lmpfr -lgmp 
/usr/bin/ld: CMakeFiles/Simplex_tree_example_mini_simplex_tree.dir/mini_simplex_tree.cpp.o: in function `tbb::interface7::task_arena::current_thread_index()':
/usr/include/tbb/task_arena.h:372: undefined reference to `tbb::interface7::internal::task_arena_base::internal_current_slot()'
collect2: error: ld returned 1 exit status</pre>

Even if we are not really using TBB, if TBB is found, we compile this file with GUDHI_USE_TBB and include TBB headers, so we should link with TBB for consistency.
I get the same error for example_witness_complex_off.cpp, example_strong_witness_complex_off.cpp and example_witness_complex_sphere.cpp.

VincentRouvreau:


src/common/doc/installation.h
```diff
@@ -204,6 +204,8 @@ make doxygen
  * Nerve_GIC/Nerve.cpp</a>
  * \li <a href="_nerve__g_i_c_2_voronoi_g_i_c_8cpp-example.html">
  * Nerve_GIC/VoronoiGIC.cpp</a>
+ * \li <a href="_simplex_tree_2mini_simplex_tree_8cpp-example.html">
+ * Simplex_tree/mini_simplex_tree.cpp</a>
```
mglisse: I wonder about this. The heading says "Having Intel&reg; TBB installed is recommended to parallelize and accelerate some GUDHI computations. The following examples/utilities are using Intel&reg; TBB if installed". This example gets linked with TBB, but that's artificial, it doesn't actually use TBB. So it depends what the goal of this list is. If the goal was to see what may have a dependency on TBB, then we should include it. If the goal is to see what examples benefit from TBB, which seems to be the case, then maybe we should not list it? What do you think?
(the rest of the patch looks fine, thanks)

VincentRouvreau: Ok, I see what you mean.
I am ok to remove this part.

PR #220, opened 2020-02-05
Add gudhi.representations in setup.py packages

### LINKED ISSUE #219: [Python - representations] Python representations module is not installed by make install
It includes conda and debian package.

VincentRouvreau:
Fix #219

### DIFF
--- a/src/python/setup.py.in
+++ b/src/python/setup.py.in
@@ -44,7 +44,7 @@ for module in modules:
 
 setup(
     name = 'gudhi',
-    packages=["gudhi",],
+    packages=["gudhi","gudhi.representations"],
     author='GUDHI Editorial Board',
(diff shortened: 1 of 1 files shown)


mglisse:
ok, although I wonder if we should use something like find_packages.

VincentRouvreau:
> ok, although I wonder if we should use something like find_packages (later).

If we go with `find_packages` we are moving away from rolling back to `distutils`.
Accordingly to this [Python packaging guide](https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/introduction.html#current-state-of-packaging), they advise to use preferably `distutils`.
I don't know if the advise is good as the guide is from 2010...

mglisse:
https://packaging.python.org/guides/tool-recommendations/ seems more up to date.

PR #223, opened 2020-02-08
perslay

MathieuCarriere:
Integration of PersLay to Gudhi


MathieuCarriere:
I have an issue with the doc. `perslay_channel` takes a huge number of parameters (described in the `perslay.md` file, so I don't know what is the best option for rendering. Maybe something like what `matplotlib` does could be nice: see "Other parameters" in https://matplotlib.org/api/_as_gen/matplotlib.pyplot.plot.html

mglisse:
Shouldn't this be inside `representations/`? It takes diagrams as input and learns some kind of vectorization on them...

src/python/doc/perslay.rst
```diff
@@ -0,0 +1,65 @@
...
+
+==============
+PersLay manual
+==============
+
+.. include:: perslay_sum.inc
+
+This module, originally available at https://github.com/MathieuCarriere/perslay, aims at bridging the gap between persistence diagrams and deep learning, by providing the implementation in tensorflow of a neural network layer for persistence diagrams, called PersLay, defined in https://arxiv.org/abs/1904.09378. More specifically, it provides code for reproducing most of the persistence diagram representations in the literature, and optimize their parameters with a neural network.
+
+Since tensorflow requires its inputs to be fixed-size matrices, persistence diagrams are padded  to a fixed number of points, with a third boolean coordinate for each points assessing whether the point is real or just a dummy point added after padding. A diagram is thus represented as a tensorflow tensor of shape (n,3), as can be obtained from :func:`~gudhi.SimplexTree.persistence_intervals_in_dimension` and :func:`~gudhi.representations.preprocessing.Padding` for instance. Points at infinity are represented as a tensorflow tensors of shape (n,2), storing only the birth times.
```
mglisse: You should move to pytorch, it doesn't have such a strict structure and lets you use diagrams of varying size :wink:

src/python/gudhi/perslay.py
```diff
@@ -0,0 +1,263 @@
...
+        if c == "b":
+            vector = tf.layers.batch_normalization(vector)
+        if c == "d":
+            vector = tf.nn.dropout(vector, dropout_value)
+        if c == "r":
+            vector = tf.nn.relu(vector)
+    return vector
+
+# Vectorization implementing DeepSet architecture
+def permutation_equivariant_layer(inp, dimension, perm_op, L_init, G_init, bias_init, L_const, G_const, bias_const, train_vect):
```
mglisse: As an outsider, I usually see layers described as classes, although functions work as well. Are classes only useful if you store weights? Is there some other reason to pick one over the other?

src/python/gudhi/perslay.py
```diff
@@ -0,0 +1,263 @@
...
+    mu = tf.get_variable("m", shape=[1, 1, dimension_before, num_elements], initializer=mean_init, trainable=train_vect)      if not mean_const      else tf.get_variable("m", initializer=tf.reshape(mean_init, [1, 1, dimension_before, num_elements]))
+    sg = tf.get_variable("s", shape=[1, 1, dimension_before, num_elements], initializer=variance_init, trainable=train_vect)  if not variance_const  else tf.get_variable("s", initializer=tf.reshape(variance_init, [1, 1, dimension_before, num_elements]))
+    bc_inp = tf.expand_dims(inp, -1)
+    return tf.exp(tf.reduce_sum(-tf.multiply(tf.square(bc_inp - mu), tf.square(sg)), axis=2))
+
+# Vectorizations implementing persistence landscapes
+def landscape_layer(inp, num_samples, sample_init, sample_const, train_vect):
+    """ Landscape PersLay """
+    sp = tf.get_variable("s", shape=[1, 1, num_samples], initializer=sample_init, trainable=train_vect) if not sample_const else tf.get_variable("s", initializer=tf.reshape(sample_init, [1,1,num_samples]))
+    return np.sqrt(2) * tf.maximum( .5 * (inp[:, :, 1:2] - inp[:, :, 0:1]) - tf.abs(sp - .5 * (inp[:, :, 1:2] + inp[:, :, 0:1])), np.array([0]))
```
mglisse: As I feared, we need to duplicate a lot of code for tensorflow, and most likely we will need yet more copies for other frameworks :disappointed: 
The implementation here looks much shorter than your other one in vector_methods. What drawbacks does it have? Is it just slower because it always evaluates all diagram points at every sample position (bad with many points close to the diagonal)? On the other hand, using the vector syntax makes it win a big constant factor over the python double loop.

src/python/gudhi/perslay.py
```diff
@@ -0,0 +1,263 @@
...
+def entropy_layer(inp, theta, num_samples, sample_init, sample_const, train_vect):
+    """ Entropy PersLay
+    WARNING: this function assumes that padding values are zero
+    """
+    bp_inp = tf.einsum("ijk,kl->ijl", inp, tf.constant(np.array([[1.,-1.],[0.,1.]], dtype=np.float32)))
+    sp = tf.get_variable("s", shape=[1, 1, num_samples], initializer=sample_init, trainable=train_vect) if not sample_const else tf.get_variable("s", initializer=tf.reshape(sample_init, [1,1,num_samples]))
+    L, X, Y = bp_inp[:, :, 1:2], bp_inp[:, :, 0:1], bp_inp[:, :, 0:1] + bp_inp[:, :, 1:2]
+    LN = tf.multiply(L, 1. / tf.expand_dims(tf.matmul(L[:,:,0], tf.ones([L.shape[1],1])), -1))
+    entropy_terms = tf.where(LN > 0., -tf.multiply(LN, tf.log(LN)), LN)
+    return  tf.multiply(entropy_terms, 1. / ( 1. + tf.exp( -theta * (.5*(Y-X) - tf.abs(sp - .5*(Y+X))) )  ))
```
mglisse: The implementation of several functions is hard to follow, could you add some comments along the way, so that future readers can understand what the magic incantations (with einsum for instance) yield?

src/python/gudhi/perslay.py
```diff
@@ -0,0 +1,263 @@
...
+
+# Vectorizations implementing persistence landscapes
+def landscape_layer(inp, num_samples, sample_init, sample_const, train_vect):
+    """ Landscape PersLay """
+    sp = tf.get_variable("s", shape=[1, 1, num_samples], initializer=sample_init, trainable=train_vect) if not sample_const else tf.get_variable("s", initializer=tf.reshape(sample_init, [1,1,num_samples]))
+    return np.sqrt(2) * tf.maximum( .5 * (inp[:, :, 1:2] - inp[:, :, 0:1]) - tf.abs(sp - .5 * (inp[:, :, 1:2] + inp[:, :, 0:1])), np.array([0]))
+
+# Vectorizations implementing Betti curves
+def betti_layer(inp, theta, num_samples, sample_init, sample_const, train_vect):
+    """ Betti PersLay """
```
mglisse: The documentation of these functions is very minimalistic. If you are just repeating the name of the function, you might as well omit the docstring.
The question is: are those functions private implementation details, and perslay is the only interface? Or should those functions be documented bricks, that users can connect as they want in their own pipeline (possibly involving derivatives of diagrams)?
It might be ok to start with them as private and promote them to public later, I haven't thought about it much, but if you have some idea about longer term goals...

src/python/gudhi/perslay.py
```diff
@@ -0,0 +1,263 @@
...
+        if c == "r":
+            vector = tf.nn.relu(vector)
+    return vector
+
+# Vectorization implementing DeepSet architecture
+def permutation_equivariant_layer(inp, dimension, perm_op, L_init, G_init, bias_init, L_const, G_const, bias_const, train_vect):
+    """ DeepSet PersLay """
+    dimension_before, num_pts = inp.shape[2].value, inp.shape[1].value
+    lbda = tf.get_variable("L", shape=[dimension_before, dimension], initializer=L_init, trainable=train_vect)   if not L_const     else tf.get_variable("L", initializer=L_init)
+    b    = tf.get_variable("b", shape=[1, 1, dimension], initializer=bias_init, trainable=train_vect)            if not bias_const  else tf.get_variable("b", initializer=tf.reshape(bias_init, [1, 1, dimension]))
```
mglisse: You have a number of very long lines like that. While the PEP8 limit of 79 characters per line seems too small to me, 211 on this line is too much. Could you consider ways to reduce super-long lines? For instance, this one could easily be split, although it loses the alignment with `A = ...` below. Running `black -l N perslay.py` (where N might be 79, 120, whatever line length) is another way to avoid long lines.
```suggestion
    if bias_const:
        b = tf.get_variable("b", initializer=tf.reshape(bias_init, [1, 1, dimension]))
    else:
        b = tf.get_variable("b", shape=[1, 1, dimension], initializer=bias_init, trainable=train_vect)
```

src/python/doc/perslay.md
```diff
@@ -0,0 +1,135 @@
...
+ 
+In the following description of PersLay parameters, each parameter, or dictionary key, that contains "_init" in its name is optimized and learned by PersLay during training. If you do not want to optimize the vectorization, set the (optional) keys **train_vect** and **train_weight** to False.
+
+  * The following keys are mandatory:
+
+    | **name** | **description** |
+    | --- | --- |
+    | **layer**              | Either "pm", "im", "ls", "bc", "en", "ex", "rt", "rh". Type of the PersLay layer. "im" is for [persistence images](https://arxiv.org/abs/1507.06217), "ls" is for [persistence landscapes](http://www.jmlr.org/papers/volume16/bubenik15a/bubenik15a.pdf), "ex", "rt" and "rh" are for [structure elements](http://jmlr.org/beta/papers/v20/18-358.html), "pm" is for the original DeepSet layer, defined in [this article](https://arxiv.org/abs/1703.06114), "bc" is for [Betti curves](https://www.jstage.jst.go.jp/article/tjsai/32/3/32_D-G72/_pdf) and "en" is for [entropy](https://arxiv.org/abs/1803.08304). |
+    | **perm_op**            | Either "sum", "mean", "max", "topk". Permutation invariant operation. |
```
mglisse: Isn't "reduction" a standard name for this kind of operation?

src/python/doc/perslay.md
```diff
@@ -0,0 +1,135 @@
...
+
+  * the fourth argument is a python dictionary containing the parameters of PersLay. Examples can be found in `tutorialPersLay.ipynb`.
+ 
+In the following description of PersLay parameters, each parameter, or dictionary key, that contains "_init" in its name is optimized and learned by PersLay during training. If you do not want to optimize the vectorization, set the (optional) keys **train_vect** and **train_weight** to False.
+
+  * The following keys are mandatory:
+
+    | **name** | **description** |
+    | --- | --- |
+    | **layer**              | Either "pm", "im", "ls", "bc", "en", "ex", "rt", "rh". Type of the PersLay layer. "im" is for [persistence images](https://arxiv.org/abs/1507.06217), "ls" is for [persistence landscapes](http://www.jmlr.org/papers/volume16/bubenik15a/bubenik15a.pdf), "ex", "rt" and "rh" are for [structure elements](http://jmlr.org/beta/papers/v20/18-358.html), "pm" is for the original DeepSet layer, defined in [this article](https://arxiv.org/abs/1703.06114), "bc" is for [Betti curves](https://www.jstage.jst.go.jp/article/tjsai/32/3/32_D-G72/_pdf) and "en" is for [entropy](https://arxiv.org/abs/1803.08304). |
```
mglisse: Is there a particular reason for choosing such short names? For weights, you don't seem to have trouble spelling "linear" with all its letters.

src/python/doc/perslay.md
```diff
@@ -0,0 +1,135 @@
...
+    | **keep**               | Number of top values to keep. Used only if **perm_op** is "topk". |
+    | **persistence_weight** | Either "linear", "grid", "gmix" or None. Weight function to be applied on persistence diagram points. If "linear", this function is linear with respect to the distance to the diagonal of the point. If "grid", this function is piecewise-constant and defined with pixel values of a grid. If "gmix", this function is defined as a mixture of Gaussians. If None, no weighting is applied. |
+
+Depending on what **persistence_weight** is, the following additional keys are requested:
+
+  * if **persistence_weight** is "linear":
+
+    | **name** | **description** |
+    | --- | --- |
+    | **coeff_init**         | Initializer of the coefficient of the linear weight function. It can be either a single value, or a random initializer from tensorflow, such as tensorflow.random_uniform_initializer(1., 1.). |
```
mglisse: Does the coefficient of the linear weight matter? I would have expected it to just multiply everything by a constant.

src/python/doc/perslay.md
```diff
@@ -0,0 +1,135 @@
...
+Depending on what **persistence_weight** is, the following additional keys are requested:
+
+  * if **persistence_weight** is "linear":
+
+    | **name** | **description** |
+    | --- | --- |
+    | **coeff_init**         | Initializer of the coefficient of the linear weight function. It can be either a single value, or a random initializer from tensorflow, such as tensorflow.random_uniform_initializer(1., 1.). | 
+    | **coeff_const**        | Boolean specifying if **coeff_init** is initialized with a value (True) or randomly with tensorflow (False). |
+ 
+  * if **persistence_weight** is "grid":
```
mglisse: Is the grid parallel to the birth and death axis, or rotated and parallel to the diagonal?

src/python/doc/perslay.md
tlacombe: Approved

src/python/doc/perslay.md
tlacombe: Well, "sort" could be an acceptable `perm_op` I think. `top-k` is just a "take the k largest entries" so it already looks like a sort (if I remember correctly).

src/python/gudhi/perslay.py
tlacombe: I agree that (at least for a first release) these should be private (which does not prevent from providing some documentation).

src/python/doc/perslay.md
```diff
@@ -0,0 +1,135 @@
...
+    | **gmix_v_init**        | Initializer of the bandwidths of the mixture of Gaussians weight function. It can be either a numpy array of values, or a random initializer from tensorflow, such as tensorflow.random_uniform_initializer(1., 1.). |
+    | **gmix_v_const**       | Boolean specifying if **gmix_v_init** is initialized with an array (True) or randomly with tensorflow (False). Used only if **persistence_weight** is "gmix". |
+
+Depending on what **layer** is, the following additional keys are requested:
+
+  * if **layer** is "pm":
+
+    | **name** | **description** |
+    | --- | --- |
+    | **peq**                | Sequence of permutation equivariant operations, as defined in [the DeepSet article](). It is a list of tuples of the form (*dim*, *operation*). Each tuple defines a permutation equivariant function of dimension *dim* and second permutation operation *operation* (string, either "max", "min", "sum" or None). Second permutation operation is optional and is not applied if *operation* is set to None. Example: [(150, "max"), (75, None)]. |
```
tlacombe: Similarly to previous comment from Marc, I feel the names could be more explicit here.

src/python/doc/perslay.md
```diff
@@ -0,0 +1,135 @@
...
+    | **mean_const**        | Boolean specifying if **mean_init** is initialized with a value (True) or randomly with tensorflow (False). |
+    | **r_init**            | Initializer of the threshold of the rational hat structure elements. It can be either a numpy array of values, or a random initializer from tensorflow, such as tensorflow.random_uniform_initializer(3., 3.). | 
+    | **r_const**           | Boolean specifying if **r_init** is initialized with a value (True) or randomly with tensorflow (False). |
+    | **q**                 | Norm parameter. |
+    | **fc_layers**         | Sequence of fully-connected operations to be applied after the permutation invariant operation. It is a list of tuples of the form (*dim*, *pro*, *dropout*). Each tuple defines a fully-connected operation, with dimension *dim* (integer) and processing *pro* (string, e.g. "bdr" ---> batch-norm, dropout, relu). If there is a "d" in string, i.e. dropout, the dropout value can be specified with *dro* (float, default 0.9). Example: [(150,"br"), (75, "bd", 0.85)].|
```
tlacombe: Overall, I think we should redefine what a layer is, to avoid having unrelated variables at the same level ; in my mind this should rather have a tree structure, like
`perslay_param` might decompose in `permutation_operator`, `weight_function` and `vectorization_function`, each of them being independently instantiated.
As such, maybe using a Class would be better.
It might also help to have a clearer doc.

src/python/doc/perslay.rst
```diff
@@ -0,0 +1,65 @@
...
+This example computes the first landscape associated to a persistence diagram with four points. The landscape is evaluated evenly on [-1,11].
+
+.. testcode::
+
+    import numpy as np
+    import tensorflow as tf
+    from gudhi.representations.preprocessing import Padding
+    from gudhi.representations import Landscape, PersistenceImage
+    from gudhi.perslay import perslay_channel
+    import matplotlib.pyplot as plt
```
tlacombe: `matplotlib` import is not needed for the `testcode` is it?

src/python/doc/perslay.rst
```diff
@@ -0,0 +1,65 @@
...
+    # A single diagram with 4 points
+    diags = [np.array([[0.,4.],[1.,2.],[3.,8.],[6.,8.]])]
+    diags = Padding(use=True).fit_transform(diags)
```
tlacombe: Should add some comments in these lines to give more intuitions/information about what we are doing there (the user might not be familiar with `representations.preprocessing`.

src/python/example/perslay_visu.py
```diff
@@ -0,0 +1,161 @@
...
+diagram = tf.placeholder(tf.float32, shape=D.shape)
+feed    = {diagram: D}
+
+perslayParameters = {}
+
+#####################
+# Persistence image #
+#####################
+
+def func(x, y):
```
tlacombe: Perhaps these functions should be private, using `_func`.

src/python/gudhi/perslay.py
tlacombe: After few thoughts, I also agree that a Class formalism might be better (cf comments above in the doc). I am not sure yet.

src/python/gudhi/perslay.py
tlacombe: Agreed. Also, `einsum("ijk, kl-> ijl") is used multiple times, maybe it could be given a name? (e.g. `tfdot` ?)

src/python/doc/perslay.rst
```diff
@@ -0,0 +1,65 @@
...
+
+.. To get rid of WARNING: document isn't included in any toctree
+
+==============
+PersLay manual
+==============
+
+.. include:: perslay_sum.inc
+
+This module, originally available at https://github.com/MathieuCarriere/perslay, aims at bridging the gap between persistence diagrams and deep learning, by providing the implementation in tensorflow of a neural network layer for persistence diagrams, called PersLay, defined in https://arxiv.org/abs/1904.09378. More specifically, it provides code for reproducing most of the persistence diagram representations in the literature, and optimize their parameters with a neural network.
```
VincentRouvreau, line 11: https://arxiv.org/abs/1904.09378
Maybe a new entry in `biblio/bibliography.bib` and a `:cite:` tag in the documentation would have a better documentation rendering.

VincentRouvreau:
Maybe this PR is obsolete because of #402 ?

MathieuCarriere:
YesI think this PR can be closed/deleted (if that's possible)

mglisse:
At the bottom, next to "comment", you should also see a button "close pull request" or "close with comment".

VincentRouvreau:
Closed, as this PR has been reworked on #402

PR #227, opened 2020-02-14
Global improvement of rendering with Python tools

### LINKED ISSUE #27: [python - persistence_graphical tools] It would be nice to accept list of tuple(birth, death) as input
It would be nice to accept list of tuple(birth, death) as input and not only tuples(dimension, tuple(birth, death))

tlacombe:
Did some (hopefully) nice improvements on persistence diagrams/barcode/density plot rendering. Main changes:

- Added option for fontsize (crucial when it comes to produce plot for a paper).
- Added option to change the title (but set the default titles to previous ones).
- Used Latex font as default
- Added a (optional) lightgrey block on the lower half plane. I find the rendering quite nice and it seems commonly used in TDA litterature. 
- Added option to set plot aspects to "equal" for dgm and density.
- [plot dgm] Turned the $\infty$ text into a "true" `axis.yticklabel`, improving readability. 
- [dgm and density] Modified the line draw methods (why did it plot (x, x) for 1000 values of x instead of just a line between (min, min) and (max, max) ?). It doesn't change the rendering but might (slightly) improve the running time.

What could be also done:

- [x] Provide an option to plot diagrams that comes from a different format, such as (N x 2) numpy arrays, as Python implementations in `representations` are mostly using this format if I am correct. [EDIT] Did a quick but hopefully useful change to handle (N x 2) numpy arrays as input for the plot functions. [EDIT VR] Fix #27 
- [x] Add more flexibility features (removing xticklabels, etc.). [EDIT] We actually removed these as the functions return axis ==> the user can do these modifications by him/herself.
- [x] Remove the +\infty line if there is no point at infty.

### DIFF
--- a/src/python/doc/persistence_graphical_tools_user.rst
+++ b/src/python/doc/persistence_graphical_tools_user.rst
@@ -57,6 +57,18 @@ This function can display the persistence result as a diagram:
     ax.set_aspect("equal")  # forces to be square shaped
     plt.show()
 
+Note that (as barcode and density) it can also take a simple `np.array` 
+of shape (N x 2) encoding a persistence diagram (in a given dimension).
+
+.. plot::
+    :include-source:
+    
+    import matplotlib.pyplot as plt
+    import gudhi
+    import numpy as np
+    d = np.array([[0, 1], [1, 2], [1, np.inf]])
+    gudhi.plot_persistence_diagram(d)
+    plt.show()
 
 Persistence density
 -------------------
--- a/src/python/doc/persistence_graphical_tools_user.rst
+++ b/src/python/doc/persistence_graphical_tools_user.rst
@@ -20,7 +20,7 @@ This function can display the persistence result as a barcode:
 .. plot::
    :include-source:
 
-    import matplotlib.pyplot as plot
+    import matplotlib.pyplot as plt
     import gudhi
 
     off_file = gudhi.__root_source_dir__ + '/data/points/tore3D_300.off'
@@ -31,7 +31,7 @@ This function can display the persistence result as a barcode:
     diag = simplex_tree.persistence(min_persistence=0.4)
 
     gudhi.plot_persistence_barcode(diag)
-    plot.show()
+    plt.show()
 
 Show persistence as a diagram
 -----------------------------
@@ -44,15 +44,19 @@ This function can display the persistence result as a diagram:
 .. plot::
    :include-source:
 
-    import matplotlib.pyplot as plot
+    import matplotlib.pyplot as plt
     import gudhi
 
     # rips_on_tore3D_1307.pers obtained from write_persistence_diagram method
     persistence_file=gudhi.__root_source_dir__ + \
         '/data/persistence_diagram/rips_on_tore3D_1307.pers'
-    gudhi.plot_persistence_diagram(persistence_file=persistence_file,
+    ax = gudhi.plot_persistence_diagram(persistence_file=persistence_file,
         legend=True)
-    plot.show()
+    # We can modify the title, aspect, etc.
+    ax.set_title("Persistence diagram of a torus")
+    ax.set_aspect("equal")  # forces to be square shaped
+    plt.show()
+
 
 Persistence density
 -------------------
@@ -65,7 +69,7 @@ If you want more information on a specific dimension, for instance:
 .. plot::
    :include-source:
 
-    import matplotlib.pyplot as plot
+    import matplotlib.pyplot as plt
     import gudhi
     # rips_on_tore3D_1307.pers obtained from write_persistence_diagram method
     persistence_file=gudhi.__root_source_dir__ + \
@@ -75,9 +79,9 @@ If you want more information on a specific dimension, for instance:
         only_this_dim=1)
     pers_diag = [(1, elt) for elt in birth_death]
     # Use subplots to display diagram and density side by side
-    fig, axes = plot.subplots(nrows=1, ncols=2, figsize=(12, 5))
+    fig, axes = plt.subplots(nrows=1, ncols=2, figsize=(12, 5))
     gudhi.plot_persistence_diagram(persistence=pers_diag,
         axes=axes[0])
     gudhi.plot_persistence_density(persistence=pers_diag,
         dimension=1, legend=True, axes=axes[1])
-    plot.show()
+    plt.show()
--- a/src/python/gudhi/persistence_graphical_tools.py
+++ b/src/python/gudhi/persistence_graphical_tools.py
@@ -296,17 +296,6 @@ def plot_persistence_diagram(
         axis_end = max_death + delta / 2
         axis_start = min_birth - delta
 
-        # infinity line and text
-        axes.plot([axis_start, axis_end], [axis_start, axis_end], linewidth=1.0, color="k")
-        axes.plot([axis_start, axis_end], [infinity, infinity], linewidth=1.0, color="k", alpha=alpha)
-        # Infinity label
-        yt = axes.get_yticks()
-        yt = yt[np.where(yt < axis_end)] # to avoid ploting ticklabel higher than infinity
-        yt = np.append(yt, infinity)
-        ytl = ["%.3f" % e for e in yt]  # to avoid float precision error
-        ytl[-1] = r'$+\infty$'
-        axes.set_yticks(yt)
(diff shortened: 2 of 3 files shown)


mglisse:
>     * Provide an option to plot diagrams that comes from a different format, such as (N x 2) numpy arrays, as Python implementations in `representations` are mostly using this format if I am correct.

Issue #27 asks for that. I think @yagu0 (Benjamin) was planning to work on it, you may want to check with him.

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -81,11 +85,18 @@ def plot_persistence_barcode(
     :param axes: A matplotlib-like subplot axes. If None, the plot is drawn on
         a new set of axes.
     :type axes: `matplotlib.axes.Axes`
+    :param fontsize: Fontsize to use in axis.
+    :type fontsize: int
+    :param title: title for the plot.
+    :type title: string
```
VincentRouvreau: Title was one of the reason why I was returning the `axes` (the other reason was for subplotting). One can do:
```python
import matplotlib.pyplot as plot
import gudhi

persistence_file= '/data/persistence_diagram/rips_on_tore3D_1307.pers'
ax = gudhi.plot_persistence_diagram(persistence_file=persistence_file, legend=True)
ax.set_title('pouet')
plot.show()
```
I am ok for both solutions, but I think there are already a lot of arguments for these methods. Maybe we could just document this.

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -183,7 +194,11 @@ def plot_persistence_diagram(
     inf_delta=0.1,
     legend=False,
     colormap=None,
-    axes=None
+    axes=None,
+    aspect_equal=False,
+    fontsize=16,
+    title="Persistence diagram",
+    greyblock=True
```
VincentRouvreau: aspect_equal, font_size and grey_block.
I am clearly faulty on this point but we shall homogenize the names.

src/python/gudhi/persistence_graphical_tools.py
tlacombe: Ah, I didn't noticed `ax` was returned ; it is not straightforward when looking at the user manual. Indeed it seems natural to go the way you suggest, and remove `title` as an argument of the function (I agree it is a bit overloaded).

src/python/gudhi/persistence_graphical_tools.py
tlacombe: Could we keep `fontsize` this way? This is the variable name used by matplotlib functions by default, and I think we should be consistent with it. Having `fontsize=font_size` is a bit weird to me, and a someone used to matplotlib would naturally go for `fontsize=...` by default.

tlacombe:
> Some plots are quite strange with the LaTeX feature:
> 
>     * scales are strange (0.050000000000002, 0.075000000000002, ...) - **Death** is no more displayed
> 
>     * all the titles are not in LaTeX
>       I will send it to you by email.

Should be improved now. (However, Idk how to solve this "latex is not fully rendered" yet. Will investigate... or remove the option).

tlacombe:
> Some plots are quite strange with the LaTeX feature:
> 
>     * scales are strange (0.050000000000002, 0.075000000000002, ...) - **Death** is no more displayed
> 
>     * all the titles are not in LaTeX
>       I will send it to you by email.

Should be improved now. (However, Idk how to solve this "latex is not fully rendered" yet. Will investigate... or remove the option).

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -45,6 +45,13 @@ def __min_birth_max_death(persistence, band=0.0):
     return (min_birth, max_death)
 
 
+def _array_handler(a):
+    if isinstance(a, np.ndarray):
```
VincentRouvreau:
```suggestion
    if isinstance(a[0][1], np.float64) or isinstance(a[0][1], float):
```
Historical accepted format is tuples(dimension, tuple(birth, death)).
What I propose would allow tuple(birth, death) but also numpy.array of shape (n,2).

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -65,9 +70,12 @@ def plot_persistence_barcode(
...
 ):
     """This function plots the persistence bar code from persistence values list
+    , a np.array of shape (N x 2) (representing a diagram 
+    in a single homology dimension), 
     or from a :doc:`persistence file <fileformats>`.
 
-    :param persistence: Persistence intervals values list grouped by dimension.
+    :param persistence: Persistence intervals values list grouped by dimension, 
+                        or np.array of shape (N x 2).
     :type persistence: list of tuples(dimension, tuple(birth, death)).
```
VincentRouvreau: I am proposing this documentation (to modified for the 2 other functions too), to explain that it can be an numpy.ndarray (N x 2), or a classical array:
```suggestion
    :param persistence: Persistence intervals values list. Can be grouped by dimension or not.
    :type persistence: an array of (dimension, array of (birth, death)) or an array of (birth, death).
```

tlacombe: incorporated in last commit

src/python/doc/persistence_graphical_tools_user.rst
```diff
@@ -75,9 +91,9 @@ If you want more information on a specific dimension, for instance:
         only_this_dim=1)
     pers_diag = [(1, elt) for elt in birth_death]
     # Use subplots to display diagram and density side by side
-    fig, axes = plot.subplots(nrows=1, ncols=2, figsize=(12, 5))
+    fig, axes = plt.subplots(nrows=1, ncols=2, figsize=(12, 5))
     gudhi.plot_persistence_diagram(persistence=pers_diag,
         axes=axes[0])
     gudhi.plot_persistence_density(persistence=pers_diag,
         dimension=1, legend=True, axes=axes[1])
```
VincentRouvreau: Maybe remove the greyblock on density (nicer I found) and it will be a use case of this functionnality
```suggestion
    gudhi.plot_persistence_density(persistence=pers_diag,
        dimension=1, legend=True, greyblock=True, axes=axes[1])
```

tlacombe: Do you mean set the Default to `False` ?

VincentRouvreau: My point was only for this particular example, but as you want:
* only for this example
* or default greyblock value is False for density
On this particular example, some colors are hidden by the greyblock.

tlacombe: Ok I get it, maybe it is better with density to set greyblock to `False` by default, as indeed, a density is not requested to be located above the diagonal so the greyblock is not especially meaningful.

mglisse: I didn't follow or test, but is the grey block drawn above or below the density, if both try to paint the same point? (probably related to the order in which they appear in the function)

tlacombe: It is printed above. I think it is just nice to left this option accessible but safer to set it to False by default.

mglisse: Whether it is True of False by default, would printing it below look nicer?

tlacombe: I don't think it makes sense this way: when the density is printed, is overlap the whole square (so it would erase the grey block if the latter is printed before). Furthermore, the density is a gaussian KDE, so it never equals zero ; we would have to pick a threshold (or a `alpha` opacity parameter) and I don't think the rendering would be satisfactory in any case.

Also, it is a matter of taste, but I prefer to have the grey block above so that you clearly see the diagonal (I don't see the point of printing a grey block if below, actually).

mglisse: thanks for the explanations.

tlacombe:
Not ready for merge yet: I noticed that the +infty line is printed even if there is no point at infty in the diagram, which I think can be improved. Will be done asap.

VincentRouvreau:
It seems your development is done ? Tell me if I can merge (and maybe delete your branch ?).

tlacombe:
Yes I think it can be merged (and we can delete the branch).

PR #230, opened 2020-02-24
Modernize ci

VincentRouvreau:


VincentRouvreau:
> All the set_tests_properties really mean "wait until B is finished to run A", right?

Yes that's it. Some tests fail in parallel because a test A requires a test B, that creates a file required by test A for example. This will also allow us to use ninja instead of make.

PR #232, opened 2020-02-26
Try to build CI faster, fix parallel tests and travis doxygen install

VincentRouvreau:


VincentRouvreau:
Cannot repair travis. Let's go with Azure on another PR

PR #233, opened 2020-02-28
Bottleneck and Wasserstein in a metrics module

VincentRouvreau:
A prototype for a metrics Python module.
This module contains :
* Bottleneck
* POT Wasserstein
* Hera Wasserstein

Barycenter can be added the same way.

I will see later how to do some DRY.


tlacombe:
how is this PR supposed to interact with #205 ?

VincentRouvreau:
> how is this PR supposed to interact with #205 ?

@tlacombe Not really related with #205 as it won't modify the Wasserstein module, it is just a documentation reorganization proposal.
Maybe more related with #167 as Wasserstein barycenters will come in this section.

mglisse:
I looked at the compiled doc and it looks strange, with what was supposed to be the rightmost column appearing as the topmost line instead.

PR #236, opened 2020-03-03
optimal matching in Wasserstein distance (pot) (Fix #203)

### LINKED ISSUE #203: Associated matching to Wasserstein distance
@tlacombe 

I think it would be great if `gudhi.wasserstein.wasserstein_distance()` could optionally return the optimal partial matching associated to the distance. 

There are probably a couple of ways to implement this, the most naive one being to take, for each row `i` of the cost matrix `Gamma`, the maximum of `Gamma[i,:]` (although this could potentially return something that is not a matching if we ask for it before convergence).

tlacombe:
Updated Wasserstein distance (`pot` version) from `master` so that it can now return the optimal matching if passed `matching=True` (set to `False` by default).

Fix #203.

I don't really know how it should interact with #205 

@MathieuCarriere : does it correspond to your needs ?


src/python/gudhi/wasserstein.py
```diff
@@ -62,36 +62,79 @@ def _perstot(X, order, internal_p):
...
+    :param n: int, size of the first dgm
+    :param m: int, size of the second dgm
+    :return: a modified version of match where indices greater than n, m are replaced by -1, encoding the diagonal.
+                and (-1, -1) are removed
+    '''
+    new_match = []
+    for i,j in match:
+        if i >= n:
+            if j < m:
+               new_match.append((-1, j))
```
mglisse: Can't you do something like this?
```
P[-1,-1]=0
new=match.nonzero()
new[0][new[0]>=n]=-1
new[1][new[1]>=m]=-1
```
(or some variant to give whatever output we decide is the right one)

src/python/gudhi/wasserstein.py
```diff
@@ -62,36 +62,79 @@ def _perstot(X, order, internal_p):
...
-    b[-1] = b[-1] * n                # so that we have a probability measure, required by POT
+    a = np.ones(n+1) # weight vector of the input diagram. Uniform here.
+    a[-1] = m
+    b = np.ones(m+1) # weight vector of the input diagram. Uniform here.
+    b[-1] = n
+
+    if matching:
+        P = ot.emd(a=a,b=b,M=M, numItermax=2000000)
+        ot_cost = np.sum(np.multiply(P,M))
+        P[P < 0.5] = 0  # trick to avoid numerical issue, could it be improved?
```
mglisse: Do you still need this now that you are passing integer weights?

tlacombe: Well, I did some trials and it seems the minimum is always `1.` so I think could remove this ; however I don't know how to be 100% sure that I won't have some 1E-10 numerical issue happening that could lead to weird bugs in further steps.

How bad is it to left this?

src/python/gudhi/wasserstein.py
```diff
@@ -62,36 +62,79 @@ def _perstot(X, order, internal_p):
...
+            new_match.append((i,j))
+    return new_match
+
+
+def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2.):
+    '''
+    :param X: (n x 2) numpy.array encoding the (finite points of the) first diagram. Must not contain essential points 
+                (i.e. with infinite coordinate).
     :param Y: (m x 2) numpy.array encoding the second diagram.
+    :param matching: if True, computes and returns the optimal matching between X and Y, encoded as...
```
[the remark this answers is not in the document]
tlacombe: haha :+1:  completed.

src/python/gudhi/wasserstein.py
tlacombe: Indeed it seems much more compact :) It changes the output type, but I have no reason to prefer one over the other.

src/python/gudhi/wasserstein.py
mglisse: From an algorithmic point of view, this method does not invent weights, it takes the weight of a point and adds/subtracts it to other weights. Integers should then remain integers quite safely. Actually, by default, the C++ code used by POT uses integral weights, and POT explicitly tells it to use double instead.
One issue with keeping the rounding is that if there are indeed numerical issues, we will never learn of it. And reciprocally, if the line is useless, we will not know, so we can never remove it.
If you really want to keep it, we can (I won't block the PR for that), although I'd rather replace it with `assert (P[P<1]==0).all()`.

src/python/gudhi/wasserstein.py
mglisse: If we output everything at once, with -1 to represent the diagonal, then either a numpy array (n,2) or a pair of arrays would be fine (possibly some other variants as well). It could also be convenient to return an array (n,2) for regular pairs, and 2 1D arrays for matchings with the diagonal. I can get those from the combined array using `np.compress(a[:,0]==-1,a[:,1])` or similar, but if most users are going to do this anyway, we could do it for them (if it is easier for us than it is for them). Then the question is back to what users want it for...

tlacombe: The advantage of returning a (n x 2) np.array is that it is rather compact ; also as you mentioned earlier using `-1` to represent the diagonal is convenient as, say, if one has `(- 1 , j )` in the matching, the corresponding cost can actually be accessed by `C[-1, j]`. 


(It makes me realize that we do not provide a way for the user to build the distance matrix, although it might be useful... Should we do that? It would just consist in removing the `_` in `_build_dist_matrix` I think.)

src/python/gudhi/wasserstein.py
tlacombe: I understand your point ; I will remove it.

src/python/gudhi/wasserstein.py
mglisse: > The advantage of returning a (n x 2) np.array is that it is rather compact

Yes. The drawback is if "most" users split it in 3 afterwards... Of course there is the possibility of having an option to choose which type of output we want. That could be overkill.

> also as you mentioned earlier using `-1` to represent the diagonal is convenient as, say, if one has `(- 1 , j )` in the matching, the corresponding cost can actually be accessed by `C[-1, j]`.

Keeping `n` and `m` also works for that, although it is less convenient for other purposes.

> (It makes me realize that we do not provide a way for the user to build the distance matrix, although it might be useful... Should we do that? It would just consist in removing the `_` in `_build_dist_matrix` I think.)

What use do you see for that distance matrix? If I use the distance matrix, it will be one built with pytorch, jax, or some other autodifferentiable type, so I won't be able to use your function that explicitly uses scipy.spatial.distance.cdist. But maybe there are other reasons to want that matrix?

tlacombe: Well it is hard for me to anticipate what people would do with the matching... I think the compact (n x 2) output is the way to go. 

About returning the matrix, I don't have much idea. It is just cheap to provide this option to the user, although it make the doc heavier. Let's forget that for now.

mglisse: > Well it is hard for me to anticipate what people would do with the matching... I think the compact (n x 2) output is the way to go.

ok for me.

src/python/gudhi/wasserstein.py
```diff
@@ -30,7 +30,9 @@ def _build_dist_matrix(X, Y, order=2., internal_p=2.):
     :param order: exponent for the Wasserstein metric.
     :param internal_p: Ground metric (i.e. norm L^p).
     :returns: (n+1) x (m+1) np.array encoding the cost matrix C. 
-                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) and its orthogonal proj onto the diagonal.
+                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j],
```
mglisse: 0<=i<n ?

src/python/gudhi/wasserstein.py
```diff
@@ -30,7 +30,9 @@ def _build_dist_matrix(X, Y, order=2., internal_p=2.):
     :param order: exponent for the Wasserstein metric.
     :param internal_p: Ground metric (i.e. norm L^p).
     :returns: (n+1) x (m+1) np.array encoding the cost matrix C. 
-                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) and its orthogonal proj onto the diagonal.
+                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], 
+                while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j])
```
mglisse: m+1 -> m and n+1 -> n here as well?

src/python/gudhi/wasserstein.py
```diff
@@ -30,7 +30,9 @@ def _build_dist_matrix(X, Y, order=2., internal_p=2.):
     :param order: exponent for the Wasserstein metric.
     :param internal_p: Ground metric (i.e. norm L^p).
     :returns: (n+1) x (m+1) np.array encoding the cost matrix C. 
-                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) and its orthogonal proj onto the diagonal.
+                For 1 <= i <= n, 1 <= j <= m, C[i,j] encodes the distance between X[i] and Y[j], 
+                while C[i, m+1] (resp. C[n+1, j]) encodes the distance (to the p) between X[i] (resp Y[j]) 
+                and its orthogonal proj onto the diagonal.
```
```suggestion
                and its orthogonal projection onto the diagonal.
```

src/python/gudhi/wasserstein.py
tlacombe: Indeed !

src/python/gudhi/wasserstein.py
```diff
@@ -62,36 +64,61 @@ def _perstot(X, order, internal_p):
...
-    b = np.full(m+1, 1. / (n + m) )  # weight vector of the input diagram. Uniform here.
-    b[-1] = b[-1] * n                # so that we have a probability measure, required by POT
+    a = np.ones(n+1) # weight vector of the input diagram. Uniform here.
+    a[-1] = m
+    b = np.ones(m+1) # weight vector of the input diagram. Uniform here.
+    b[-1] = n
+
+    if matching:
+        P = ot.emd(a=a,b=b,M=M, numItermax=2000000)
+        ot_cost = np.sum(np.multiply(P,M))
```
mglisse: I wonder if it would be better or worse to compute ot_cost later as `M[match[:,0],match[:,1]].sum()` (untested). It probably doesn't matter.

src/python/gudhi/wasserstein.py
tlacombe: isn't it corrected in the last commit ?

src/python/test/test_wasserstein_distance.py
```diff
@@ -69,7 +69,7 @@ def approx(x):
         match = wasserstein_distance(diag2, emptydiag, matching=True, internal_p=np.inf, order=2.24)[1]
         assert np.array_equal(match , np.array([[0, -1], [1, -1]]))
         match = wasserstein_distance(diag1, diag2, matching=True, internal_p=2., order=2.)[1]
-        assert np.array_equal(match, np.array_equal([[0, 0], [1, 1], [2, -1]]))
+        assert np.array_equal(match, np.array([[0, 0], [1, 1], [2, -1]]))
```
mglisse: You don't even need `np.array`, `array_equal` happily accepts a list as argument.

src/python/gudhi/wasserstein.py
mglisse: Yes, my comment was delayed because it was grouped with others.

src/python/gudhi/wasserstein.py
```diff
@@ -62,36 +64,61 @@ def _perstot(X, order, internal_p):
...
+                        the j-th point in Y, with the convention (-1) represents the diagonal.
     :param order: exponent for Wasserstein; Default value is 2.
-    :param internal_p: Ground metric on the (upper-half) plane (i.e. norm L^p in R^2); Default value is 2 (Euclidean norm).
-    :returns: the Wasserstein distance of order q (1 <= q < infinity) between persistence diagrams with respect to the internal_p-norm as ground metric.
-    :rtype: float
+    :param internal_p: Ground metric on the (upper-half) plane (i.e. norm L^p in R^2); 
+                        Default value is 2 (Euclidean norm).
+    :returns: the Wasserstein distance of order q (1 <= q < infinity) between persistence diagrams with 
+                        respect to the internal_p-norm as ground metric.
+              If matching is set to True, also returns the optimal matching between X and Y.
```
mglisse: I see the line "the Wasserstein [...] with" in bold and the next line strangely indented. I think you shouldn't indent the middle line more than the others. If you want the last line to be a separate paragraph, you can insert an empty line in between.

MathieuCarriere:
> @MathieuCarriere : does it correspond to your needs ?

This is exactly what I need :-)

src/python/test/test_wasserstein_distance.py
tlacombe: Cleaned :)

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -47,3 +47,27 @@ The output is:
...
+    cost, matching = gudhi.wasserstein.wasserstein_distance(diag1, diag2, matching=True, order=1., internal_p=2.)
+
+    message = "Wasserstein distance value = %.2f, optimal matching: %s"  %(cost, matching)
+    print(message)
+
+The output is:
+
+.. testoutput::
+    
+    Wasserstein distance value = 2.15, optimal matching: [[0, 0], [1, 2], [2, -1], [-1, 1]]
```
mglisse: <pre>
Expected:
    Wasserstein distance value = 2.15, optimal matching: [[0, 0], [1, 2], [2, -1], [-1, 1]]
Got:
    Wasserstein distance value = 2.15, optimal matching: [[ 0  0]
     [ 1  2]
     [ 2 -1]
     [-1  1]]
</pre>
You can use `:options: +NORMALIZE_WHITESPACE` to ignore the difference between spaces and newlines (although it will probably still be unhappy about the space between `[` and the digit), but the commas cannot be ignored. Either you remove them from the expected output, or you change the way you print the array (possibly using `np.array2string` or `repr`).

src/python/gudhi/wasserstein.py
```diff
@@ -62,36 +64,61 @@ def _perstot(X, order, internal_p):
...
     m = len(Y)
 
     # handle empty diagrams
     if X.size == 0:
         if Y.size == 0:
-            return 0.
+            if not matching:
+                return 0.
+            else:
+                return 0., np.array([])
```
mglisse: `np.empty((0,2))` would be a possibility, but no need to change.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -58,16 +58,29 @@ An index of -1 represents the diagonal.
...
-
-    message = "Wasserstein distance value = %.2f, optimal matching: %s"  %(cost, matching)
-    print(message)
+    dgm1 = np.array([[2.7, 3.7],[9.6, 14.],[34.2, 34.974]])
+    dgm2 = np.array([[2.8, 4.45], [5, 6], [9.5, 14.1]])
+    cost, matchings = gudhi.wasserstein.wasserstein_distance(diag1, diag2, matching=True, order=1., internal_p=2.)
+
+    message_cost = "Wasserstein distance value = %.2f" %cost
+    print(message_cost)
+    dgm1_to_diagonal = matchings[np.where(matchings[:,0] == -1)][:,1]
```
mglisse: I don't think you need `np.where`, `a[(a[:,0]==-1)][:,1]` or even simpler `a[a[:,0]==-1,1]` should do, no? Or `np.compress(a[:,0]==-1,a[:,1])` or any of many variations.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -58,16 +58,29 @@ An index of -1 represents the diagonal.
...
 
-    diag1 = np.array([[2.7, 3.7],[9.6, 14.],[34.2, 34.974]])
-    diag2 = np.array([[2.8, 4.45], [5, 6], [9.5, 14.1]])
-    cost, matching = gudhi.wasserstein.wasserstein_distance(diag1, diag2, matching=True, order=1., internal_p=2.)
-
-    message = "Wasserstein distance value = %.2f, optimal matching: %s"  %(cost, matching)
-    print(message)
+    dgm1 = np.array([[2.7, 3.7],[9.6, 14.],[34.2, 34.974]])
+    dgm2 = np.array([[2.8, 4.45], [5, 6], [9.5, 14.1]])
+    cost, matchings = gudhi.wasserstein.wasserstein_distance(diag1, diag2, matching=True, order=1., internal_p=2.)
```
mglisse: Would it work to use `1` and `2` instead of `1.` and `2.`?

src/python/doc/wasserstein_distance_user.rst
tlacombe: I think so. I will investigate asap.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -58,16 +58,29 @@ An index of -1 represents the diagonal.
...
+
+    message_cost = "Wasserstein distance value = %.2f" %cost
+    print(message_cost)
+    dgm1_to_diagonal = matchings[np.where(matchings[:,0] == -1)][:,1]
+    dgm2_to_diagonal = matchings[np.where(matchings[:,1] == -1)][:,0]
+    off_diagonal_match = np.delete(matchings, np.where(matchings == -1)[0], axis=0)
```
mglisse: An alternative would be `a[(a!=-1).all(axis=1)]`, but it isn't obviously better or worse.

mglisse: Ok to keep your version with delete here, but if you write an example or tutorial elsewhere, please use the variant `a[(a!=-1).all(axis=1)]` so users can see that one as well. Deal? (If you don't write one of those, then you have nothing to do)

tlacombe: Deal. I plan to write a tutorial about barycenters once #167 got merged, it could be such an opportunity.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -36,14 +36,51 @@ Note that persistence diagrams must be submitted as (n x 2) numpy arrays and mus
...
+    import numpy as np
+
+    dgm1 = np.array([[2.7, 3.7],[9.6, 14.],[34.2, 34.974]])
+    dgm2 = np.array([[2.8, 4.45], [5, 6], [9.5, 14.1]])
+    cost, matchings = gudhi.wasserstein.wasserstein_distance(dgm1, dgm2, matching=True, order=1, internal_p=2)
+
+    message_cost = "Wasserstein distance value = %.2f" %cost
+    print(message_cost)
+    dgm1_to_diagonal = matchings[matchings[:,0] == -1, 1]
+    dgm2_to_diagonal = matchings[matchings[:,1] == -1, 0]
```
mglisse: the 2 seem to be swapped: if the first column is -1, it is the second diagram that is matched to the diagonal.

tlacombe: Yes. Corrected.

PR #237, opened 2020-03-03
Fix #172 and add a proper comment on the modification

### LINKED ISSUE #172: [Simplex_tree] Handle NaN in make_filtration_non_decreasing
https://github.com/GUDHI/gudhi-devel/blob/3e829fd6f4a3a122da9df35a88e5c51122860bf6/src/Simplex_tree/include/gudhi/Simplex_tree.h#L1350

Replacing `if(f<max)` with `if(!(f>=max))` would mean that if `f` is NaN, we replace it with the max of the children. That seems more useful than keeping NaN.

VincentRouvreau:


src/Simplex_tree/test/simplex_tree_make_filtration_non_decreasing_unit_test.cpp
```diff
@@ -0,0 +1,148 @@
...
+  /*  2   0   3\X/4        */
+  /*            o          */
+  /*            5          */
+
+  std::cout << "SPECIFIC CASE:" << std::endl;
+  std::cout << "Insertion with NaN values does not ensure the filtration values are non decreasing" << std::endl;
+  st.make_filtration_non_decreasing();
+
+  std::cout << "Check all filtration values are NaN" << std::endl;
+  for (auto f_simplex : st.filtration_simplex_range()) {
```
mglisse: Why not complex_simplex_range? filtration_simplex_range is a more costly, more specific tool when you need the ordering.

VincentRouvreau: You are right. Fixed on 19ea0c1.
complex_simplex_range is not binded. How could we call it ? We have `get_skeleton` and `get_filtration`. `get_simplices` ? May be confusing as `get_filtration` returns simplex and its filtration value ?

mglisse: > Fixed on [19ea0c1](https://github.com/GUDHI/gudhi-devel/commit/19ea0c10f283188282a78ebebf4c1a51f2f40040).

Thanks.

> `get_simplices` ?

Sure.

> May be confusing as `get_filtration` returns simplex and its filtration value ?

In some sense, it is slightly confusing that get_skeleton returns the filtration values. But if we consider that simplices always come with their filtration value, then get_simplices could be equivalent to get_skeleton(max), i.e. return the same objects as get_filtration, but without a specified order, and thus more cheaply (at least if you only iterate once), without initialize_filtration. Or we could have get_simplices to get just the simplices, and get_simplices_with_filtration to get the pairs. Or some other variant. We should discuss this elsewhere though, it isn't related to this PR anymore...

PR #238, opened 2020-03-05
remove ext (submodule directory) from doxygen parsing

VincentRouvreau:


VincentRouvreau:
> Oops, I didn't notice doxygen was looking there...

No pb, it was easy too fix :wink:

PR #242, opened 2020-03-06
Generators for Simplex_tree

### LINKED ISSUE #235: [Simplex_tree] Find generator in lower-star filtration
When we call `make_filtration_non_decreasing`, the value of many simplices will be a simple copy of the value of a lower-dimensional simplex (usually an edge). In order to compute a gradient, we need to find where the filtration value of a simplex originates from.
One version would be to always look for a face of dimension 1 (an edge) of the simplex that has the same filtration value.
Another one would be to look for the lowest-dimensional face of the simplex that has the same filtration value.
In case of ambiguity, I think it is ok to make arbitrary choices.

mglisse:
This provides some functions to "reverse engineer" `expansion` and `make_filtration_non_decreasing`. It is a brick, to be used later in #243 on the simplices in persistence pairs.
Fix #235.
This is related to #204.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -24,6 +24,7 @@
 #include <boost/iterator/transform_iterator.hpp>
 #include <boost/graph/adjacency_list.hpp>
 #include <boost/range/adaptor/reversed.hpp>
+#include <boost/container/static_vector.hpp>
 
 #ifdef GUDHI_USE_TBB
 #include <tbb/parallel_sort.h>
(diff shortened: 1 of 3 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1465,6 +1475,74 @@ class Simplex_tree {
...
+  Simplex_handle edge_with_same_filtration(Simplex_handle sh) {
+#if 0
+    // FIXME: Only do this if dim >= 2, since we don't want to return a vertex...
+    // Test if we are lucky and the parent has the same filtration value.
+    Siblings* sib = self_siblings(sh);
+    Vertex_handle v_par = sib->parent();
+    sib = sib->oncles();
+    Simplex_handle par = sib->find(v_par);
+    if(filtration_(par) == filt) return edge_with_same_filtration(par);
+#endif
```
VincentRouvreau: Do we have to keep this dead code ?

mglisse: I'd like to keep at least a comment saying that it may be worth investigating if this type of thing helps. And it is convenient if that code is available somewhere easy to find, so that when we do decide to investigate, we don't have to reinvent it. But I agree that keeping dead code like that is not so good. I think I could open an issue and put that code there, with a comment in the code linking to the issue.

PR #243, opened 2020-03-07
persistence generators for lower-star and flag filtrations

mglisse:
This branch uses #242.
It is the Simplex_tree equivalent of #204.
The interface is open for discussion.
Cython sucks.
The other changes (get_persistence for instance) are small simplifications and speed-ups, they are not supposed to change the result.


mglisse:
For the interface, and in particular what the functions return, I can think of several possibilities. Currently, I first separate finite and infinite points. Then I make a list, with one entry per dimension, and each such entry is an array of vertex indices.

- Arrays don't have the same width. A dimension 0 finite interval in a flag complex is described by 3 vertices, one for birth, and two for the death edge. On the other hand, in dimension 1+, there are 4 vertices, 2 for the edge that gives its filtration value to the birth simplex, and the same for death. To avoid confusion, I currently return the 0-dim array separately from the list of the k-dim arrays. This way it is clear for users that they should handle this dimension differently.
- It would be possible to "pad" to 4 vertices for dimension 0, for instance writing the birth vertex twice, or introducing a dummy second birth vertex -1. If we see the output as a potential index for the distance matrix and the vertex weights, duplicating v corresponds to writing the vertex weights on the diagonal of the distance matrix, while the dummy -1 means adding the vertex weights as an extra row or column. This would allow the "flag" function to return a tuple of size 2, with the 0-dim arrays reintegrated in the lists.
- It would even be possible to pad infinite intervals so they can fit with the finite ones, but the benefits don't seem as important.
- Instead of a list of arrays, 1 per dimension, it is possible to have a single array, with an extra column giving the dimension of each interval. I don't yet have an opinion whether that's better or worse than a list (or tuple, dict).

MathieuCarriere:
I have mixed feelings. Overall, I think the list of arrays is a better format. However, if I want to use Tensorflow, padding is going to make life much easier. So I don't know, I guess if you manage to use PyTorch, then we can go for the list of arrays and I can figure out a way to use it in Tensorflow.

mglisse:
@MathieuCarriere I am not opposed to having 2 versions of the function if different applications have different needs. I expect the different frameworks will force us to duplicate a lot more code than that anyway (perslay already shows that a bit).
I manage to use pytorch (and jax, if you prefer google stuff), but I can't promise that I am doing things the right way, I am just discovering it. I am happy that I don't have to write any explicit gradient for now (alpha-complex will wait).
With the list of arrays, there are still a few options. The current version splits everything that behaves differently (dimension 0 comes outside of the list with a different shape for flag complexes, infinite intervals are in a different list, etc). Do you have some preference?
If you think we should go with a list of arrays, will you change your branch on generators for cubical complexes to return a list of arrays as well?

MathieuCarriere:
I think that the list of arrays is perfect as it is. As you mentioned, people will want to separate things by dimension and "finiteness", so we might as well do it in advance. And I've thought about it again, and I think Tensorflow can actually work with a list of arrays (I have figured how to wrap python functions into Tensorflow layers), so no need to implement padding! Of course, I'll update the output of CubicalComplex to make everything consistent.

src/python/test/test_simplex_generators.py
```diff
@@ -0,0 +1,57 @@
...
+    st = r.create_simplex_tree(max_dimension=50)
+    st.persistence()
+    g = st.flag_persistence_generators()
+    assert np.array_equal(g[0], [[2, 2, 0], [1, 1, 0], [3, 3, 1], [6, 6, 4], [5, 5, 4], [7, 7, 5]])
+    assert len(g[1]) == 1
+    assert np.array_equal(g[1][0], [[3, 2, 2, 1]])
+    assert np.array_equal(g[2], [0, 4])
+    assert len(g[3]) == 1
+    assert np.array_equal(g[3][0], [[7, 6]])
+
```
VincentRouvreau, line 27: Here, maybe we could test with the result of persistence pairs. I don't know if it would be clearer.

mglisse: Could you clarify what we would test with persistence_pairs? Would we check if the generator is indeed an edge of the simplex given by persistence_pairs, with the same filtration value?

VincentRouvreau: I was thinking checking, for instance, that `st.flag_persistence_generators()[0]` and `list([pair[0][0], pair[1][0], pair[1][1]] for pair in st.persistence_pairs() if len(pair[1])==2)` are equal. But that's maybe not a good idea.

mglisse: Ah, checking that the pairs that are already of the right dimension, where the generators are trivial, are the same? That makes sense, I'll see what I can do.

mglisse: Like this?

VincentRouvreau: :+1:

MathieuCarriere:
The output format in this PR is great, and I modified #204 so that the format is the same.

VincentRouvreau:
@mglisse could you resolve conflicts on this ? I think your work is ready to be merged.

PR #250, opened 2020-03-17
Merge filtered edge container from strong collapse and fix code review

VincentRouvreau:


src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +32,133 @@ struct vertex_filtration_t {
...
  */
-template< typename SimplicialComplexForProximityGraph
-          , typename ForwardPointRange
-          , typename Distance >
-Proximity_graph<SimplicialComplexForProximityGraph> compute_proximity_graph(
-    const ForwardPointRange& points,
-    typename SimplicialComplexForProximityGraph::Filtration_value threshold,
+template <template <class> class Edge_graph, class SimplicialComplexForEdgeGraph, typename ForwardPointRange,
+          typename Distance>
+Edge_graph<SimplicialComplexForEdgeGraph> compute_edge_graph(
```
mglisse: I can't find this function in doxygen.

src/common/include/gudhi/graph_simplicial_complex.h
```diff
@@ -30,40 +32,133 @@ struct vertex_filtration_t {
...
+                                   typename SimplicialComplexForFilteredEdges::Vertex_handle,
+                                   typename SimplicialComplexForFilteredEdges::Vertex_handle>;
+  /** \brief Type of the `Gudhi::Filtered_edges_container` data structure.
+   */
+  using Filtered_edge_set = std::vector<Filtered_edge>;
+
+  using Filtered_edge_set_iterator = typename Filtered_edge_set::const_iterator;
+  /** \brief Range over the `Gudhi::Filtered_edges_container` data structure.
+   */
+  using Filtered_edge_range = boost::iterator_range<Filtered_edge_set_iterator>;
```
mglisse: Why do we need this when we have Filtered_edge_set which is already a range?

mglisse: It looks like there actually was a good reason for Filtered_edge_range, it acts as a view (std::span in C++20) so you can return a cheap object that references a sub-range of edges_, while returning a vector forces a copy, and we cannot return a vector by reference since we only want a subrange.

Issue #241, opened 2020-03-05
GudhUI will not compile because of alleged missing Qt5 components, despite Qt being installed.


ChemGuy88:
I compiled GUDHI, but the UI part did not compile.

```
++ GudhUI will not be compiled because Qt5 COMPONENTS Widgets Xml OpenGL are not found
```

I've installed `qt` via `brew`, which I thought would satisfy the requirements, but it did not. 

```
brew install qt
```

I'm asking for clarification on whether ** Widgets**, ** Xml**, and **OpenGL** are included in Qt5, and if I can download Qt5 through `brew`.

Also, I have done the following, but none make the error go away:

```
brew cask install qt-design-studio
brew cask install qt-creator
```

mglisse:
Hello,
thank you for your report.
First, if you don't specifically mean to use the edge-contraction algorithm on skeleton-blocker datastructure, you can probably ignore GudhUI.
Did cmake say anything interesting before that?
Did cmake find parts of Qt5 but not some components, or did it not find it at all? You can use cmake-gui or ccmake to examine the result. Maybe you just need to tell it where Qt is with Qt5_DIR.
You could check if there are files with a name similar to libQt5Widgets.so.5 (maybe .dylib on mac?) in your brew directory, same with xml and opengl.
I am not a pro of cmake, the first thing that comes to mind is running cmake with the --trace option, and looking at the parts that talk about Qt...

VincentRouvreau:
I reproduced the issue on a OSx virtual machine.
I am investigating...

ChemGuy88:
Thanks to both @mglisse and @VincentRouvreau for their replies. First comes my response to [mglisee](#CMake), and then a response to [VincentRouvreau](#Fixing Homebrew links).

## <a name="CMake"></a> CMake

I reduced the number of missing Qt5 components to just QGLViewer since my last post by erasing `CMakeCache.txt` and refreshing `$PATH$`. I installed QGLViewer according to the Mac OS X instructions in the [QGLViewer website](http://libqglviewer.com/installUnix.html).

> Did cmake say anything interesting before that?

Previously I got the following output with some CMake warnings, most of which I was able to avoid by implementing this [solution](https://stackoverflow.com/a/58085634/5478086) from StackOverflow.

```
(base) 06:00 PM [4] ~/Library/Python/3.7/lib/python/site-packages/gudhi/build --> cmake -DPython_ADDITIONAL_VERSIONS=3 ..
[...]

PR #253, opened 2020-03-17
Remove workarounds

### LINKED ISSUE #224: 1.#INF
Since we only support Visual Studio 2015 or more recent, I think we could remove the workarounds we put in place for when it used to print infinity as `1.#INF`.

VincentRouvreau:
* Fix #224 
* Remove thread_local workaround


mglisse:
I notice in #241 
> -- Performing Test CGAL_CAN_USE_CXX11_THREAD_LOCAL_RESULT - Failed

I hope that's because of some installation issue, and that all recent-enough compilers handle thread_local.

PR #260, opened 2020-03-24
merge https://github.com/mglisse/gudhi-devel/tree/alpha-cache

### LINKED ISSUE #133: [Alpha_complex] Inefficient point comparison
In `Alpha_complex::propagate_alpha_filtration`, we have:
```c++
  std::find(pointVector.begin(), pointVector.end(), point_for_gabriel)
```
which compares points with `operator==`. This is a bad idea. If we need to compare something, it should be vertices (combinatorial, not geometric).

### LINKED ISSUE #135: [Alpha_complex] Cache circumcenters
The way `Compute_squared_radius_d` and `Side_of_bounded_sphere_d` are currently implemented in CGAL, they start by computing the circumcenter. We end up doing that many times for the same simplex. One option to speed things up would be to cache those circumcenters somewhere. For Epeck_d, things would remain just as exact. Epick_d would become even more unreliable than now: `Compute_squared_radius_d` would be the same, but `Side_of_bounded_sphere_d` which is exact would be replaced by something inexact. I could try and devise a mechanism in CGAL so we could cache the circumcenter of AK (and EK) instead of doing a construction with the main kernel, but that will wait until a more motivating use case appears.

VincentRouvreau:
and fix conflicts.
* Fix #133 
* Fix #135

### DIFF
--- a/src/Alpha_complex/concept/SimplicialComplexForAlpha.h
+++ b/src/Alpha_complex/concept/SimplicialComplexForAlpha.h
@@ -72,6 +72,24 @@ struct SimplicialComplexForAlpha {
   /** \brief Return type of an insertion of a simplex
    */
   typedef unspecified Insertion_result_type;
+
+  /** \name Map interface
+   * Conceptually a `std::unordered_map<Simplex_handle,std::size_t>`.
+   *  @{ */
+  /** \brief Data stored for each simplex.
+   *
+   * Must be an integer type. */
+  typedef unspecified      Simplex_key;
+  /** \brief Returns a constant dummy number that is either negative,
+   * or at least as large as the number of simplices.  Suggested value: -1.  */
+  Simplex_key              null_key ();
+  /** \brief Returns the number stored for a simplex by `assign_key()`.
+   *
+   * If `assign_key()` has not been called, it must return `null_key()`.  */
+  Simplex_key              key      ( Simplex_handle sh );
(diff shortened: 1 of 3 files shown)


mglisse:
While I was looking at the profile of the safe, dynamic version on spiral_3d_10k, I found

1. https://github.com/CGAL/cgal/pull/4606
2. Could you add a private version of `get_point` (`get_point_`?) for internal use, that uses operator[] instead of at()? The gain is very small, but it seemed to be above the level of noise.

mglisse:
The approach makes the fast (Epick_d) version even more unreliable, but I guess if you can't rely on it anyway, it doesn't matter that much, and it becomes a bit faster.
I am wondering what it does to the unpublished Epack_d (same objects and predicates as Epick_d, but constructions are done with a bounded error), I'll update this comment later.

mglisse:
I pushed a patch to try and reduce the memory use a bit. I didn't measure. Except in very large dimension, the difference is likely not that big, since we keep the cache for 2 dimensions at the same time.
I think the main thing missing is updating the various concepts so they match what the class really uses.

mglisse:
I had forgotten that we don't document a concept for kernels, we just say Ep[ie]ck_d, so that's less concept updating needed, I did the simplicial complex.

VincentRouvreau:
Thank you for the documentation !

PR #262, opened 2020-03-26
Add 3 pip runtime dependencies

mglisse:
I am using those for nearest neighbors in a branch I'll submit soon.
I am not sure how well things that use CUDA on my computer will work on CI...
Do we want to limit the number of pip packages we install, or is it fine to keep adding more?

### DIFF
--- a/.github/test-requirements.txt
+++ b/.github/test-requirements.txt
@@ -7,3 +7,6 @@ scipy
 scikit-learn
 POT
 tensorflow
+torch
+pykeops
+hnswlib


mglisse:
Since circleci also runs on my branches that do not have a PR, the docker image could be used right now :smiley:

mglisse:
Actually, docker fails in `pip install` with MemoryError, while handling pytorch :cry: 
Some people on the net claim that passing `--no-cache-dir` to pip helps, although it isn't obvious to me why it should.

PR #265, opened 2020-03-28
DTM

mglisse:
One class to wrap external libraries to compute k nearest neighbors.
One class to compute the DTM.
(Missing is an extra class that returns a density estimation based on the DTM, that will be a separate PR)
I am using KNearestNeighbors and DistanceToMeasure, because KNN and DTM seem a bit short, not explicit enough. But I am using short names for the files. You may complain :wink:
In some cases where base points and query points are the same, we may want to ignore each point as its own nearest neighbor, that's not done currently.
There is a mysterious random failure for sphinx_py_test on circleci ("Child killed"...) while it works fine here, and I didn't add any new sphinx tests, just regular doc. Restarting the test a few times seems to work, but...

https://830-175032741-gh.circle-artifacts.com/0/tmp/sphinx/point_cloud.html


src/python/gudhi/point_cloud/knn.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def __init__(self, k, return_index=True, return_distance=False, metric="euclidean", **kwargs):
+        """
+        Args:
+            k (int): number of neighbors (possibly including the point itself).
+            return_index (bool): if True, return the index of each neighbor.
+            return_distance (bool): if True, return the distance to each neighbor.
+            implementation (str): choice of the library that does the real work.
+
+                * 'keops' for a brute-force, CUDA implementation through pykeops. Useful when the dimension becomes large (10+) but the number of points remains low (less than a million). Only "minkowski" and its aliases are supported.
```
VincentRouvreau, line 28: Some documentation are really long lines. Please let lines <= 120.
Here I think you can jump to the next line with 4 spaces at the beginning, it shall work.

mglisse: I am quite sure I tried that already and it doesn't work, not for a list nested in a param. That's why I kept those very long lines. Théo hit a similar issue IIRC. If you can find a way to split the lines without breaking the rendering by sphinx, I'd be happy to split those lines indeed.

VincentRouvreau, line 28:
```suggestion
                * 'keops' for a brute-force, CUDA implementation through pykeops. Useful when the dimension becomes
                    large (10+) but the number of points remains low (less than a million). Only "minkowski" and its
                    aliases are supported.
```
seems to work fine for me.

VincentRouvreau: Oh no... You are right it do not work... argh...

src/python/gudhi/point_cloud/knn.py
```diff
@@ -0,0 +1,317 @@
...
+#
+# Copyright (C) 2020 Inria
+#
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
+
+import numpy
+
+# TODO: https://github.com/facebookresearch/faiss
+
```
VincentRouvreau, line 13: Here, or in the class, is missing some:
```
__license__ = "blablabla"
```

src/python/gudhi/point_cloud/dtm.py
```diff
@@ -0,0 +1,66 @@
...
+# See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+# Author(s):       Marc Glisse
+#
+# Copyright (C) 2020 Inria
+#
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
+
+from .knn import KNearestNeighbors
+
```
VincentRouvreau, line 11: Here, or in the class, is missing some:
```
__license__ = "blablabla"
```

mglisse:
> It would be perfect with a test using `sort_results = True` !

You mean using it explicitly? True is the default value, so it should already be used by several tests (and False it used via the DTM).

VincentRouvreau:
> > It would be perfect with a test using `sort_results = True` !
> 
> You mean using it explicitly? True is the default value, so it should already be used by several tests (and False it used via the DTM).

Ok, sorry, I think I misunderstood the result of the unitary tests coverage. But your change is ok.

mglisse:
> I think I misunderstood the result of the unitary tests coverage

Ah, looking at them (by the way, what is the most convenient way to see them?), I see indeed that some combinations of return_index/sort_results/n_jobs are not covered. The combinatorics are killing me :tired_face:

VincentRouvreau:
> Ah, looking at them (by the way, what is the most convenient way to see them?), I see indeed that some combinations of return_index/sort_results/n_jobs are not covered. The combinatorics are killing me tired_face

No, it's ok, you can keep it like it is. 95% is already a great achievement :smiley:

mglisse:
> No, it's ok, you can keep it like it is. 95% is already a great achievement :smiley:

I'll merge to unblock other things, but I should still try to add them later.

mglisse:
Hmm, it seems to have passed circleci on master, good, but now that I think about it, maybe I should have waited to understand better what the random issue is with circleci. Maybe memory and lowering -j4 to -j3 for the python tests would help?

PR #268, opened 2020-03-30
Streamline initialize_filtration

mglisse:
I think it is now 100% automatic in python.
In C++, cheap functions like insert / remove_maximal / assign_filtration are not automatic though. Clearing the vector is cheap but not free, and I didn't feel like paying for it. Expensive functions do clear the vector.
I kept a few calls to initialize_filtration here and there in some C++ tests, most of them are likely unnecessary.


mglisse:
Heads up: I don't think this is controversial, and I intend to merge it soon.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -90,24 +90,15 @@ cdef class SimplexTree:
...
-            :func:`persistence()<gudhi.SimplexTree.persistence>`,
-            :func:`betti_numbers()<gudhi.SimplexTree.betti_numbers>`,
-            :func:`persistent_betti_numbers()<gudhi.SimplexTree.persistent_betti_numbers>`,
-            or :func:`get_filtration()<gudhi.SimplexTree.get_filtration>`
-            after :func:`inserting<gudhi.SimplexTree.insert>` or
-            :func:`removing<gudhi.SimplexTree.remove_maximal_simplex>`
-            simplices.
+        .. deprecated:: 3.2.0
         """
         self.get_ptr().initialize_filtration()
```
VincentRouvreau, line 103: If deprecated, shouldn't we display a message ?

mglisse: We could. One reason to warn would be because we may remove it in a later release, but I don't really have plans for that. On the other hand, for downstream users who try to be compatible with several releases of gudhi, these warnings can be a bit irritating. Maybe we can have several steps: one or more releases where the function is unnecessary, then one or more releases where it warns, and finally a release where it is removed?
```python
if getattr(gudhi, "_warn_deprecated_ST_initialize_filtration", True):
    print("SimplexTree.initialize_filtration() is now unnecessary and deprecated", file=sys.stderr)
    gudhi._warn_deprecated_ST_initialize_filtration = False
```

PR #272, opened 2020-04-02
Quantization of persistence diagrams

tlacombe:
Propose two methods to quantize persistence diagrams, that is turning a persistance diagram of size N (with N large) into a persistence diagram of size k, with k chosen by the user. 

- `balanced_quantization` : split the input diagram `Y` into k cells such that the total mass of each cell is the same. Here, the total mass of a cell is the sum of the masses of the points of ``Y`` , where the mass of a point can be anything. By default it is the distance to the diagonal (so the cells have the same "total persistence").

- `kmeans_quantization` : Solve the following problem : given `Y` of mass `N` (by default, the mass is the number of points), find a diagram with `k` points with the same total mass `N` that is close to the input `Y`. Here, closeness is measures in terms of "Wasserstein distance between persistence diagrams" (i.e. it takes the diagonal into account).

TODO: 
- [ ] Write documentation
- [ ] Write a proper test file
- [ ] Check numerical dependencies on parameters (in particular gamma), include that in the doc.
- [ ] Check that 0 entries in the weights do not break everything. Handle that in a better way.
- [x] Discuss with @martinroyer about interactions with ATOL.

### DIFF
--- a/src/python/gudhi/quantization.py
+++ b/src/python/gudhi/quantization.py
@@ -1,0 +1,264 @@
+""" This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+    Author(s):       Theo Lacombe
+
+    Copyright (C) 2019 Inria
+
+    Modification(s):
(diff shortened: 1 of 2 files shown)


tlacombe:
Closing this PR, a new one witha better algorithm will come.

Issue #270, opened 2020-04-01
Bibliography location with sphinx


mglisse:
The bibliography entries currently appear in rather random places, with most of them in witness_complex_user.html. This seems to be in large part due to the bibtex plugin, but some workarounds may be possible.

1. The easiest seems to be having a separate zbibliography.rst file and have a single `.. bibliography::` entry there, instead of trying to have a biblio at the bottom of each page.
2. Not all files that use `:cite:` have a `.. bibliography::` in them (wasserstein, nerve_gic, bottleneck, rips, etc). If we keep the per-file biblio, we should add it to those to be consistent.
3. There are some filtering and prefixing options for `.. bibliography::` which could help put the right references in the right places, and remove some warnings. It may be a bit complicated for summary files which get included multiple times, but that's a minority of cases.

https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html

Of course, switching to a different way to handle the bibliography than this plugin is also a possibility.

PR #275, opened 2020-04-04
Improve local bibliographies in python

mglisse:
I also experimented with changes that remove most warnings, but didn't like the results.
I still get some weirdness, for instance in Wasserstein both references appear as [2], but the links work properly...
Helps with #270.

### DIFF
--- a/src/python/doc/alpha_complex_user.rst
+++ b/src/python/doc/alpha_complex_user.rst
@@ -209,4 +209,3 @@ CGAL citations
 .. bibliography:: ../../biblio/how_to_cite_cgal.bib
    :filter: docname in docnames
    :style: unsrt
-   :labelprefix: A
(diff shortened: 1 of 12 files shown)


mglisse:
On the circleci build, the witness reference has label 2, not 22.
Yes, each paper may appear in several bibliographies, but all hyperlinks for that paper point to the last page that cites it. That bug isn't fixed, I don't know why I thought that links worked... It is even more confusing when using label prefixes, because the label seems to come from the *first* citation and still links to the last one...
Well, still progress I say.

VincentRouvreau:
> On the circleci build, the witness reference has label 2, not 22.
> Yes, each paper may appear in several bibliographies, but all hyperlinks for that paper point to the last page that cites it. That bug isn't fixed, I don't know why I thought that links worked... It is even more confusing when using label prefixes, because the label seems to come from the _first_ citation and still links to the last one...
> Well, still progress I say.

I agree that it is better. Removing all warnings and being able to see new ones would be awesome !

mglisse:
> Removing all warnings and being able to see new ones would be awesome !

That's easy, instead of having one biblio per file, have a single one in bibliography.rst.

PR #276, opened 2020-04-04
Fix links and remove warnings in biblio

### LINKED ISSUE #270: Bibliography location with sphinx

mglisse:
This is super fragile, if the package is ever updated...
Fix #270.

### DIFF
--- a/src/python/doc/conf.py
+++ b/src/python/doc/conf.py
@@ -203,3 +203,75 @@ html_last_updated_fmt = '%b %d, %Y'
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'GUDHIdoc'
 
+# Vincent hates those warnings
+import sphinxcontrib.bibtex
+sphinxcontrib.bibtex.check_duplicate_labels = lambda app, env: None
+
+# Automatically prefix citation keys with the file name
+
+import sphinxcontrib.bibtex.cache
+
+def my_get_bibliography_entries(self, docname, id_, warn):
+    """Return filtered bibliography entries, sorted by occurence
+    in the bib file.
+    Modified to use docname as keyprefix
+    """
+    import six, copy
+    from sphinxcontrib.bibtex.cache import _FilterVisitor
+    # get the information of this bibliography node
+    bibcache = self.get_bibliography_cache(docname=docname, id_=id_)
+    # generate entries
+    for bibfile in bibcache.bibfiles:
+        data = self.bibfiles[bibfile].data
+        for entry in six.itervalues(data.entries):
+            # beware: the prefix is not stored in the data
+            # to allow reusing the data for multiple bibliographies
+            cited_docnames = self.get_cited_docnames(
+                docname + entry.key)
+            visitor = _FilterVisitor(
+                entry=entry,
+                docname=docname,
+                cited_docnames=cited_docnames)
+            try:
+                success = visitor.visit(bibcache.filter_)
+            except ValueError as err:
+                warn("syntax error in :filter: expression; %s" % err)
+                # recover by falling back to the default
+                success = bool(cited_docnames)
+            if success:
+                # entries are modified in an unpickable way
+                # when formatting, so fetch a deep copy
+                # and return this copy with prefixed key
+                # we do not deep copy entry.collection because that
+                # consumes enormous amounts of memory
+                entry.collection = None
+                entry2 = copy.deepcopy(entry)
+                entry2.key = docname + entry.key
+                entry2.collection = data
+                entry.collection = data
+                yield entry2
+
+sphinxcontrib.bibtex.cache.Cache._get_bibliography_entries = my_get_bibliography_entries
+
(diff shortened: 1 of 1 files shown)


VincentRouvreau:
Here is the result I have when performing `make sphinx`: https://gist.github.com/VincentRouvreau/85c272676eea4d2dfd56e81deec99226
I do not really understand how you did that, but it seems to be part of the [sphinxcontrib-bibtex API](https://sphinxcontrib-bibtex.readthedocs.io/en/0.2.8/api/interface.html).
Do you think removing the `duplicate labels` warnings is because we need a biblio page ? Or we need some kind of magic like you did ?

VincentRouvreau:
Oops, sorry for the closure.

mglisse:
https://app.circleci.com/pipelines/github/mglisse/gudhi-devel/150/workflows/54cedcb3-80d2-43c8-a4c7-738071bd15c3/jobs/750 does **not** show the same warnings as your gist.

How I did it: basically I patched sphinxcontrib.bibtex. Looking at a diff between the functions in this patch and the ones they replace in that package would be clearer:

```diff
-def result_nodes(self, document, env, node, is_ref):
+def my_result_nodes(self, document, env, node, is_ref):
     """Transform reference node into a citation reference,
     and note that the reference was cited.
+    Modified to use docname as keyprefix
     """
-    keys = node['reftarget'].split(',')
+    keys = [env.docname+r for r in node['reftarget'].split(',')]
     # Note that at this point, usually, env.bibtex_cache.bibfiles
     # is still empty because the bibliography directive may not
     # have been processed yet, so we cannot get the actual entry.
     # Instead, we simply fake an entry with the desired key, and
     # fix the label at doctree-resolved time. This happens in
     # process_citation_references.
     refnodes = [
         self.backend.citation_reference(_fake_entry(key), document)
         for key in keys]
     for key in keys:
         env.bibtex_cache.add_cited(key, env.docname)
     return refnodes, []
```
```diff
-def _get_bibliography_entries(self, docname, id_, warn):
+def my_get_bibliography_entries(self, docname, id_, warn):
     """Return filtered bibliography entries, sorted by occurence
     in the bib file.
+    Modified to use docname as keyprefix
     """
     # get the information of this bibliography node
     bibcache = self.get_bibliography_cache(docname=docname, id_=id_)
     # generate entries
     for bibfile in bibcache.bibfiles:
         data = self.bibfiles[bibfile].data
         for entry in six.itervalues(data.entries):
             # beware: the prefix is not stored in the data
             # to allow reusing the data for multiple bibliographies
             cited_docnames = self.get_cited_docnames(
-                bibcache.keyprefix + entry.key)
+                docname + entry.key)
             visitor = _FilterVisitor(
                 entry=entry,
                 docname=docname,
                 cited_docnames=cited_docnames)
             try:
                 success = visitor.visit(bibcache.filter_)
             except ValueError as err:
                 warn("syntax error in :filter: expression; %s" % err)
                 # recover by falling back to the default
                 success = bool(cited_docnames)
             if success:
                 # entries are modified in an unpickable way
                 # when formatting, so fetch a deep copy
                 # and return this copy with prefixed key
                 # we do not deep copy entry.collection because that
                 # consumes enormous amounts of memory
                 entry.collection = None
                 entry2 = copy.deepcopy(entry)
-                entry2.key = bibcache.keyprefix + entry.key
+                entry2.key = docname + entry.key
                 entry2.collection = data
                 entry.collection = data
                 yield entry2
```
plus various `import` because of where the function is defined. This part was about getting references to the biblio in the same page.

And I replaced the function `check_duplicate_labels` which was printing warnings with a function that does nothing...

mglisse:
> we need a biblio page ?

I think I tried that (not very hard) and was surprised that it didn't seem to work at all. I was probably doing it wrong...

VincentRouvreau:
I will have a closer look at this [sphinxcontrib bibtex issue](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/62). I think this is the good way to fix it.

mglisse:
I'll close this. It works, but the single-biblio #292 is less likely to break and easier to maintain (we don't need to add a biblio section to every page).

Issue #255, opened 2020-03-18
[Python] Split compute_persistence() from persistence()


mglisse:
It is strange that computing the persistence diagram is done in persistence, which is one of the functions giving some view into the diagram.
I am posting this suggestion here without thinking, we need to consider how much trouble it would cause.
We could have a function compute_persistence, that does just that, and for instance in intervals_in_dimension we would ask users to call compute_persistence first, instead of asking them to call persistence().
For compatibility, persistence would call compute_persistence if the diagram wasn't precomputed. Actually, intervals_in_dimension could do that too. However, currently, persistence recomputes the diagram from scratch if it is called again, and I would be tempted to restrict that behavior to the new compute_persistence.

PR #277, opened 2020-04-06
compute_persistence in python

mglisse:
Based on #255.
I am adding a new function `compute_persistence` in python for when we were calling `persistence` but don't care about the result, because we want to read the diagram with `persistence_intervals_in_dimension` for instance.
Since persistence computation requires some parameters, it seems hard to do it implicitly from `persistent_betti_numbers`, so we still have to call (compute_)persistence first. In the reverse direction, dropping the parameters of `persistence` would break too much. Even just ignoring them and using an already computed persistence if one exists (and building one otherwise) seemed to change a bit much.
The performance gain is non-existent. I find it a bit less disturbing to call compute_persistence than persistence when I won't use the list it returns. But that's subjective, I expect some people will find it simpler if there are fewer functions that do similar things (i.e. no `compute_persistence`).
I took the chance to do a number of small clean-ups. If we don't merge compute_persistence, I am still interested in merging those.


src/python/gudhi/simplex_tree.pyx
```diff
@@ -427,11 +427,11 @@ cdef class SimplexTree:
             0.0.
             Sets min_persistence to -1.0 to see all values.
         :type min_persistence: float.
-        :returns: A list of four persistence diagrams in the format described in :func:`persistence()<gudhi.SimplexTree.persistence>`. The first one is Ordinary, the second one is Relative, the third one is Extended+ and the fourth one is Extended-. See https://link.springer.com/article/10.1007/s10208-008-9027-z and/or section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z for a description of these subtypes.
+        :returns: A list of four persistence diagrams in the format described in :func:`persistence`. The first one is Ordinary, the second one is Relative, the third one is Extended+ and the fourth one is Extended-. See https://link.springer.com/article/10.1007/s10208-008-9027-z and/or section 2.2 in https://link.springer.com/article/10.1007/s10208-017-9370-z for a description of these subtypes.
```
VincentRouvreau, line 430: A bibtex link would be better, no ?

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -571,6 +571,7 @@ class Persistent_cohomology {
 
   void write_output_diagram(std::string diagram_name) {
     std::ofstream diagram_out(diagram_name.c_str());
+    diagram_out.exceptions(diagram_out.failbit);
```
VincentRouvreau, line 574: I did not know this one. Will it be catched at python level ?

mglisse: It gets converted to a python exception (because I added `except +` in the pxd file). Then it is up to the user to catch the exception if they call write_persistence_diagram with a bad file name, as it would be if they just called `open`.

src/python/gudhi/simplex_tree.pyx
mglisse: Maybe, but that's independent of this patch. The docstring for SimplexTree already has an inline citation for Clément's paper, so it is not the only place. One possible reason not to use bibtex in the python files (I don't know if that's why people did it) is that the text is not just used to generate the sphinx doc, where bibtex runs, it also appears when you do `help(gudhi.SimplexTree)` in python, where there is no bibtex. That may not be a good enough reason not to use bibtex, but I'd say, separate issue/PR to convert them all at once (if people agree that we should change them).

PR #281, opened 2020-04-14
Edge collapse integration vincent

VincentRouvreau:


src/Collapse/utilities/distance_matrix_edge_collapse_rips_persistence.cpp
```diff
@@ -0,0 +1,177 @@
...
+  Filtration_value threshold;
+  int dim_max = 2;
+  int p;
+  Filtration_value min_persistence;
+
+  program_options(argc, argv, csv_matrix_file, filediag, threshold, dim_max, p, min_persistence);
+
+  Distance_matrix distances;
+
+  distances = Gudhi::read_lower_triangular_matrix_from_csv_file<Filtration_value>(csv_matrix_file);
```
mglisse, line 84: Could you put that on the same line as the declaration, to avoid copying?

src/common/doc/main_page.md
```diff
@@ -242,6 +242,36 @@
  </tr>
 </table>
 
+#### Strong collapse
+
+<table>
+  <tr>
+    <td width="35%" rowspan=2>
+      \image html "edge_collapse_representation.png"
```
mglisse, line 250: Image missing.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+namespace collapse {
+
+/**
+ * \class Flag_complex_sparse_matrix
+ * \brief Flag complex sparse matrix data structure.
+ *
+ * \ingroup collapse
+ *
+ * \details
+ * A class to store the vertices v/s MaxSimplices Sparse Matrix and to perform collapse operations using the N^2()
```
mglisse, line 47: Strange sentence.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  using Sparse_row_matrix = Eigen::SparseMatrix<Filtration_value, Eigen::RowMajor>;
+
+  // A range of row indices
+  using Row_indices_vector = std::vector<Row_index>;
+
+ public:
+  /** \brief A Filtered_edge is a std::pair<std::pair<`Vertex_handle`, `Vertex_handle`>, `Filtration_value`>. */
+  using Filtered_edge = std::pair<Edge, Filtration_value>;
+  /** \brief Proximity_graph is a type that can be used to construct easily a Flag_complex_sparse_matrix. */
+  using Proximity_graph = Gudhi::Proximity_graph<Flag_complex_sparse_matrix>;
```
mglisse, line 79: I can't find Gudhi::Proximity_graph in the doc.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  using Row_index = std::size_t;
+
+  // The sparse matrix data type
+  using Sparse_row_matrix = Eigen::SparseMatrix<Filtration_value, Eigen::RowMajor>;
+
+  // A range of row indices
+  using Row_indices_vector = std::vector<Row_index>;
+
+ public:
+  /** \brief A Filtered_edge is a std::pair<std::pair<`Vertex_handle`, `Vertex_handle`>, `Filtration_value`>. */
```
mglisse, line 76: The description seems redundant?

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+    for (auto edge_it = boost::edges(one_skeleton_graph);
+         edge_it.first != edge_it.second; ++edge_it.first) {
+      auto edge = *(edge_it.first);
+      Vertex_handle u = source(edge, one_skeleton_graph);
+      Vertex_handle v = target(edge, one_skeleton_graph);
+      f_edge_vector_.push_back({{u, v}, boost::get(Gudhi::edge_filtration_t(), one_skeleton_graph, edge)});
+    }
+  }
+
+  /** \brief Performs edge collapse in a decreasing sequence of the filtration value.
```
mglisse, line 361: The user manual says increasing?

src/Collapse/utilities/distance_matrix_edge_collapse_rips_persistence.cpp
```diff
@@ -0,0 +1,177 @@
...
+                                                                                   return distances[j][i];
+                                                                                 });
+
+  // Now we will perform filtered edge collapse to sparsify the edge list edge_t.
+  Flag_complex_sparse_matrix flag_complex(proximity_graph);
+
+  Simplex_tree stree;
+  flag_complex.filtered_edge_collapse(
+    [&stree](std::vector<Vertex_handle> edge, Filtration_value filtration) {
+        // insert the 2 vertices with a 0. filtration value just like a Rips
```
mglisse, line 100: IIUC edge collapses preserve the 0-skeleton. It seems that it would be better to insert first all the vertices before doing the collapses, and then the callback can insert just the edge.

biblio/bibliography.bib
```diff
@@ -1219,3 +1219,16 @@ @article{turner2014frechet
...
 }
+
+@unpublished{edgecollapsesocg2020,
+  title = {{Edge Collapse and Persistence of Flag Complexes}},
+  author = {Boissonnat, Jean-Daniel and Pritam, Siddharth},
+  url = {https://hal.inria.fr/hal-02395227},
+  note = {working paper or preprint},
+  year = {2019},
+  month = Dec,
+  keywords = {Persistent homology ; Strong Collapse ; Computational geometry ; Topological Data Analysis ; 2012 ACM Subject Classification Mathematics of computing ; Computational Topology},
```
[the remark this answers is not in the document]
VincentRouvreau: You want me to replace `strong collapse` with `edge collapse` ?
And yes it was a copy / paste from hal.

src/common/doc/main_page.md
VincentRouvreau: I am waiting for it from Siddharth

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: compute_proximity_graph is now available in the documentation and search. If you want a link to it, I have to set `EXTRACT_ALL = YES`in `Doxyfile.in`. The side effect of it is that it generates documentation even for `EXCLUDE` list (hera, python, ...)

src/Collapse/doc/intro_edge_collapse.h
```diff
@@ -0,0 +1,101 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+ *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+ *    Author(s):       Siddharth Pritam
+ *
+ *    Copyright (C) 2019 Inria
```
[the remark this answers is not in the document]
VincentRouvreau: Fix on 4fdc721

src/Collapse/utilities/distance_matrix_edge_collapse_rips_persistence.cpp
VincentRouvreau: Fix on 4fdc721

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Fix on 4fdc721

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Fix on 4fdc721

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Fix on 4fdc721. I checked on user manual and it is increasing.

biblio/bibliography.bib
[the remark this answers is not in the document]
mglisse: I was referring to "2012 ACM Subject Classification Mathematics of computing" which doesn't really look like a keyword to me. But never mind, let's keep it as a copy-paste, and we'll update it to a copy-paste of the official SoCG version when that appears.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+          // If idx is affected
+          if (effected_indices.find(idx) != effected_indices.end()) {
+            if (!check_edge_domination(edge)) {
+#ifdef DEBUG_TRACES
+              std::cout << "The curent index became critical " << idx  << std::endl;
+#endif  // DEBUG_TRACES
+              critical_edge_indicator_[idx] = true;
+              filtered_edge_insert({u, v}, filt);
+              std::set<Row_index> inner_effected_indcs = three_clique_indices(idx);
+              for (auto inr_idx = inner_effected_indcs.rbegin(); inr_idx != inner_effected_indcs.rend(); inr_idx++) {
```
mglisse, line 212: Any particular reason to do this backwards? If not, a for range loop would make sense.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  std::unordered_map<Row_index, Vertex_handle> row_to_vertex_;
+
+  // Vertices stored as an unordered_set
+  std::unordered_set<Vertex_handle> vertices_;
+
+  // Unordered set of removed edges. (to enforce removal from the matrix)
+  std::unordered_set<Edge, boost::hash<Edge>> u_set_removed_edges_;
+
+  // Unordered set of dominated edges. (to inforce removal from the matrix)
+  std::unordered_set<Edge, boost::hash<Edge>> u_set_dominated_edges_;
```
mglisse, line 92: I am trying to understand if this is redundant with critical_edge_indicator_ or not. It seems that filling this structure is what prevents us in set_edge_critical from iterating on effected_indices, instead of going through all edges.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+#ifdef DEBUG_TRACES
+              std::cout << "The curent index became critical " << idx  << std::endl;
+#endif  // DEBUG_TRACES
+              critical_edge_indicator_[idx] = true;
+              filtered_edge_insert({u, v}, filt);
+              std::set<Row_index> inner_effected_indcs = three_clique_indices(idx);
+              for (auto inr_idx = inner_effected_indcs.rbegin(); inr_idx != inner_effected_indcs.rend(); inr_idx++) {
+                if (*inr_idx < idx) effected_indices.emplace(*inr_idx);
+              }
+              inner_effected_indcs.clear();
```
mglisse, line 215: Useless just before destruction.

biblio/bibliography.bib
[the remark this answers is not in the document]
VincentRouvreau: You are right it is strange. Done on #08967b0d

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+
+ public:
+  /** \brief Filtered_edge is a type to store an edge with its filtration value. */
+  using Filtered_edge = std::pair<Edge, Filtration_value>;
+  /** \brief Proximity_graph is a type that can be used to construct easily a Flag_complex_sparse_matrix. */
+  using Proximity_graph = Gudhi::Proximity_graph<Flag_complex_sparse_matrix>;
+
+ private:
+  // Map from row index to its vertex handle
+  std::unordered_map<Row_index, Vertex_handle> row_to_vertex_;
```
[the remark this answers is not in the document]
VincentRouvreau: You are right. I fixed the review on 8917c9bb

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+    // Sort edges
+    auto sort_by_filtration = [](const Filtered_edge& edge_a, const Filtered_edge& edge_b) -> bool
+    {
+      return (get<1>(edge_a) < get<1>(edge_b)); 
+    };
+
+#ifdef GUDHI_USE_TBB
+    tbb::parallel_sort(f_edge_vector_.begin(), f_edge_vector_.end(), sort_by_filtration);
+#else
+    std::stable_sort(f_edge_vector_.begin(), f_edge_vector_.end(), sort_by_filtration);
```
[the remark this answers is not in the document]
VincentRouvreau: Sorry about that, it was just a copy/paste without thinking about the thing you mention: 2de21960

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  using Filtered_edge = std::pair<Edge, Filtration_value>;
+  /** \brief Proximity_graph is a type that can be used to construct easily a Flag_complex_sparse_matrix. */
+  using Proximity_graph = Gudhi::Proximity_graph<Flag_complex_sparse_matrix>;
+
+ private:
+  // Map from row index to its vertex handle
+  std::unordered_map<Row_index, Vertex_handle> row_to_vertex_;
+
+  // Vertices stored as an unordered_set
+  std::unordered_set<Vertex_handle> vertices_;
```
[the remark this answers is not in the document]
VincentRouvreau: You are right there is no added value to store such values. I fixed it on 9273dd8d

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+      sparse_row_adjacency_matrix_.insert(rows_, rows_) = filt_val;
+      domination_indicator_.push_back(false);
+      vertex_to_row_.insert(std::make_pair(vertex, rows_));
+      row_to_vertex_.insert(std::make_pair(rows_, vertex));
+      rows_++;
+    }
+  }
+
+  // Insert an edge in the data structure
+  void insert_new_edges(Vertex_handle u, Vertex_handle v, Filtration_value filt_val)
```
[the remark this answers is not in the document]
VincentRouvreau: Rename insert_new_edge method on 71af6855

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  //! Stores the number of vertices in the original graph (which is also the number of rows in the Matrix).
+  Row_index rows_;
+
+  // Edge e is the actual edge (u,v). Not the row ids in the matrixs
+  bool check_edge_domination(const Edge& edge) const
+  {
+    Vertex_handle u = std::get<0>(edge);
+    Vertex_handle v = std::get<1>(edge);
+
+    const Row_index rw_u = vertex_to_row_.at(u);
```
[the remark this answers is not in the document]
VincentRouvreau: Here this is more a matter of const. `.at()` has a const signature in `std::unordered_map`, whereas `[]`does not have one.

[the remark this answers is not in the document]
mglisse: Uh, I had never noticed that. I don't understand why the standard doesn't provide a const version of operator[], but that's how it is...

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+    non_zero_indices_v = closed_neighbours_row_index(rw_v);
+    std::set_intersection(non_zero_indices_u.begin(), non_zero_indices_u.end(), non_zero_indices_v.begin(),
+                          non_zero_indices_v.end(), std::inserter(common, common.begin()));
+
+    return common;
+  }
+
+  // Insert a vertex in the data structure
+  void insert_vertex(Vertex_handle vertex, Filtration_value filt_val) {
+    auto rw = vertex_to_row_.find(vertex);
```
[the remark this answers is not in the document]
VincentRouvreau: Use of emplace on 1269de8a

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -273,12 +273,12 @@ class Flag_complex_sparse_matrix {
 
   // Insert a vertex in the data structure
   void insert_vertex(Vertex_handle vertex, Filtration_value filt_val) {
-    auto rw = vertex_to_row_.find(vertex);
-    if (rw == vertex_to_row_.end()) {
+    auto result = vertex_to_row_.emplace(std::make_pair(vertex, rows_));
```
mglisse, line 276: without `make_pair` please.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  // Subsequent removal of dominated vertices is reflected by concerned entries changing to <I>true</I>
+  // in this vector.
+  std::vector<bool> domination_indicator_;
+
+  // Vector of filtered edges, for edge-collapse, the indices of the edges are the row-indices.
+  std::vector<Filtered_edge> f_edge_vector_;
+
+
+  //! Stores the number of vertices in the original graph (which is also the number of rows in the Matrix).
+  Row_index rows_;
```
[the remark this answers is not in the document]
VincentRouvreau: Yes you are completely right. I removed it on b9e6a2d6

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+      auto rw_v = vertex_to_row_.find(v);
+#ifdef DEBUG_TRACES
+      std::cout << "Inserting the edge " << u <<", " << v << std::endl;
+#endif  // DEBUG_TRACES
+      sparse_row_adjacency_matrix_.insert(rw_u->second, rw_v->second) = filt_val;
+      sparse_row_adjacency_matrix_.insert(rw_v->second, rw_u->second) = filt_val;
+    }
+#ifdef DEBUG_TRACES
+    else {
+     	std::cout << "Already a member simplex,  skipping..." << std::endl;
```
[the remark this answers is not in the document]
VincentRouvreau: Well seen ! This trace is not correct. It happens when we try to insert an edge `[u, u]` for instance, but this is not an edge, no ? I will try to think of what would be a good behaviour, maybe a `GUDHI_CHECK` would be better.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+
+  // Vector of filtered edges, for edge-collapse, the indices of the edges are the row-indices.
+  std::vector<Filtered_edge> f_edge_vector_;
+
+
+  //! Stores the number of vertices in the original graph (which is also the number of rows in the Matrix).
+  Row_index rows_;
+
+  // Edge e is the actual edge (u,v). Not the row ids in the matrixs
+  bool check_edge_domination(const Edge& edge) const
```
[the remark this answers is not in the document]
VincentRouvreau: I agree, I renamed it on 6eb48ea4

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+    Vertex_handle u = std::get<0>(edge);
+    Vertex_handle v = std::get<1>(edge);
+
+    const Row_index rw_u = vertex_to_row_.at(u);
+    const Row_index rw_v = vertex_to_row_.at(v);
+    auto rw_e = std::make_pair(rw_u, rw_v);
+#ifdef DEBUG_TRACES
+    std::cout << "The edge {" << u << ", " << v <<  "} is going for domination check." << std::endl;
+#endif  // DEBUG_TRACES
+    auto common_neighbours = closed_common_neighbours_row_index(rw_e);
```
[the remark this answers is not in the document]
VincentRouvreau: Passing 2 vertices as an argument is a good idea. Done on 2795e009

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+      std::cout << row_to_vertex_[neighbour] << ", " ;
+    }
+    std::cout<< std::endl;
+#endif  // DEBUG_TRACES
+    if (common_neighbours.size() > 2) {
+      if (common_neighbours.size() == 3)
+        return true;
+      else
+        for (auto rw_c : common_neighbours) {
+          if (rw_c != rw_u and rw_c != rw_v) {
```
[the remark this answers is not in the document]
VincentRouvreau: I modified it on 1fc68dc as it was also making compilation fail on Windows

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+  // Stores the Sparse matrix of Filtration values representing the original graph.
+  // This is row-major version of the same sparse-matrix, to facilitate easy access
+  // to elements when traversing the matrix row-wise.
+  Sparse_row_matrix sparse_row_adjacency_matrix_;
+
+  // Stores <I>true</I> for dominated rows and <I>false</I> otherwise.
+  // Initialised to a vector of length equal to the value of the variable <B>rows_</B> with all <I>false</I> values.
+  // Subsequent removal of dominated vertices is reflected by concerned entries changing to <I>true</I>
+  // in this vector.
+  std::vector<bool> domination_indicator_;
```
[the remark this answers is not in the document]
VincentRouvreau: Oops, yes: 26f396ce

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+    Row_indices_vector non_zero_indices;
+#ifdef DEBUG_TRACES
+    std::cout << "The neighbours of the vertex: " << row_to_vertex_[rw_u] << " are. " << std::endl;
+#endif  // DEBUG_TRACES
+    if (!domination_indicator_[rw_u]) {
+      // Iterate over the non-zero columns
+      for (typename Sparse_row_matrix::InnerIterator it(sparse_row_adjacency_matrix_, rw_u); it; ++it) {
+        Row_index rw_v = it.index();
+        // If the vertex v is not dominated and the edge {u,v} is still in the matrix
+        if (!domination_indicator_[rw_v] and u_set_removed_edges_.find(std::minmax(rw_u, rw_v)) == u_set_removed_edges_.end() and
```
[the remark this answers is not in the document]
VincentRouvreau: Thanks for the tips. I let it as is.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+#endif  // DEBUG_TRACES
+            } else
+              u_set_dominated_edges_.emplace(std::minmax(vertex_to_row_[u], vertex_to_row_[v]));
+          } else
+            // Idx is not affected hence dominated.
+            u_set_dominated_edges_.emplace(std::minmax(vertex_to_row_[u], vertex_to_row_[v]));
+        }
+      }
+    }
+    effected_indices.clear();
```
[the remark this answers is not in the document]
VincentRouvreau: d10b2511

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+            if (std::includes(neighbours_c.begin(), neighbours_c.end(), common_neighbours.begin(),
+                              common_neighbours.end()))
+              return true;
+          }
+        }
+    }
+    return false;
+  }
+
+  std::set<Row_index> three_clique_indices(Row_index crit) {
```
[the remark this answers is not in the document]
VincentRouvreau: I took your proposal ac8f1911
I will check with Siddharth

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+      auto edge = *(edge_it.first);
+      Vertex_handle u = source(edge, one_skeleton_graph);
+      Vertex_handle v = target(edge, one_skeleton_graph);
+      f_edge_vector_.push_back({{u, v}, boost::get(Gudhi::edge_filtration_t(), one_skeleton_graph, edge)});
+    }
+  }
+
+  /** \brief Performs edge collapse in a increasing sequence of the filtration value.
+   *
+   * \tparam FilteredEdgeInsertion is an output iterator that furnishes
```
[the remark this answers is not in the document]
VincentRouvreau: You are right. Done on ed550043

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: You're correct, done on 1ab1d498

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Oops, sorry: 55aba07

src/common/doc/main_page.md
```diff
@@ -242,6 +242,36 @@
  </tr>
 </table>
 
+#### Edge collapse
```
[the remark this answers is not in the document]
VincentRouvreau: I moved it right after the Rips for the moment. We will see where it fits the best.

src/Collapse/doc/intro_edge_collapse.h
```diff
@@ -0,0 +1,101 @@
...
+ * -- For general simplicial complex: An edge \f$e \in K\f$ is dominated by another vertex \f$v^{\prime} \in K\f$,
+ * <i>if and only if</i> all the maximal simplices of \f$K\f$ that contain \f$e\f$ also contain \f$v^{\prime}\f$
+ * 
+ * -- For a flag complex: An edge \f$e \in K\f$ is dominated by another vertex \f$v^{\prime} \in K\f$, <i>if and only
+ * if</i> all the vertices in \f$K\f$ that has an edge with both vertices of \f$e\f$  also has an edge with
+ * \f$v^{\prime}\f$.
+
+ * This module implements edge collapse of a filtered flag complex, in particular  it reduces a filtration of
+ * Vietoris-Rips complex from its graph 
+ * to another smaller flag filtration with same persistence. Where a filtration is a sequence of simplicial
```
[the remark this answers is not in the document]
VincentRouvreau: I moved a sentence on 50f9bb7d to sum up the module at the beginning.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+   * `({Vertex_handle u, Vertex_handle v}, Filtration_value f)` that will fill the user defined data structure.
+   */
+  template<typename FilteredEdgeInsertion>
+  void filtered_edge_collapse(FilteredEdgeInsertion filtered_edge_insert) {
+    Row_index endIdx = 0;
+
+    u_set_removed_edges_.clear();
+    u_set_dominated_edges_.clear();
+    critical_edge_indicator_.clear();
```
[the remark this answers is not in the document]
VincentRouvreau: Removed on d46d9ef6

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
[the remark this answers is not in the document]
VincentRouvreau: GUDHI_CHECK if u==v on insert_edge method 08ffcedc

src/Collapse/test/collapse_unit_test.cpp
```diff
@@ -0,0 +1,191 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
```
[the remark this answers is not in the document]
VincentRouvreau: I added an example to test the persistence is exactly the same than the Rips presistence on e41edaed
It does not work if you set a small value for threshold (0.2 for instance with torus on 300 points)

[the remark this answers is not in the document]
mglisse: Is that a bug in edge collapse ?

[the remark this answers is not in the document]
VincentRouvreau: It was my fault. I fixed it on a106902b
Maybe setting inserting all vertices with a `0.` filtration value would require some documentation ?

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Range loop on 9ae3c860

src/Collapse/utilities/distance_matrix_edge_collapse_rips_persistence.cpp
VincentRouvreau: If I would have read this comment more carefully it would have fixed my bug :disappointed:

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: I removed it on 85dc5c1e
It is "strange" as I have to catch exception to know if it is present or not.
Do not hesitate if you want me to rollback.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,424 @@
...
+#ifdef GUDHI_USE_TBB
+    tbb::parallel_sort(f_edge_vector_.begin(), f_edge_vector_.end(), sort_by_filtration);
+#else
+    std::stable_sort(f_edge_vector_.begin(), f_edge_vector_.end(), sort_by_filtration);
+#endif
+
+    // Initializing sparse_row_adjacency_matrix_, This is a row-major sparse matrix.
+    sparse_row_adjacency_matrix_ = Sparse_row_matrix(vertices_.size(), vertices_.size());
+
+    while (endIdx < f_edge_vector_.size()) {
```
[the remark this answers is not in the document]
VincentRouvreau: Better use a for loop. Done on a81c5adc

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -224,9 +217,17 @@ class Flag_complex_sparse_matrix {
     // Iterate over the non-zero columns
     for (typename Sparse_row_matrix::InnerIterator it(sparse_row_adjacency_matrix_, rw_u); it; ++it) {
       Row_index rw_v = it.index();
+
+      bool dominated_edge_not_found = false;
+      try {
+        edge_to_index_map_.at(std::minmax(row_to_vertex_[rw_u], row_to_vertex_[rw_v]));
+      } catch (const std::out_of_range& oor) {
```
mglisse, line 224: Uh, you have `find` or `count` if you want to test if an element is present. It is surprising that u_set_dominated_edges_ gets removed for being redundant with critical_edge_indicator_ and the diff does not mention critical_edge_indicator_ anywhere, but that's possible, I didn't read it.

VincentRouvreau: My modification was not working. We do not have the same information in both structures...
Rollback on df758d75

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
mglisse:
```c++
      if (rw_u != rw_v) {
        bool b1 = u_set_dominated_edges_.count(std::minmax(rw_u, rw_v)) != 0;
        Row_index ei = edge_to_index_map_.at(std::minmax(row_to_vertex_[rw_u], row_to_vertex_[rw_v]));
        bool b2 = ei > current_backwards && !critical_edge_indicator_[ei];
        assert(b1==b2);
      }
```
does not fail while running the tests.

mglisse:
Hmm, probably  because of the use of unordered_map, this last patch, which avoids iterating on many useless edges, has made the code slower... :slightly_frowning_face: I'll look at that.

mglisse:
I am trying to think if we should give this more of a streaming kind of interface. It doesn't seem necessary to know all the edges in advance, we can add them as we go. We would create the class with something like `auto fc = makeFlagCollapser(output_function)` (we might also still have to specify some integer type). And then in a loop call `fc.insert_edge(u, v)` which itself may call output_function 0, 1, or many more times. I am not sure if fc should call `output_function(i)` where i is the index of the edge in the insertion order, or `output_function(u,v)` where u and v are the 2 vertices at the extremities of the edge.
To achieve something similar to the current code using that interface could look like
```c++
std::vector<tuple<int,int,double>> edges; // sorted
Filtration_value filt;
// insert vertices first
auto of=[&](int e){auto edge=edges[e];st.insert_simplex({get<0>(edge),get<1>(edge)},filt);};
// auto of=[&](int u, int v){st.insert_simplex({u,v},filt);};
auto fc = makeFlagCollapser(of)
for(auto&edge:edges){
  filt = get<2>(edge);
  fc.insert_edge(get<0>(edge),get<1>(edge));
}
```
which we can wrap in whatever class for convenience if that seems useful.
A drawback of this approach is that it stores the list of edges one more time. But that would be a plain vector, still smaller than the matrix, the hashmap, etc.
Does that make sense?

mglisse:
For memory (we won't do that now, but it could make sense in the future), we can wrap SparseVector in a range interface with
```c++
  class Sparse_vector_range : public ranges::view_facade<Sparse_vector_range> {
    private:
      friend ranges::range_access;
      typename Sparse_vector::InnerIterator it;
      void next() { ++it; }
      bool equal(ranges::default_sentinel_t) const { return !it; }
      std::pair<Row_index, Row_index> read() const { return {it.index(), it.value()}; }
    public:
      Sparse_vector_range() = default;
      explicit Sparse_vector_range(Sparse_vector const&v) : it(v) {}
  };
```
and modify some functions to return ranges
```c++
  auto neighbours_row_index(Row_index rw_u, bool closed) const
  {
    return Sparse_vector_range(sparse_row_adjacency_matrix_[rw_u])
      | ranges::view::remove_if([=](auto const&e){return !closed && e.first == rw_u;})
      | ranges::view::filter([=](auto const&e){
          return (closed && e.first == rw_u) || e.second <= current_backward || critical_edge_indicator_[e.second];})
      | ranges::view::transform([](auto const&e){return e.first;});
  }
```
(remove_if and filter should be merged, I was experimenting)
```c++
  Row_indices_vector open_common_neighbours_row_index(Row_index rw_u, Row_index rw_v) const
  {
    auto non_zero_indices_u = neighbours_row_index(rw_u, false);
    auto non_zero_indices_v = neighbours_row_index(rw_v, false);
    auto inter = ranges::view::set_intersection(non_zero_indices_u, non_zero_indices_v) | ranges::view::common;
    Row_indices_vector common;
    common.insert(common.end(), inter.begin(), inter.end());
    return common;
  }
```
(it could return `inter` without `common` and let edge_is_dominated create a vector, three_clique_indices is fine with a range)
```c++
        if (ranges::includes(neighbours_c, common_neighbours))
```
in edge_is_dominated.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,376 @@
...
+ *
+ * \details
+ * This class stores a <a target="_blank" href="https://en.wikipedia.org/wiki/Clique_complex">Flag complex</a>
+ * in an <a target="_blank" href="https://eigen.tuxfamily.org/dox/group__TutorialSparse.html">Eigen sparse matrix</a>.
+ *
+ * \tparam Vertex type must be a signed integer type. It admits a total order <.
+ * \tparam Filtration type for the value of the filtration function. Must be comparable with <.
+ */
+template<typename Vertex, typename Filtration>
+class Flag_complex_sparse_matrix {
```
mglisse, line 55: I am not fond of the name of the class (and file). I know it is already in namespace `collapse`, but this isn't a datastructure to store flag complexes, this is a utility to collapse edges. To try and keep part of the current name, maybe Flag_complex_edge_collapser? Feel free to suggest other names.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,376 @@
...
+  }
+
+  /** \brief Flag_complex_sparse_matrix constructor from a proximity graph, cf. `Gudhi::compute_proximity_graph`.
+   *
+   * @param[in] one_skeleton_graph The one skeleton graph. The graph must be in
+   * `Flag_complex_sparse_matrix::Proximity_graph`.
+   *
+   * The constructor is computing and filling a vector of `Flag_complex_sparse_matrix::Filtered_edge`
+   */
+  Flag_complex_sparse_matrix(const Proximity_graph& one_skeleton_graph)
```
mglisse, line 316: Why hardcode the type of the graph instead of doing it like `Simplex_tree::insert_graph`?

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,376 @@
...
+      auto edge = *(edge_it.first);
+      Vertex_handle u = source(edge, one_skeleton_graph);
+      Vertex_handle v = target(edge, one_skeleton_graph);
+      f_edge_vector_.push_back({{u, v}, boost::get(Gudhi::edge_filtration_t(), one_skeleton_graph, edge)});
+    }
+  }
+
+  /** \brief Performs edge collapse in a increasing sequence of the filtration value.
+   *
+   * \tparam FilteredEdgeOutput is a functor that furnishes `({Vertex_handle u, Vertex_handle v}, Filtration_value f)`
```
mglisse, line 331: Why not the flat (Vertex_handle u, Vertex_handle v, Filtration_value f)?

src/common/doc/main_page.md
VincentRouvreau: Done on 0a26e55a

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: You are right. I also renamed variable names when constructing a Flag_complex_edge_collapser on fcd06dde

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Done on b525bb68
Tell me if the documentation is ok.
I had to modify the constructor from filtered edges as it was conflicting.

mglisse: `end` doesn't point to the last element, it points one step after it.
That's one reason why I don't like designs with constructors so much, you cannot easily give 2 different names to 2 different constructors like you can with normal functions (member or not).
One possibility is to dispatch automatically depending on if the template type accepts begin/end, if boost::graph_traits is specialized, etc.
Another possibility is to add dummy parameters, so one would have for instance
```c++
typedef Flag_complex_edge_collapser<...> EC;
EC collapser(EC::from_range, the_range);
EC collapser2(EC::from_graph, the_graph);
```
Or we can make the constructor do almost nothing and have users call a function init_from_graph or init_from_edge_range.
Or as you did we can pass the range old-style.
Etc.
Let's forget about it for now, we'll come back to it later.

src/Collapse/example/edge_collapse_basic_example.cpp
```diff
@@ -0,0 +1,45 @@
...
+  Filtered_edge_list graph = {{{0, 1}, 1.},
+                              {{1, 2}, 1.},
+                              {{2, 3}, 1.},
+                              {{3, 0}, 1.},
+                              {{0, 2}, 2.},
+                              {{1, 3}, 2.}};
+
+  Flag_complex_sparse_matrix flag_complex_sparse_matrix(graph);
+
+  Filtered_edge_list collapse_edges;
```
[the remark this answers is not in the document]
VincentRouvreau: You are right, `collapse_edges` was a bad name.
`remaining_edges` on 0a9cadf4

src/Collapse/example/edge_collapse_basic_example.cpp
```diff
@@ -0,0 +1,45 @@
...
+  // Retrieve collapse edges from the output iterator
+  flag_complex_sparse_matrix.filtered_edge_collapse(
+    [&collapse_edges](std::pair<Vertex_handle, Vertex_handle> edge, Filtration_value filtration) {
+        collapse_edges.push_back({edge, filtration});
```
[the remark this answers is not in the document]
VincentRouvreau: Sorry, it is a bad habit. I was not confident with the difference between push_back and emplace_back. Done on 08307b23

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: And what about removing the constructor with `Filtered_edges` ?
The simplex_tree insert_graph has only one signature and it is ok.

src/Collapse/example/edge_collapse_conserve_persistence.cpp
```diff
@@ -0,0 +1,156 @@
...
+using Proximity_graph = Flag_complex_sparse_matrix::Proximity_graph;
+
+using Field_Zp = Gudhi::persistent_cohomology::Field_Zp;
+using Persistent_cohomology = Gudhi::persistent_cohomology::Persistent_cohomology<Simplex_tree, Field_Zp>;
+
+using Persistence_pair = std::tuple<int, Filtration_value, Filtration_value>;
+/*
+ * Compare two intervals by dimension, then by length.
+ */
+struct cmp_intervals_by_dim_then_length {
```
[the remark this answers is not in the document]
VincentRouvreau: Done on 4da6454b
Sorting persistence pairs is mainly to ease the debug if happens.

src/Collapse/example/edge_collapse_conserve_persistence.cpp
```diff
@@ -0,0 +1,156 @@
...
+using Point = std::vector<Filtration_value>;
+using Vector_of_points = std::vector<Point>;
+
+using Flag_complex_sparse_matrix = Gudhi::collapse::Flag_complex_sparse_matrix<Vertex_handle, Filtration_value>;
+using Proximity_graph = Flag_complex_sparse_matrix::Proximity_graph;
+
+using Field_Zp = Gudhi::persistent_cohomology::Field_Zp;
+using Persistent_cohomology = Gudhi::persistent_cohomology::Persistent_cohomology<Simplex_tree, Field_Zp>;
+
+using Persistence_pair = std::tuple<int, Filtration_value, Filtration_value>;
```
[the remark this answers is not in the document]
VincentRouvreau: Yes you are right. I renamed it on fc14ad2d

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Interface seems better : 7e92bb3a

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,376 @@
...
+   * that will get called on the output edges, in non-decreasing order of filtration.
```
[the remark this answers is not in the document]
VincentRouvreau: Done on 7e92bb3a

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,383 @@
...
+          if (effected_indices.find(idx) != effected_indices.end()) {
+            if (!edge_is_dominated(edge)) {
+#ifdef DEBUG_TRACES
+              std::cout << "The curent index became critical " << idx  << std::endl;
+#endif  // DEBUG_TRACES
+              critical_edge_indicator_[idx] = true;
+              filtered_edge_output({u, v}, filt);
+              std::set<Row_index> inner_effected_indcs = three_clique_indices(idx);
+              for (auto inr_idx : inner_effected_indcs) {
+                effected_indices.emplace(inr_idx);
```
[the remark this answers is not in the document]
VincentRouvreau: On this commit 26036ee you modified the code as:
```cpp
          for (auto inr_idx : inner_effected_indcs) {
            if(inr_idx < current_backward) // && !critical_edge_indicator_[inr_idx]
              effected_indices.emplace(inr_idx);
          }
```
It seems to me your proposal is not working anymore.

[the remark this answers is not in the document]
mglisse: indeed

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
```diff
@@ -0,0 +1,383 @@
...
+ */
+template<typename Vertex, typename Filtration>
+class Flag_complex_sparse_matrix {
+ public:
+  /** \brief Re-define Vertex as Vertex_handle type to ease the interface with compute_proximity_graph. */
+  using Vertex_handle = Vertex;
+  /** \brief Re-define Filtration as Filtration_value type to ease the interface with compute_proximity_graph. */
+  using Filtration_value = Filtration;
+  /** \brief This is an ordered pair, An edge is stored with convention of the first element being the smaller i.e
+   * {2,3} not {3,2}. However this is at the level of row indices on actual vertex lables.
```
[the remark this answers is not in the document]
VincentRouvreau: I removed Edge typedef and its documentation on 7e92bb3

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
VincentRouvreau: Ok, let's precise on 716085ac

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,364 @@
...
+   */
+  template<class OneSkeletonGraph>
+  Flag_complex_edge_collapser(const OneSkeletonGraph& one_skeleton_graph) {
+    // Insert all edges
+    for (auto edge_it = boost::edges(one_skeleton_graph);
+         edge_it.first != edge_it.second; ++edge_it.first) {
+      auto edge = *(edge_it.first);
+      Vertex_handle u = source(edge, one_skeleton_graph);
+      Vertex_handle v = target(edge, one_skeleton_graph);
+      f_edge_vector_.emplace_back(Filtered_edge(u, v, boost::get(Gudhi::edge_filtration_t(), one_skeleton_graph, edge)));
```
mglisse, line 313: I think we are not supposed to write `boost::` in front for `edges` and `get`, according to the boost EdgeListGraph and PropertyGraph concepts.

src/Collapse/include/gudhi/Flag_complex_sparse_matrix.h
mglisse: If we keep only one of them, I think the one taking a range is more important. With an EdgeListGraph, I can write a one-liner to give it the interface of a range:
```c++
boost::adaptors::transform(edges(graph), [&](auto&&x){return std::make_tuple(source(x, graph), target(x, graph), get(Gudhi::edge_filtration_t(), graph, x));})
```
while if I have a range and need to give it the interface of an EdgeListGraph, I need
- boost::adaptors::transform so that the value_type is a type with .first and .second
- boost::edge_list to wrap it with a graph interface
- associate a property to this graph, which seems to require different code depending if the range was random-access, using undocumented implementation details of edge_list.

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,364 @@
...
+ *
+ * \details
+ * This class stores a <a target="_blank" href="https://en.wikipedia.org/wiki/Clique_complex">Flag complex</a>
+ * in an <a target="_blank" href="https://eigen.tuxfamily.org/dox/group__TutorialSparse.html">Eigen sparse matrix</a>.
+ *
+ * \tparam Vertex type must be a signed integer type. It admits a total order <.
+ * \tparam Filtration type for the value of the filtration function. Must be comparable with <.
+ */
+template<typename Vertex, typename Filtration>
+class Flag_complex_edge_collapser {
```
mglisse, line 55: This is a general design comment. From the perspective of a user, what it the advantage of having a class, with the parameters arbitrarily split between template parameters, constructor arguments and executor arguments, instead of a simple function like
```c++
template<class FilteredEdgeRange, class EdgeOutput> void flag_complex_collapse_edges(FilteredEdgeRange const&, EdgeOutput);
```
or
```c++
template<class FilteredEdgeRange> auto flag_complex_collapse_edges(FilteredEdgeRange const&);
```
(which returns a range)?
That wouldn't prevent from using a class internally, as an undocumented implementation detail.

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
VincentRouvreau: Done on ddd12a6c

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,364 @@
...
+   * `Flag_complex_edge_collapser::process_edges`.
+   */
+  template<typename Filtered_edge_iterator>
+  Flag_complex_edge_collapser(Filtered_edge_iterator begin, Filtered_edge_iterator end)
+  : f_edge_vector_(begin, end) { }
+
+  /** \brief Inserts all edges given by a OneSkeletonGraph into a vector of
+   * `Flag_complex_edge_collapser::Filtered_edge`.
+   * OneSkeletonGraph must be a model of
+   * <a href="http://www.boost.org/doc/libs/1_65_1/libs/graph/doc/EdgeListGraph.html">boost::EdgeListGraph</a>
```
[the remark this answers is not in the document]
VincentRouvreau: Done with 1.73.0. It is strange they have no `latest` documentation...

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,364 @@
...
+   * The edge filtration value is accessible through the property tag
+   * edge_filtration_t.
+   *
+   * boost::graph_traits<OneSkeletonGraph>::vertex_descriptor
+   *                                    must be Vertex_handle.
+   * boost::graph_traits<OneSkeletonGraph>::directed_category
+   *                                    can be directed_tag (the fastest, the least RAM use), undirected_tag or even
+   *                                    bidirected_tag.
+   *
+   * If an edge appears with multiplicity, the function will arbitrarily pick one representative to read the filtration
```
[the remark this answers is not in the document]
VincentRouvreau: No duplicate edges on 2610ce80

src/Collapse/utilities/point_cloud_edge_collapse_rips_persistence.cpp
```diff
@@ -78,24 +81,37 @@ int main(int argc, char* argv[]) {
...
+    edge_collapser.process_edges(
+      [&remaining_edges](Vertex_handle u, Vertex_handle v, Filtration_value filtration) {
+          // insert the edge
+          remaining_edges.emplace_back(u, v, filtration);
+        });
+    if ((iter + 1) < edge_collapse_iter_nb) {
+      edges_list.clear();
+      std::copy(remaining_edges.begin(), remaining_edges.end(), std::back_inserter(edges_list));
+      remaining_edges.clear();
+    }
```
mglisse, line 103:
```suggestion
    edges_list = std::move(remaining_edges);
    remaining_edges.clear();
```
No need for copying.
(with the current code, we don't even need remaining_edges, we could reuse edges_list directly, but let's wait until after the interface change, where it may not be possible anymore)

mglisse: Ah, but then you also need to update the following to use edges_list instead of remaining_edges, sorry.

VincentRouvreau: Yes, I had it :wink:

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,375 @@
...
+ *
+ * \tparam FilteredEdgeRange furnishes `std::begin` and `std::end` methods and returns an iterator on a
+ * FilteredEdge of type `std::tuple<Vertex_handle, Vertex_handle, Filtration_value>`
+ *
+ * \ingroup edge_collapse
+ * 
+ */
+template<class FilteredEdgeRange> auto flag_complex_collapse_edges(const FilteredEdgeRange& edges) {
+  auto first_edge_itr = std::begin(edges);
+  auto first_vertex = std::get<0>(*first_edge_itr);
```
mglisse, line 354: Trouble if the range is empty?
`using Vertex_handle = std::decay_t<decltype(std::get<0>(*first_edge_itr))>;`
has the advantage that it isn't actually evaluated, only the types are considered.

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,375 @@
...
+        filtered_edge_output(u, v, filt);
+        if (endIdx > 1)
+          set_edge_critical(endIdx, filt, filtered_edge_output);
+      }
+    }
+  }
+
+};
+
+/** \brief Constructs a Flag complex from edges as an input, collapse edges and returns remaining edges.
```
mglisse, line 342:
```suggestion
/** \brief Implicitly constructs a flag complex from edges as an input, collapses edges while preserving the persistent homology and returns the remaining edges as a range.
```

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,375 @@
...
+ *
+ * \fn auto Gudhi::collapse::flag_complex_collapse_edges(FilteredEdgeRange const& edges)
+ *
+ * \tparam FilteredEdgeRange furnishes `std::begin` and `std::end` methods and returns an iterator on a
+ * FilteredEdge of type `std::tuple<Vertex_handle, Vertex_handle, Filtration_value>`
```
mglisse, line 347:
```suggestion
 * FilteredEdge of type `std::tuple<Vertex_handle, Vertex_handle, Filtration_value>` where Vertex_handle is the index of a vertex.
```

mglisse, line 347: I think we said the doc should speak about the edges being unique. We may want to mention if the range needs to be already sorted. We have lost some information when moving from the class to the function.

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
VincentRouvreau: Thank you ! I could not find the right syntax with `std::decay`.
Done on 230bbe96

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
VincentRouvreau: I did it on d8c5a1b2

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
VincentRouvreau: I choosed the second signature

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
mglisse: Just in case, an alternative would be `using E = typename std::iterator_traits<decltype(first_edge_itr)>::value_type; using Vertex_handle = std::tuple_element_t<0, E>;`.

biblio/bibliography.bib
```diff
@@ -1279,3 +1279,15 @@ @inproceedings{dtmfiltrations
   bibsource = {dblp computer science bibliography, https://dblp.org}
 }
 
+@unpublished{edgecollapsesocg2020,
```
mglisse, line 1282: It is published: https://drops.dagstuhl.de/opus/volltexte/2020/12177/ provides a bibtex entry.

src/common/doc/main_page.md
```diff
@@ -217,6 +217,36 @@
...
+    only the 1-skeletons of a simplicial complex.
+    The reduction is exact and the persistence homology of the reduced sequence is identical to the persistence
+    homology of the input sequence. The resulting method is simple and extremely efficient.
+
+    Computation of edge collapse and persistent homology of a filtered flag complex via edge collapse as described in
+    \cite edgecollapsesocg2020.
+    </td>
+    <td width="15%">
+      <b>Author:</b> Siddharth Pritam<br>
+      <b>Introduced in:</b> GUDHI 2.4.0<br>
```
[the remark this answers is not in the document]
VincentRouvreau: Oops... 3.3.0

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -342,16 +342,14 @@ class Flag_complex_edge_collapser {
...
  * \param[in] edges Range of Filtered edges.There is no need the range to be sorted, as it will be performed.
  *
  * \tparam FilteredEdgeRange furnishes `std::begin` and `std::end` methods and returns an iterator on a
  * FilteredEdge of type `std::tuple<Vertex_handle, Vertex_handle, Filtration_value>` where `Vertex_handle` is the type
  * of a vertex index and `Filtration_value` is the type of an edge filtration value.
  *
- * \return Remaining edges after collapse of type
- * `std::vector<std::tuple<Vertex_handle, Vertex_handle, Filtration_value>>`.
+ * \return Remaining edges after collapse as a range of
+ * `<std::tuple<Vertex_handle, Vertex_handle, Filtration_value>`.
```
mglisse, line 352: No need for `<>` then.

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,380 @@
...
+      }
+    }
+  }
+
+};
+
+/** \brief Implicitly constructs a flag complex from edges as an input, collapses edges while preserving the persistent
+ * homology and returns the remaining edges as a range.
+ *
+ * \fn auto Gudhi::collapse::flag_complex_collapse_edges(FilteredEdgeRange const& edges)
```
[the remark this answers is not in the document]
VincentRouvreau: I added this when I tried to fix links to compute_proximity_graph free function, and then copied / pasted it. I removed it also on compute_proximity_graph as it did not work.

biblio/bibliography.bib
VincentRouvreau: I changed it on f7876ea08e810c57f90e0233fffbd91d57f6d037

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
```diff
@@ -0,0 +1,380 @@
...
+ * 
+ * \param[in] edges Range of Filtered edges.There is no need the range to be sorted, as it will be performed.
+ *
+ * \tparam FilteredEdgeRange furnishes `std::begin` and `std::end` methods and returns an iterator on a
+ * FilteredEdge of type `std::tuple<Vertex_handle, Vertex_handle, Filtration_value>` where `Vertex_handle` is the type
+ * of a vertex index and `Filtration_value` is the type of an edge filtration value.
+ *
+ * \return Remaining edges after collapse of type
+ * `std::vector<std::tuple<Vertex_handle, Vertex_handle, Filtration_value>>`.
```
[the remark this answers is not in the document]
VincentRouvreau: Done on f7876ea08e810c57f90e0233fffbd91d57f6d037

src/common/doc/main_page.md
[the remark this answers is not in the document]
VincentRouvreau: Done on f7876ea08e810c57f90e0233fffbd91d57f6d037

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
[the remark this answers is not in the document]
VincentRouvreau: Done on f7876ea and 02c5725

src/Collapse/include/gudhi/Flag_complex_edge_collapser.h
VincentRouvreau: Bad fix... 
eedb34f

PR #284, opened 2020-04-17
Simplify distance-to-diagonal in Wasserstein

mglisse:
This doesn't change much, but it bothered me that we were computing y-(x+y)/2 and x-(x+y)/2 when that's just ±(y-x)/2...
If we don't want to assume that the points are above the diagonal, I can add np.abs.

By the way, I was thinking of renaming wasserstein.py to distance.py...

### DIFF
--- a/src/python/gudhi/wasserstein/wasserstein.py
+++ b/src/python/gudhi/wasserstein/wasserstein.py
@@ -15,6 +15,17 @@ try:
 except ImportError:
     print("POT (Python Optimal Transport) package is not installed. Try to run $ conda install -c conda-forge pot ; or $ pip install POT")
 
+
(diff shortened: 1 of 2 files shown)


tlacombe:
I understand your point. I was using `_proj_on_diag` in `wasserstein.barycenter` at some point but finally handle this differently. So `_proj_on_diag` can be safely removed. However the function `_proj_on_diag` might be useful at some point (eg if we write a `wasserstein.plot_matching` function), is it necessary to remove it if note used ? Or can we keep it (but still use your `_dist_to_diag` as I agree, this is better) ?

mglisse:
> Or can we keep it

I think we can keep it, but then we need to add one small test for it, so we notice if it ever breaks.

PR #286, opened 2020-04-20
Tweak CircleCI parallelism

mglisse:
From my understanding, "parallelism: 4" is not used if we just run ctest -j4, it requires using circleci's specific way of running tests. And by default we get an instance with 2 vCPUs so -j2 should be enough. I am hoping that less parallelism will reduce the random infanticide in the sphinx test. On one test, it passed, and was as fast as before. But since it is random...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,7 +45,6 @@ jobs:
   python:
     docker:
       - image: gudhi/ci_for_gudhi:latest
-    parallelism: 4
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


mglisse:
> Tests were not launched on CircleCI

They were, but as usual you have to go to https://circleci.com/gh/mglisse to see them for my branches...

mglisse:
Ok, I am merging, it will be easy to revert if that doesn't help.

PR #287, opened 2020-04-20
Make sphinx test as first

### LINKED ISSUE #279: Long sphinx test
Among the python tests, sphinx_py_test is by far the longest. Sadly, it is also the last one to run. For those like me who run the tests in parallel, it would be much faster if this test could be the first one. I expect it is simply a matter of moving things in CMakeLists.txt and the tests are just run in the order they appear in that file.

VincentRouvreau:
Fix #279 
Make sphinx test as first as it is by far the longest test.
This is quite nice when running in parallel.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -242,6 +242,71 @@ if(PYTHONINTERP_FOUND)
 
     install(CODE "execute_process(COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/setup.py install)")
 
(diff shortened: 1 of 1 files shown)


VincentRouvreau:
> Je suppose que c'est un pur couper-coller.

yes. I am just experiencing circle CI `Child killed` on nerve_gic python tests. Maybe link with what you fixed on circleCI.

mglisse:
> Maybe link with what you fixed on circleCI.

I am still not sure I really fixed anything, we'll have to see in the next few days...

PR #289, opened 2020-04-22
Document several optional dependencies of knn

### LINKED ISSUE #288: Document dependencies introduced with knn (torch, eagerpy, etc)
I completely forgot to document the new dependencies when I added knn/dtm. I'll get to it.

mglisse:
Fix #288.


mglisse:
This is just doc, I am going to push it right away, but feel free to comment even if it is already merged.

PR #292, opened 2020-04-24
Fix bibliography for sphinx

VincentRouvreau:
* I had to remove cgal citations (but kept the links). I cannot deal with 2 biblio
* I added a bibliography page with all the biblio. No more duplicated. This file has to be named zbibliography.rst as it must be parsed last (cf. [this sphinxcontrib-bibtex issue](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/8) ).

### DIFF
--- a/src/python/doc/zbibliography.rst
+++ b/src/python/doc/zbibliography.rst
@@ -1,0 +1,10 @@
+:orphan:
+
+.. To get rid of WARNING: document isn't included in any toctree
+
+Bibliography
+------------
+
+.. bibliography:: ../../biblio/bibliography.bib
+    :style: unsrt
+
--- a/src/python/doc/zbibliography.rst
+++ b/src/python/doc/zbibliography.rst
@@ -6,5 +6,5 @@ Bibliography
 ------------
 
 .. bibliography:: ../../biblio/bibliography.bib
-    :style: unsrt
+    :style: plain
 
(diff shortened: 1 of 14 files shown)


mglisse:
> * I had to remove cgal citations (but kept the links). I cannot deal with 2 biblio

I guess you could always concatenate the 2 bib files (either as part of making user_version or during the build) to create the one bib file used by sphinx, if necessary.

> * I added a bibliography page with all the biblio. No more duplicated. This file has to be named zbibliography.rst as it must be parsed last (cf. [this sphinxcontrib-bibtex issue](https://github.com/mcmtroffaes/sphinxcontrib-bibtex/issues/8) ).

Yes, that's probably the easiest solution.

I notice that a number of bibliography entries don't have a URL (or a DOI, which automatically provides a URL), we may want to add some later, but that's an orthogonal point, not to be mixed in this PR.

src/python/doc/zbibliography.rst
```diff
@@ -0,0 +1,10 @@
+:orphan:
+
+.. To get rid of WARNING: document isn't included in any toctree
+
+Bibliography
+------------
+
+.. bibliography:: ../../biblio/bibliography.bib
+    :style: unsrt
```
mglisse, line 9: Does `plain` work? Sorting the biblio looks a bit more natural.
`alpha` sure looks ugly...

VincentRouvreau: I used the `plain` but I did not really understand the difference with `unsrt`. Tell if it fits your will.

mglisse: It sorts the bibliography by alphabetical order of the first author, which is the usual way in this community.

PR #304, opened 2020-05-04
Add some nogil for cython

mglisse:
`nogil` makes it possible to use multithreading in python. However, using it with cython is not super convenient. For instance here, in the construction of AlphaComplex, it causes the points to be copied one more time.
```cython
cdef vector[vector[double]] pts
pts = points
```
where points is a python type is transformed by cython into
```c++
std::vector<std::vector<double>> pts, tmp;
tmp = convert(points);
pts = tmp;
```
while if I don't name pts and pass points directly to a function that takes a vector, there is no extra copy. But the conversion cannot happen in a nogil region, so I have to choose between nogil or fewer copies...
The good thing about nogil in cython is that cython tells you whenever you are doing something it thinks is unsafe. The bad thing is that there is not much that is safe. From what I have seen, we have more control with pybind11.


mglisse:
The extra copy is fixed in cython 3.0 (no idea when that may be released), it now uses std::move, so we could ignore it. Longer term, using vector in the interface is probably not the best idea.

VincentRouvreau:
I imagine you parsed the most used ones ? And we could continue the list with tangential, witness, cover, ...

mglisse:
I added `nogil` to pretty much all the C++ functions in the .pyx files I touched, since it was obvious they didn't use any python commands, but it has no effect by itself. Then I added `with nogil:` to a few functions that seemed costly, where it didn't look too hard to add, and which people might be tempted to run in parallel. Yes, more are certainly possible, even in some files I already modified.

PR #306, opened 2020-05-06
Bottleneck cython interface

mglisse:
This comes on top of #304 and rewrites the cython interface of bottleneck to avoid copying vectors. We could avoid a copy in even more cases, accepting any numpy array and passing strides to C++, but then we might as well share the code with hera's interface.


mglisse:
I prefer #308, but I'll leave this one open as draft in case someone reviewing #308 wants to look at it.

mglisse:
#308 was merged, this one isn't needed anymore (#304 still is though).

PR #307, opened 2020-05-07
Improve dependencies documentation

### LINKED ISSUE #299: Missing dependencies in _sum.inc files
(from #285)
Most dependencies are documented in a Requires: field in the _sum.inc files. Some are missing, in particular "point cloud" can use CGAL+Eigen, numpy, scipy, sklearn, hnswlib, pykeops+torch, eagerpy (maybe I forgot one).

- There are 2 types of dependencies: some are needed when building the python module (CGAL, Eigen, pybind11), while others are needed at runtime (sklearn, etc). This difference seems quite relevant for users.
- For some packages, the dependencies are very different from one function to another (or even depending on the parameters passed to a function). We don't want to give users the impression that they need to install 10 packages for a simple functionality, but we do want them to be aware that some functions have more requirements.
- Long lists of dependencies are going to take quite a bit of space in the index, especially if we give details about them.

VincentRouvreau:
Fix #299 
Some doc improvement to link to a documentation section thanks to `sphinx.ext.autosectionlabel` in sphinx conf.py file


src/python/doc/installation.rst
```diff
@@ -174,9 +175,8 @@ Documentation
 
 To build the documentation, `sphinx-doc <http://www.sphinx-doc.org>`_ and
 `sphinxcontrib-bibtex <https://sphinxcontrib-bibtex.readthedocs.io>`_ are
-required. As the documentation is auto-tested, `CGAL`_, `Eigen`_,
-`Matplotlib`_, `NumPy`_, `POT`_, `Scikit-learn`_ and `SciPy`_ are
-also mandatory to build the documentation.
+required. As the documentation is auto-tested, all the :ref:`Optional third-party libraries`
+are also mandatory to build the documentation.
```
mglisse, line 179: Er, no, not all. You don't need torch/eagerpy/hnswlib/pykeops to build the doc.

src/python/gudhi/point_cloud/dtm.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, k, q=2, **kwargs):
         Args:
             k (int): number of neighbors (possibly including the point itself).
             q (float): order used to compute the distance to measure. Defaults to 2.
-            kwargs: same parameters as :class:`~gudhi.point_cloud.knn.KNearestNeighbors`, except that
+            kwargs: same parameters as :ref:`K nearest neighbors`, except that
```
mglisse, line 27: I think the link to the class was more precise (could even link to `__init__`), it is really "same arguments as this function". We could have more than just this one class in the section, and then the link wouldn't be so clear.

src/python/doc/installation.rst
```diff
@@ -12,16 +12,16 @@ The easiest way to install the Python version of GUDHI is using
...
 the GUDHI Python module.
 It is a multi-platform library and compiles on Linux, Mac OSX and Visual
 Studio 2017.
 
 On `Windows <https://wiki.python.org/moin/WindowsCompilers>`_ , only Python
-≥ 3.5 are available because of the required Visual Studio version.
+:math:`\geq` 3.5 are available because of the required Visual Studio version.
 
 On other systems, if you have several Python/python installed, the version 2.X
 will be used by default, but you can force it by adding
```
mglisse, line 27: Maybe we should specify there that we don't actually support 2.X anymore? (even if because of cmake weirdness we default to it on some platforms...)

src/python/doc/installation.rst
```diff
@@ -30,7 +30,8 @@ will be used by default, but you can force it by adding
 GUDHI Python module compilation
 ===============================
 
-To build the GUDHI Python module, run the following commands in a terminal:
+After making sure that the :ref:`Compilation dependencies` are properly installed,
+one can build the GUDHI Python module, by running the following commands in a terminal:
```
mglisse, line 34: (it isn't related to this PR, but while I am reading this text... Feel free to note that somewhere and ignore it for this PR)
Should we have a not explaining that if we build directly from a git checkout, it is `cd src/python`? Do we document user_version anywhere?

src/python/doc/installation.rst
```diff
@@ -185,11 +185,17 @@ Run the following commands in a terminal:
...
-Optional third-party library
-****************************
+Optional third-party libraries
+******************************
+
+Compilation dependencies
+========================
+
+These third party dependencies are detected by `CMake <https://www.cmake.org/>`_.
+They have to be installed before to perform the :ref:`GUDHI Python module compilation`.
```
mglisse, line 195:
```suggestion
They have to be installed before performing the :ref:`GUDHI Python module compilation`.
```
I think

src/python/doc/installation.rst
```diff
@@ -30,7 +30,8 @@ will be used by default, but you can force it by adding
 GUDHI Python module compilation
 ===============================
 
-To build the GUDHI Python module, run the following commands in a terminal:
+After making sure that the :ref:`Compilation dependencies` are properly installed,
```
mglisse, line 33: Just for curiosity, I think `` `Compilation dependencies`_`` would have worked without the new plugin.

src/python/gudhi/point_cloud/dtm.py
VincentRouvreau: I did it this, because when you click on the link, you will see the `requires` field.

src/python/doc/installation.rst
VincentRouvreau: Ok, but I will have to improve the documentation and maybe making cmake use python3 by default

src/python/doc/installation.rst
VincentRouvreau: It is true we do not talk about user and developer versions anywhere... Neither user_version

src/python/doc/installation.rst
VincentRouvreau: Yes I think so.

VincentRouvreau:
> Do we get warnings if different files use a heading with the same name? I am just a bit worried that some new package comes in, adds a subsection titled CGAL or "Persistence diagram" to its page, and all of a sudden some links everywhere may now point here instead of the installation page.
> Mostly looks good otherwise.

You are right. I will check if it produces warnings or if it fails miserably.

src/python/doc/installation.rst
VincentRouvreau: I opened #310

VincentRouvreau:
> Do we get warnings if different files use a heading with the same name? I am just a bit worried that some new package comes in, adds a subsection titled CGAL or "Persistence diagram" to its page, and all of a sudden some links everywhere may now point here instead of the installation page.
> Mostly looks good otherwise.

We do have warnings. It loooks like this:
```
python/doc/installation.rst:198: WARNING: duplicate label cgal, other instance in python/doc/alpha_complex_user.rst
python/doc/fileformats.rst:38: WARNING: duplicate label persistence diagram, other instance in python/doc/alpha_complex_user.rst
```
hence the importance of correcting all warnings.

src/python/doc/installation.rst
VincentRouvreau: Sorry about that. I thought the list had grown but didn't check. I rolled back on 260570c

src/python/gudhi/point_cloud/dtm.py
mglisse: Hmm, I think the priority is for the user to see what the arguments are, not the requires thing (and the argument description kind of mentions the dependencies already), but if you prefer your version, ok.

src/python/doc/installation.rst
mglisse: I am just trying to judge the advantages of the plugin versus its drawbacks.

mglisse:
It produces warnings, so at least we will know about it. On the other hand, it means that section titles need to be unique across the whole project? Good thing we don't have "Bibliography" in each page anymore. But it also means we cannot have generic section names like "Example", "Description", "Algorithm", etc because most likely several pages will have them?

src/python/gudhi/point_cloud/dtm.py
VincentRouvreau: Ok, I rolled it back

src/python/doc/installation.rst
VincentRouvreau: Without the plugin, we have to do for instance:
```
:ref:`Perseus file format`
# and at the point of the documentation we want to reach
.. _Perseus file format:
```

We can also do:
```
`CGAL <installation.html#cgal>`__
```

With the plugin we have a behaviour much more like doxygen, where you have warnings on duplicate sections names. Tell me if you really dislike it I can remove it. It was maybe a bad choice.

mglisse: > Without the plugin, we have to do for instance:

Except when the reference is to a section in the same file, where section names are already automatic references without the plugin. That was the point of my comment.

> With the plugin we have a behaviour much more like doxygen, where you have warnings on duplicate sections names.

Oh, I didn't know we had that for doxygen. And it hasn't caused trouble yet? Hmm, maybe that's a sign that it isn't that bad...

> Tell me if you really dislike it I can remove it. It was maybe a bad choice.

I don't have a strong opinion. I am slightly worried, but if we have the same thing with doxygen, my worries may be unfounded. If you like the plugin, we can use it.

VincentRouvreau: The difference is that you can name section with doxygen, for instance:
```
\section mymoduleexample Example
```
You cannot do this with sphinx, you have to use anchors.

VincentRouvreau:
We will keep #311 and not use `sphinx.ext.autosectionlabel`

PR #311, opened 2020-05-11
Improve dependencies doc

### LINKED ISSUE #299: Missing dependencies in _sum.inc files

VincentRouvreau:
Fix #299
Also fixes some link in between the documentation.
Contrary to #307 this fix do not use `sphinx.ext.autosectionlabel`


src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -72,11 +72,11 @@ def plot_persistence_barcode(
     """This function plots the persistence bar code from persistence values list
     , a np.array of shape (N x 2) (representing a diagram 
     in a single homology dimension), 
-    or from a :doc:`persistence file <fileformats>`.
+    or from a `persistence diagram <fileformats.html#persistence-diagram>`__ file.
```
mglisse, line 75: What does the double __ do? Do you need it because something strips one of them before is reaches the RST processor?

mglisse: Also, what is the advantage of this version compared to the one with `:doc:`, is it because it is shorter? Does it allow the `#label` thing while doc doesn't? Was the doc link broken? For uniformity? I don't care which one we use, just trying to understand the choices.

VincentRouvreau: Here, I wanted `persistence file` to link to the `Persistence diagram` section of  `fileformats` page.
If I say:
```
:doc:`persistence file <fileformats#persistence-diagram>`
or
:doc:fileformats#persistence-diagram
```
I have the following warning:
```
 WARNING: unknown document: fileformats#persistence-diagram
```

mglisse: Ok, good reason.

src/python/doc/alpha_complex_user.rst
```diff
@@ -11,8 +11,8 @@ Definition
 
 `AlphaComplex` is constructing a :doc:`SimplexTree <simplex_tree_ref>` using
 `Delaunay Triangulation  <http://doc.cgal.org/latest/Triangulation/index.html#Chapter_Triangulations>`_
-:cite:`cgal:hdj-t-19b` from `CGAL <http://www.cgal.org/>`_ (the Computational Geometry Algorithms Library
-:cite:`cgal:eb-19b`).
+:cite:`cgal:hdj-t-19b` from the `Computational Geometry Algorithms Library <http://www.cgal.org/>`_
```
mglisse, line 14: What was it a duplicate of? I thought this was the branch without the new plugin, so I am confused...

VincentRouvreau: sphinx was complaining:
```
alpha_complex_user.rst:6: WARNING: Duplicate explicit target name: "cgal".
```
because `'CGAL <http://www.cgal.org/>'_` is also creating a cgal target.

mglisse: Yes, but where is the **other** one? Is it the one in installation.rst? Ah, I guess this is the same problem as with the bibliography, everything is global... I am starting to understand, slowly.

VincentRouvreau: Yes that's it. Everything is global, but you can't reach the link...
At least, when you use `sphinx.ext.autosectionlabel` it may be clearer that it is global, no ?

mglisse: But the target in installation.rst is also towards www.cgal.org (assuming it uses the explicit `` `CGAL <https://www.cgal.org/>`_`` and not the section title), so here we could just write `CGAL_` and it would find the right link without duplicating the target?

mglisse: Ah, no, the other one is from the .inc file, that makes more sense. And that one links to the installation page. Well, it wouldn't be so bad if this one also linked to installation.html#cgal, but ok, things make more sense now.

src/python/doc/alpha_complex_user.rst
```diff
@@ -11,8 +11,8 @@ Definition
...
+(CGAL Library :cite:`cgal:eb-19b`).
```
mglisse, line 15: I think you can even remove "CGAL Library" (it is a bit redundant), and the parentheses.

src/python/doc/installation.rst
```diff
@@ -200,7 +207,7 @@ The procedure to install this library
 according to your operating system is detailed
 `here <http://doc.cgal.org/latest/Manual/installation.html>`_.
 
-The following examples requires CGAL version ≥ 4.11.0:
+The following examples requires CGAL version :math:`\geq` 4.11.0:
```
mglisse, line 210:
```suggestion
The following examples require CGAL version :math:`\geq` 4.11.0:
```

VincentRouvreau:
> Thanks for handling this. Do you still want me to look at #307?

No it's ok, I will close it.

PR #315, opened 2020-05-14
Try to install from conda

VincentRouvreau:


azure-pipelines.yml
```diff
@@ -9,7 +9,6 @@ jobs:
     variables:
       pythonVersion: '3.6'
       cmakeBuildType: Release
-      customInstallation: 'brew update && brew install graphviz doxygen boost eigen gmp mpfr tbb cgal'
```
mglisse, line 12: Does graphviz come as a dependency of something else, or is it actually unnecessary? (I still see it in the dockerfiles)

azure-pipelines.yml
```diff
@@ -21,9 +20,10 @@ jobs:
     - bash: |
         source activate gudhi_build_env
         sudo conda install --yes --quiet --name gudhi_build_env python=$(pythonVersion)
+        sudo conda update --yes --quiet -n base -c defaults conda
+        sudo conda install --yes -c conda-forge doxygen cgal-cpp=5.0.1 eigen
```
mglisse, line 24: Would replacing `=` with `>=` work? It seems less likely to force us to update it soon.

VincentRouvreau: Yes you are right it will be less restrictive

PR #324, opened 2020-05-21
DTM Rips complex

yuichi-ike:
A class to build a DTM Rips complex, which is (for now) a wrapper of WeightedRipsComplex. I'd like to start with this straightforward implementation.
To be discussed:
- Full/triangular distance matrix: now it is assumed to be a full matrix to compute DistanceToMeasure with `metric="precomputed"`
- Documentation
- More efficient implementation (cf. the discussion in https://github.com/GUDHI/gudhi-devel/pull/295)


src/python/gudhi/dtm_rips_complex.py
```diff
@@ -0,0 +1,46 @@
...
+    def __init__(self, 
+                 points=None, 
+                 distance_matrix=None, 
+                 k=1, 
+                 q=2,
+                 max_filtration=float('inf')):
+        """
+        Args:
+            points (Sequence[Sequence[float]]): list of points.
+            distance_matrix (ndarray): full distance matrix.
```
mglisse, line 31: We *could* use a name other than `distance_matrix` (`full_distance_matrix`? `distances`?) to reflect the fact that it has stricter expectations than `distance_matrix` in `RipsComplex`, but I don't think that's necessary, `distance_matrix` looks good enough to me.
(just posting this in case someone disagrees)

src/python/gudhi/dtm_rips_complex.py
```diff
@@ -0,0 +1,46 @@
...
+#
+# Modification(s):
+#   - YYYY/MM Author: Description of the modification
+
+
+from gudhi.weighted_rips_complex import WeightedRipsComplex
+from gudhi.point_cloud.dtm import DistanceToMeasure
+from scipy.spatial.distance import cdist
+
+class DtmRipsComplex(WeightedRipsComplex):
```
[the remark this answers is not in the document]
yuichi-ike: I have no strong opinion on this point. Then I will change it to `DTMRipsComplex`.

src/python/gudhi/dtm_rips_complex.py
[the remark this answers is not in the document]
yuichi-ike: Yes, I think so too. But I don't have a strong opinion either.

yuichi-ike: `distance_matrix` looks good enough also to me.

src/python/test/test_dtm_rips_complex.py
```diff
@@ -0,0 +1,40 @@
...
+
+def test_dtm_rips_complex():
+    pts = np.array([[2.0, 2.0], [0.0, 1.0], [3.0, 4.0]])
+    dtm_rips = DTMRipsComplex(points=pts, k=2)
+    st = dtm_rips.create_simplex_tree(max_dimension=2)
+    st.persistence()
+    persistence_intervals0 = st.persistence_intervals_in_dimension(0)
+    assert persistence_intervals0 == pytest.approx(np.array([[3.16227766, 5.39834564],[3.16227766, 5.39834564], [3.16227766, float("inf")]]))
+    
+def test_compatibility_with_rips():
```
VincentRouvreau, line 24: Here you do not really test the compatibility.
```python
from gudhi import RipsComplex
rips_complex = RipsComplex(distance_matrix=distance_matrix, max_edge_length=42)
st_from_rips = rips_complex.create_simplex_tree(max_dimension=1)
assert list(st.get_filtration()) == list(st_from_rips.get_filtration())
```
This test would fit better the name of the test.

src/python/doc/rips_complex_user.rst
```diff
@@ -378,6 +378,7 @@ Example from a point cloud combined with DistanceToMeasure
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Combining with DistanceToMeasure, one can compute the DTM-filtration of a point set, as in `this notebook <https://github.com/GUDHI/TDA-tutorial/blob/master/Tuto-GUDHI-DTM-filtrations.ipynb>`_. 
+Remark that DTMRipsComplex class provides exactly this function.
```
VincentRouvreau, line 381:
```suggestion
Remark that `DTMRipsComplex <rips_complex_user.html#dtm-rips-complex>`_ class provides exactly this function.
```
Maybe a link would be appreciated.

src/python/gudhi/dtm_rips_complex.py
```diff
@@ -0,0 +1,50 @@
...
+                # Empty Rips construction
+                points=[]
+            distance_matrix = cdist(points,points)
+        self.distance_matrix = distance_matrix
+
+        # TODO: address the error when k is too large 
+        if k <= 1:
+            self.weights = [0] * len(distance_matrix)
+        else:
+            dtm = DistanceToMeasure(k, q=q, metric="precomputed")
```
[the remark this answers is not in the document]
mglisse: Yes, it uses scipy by default. I could have made it use brute force when scipy is not present, but that was getting tiring. I wonder if we should make scipy a hard requirement for the whole gudhi python (like numpy)...

src/python/test/test_dtm_rips_complex.py
yuichi-ike: Exactly as you say. I will fix it.

src/python/gudhi/dtm_rips_complex.py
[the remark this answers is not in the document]
mglisse: Let's require a numpy.ndarray for points as well. Then I think the branch is ok.

PR #327, opened 2020-05-26
Cmake user version copy cpp cython generated

### LINKED ISSUE #206: [cython] cython generates .cpp files in the source tree instead of the build tree
Cython generates .cpp files from .pyx files, and writes them in the source tree instead of the build tree. Ideally, it would write them in the build tree instead.
The reason I noticed is that the following scenario happened several times to me: I build the python plugin directly from the git sources, then I modify the sources (possibly switch branch), create a user_version, build the python plugin from that user_version, and get confused because my modifications were not taken into account. Indeed, the user_version includes the stale .cpp that were lying around in src/python/gudhi, probably with a modification time just as recent as the .pyx so the build doesn't feel the need to regenerate them.
It looks like I should be more consistent, either always build a user_version, or never. Still, it may not be too hard to make the user_version generation explicitly exclude src/python/gudhi/*.cpp from the list of files to copy.

VincentRouvreau:
Work around #206


VincentRouvreau:
> Ok, but please leave the issue open. It may be hard and we may ignore it for now, but a build should not touch the source directory.

You are completely right, this is definitely a work around and not a fix.

PR #330, opened 2020-05-28
Entropic regularization for wasserstein distances

tlacombe:
Added an option to use `ot.sinkhorn` instead of `ot.emd` when computing the Wasserstein distance between persistence diagrams. 

Tried to manage as well as possible the numerical issues that could be induced by used a too small value for `reg`.

### DIFF
--- a/src/python/doc/wasserstein_distance_user.rst
+++ b/src/python/doc/wasserstein_distance_user.rst
@@ -153,6 +153,28 @@ Morozov, and Arnur Nigmetov.
 
 .. autofunction:: gudhi.hera.wasserstein_distance
 
+This example computes the 1-Wasserstein distance with infinity-norm as ground metric
+between 2 persistence diagrams with Euclidean ground metric.
+Note that persistence diagrams must be submitted as (n x 2) numpy arrays and can contain inf values.
+
+.. testcode::
+
+    import gudhi.hera
+    import numpy as np
+
+    dgm1 = np.array([[2.7, 3.7],[9.6, 14.],[34.2, 34.974], [1, np.inf]])
+    dgm2 = np.array([[2.8, 4.45],[9.5, 14.1], [2, np.inf]])
+
+    message = "Wasserstein distance value (hera) = " + '%.2f' % gudhi.hera.wasserstein_distance(dgm1, dgm2, order=1., internal_p=2.)
+    print(message)
+
+The output is:
+
+.. testoutput::
+
+    Wasserstein distance value (hera) = 2.45
+
+
 
 Barycenters
 -----------
--- a/src/python/doc/wasserstein_distance_user.rst
+++ b/src/python/doc/wasserstein_distance_user.rst
@@ -100,6 +100,57 @@ The output is:
     point 2 in dgm1 is matched to the diagonal
     point 1 in dgm2 is matched to the diagonal
 
+
+Entropic regularization
+***********************
+
+When using the Optimal Transport based version of wasserstein distances, 
+we also allow for the use of entropic regularization of optimal transport distance via the parameter ``reg``.
+If ``reg > 0``, we solve 
+:math:`\argmin_P \sum P_{ij} M_{ij} + reg * h(P)`
+under marginal constraints on :math:`P` , which can be understood as a generalized matching between the diagrams.
+:math:`P_{ij}` denotes the quantity of mass from the :math:`i`-th points in the first diagram that is
+transported to the :math:`j`-th point in the second diagram. Last row and column of :math:`P` denotes the diagonal
+by convention.
+Here :math:`M` is a matrix encoding the pairwises distances between the points of the two diagrams along with
+distances from points to the diagonal, and :math:`h(P) = \sum P_{ij} \log(P_{ij})` is the entropic regularization term.
+
+If ``matching=True``, we return both the estimated transport cost :math:`sum P_{ij} C_{ij}` and the computed :math:`P`, 
+otherwise we simply return the estimated transport cost.
+
+The closer ``reg`` is to 0, the closer the transport cost will be from the exact distance between the diagrams.
+The larger the regularization parameter, the faster the computation.
+Note however that taking too low values for ``reg`` can run into numerical issues.
+
+.. testcode::
+
+    import gudhi.wasserstein
+    import numpy as np
+
+    # regularization parameter
+    reg = 0.1
+
+    dgm1 = np.array([[2.7, 3.7], [4., 4.1]])
+    dgm2 = np.array([[2.8, 4.45]])
+    cost, P = gudhi.wasserstein.wasserstein_distance(dgm1, dgm2, matching=True, order=1, internal_p=2, reg=reg)
+
+    message_cost = "Regularized Wasserstein distance value = %.2f" %cost
+    print(message_cost)
+    print("Mass located in the first point of dgm1 transported to the first point of dgm2:")
+    print("%.3f" %P[0,0]) 
+    print("Mass located in the first point of dgm1 transported to the diagonal:")
+    print("%.3f" %P[0,1]) 
+
+
+.. testoutput::
+
+   Regularized Wasserstein distance value = 0.83
+   Mass located in the first point of dgm1 transported to the first point of dgm2:
+   0.995
+   Mass located in the first point of dgm1 transported to the diagonal:
+   0.004
+
+
 Barycenters
 -----------
 
--- a/src/python/doc/wasserstein_distance_user.rst
+++ b/src/python/doc/wasserstein_distance_user.rst
@@ -107,7 +107,7 @@ Entropic regularization
 When using the Optimal Transport based version of wasserstein distances, 
 we also allow for the use of entropic regularization of optimal transport distance via the parameter ``reg``.
 If ``reg > 0``, we solve 
-:math:`\argmin_P \sum P_{ij} M_{ij} + reg * h(P)`
+:math:`\min_P \sum P_{ij} M_{ij} + reg * h(P)`
 under marginal constraints on :math:`P` , which can be understood as a generalized matching between the diagrams.
 :math:`P_{ij}` denotes the quantity of mass from the :math:`i`-th points in the first diagram that is
 transported to the :math:`j`-th point in the second diagram. Last row and column of :math:`P` denotes the diagonal
@@ -115,7 +115,7 @@ by convention.
 Here :math:`M` is a matrix encoding the pairwises distances between the points of the two diagrams along with
 distances from points to the diagonal, and :math:`h(P) = \sum P_{ij} \log(P_{ij})` is the entropic regularization term.
 
-If ``matching=True``, we return both the estimated transport cost :math:`sum P_{ij} C_{ij}` and the computed :math:`P`, 
+If ``matching=True``, we return both the estimated transport cost :math:`\sum P_{ij} C_{ij}` and the computed :math:`P`, 
 otherwise we simply return the estimated transport cost.
 
 The closer ``reg`` is to 0, the closer the transport cost will be from the exact distance between the diagrams.
@@ -139,7 +139,7 @@ Note however that taking too low values for ``reg`` can run into numerical issue
     print("Mass located in the first point of dgm1 transported to the first point of dgm2:")
     print("%.3f" %P[0,0]) 
     print("Mass located in the first point of dgm1 transported to the diagonal:")
-    print("%.3f" %P[0,1]) 
+    print("%.3f" %P[0,-1]) 
 
 
 .. testoutput::
--- a/src/python/doc/wasserstein_distance_user.rst
+++ b/src/python/doc/wasserstein_distance_user.rst
@@ -148,7 +148,7 @@ Note however that taking too low values for ``reg`` can run into numerical issue
    Mass located in the first point of dgm1 transported to the first point of dgm2:
    0.995
    Mass located in the first point of dgm1 transported to the diagonal:
-   0.004
+   0.005
 
 
 Barycenters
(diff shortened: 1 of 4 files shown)


src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -99,6 +99,10 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
     :param order: exponent for Wasserstein; Default value is 2.
     :param internal_p: Ground metric on the (upper-half) plane (i.e. norm L^p in R^2);
                        Default value is 2 (Euclidean norm).
+    :param reg: entropic regularization paramter. If 0. (default), exact distance is computed. Beware that low
```
mglisse, line 102:
```suggestion
    :param reg: entropic regularization parameter. If 0. (default), exact distance is computed. Beware that low
```

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -99,6 +99,10 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
...
+                values of ``reg`` might lead to numerical instability.
+                Note that using ``reg>0`` is incompatible with ``enable_autodiff=True`` for now.
+                Using ``reg>0`` with ``matching=True`` returns a transport plan, that is a (n+1)x(m+1) matrix.
     :param enable_autodiff: If X and Y are torch.tensor, tensorflow.Tensor or jax.numpy.ndarray, make the computation
         transparent to automatic differentiation. This requires the package EagerPy and is currently incompatible
         with `matching=True`.
```
mglisse, line 108:
```suggestion
        with `matching=True` or ``reg>0``.
```

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -100,6 +100,57 @@ The output is:
...
     point 1 in dgm2 is matched to the diagonal
 
+
+Entropic regularization
+***********************
+
+When using the Optimal Transport based version of wasserstein distances, 
+we also allow for the use of entropic regularization of optimal transport distance via the parameter ``reg``.
+If ``reg > 0``, we solve 
+:math:`\argmin_P \sum P_{ij} M_{ij} + reg * h(P)`
```
mglisse, line 110: I am seeing the `\` before argmin in the sphinx doc.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -100,6 +100,57 @@ The output is:
...
+under marginal constraints on :math:`P` , which can be understood as a generalized matching between the diagrams.
+:math:`P_{ij}` denotes the quantity of mass from the :math:`i`-th points in the first diagram that is
+transported to the :math:`j`-th point in the second diagram. Last row and column of :math:`P` denotes the diagonal
+by convention.
+Here :math:`M` is a matrix encoding the pairwises distances between the points of the two diagrams along with
+distances from points to the diagonal, and :math:`h(P) = \sum P_{ij} \log(P_{ij})` is the entropic regularization term.
+
+If ``matching=True``, we return both the estimated transport cost :math:`sum P_{ij} C_{ij}` and the computed :math:`P`,
```
mglisse, line 118:
```suggestion
If ``matching=True``, we return both the estimated transport cost :math:`\sum P_{ij} C_{ij}` and the computed :math:`P`, 
```
?

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -100,6 +100,57 @@ The output is:
...
+    dgm1 = np.array([[2.7, 3.7], [4., 4.1]])
+    dgm2 = np.array([[2.8, 4.45]])
+    cost, P = gudhi.wasserstein.wasserstein_distance(dgm1, dgm2, matching=True, order=1, internal_p=2, reg=reg)
+
+    message_cost = "Regularized Wasserstein distance value = %.2f" %cost
+    print(message_cost)
+    print("Mass located in the first point of dgm1 transported to the first point of dgm2:")
+    print("%.3f" %P[0,0]) 
+    print("Mass located in the first point of dgm1 transported to the diagonal:")
+    print("%.3f" %P[0,1])
```
mglisse, line 142: I wonder if `P[0,-1]` would be clearer or more confusing.

src/python/doc/wasserstein_distance_user.rst
tlacombe: ah yes my bad, I am so used to defined `argmin` as a macro that I forgot it wasn't by default, will fix this.

src/python/doc/wasserstein_distance_user.rst
tlacombe: Probably, that's a nice idea !

mglisse:
ISTR that your earlier experiments with entropic regularization hadn't been very convincing. Did things change? Do you have some numbers on how much faster this is than without regularization, and how far the value computed is?
Is it clear that this is **the** way to define entropic regularization for diagrams, or are there other ways the diagonal could be handled?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -173,9 +173,40 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
         return ep.concatenate(dists).norms.lp(order).raw
         # We can also concatenate the 3 vectors to compute just one norm.
 
-    # Comptuation of the otcost using the ot.emd2 library.
+    # Comptuation of the otcost using the ot.emd2 library, and matching/transport plan using ot.emd.
```
mglisse, line 176:
```suggestion
    # Computation of the otcost using the ot.emd2 library, and matching/transport plan using ot.emd.
```
might as well fix the typo at the same time

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -173,9 +173,40 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
...
+            match = np.argwhere(P)
+            # Now we turn to -1 points encoding the diagonal
+            match[:,0][match[:,0] >= n] = -1
+            match[:,1][match[:,1] >= m] = -1
+            return ot_cost ** (1./order) , match
+        else:
+            ot_cost = ot.emd2(a, b, M, numItermax=2000000)
+
+    elif reg > 0:
+        np.seterr(divide='raise')  # For a better handling of numerical issues due to reg being too low
```
mglisse, line 193: Do you reset it to its previous setting afterwards? `numpy.errstate` looks like a convenient way to set it only for the duration of some block of code.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -173,9 +173,40 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
...
+        try:
+            if matching:
+                P = ot.sinkhorn(a=a, b=b, M=M, reg=reg)
```
mglisse, line 196: Should the documentation have the word sinkhorn somewhere in it, as a relevant keyword?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -173,9 +173,40 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
...
+                ot_cost = np.sum(np.multiply(P, M))
+                return ot_cost ** (1./order), P
+            else:
+                ot_cost = ot.sinkhorn2(a, b, M, reg=reg)[0]
+                return ot_cost ** (1. / order)
+        except FloatingPointError:
+            raise FloatingPointError('Error when computing regularized wasserstein distance (Sinkhorn loop).' \
+                                     'This is probably due to reg being too low. Try' \
```
mglisse, line 204: Do we need the `\`, seeing as this is happening inside parentheses?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -173,9 +173,40 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
...
+                                     'to increase reg or set it to 0 for exact computation')
+    else:
+        raise ValueError('Negative value for reg')
+        # It seems ot.sinkhorn2 returns something when reg<0, which is reasonnable but probably unexpected
```
mglisse, line 208: Interesting. If you put a very large negative reg, what should it converge to, should it stay at the same value as reg=0 or send every point to the diagonal?

tlacombe: If reg is super negative, it seems you obtain a similar behavior as reg being super positive, that is you obtain the "trivial transport plan", where each point is transported uniformly to all other points and the diagonal. 
The reg parameter only appears through the matrix K = exp( - M / reg), (term-wise exponentiation) where M is the pairwise cost matrix (including the diagonal), and then the only requirement you have about this matrix is to have positive entries (which is the case, even with reg < 0). Using reg < 0 is somewhat equivalent to use negative distances as your transport cost.

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Yes, that would be great indeed.

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Ok I will check that ! (indeed, I did not revert to standard settings afterwards, need to improve on this).

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -100,6 +100,57 @@ The output is:
...
+by convention.
+Here :math:`M` is a matrix encoding the pairwises distances between the points of the two diagrams along with
+distances from points to the diagonal, and :math:`h(P) = \sum P_{ij} \log(P_{ij})` is the entropic regularization term.
+
+If ``matching=True``, we return both the estimated transport cost :math:`\sum P_{ij} C_{ij}` and the computed :math:`P`, 
+otherwise we simply return the estimated transport cost.
+
+The closer ``reg`` is to 0, the closer the transport cost will be from the exact distance between the diagrams.
+The larger the regularization parameter, the faster the computation.
+Note however that taking too low values for ``reg`` can run into numerical issues.
```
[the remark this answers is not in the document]
tlacombe: What matters for numerical stability is the ratio `M / reg`, where M is the pairwise distance matrix between points (including the diagonal), and essentially because you compute at some point `exp(-M/reg)` (term-wise exponentiation), and you don't want it to be 0. So numerical issues arise when `max(M) / reg \sim 1000`. 

In this implementation, you indeed have a control which (theoretically) scales linearly with the number of points and with the entries of `M`. Roughly speaking, the more points you have and the larger the coordinate, the smaller you might want to pick `reg` to reach a fixed _additive_ error (it is not a surprise: the more points / the larger the coordinates, the larger the exact distance. Although I don't know how to prove it, I think the relative error is less impacted). Note also that these "linear additive errors" are worst-case bounds, the error made in practice is in general much smaller. 

I don't know a heuristic to pick a value for `reg`. From my intuition, this should just be considered as another notion of similarity between persistence diagrams, which might have various benefits (of interest, the optimal transport plan is more stable/differentiable). 
I'd say it is somewhat similar to using the `order=p` metric instead of the bottleneck distance to compare persistence diagrams. Although those are linked in the regime p-->infty, in practice you don't really care about this regime and use p=1 or p=2.

src/python/gudhi/wasserstein/wasserstein.py
mglisse: I see. Thanks.

src/python/doc/wasserstein_distance_user.rst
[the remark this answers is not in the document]
mglisse: Some discussion like that in the doc could be useful. At least knowing that reg is homogeneous to a distance.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -173,9 +173,40 @@ def wasserstein_distance(X, Y, matching=False, order=2., internal_p=2., enable_a
         return ep.concatenate(dists).norms.lp(order).raw
         # We can also concatenate the 3 vectors to compute just one norm.
 
-    # Comptuation of the otcost using the ot.emd2 library.
+    # Comptuation of the otcost using the ot.emd2 library, and matching/transport plan using ot.emd.
     # Note: it is the Wasserstein distance to the power q.
     # The default numItermax=100000 is not sufficient for some examples with 5000 points, what is a good value?
```
VincentRouvreau, line 178: This comment needs to be updated as default numItermax=2000000

src/python/gudhi/wasserstein/wasserstein.py
VincentRouvreau: Yes please remove the `\`, cf. [PEP8 E502](https://www.flake8rules.com/rules/E502.html)

VincentRouvreau:
I don't know if we should move `hera` after `Entropic regularization`, but the sections order is bizarre:
* Distance Functions
    * Optimal Transport
    * Hera
    * Basic example (which uses ot, but can be easily modified to use Hera)
    * Entropic regularization (that only applies to ot)
* Barycenters

src/python/gudhi/wasserstein/wasserstein.py
mglisse: IIRC, numItermax=100000 is the default in POT, we override it with a larger number, and this comment explains why.

mglisse:
`reg`: I wonder if the name might be too short, I need to guess that it is the coefficient of a (entropic) regularization term in the cost of a transport plan. Of course, `entropic_regularization_coefficient` would be a bit long :wink: . And people can just read the doc, so it may be fine.

What may the interface end up looking like if we want to give the possibility to use geomloss (also using entropic regularization) for the case p=1, or some other code specialized to use geometry without requiring a GPU? It is interesting how the webpage for geomloss says that POT is "for small problems". It also mentions code by Bernhard Schmitzer which seems interesting (there is a newer version called MultiScaleOT), but that looks really experimental.

tlacombe:
PR closed ; it seems there is room for improvement before including entropic regularization in diagram distances (issues of stability & cie). This must be investigated and a new PR will be done if deemed necessary.

PR #338, opened 2020-06-02
Parallelism in pairwise_distances

mglisse:
Implement some parallelism for pairwise distances. Don't rely on the one from sklearn for pairwise(X,X), since last time I checked it didn't take advantage of symmetry (and 0 diagonal for distances). Also, it is tuned for many cheap distances that all take the same time to compute, while we have few, expensive distances of uneven cost.

### DIFF
--- a/src/python/gudhi/representations/metrics.py
+++ b/src/python/gudhi/representations/metrics.py
@@ -12,7 +12,7 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.metrics import pairwise_distances
 from gudhi.hera import wasserstein_distance as hera_wasserstein_distance
 from .preprocessing import Padding
-from joblib import Parallel, delayed, effective_n_jobs
+from joblib import Parallel, delayed
 
 #############################################
 # Metrics ###################################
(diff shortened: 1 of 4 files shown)


src/python/gudhi/representations/metrics.py
```diff
@@ -12,6 +12,7 @@
 from sklearn.metrics import pairwise_distances
 from gudhi.hera import wasserstein_distance as hera_wasserstein_distance
 from .preprocessing import Padding
+from joblib import Parallel, delayed
```
VincentRouvreau, line 15: I wanted to ask to add joblib as a `:Requires:` in the documentation, but even if it is not clearly said in scikit-learn documentation, it seams joblib is installed when you install scikit-learn. So I am ok if it is not explicitly written in the documentation.

src/python/gudhi/representations/metrics.py
```diff
@@ -227,14 +245,16 @@ class BottleneckDistance(BaseEstimator, TransformerMixin):
...
     :Requires: `CGAL <installation.html#cgal>`_ :math:`\geq` 4.11.0
     """
-    def __init__(self, epsilon=None):
+    def __init__(self, epsilon=None, n_jobs=None):
         """
         Constructor for the BottleneckDistance class.
 
         Parameters:
             epsilon (double): absolute (additive) error tolerated on the distance (default is the smallest positive float), see :func:`gudhi.bottleneck_distance`.
+            n_jobs (int): number of jobs to use for the computation. See :func:`pairwise_persistence_diagram_distances` for details.
```
VincentRouvreau, line 254: I put this comment here but it applies to all `See :func:`pairwise_persistence_diagram_distances` for details`.
I think it would be better to add a new joblib subsection in installation.rst / Run time dependencies section and link to that new subsection.

src/python/gudhi/representations/metrics.py
mglisse: Exactly. I was going to say that in the PR, and then completely forgot :disappointed:

src/python/gudhi/representations/metrics.py
mglisse: I can add a section there, but here I am not trying to add a link to joblib in general. This is more to advertise the fact that I used `prefer="threads"` in this particular function, which is important for a user to understand how it interacts with their own use of parallelism. In other places in gudhi, we may use `prefer="processes"` (or no `prefer`), and we may know more about the functions called inside the parallel section.

mglisse: I added a paragraph about joblib in installation.rst.
(I meant to link to the section about scikit-learn below, but the link goes straight to the scikit-learn website, which I guess is fine)

PR #349, opened 2020-06-11
fix #342

### LINKED ISSUE #342: Default arguments for wasserstein_distance
We have 2 functions wasserstein_distance, one that defaults to order=1 and internal_p=inf, and the other that defaults to 2 and 2 (WassersteinDistance probably copied that one). That seems confusing.

tlacombe:
Propose to set the parameters of `gudhi.wasserstein.wasserstein_distance` to order=1, internal_p = np.inf to make it consistent with those of `gudhi.hera.wasserstein_distance`. 

I decided to go this way as it seems `internal_p = np.inf` is more common than `2` in papers, and I do not have any reason to prefer `order = 2` over `order=1`.


mglisse:
With this patch, do we still have different defaults for WassersteinDistance?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,17 +88,17 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
                 (i.e. with infinite coordinate).
     :param Y: (m x 2) numpy.array encoding the second diagram.
     :param matching: if True, computes and returns the optimal matching between X and Y, encoded as
                      a (n x 2) np.array  [...[i,j]...], meaning the i-th point in X is matched to
                      the j-th point in Y, with the convention (-1) represents the diagonal.
-    :param order: exponent for Wasserstein; Default value is 2.
+    :param order: exponent for Wasserstein; Default value is 1.
     :param internal_p: Ground metric on the (upper-half) plane (i.e. norm L^p in R^2);
-                       Default value is 2 (Euclidean norm).
+                       Default value is `np.inf`.
```
mglisse, line 101: I don't think people usually capitalize the word after `;`.

PR #357, opened 2020-06-17
No latex for matplotlib if not available

### LINKED ISSUE #351: plot_persistence_diagram depends on non-default latex packages
See https://github.com/GUDHI/TDA-tutorial/issues/16 for what happens when someone has a basic (instead of full) installation of latex. Not sure what, if anything, we should do about it, but it isn't a nice behavior.

VincentRouvreau:
Use `checkdep_usetex` from matplotlib.
It is not perfect as it does not check if `type1cm` LaTeX package is available.
Partially #351

### DIFF
--- a/src/python/gudhi/persistence_graphical_tools.py
+++ b/src/python/gudhi/persistence_graphical_tools.py
@@ -11,6 +11,7 @@
 from os import path
 from math import isfinite
 import numpy as np
+from functools import lru_cache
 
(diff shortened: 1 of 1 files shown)


mglisse:
Could we cache the result of checkdep_usetex, or is the check cheap enough that it doesn't matter?

VincentRouvreau:
> Could we cache the result of checkdep_usetex, or is the check cheap enough that it doesn't matter?

Good idea ! Done on 0f9eb77

Issue #29, opened 2019-03-07
[python - SimplexTree] boundary of a simplex


VincentRouvreau:
We have python interfaces for the inefficient access to coboundaries (get_cofaces), but we seem to be missing access to the more natural boundaries...

mglisse:
`itertools.combinations(s,len(s)-1)`

VincentRouvreau:
@mglisse now that #362 is merged, do we keep this issue to show how to get boundaries in pure python ?

mglisse:
If you want we can add to the documentation of the new function a sentence like
> If you do not need the filtration values, the boundary can also be obtained as `itertools.combinations(s,len(s)-1)`

but I don't think we need to keep this issue open.

PR #362, opened 2020-07-03
A fix proposal for boudaries of a simplex python version

VincentRouvreau:
A fix proposal for #29 
Maybe overkill because it could be done easily in pure Python, but the interface would not be the same as the other iterators.
```python
import itertools

    def get_boundaries(self, simplex):
        """This function returns a generator with the boundaries of a given N-simplex.

        :param simplex: The N-simplex, represented by a list of vertex.
        :type simplex: list of int.
        :returns:  The (simplices of the) boundaries of a simplex
        :rtype:  generator of simplices
        """
        return itertools.combinations(simplex,len(simplex)-1)
```
We wouldn't have `:rtype:  generator with tuples(simplex, filtration)` like the other generators.


mglisse:
> the interface would not be the same as the other iterators

Generators and lists are both iterable. The whole point of those concepts is that you don't have to use the same type all the time. I haven't looked at the PR yet.

src/python/include/Simplex_tree_interface.h
```diff
@@ -188,6 +189,16 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
...
   }
+
+  Boundary_simplex_iterator get_boundary_iterator_begin(const Simplex& simplex) {
+    // this specific case works because the range is just a pair of iterators - won't work if range was a vector
+    return Base::boundary_simplex_range(Base::find(simplex)).begin();
+  }
+
+  Boundary_simplex_iterator get_boundary_iterator_end(const Simplex& simplex) {
+    // this specific case works because the range is just a pair of iterators - won't work if range was a vector
+    return Base::boundary_simplex_range(Base::find(simplex)).end();
```
mglisse, line 200: Calling Base::find twice (for begin and end) is not nice.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -285,6 +285,22 @@ cdef class SimplexTree:
             ct.append((v, filtered_simplex.second))
         return ct
 
+    def get_boundaries(self, simplex):
+        """This function returns a generator with the boundaries of a given N-simplex.
+
+        :param simplex: The N-simplex, represented by a list of vertex.
+        :type simplex: list of int.
+        :returns:  The (simplices of the) boundaries of a simplex
```
mglisse, line 293:
```suggestion
        :returns:  The (simplices of the) boundary of a simplex
```

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_iterators.h
```diff
@@ -85,6 +85,11 @@ class Simplex_tree_boundary_simplex_iterator : public boost::iterator_facade<
   typedef typename SimplexTree::Vertex_handle Vertex_handle;
   typedef typename SimplexTree::Siblings Siblings;
 
+  Simplex_tree_boundary_simplex_iterator()
```
mglisse, line 88: I assume cython is forcing you to do that? I'd like to see a comment next to it saying that it is added just for cython, and that the object it initializes should be overwritten ASAP and never used before it is overwritten.

src/python/test/test_simplex_tree.py
```diff
@@ -340,3 +340,12 @@ def test_simplices_iterator():
...
         assert st.filtration(simplex[0]) == simplex[1]
+
+def test_boundaries_iterator():
+    st = SimplexTree()
+
+    assert st.insert([0, 1, 2, 3], filtration=1.0) == True
+    assert st.insert([1, 2, 3, 4], filtration=2.0) == True
+
+    assert list(st.get_boundaries([1, 2, 3])) == [([1, 2], 1.0), ([1, 3], 1.0), ([2, 3], 1.0)]
+    assert list(st.get_boundaries([2, 3, 4])) == [([2, 3], 1.0), ([2, 4], 2.0), ([3, 4], 2.0)]
```
mglisse, line 351: Maybe check also the boundary of a vertex (should be an empty list).
What happens if I do `st.get_boundaries([])`? (I don't think it is very important, I am just curious)

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_iterators.h
VincentRouvreau: Comment added on 1f31e5ce as cython is trying to do:
```cpp
new((void*)&(p->__pyx_v_it)) Gudhi::Simplex_tree_interface<Gudhi::Simplex_tree_options_full_featured>::Boundary_simplex_iterator();
```

src/python/include/Simplex_tree_interface.h
VincentRouvreau: Do you want me to save the `Simplex_handle` returned by `Base::find(simplex)` from the `get_boundary_iterator_begin` call and save it in a class attribute ? `get_boundary_iterator_end` could be without arguments.

src/python/test/test_simplex_tree.py
VincentRouvreau: I added a test on 8f9c065d
`st.get_boundaries([])` seg faults... I have a look on it

VincentRouvreau: Ok this is because `st.boundary_simplex_range(st.null_simplex())` seg faults.

If I fix it in `Simplex_tree_iterators.h` by saying:
```cpp
  template<class SimplexHandle>
  Simplex_tree_boundary_simplex_iterator(SimplexTree * st, SimplexHandle sh)
      : //last_(sh->first),
        next_(st->null_vertex()),
        sib_(nullptr),
        sh_(st->null_simplex()),
        st_(st) {
    // Only check once at the beginning instead of for every increment, as this is expensive.
    if (SimplexTree::Options::contiguous_vertices)
      GUDHI_CHECK(st_->contiguous_vertices(), "The set of vertices is not { 0, ..., n } without holes");
    if (sh != st->null_simplex()) {
      last_ = sh->first;
      Siblings * sib = st->self_siblings(sh);
      next_ = sib->parent();
      sib_ = sib->oncles();
    }
```

(I find `template<class SimplexHandle>` quite strange here)
It works fine, but now Cech complex compilation fails because of:

```
Simplex_tree.h:1262:26: error: conversion from ‘boost::container::flat_map<int, Gudhi::Simplex_tree_node_explicit_storage<Gudhi::Simplex_tree<> >, std::less<int>, boost::container::new_allocator<std::pair<int, Gudhi::Simplex_tree_node_explicit_storage<Gudhi::Simplex_tree<> > > > >::reverse_iterator {aka boost::intrusive::reverse_iterator<boost::container::container_detail::vec_iterator<std::pair<int, Gudhi::Simplex_tree_node_explicit_storage<Gudhi::Simplex_tree<> > >*, false> >}’ to non-scalar type ‘Gudhi::Simplex_tree<>::Simplex_handle {aka boost::container::container_detail::vec_iterator<std::pair<int, Gudhi::Simplex_tree_node_explicit_storage<Gudhi::Simplex_tree<> > >*, false>}’ requested
       for(Simplex_handle next = siblings->members().rbegin(); next != simplex; next++) {
                          ^~~~
```

It cannot convert a `boost::intrusive::reverse_iterator<boost::container::container_detail::vec_iterator<...>>` to a `boost::container::container_detail::vec_iterator<...>`

mglisse: I think it is ok to crash in the C++ code when violating a precondition (sh has to refer to a real simplex). It is just in python that crashing the interpreter is a bit too extreme. `assert len(simplex) != 0` would be good enough for me, or something similar in the C++ glue.
Maybe at some point we could introduce something similar to GUDHI_CHECK, for things that we want to check even in release mode (and throw an exception if that fails) but only when used from python. I don't know exactly how much we would use it, but I think it isn't the first time I've thought of that.

src/python/include/Simplex_tree_interface.h
mglisse: The cleanest would be to also call boundary_simplex_range only once. But if that's complicated, storing the Simplex_handle would be fine, although "class attribute" seems strange (what class?). Couldn't the C++ interface return a pair of iterators (I think cython understands std::pair)?

src/python/test/test_simplex_tree.py
VincentRouvreau: Hmm ok. Sounds good to me. I think I need to perform something like `assert find(simplex) != null_simplex()` because something not found would also seg fault.

mglisse: Ah, I hadn't thought of that case...

VincentRouvreau: Ok, I used exceptions on 458bc2d and seems better.

src/python/include/Simplex_tree_interface.h
VincentRouvreau: I return a pair of iterators on 458bc2d

src/python/include/Simplex_tree_interface.h
```diff
@@ -219,6 +220,14 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
     // this specific case works because the range is just a pair of iterators - won't work if range was a vector
     return Base::skeleton_simplex_range(dimension).end();
   }
+
+  std::pair<Boundary_simplex_iterator, Boundary_simplex_iterator> get_boundary_iterators(const Simplex& simplex) {
+    auto bd_sh = Base::find(simplex);
+    if (bd_sh == Base::null_simplex())
+      throw std::runtime_error("simplex not found - cannot find boundaries");
+    // this specific case works because the range is just a pair of iterators - won't work if range was a vector
+    return std::make_pair(Base::boundary_simplex_range(bd_sh).begin(), Base::boundary_simplex_range(bd_sh).end());
```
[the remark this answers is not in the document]
VincentRouvreau: `boundary_simplex_range` called only one on d9ae2cb

src/python/gudhi/simplex_tree.pyx
```diff
@@ -285,6 +285,24 @@ cdef class SimplexTree:
...
+        :rtype:  generator with tuples(simplex, filtration)
+        """
+        cdef pair[Simplex_tree_boundary_iterator, Simplex_tree_boundary_iterator] it =  self.get_ptr().get_boundary_iterators(simplex)
+
+        try:
+            while it.first != it.second:
+                yield self.get_ptr().get_simplex_and_filtration(dereference(it.first))
+                preincrement(it.first)
+        except RuntimeError:
+            print("simplex not found - cannot find boundaries")
```
mglisse, line 303: I would remove the try-except-print. Letting the exception propagate seems better.

VincentRouvreau: Let the exception propagate on 41869710

mglisse (approved):

Issue #73, opened 2019-06-20
[Simplex_tree] reset_filtration


mglisse:
A function that resets the filtration of all simplices to a given value could be useful, especially in python where iterating on all simplices is slow.
For instance, I may build a Delaunay triangulation using Alpha_complex. I set new values on the vertices and want the corresponding lower-star filtration (the filtration of a simplex is the max of its vertices). `make_filtration_non_decreasing` does that if the filtration values of edges, etc are -infinity before, but if the existing filtration values are larger, it gives something different.
I imagine something like `st.reset_filtrations(new_value=0, min_dimension=0)`.

PR #363, opened 2020-07-03
A proposal for simplex_tree reset_filtration (python & C++)

VincentRouvreau:
Maybe useless as we can now use the default_filtration_value to set to NaN an alpha complex.
This is a fix proposal for #73 
Does it worth it to set default values ? `min_dim` could be 0 and `filt_value` could be NaN by default .

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -1668,8 +1668,8 @@ class Simplex_tree {
   }
 
  public:
-  /** \brief This function resets filtration value from a given dimension. Resets all the Simplex_tree when
-   * `min_dim = 0`.
+  /** \brief This function resets the filtration value of all the simplices of dimension at least min_dim. Resets all
+   * the Simplex_tree when `min_dim = 0`.
    * `reset_filtration` may break the filtration property with `min_dim > 0`, and it is the user's responsibility to
    * make it a valid filtration (using a large enough `filt_value`, or calling `make_filtration_non_decreasing`
    * afterwards for instance).
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -1670,10 +1670,13 @@ class Simplex_tree {
  public:
   /** \brief This function resets filtration value from a given dimension. Resets all the Simplex_tree when
    * `min_dim = 0`.
+   * `reset_filtration` may break the filtration property with `min_dim > 0`, and it is the user's responsibility to
+   * make it a valid filtration (using a large enough `filt_value`, or calling `make_filtration_non_decreasing`
+   * afterwards for instance).
    * @param[in] filt_value The new filtration value.
-   * @param[in] min_dim The minimal dimension.
+   * @param[in] min_dim The minimal dimension. Default value is 0.
    */
-  void reset_filtration(Filtration_value filt_value, int min_dim) {
+  void reset_filtration(Filtration_value filt_value, int min_dim = 0) {
     for (auto& simplex : root_.members()) {
       if (min_dim <= 0) {
         simplex.second.assign_filtration(filt_value);
@@ -1689,7 +1692,7 @@ class Simplex_tree {
   /** \brief Recursively resets filtration value from a given dimension.
    * @param[in] sib Siblings to be parsed.
    * @param[in] filt_value The new filtration value.
-   * @param[in] min_dim The maximal dimension.
+   * @param[in] min_dim The minimal dimension.
    */
   void rec_reset_filtration(Siblings * sib, Filtration_value filt_value, int min_dim) {
     for (auto sh = sib->members().begin(); sh != sib->members().end(); ++sh) {
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -1677,14 +1677,7 @@ class Simplex_tree {
    * @param[in] min_dim The minimal dimension. Default value is 0.
    */
   void reset_filtration(Filtration_value filt_value, int min_dim = 0) {
-    for (auto& simplex : root_.members()) {
-      if (min_dim <= 0) {
-        simplex.second.assign_filtration(filt_value);
-      }
-      if (has_children(&simplex)) {
-        rec_reset_filtration(simplex.second.children(), filt_value, min_dim - 1);
-      }
-    }
+    rec_reset_filtration(&root_, filt_value, min_dim);
     clear_filtration(); // Drop the cache.
   }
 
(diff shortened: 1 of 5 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1667,6 +1667,36 @@ class Simplex_tree {
     return sh; // None of its faces has the same filtration.
   }
 
+ public:
+  /** \brief This function resets filtration value until a given dimension.
+   * @param[in] filt_value The new filtration value.
+   * @param[in] max_dim The maximal dimension.
+   */
+  void reset_filtration(Filtration_value filt_value, int max_dim) {
```
mglisse, line 1675: I think a minimal dimension may be more useful than a maximum. For instance, I set some values values on the vertices, reset all other simplices, and call make_filtration_non_decreasing. We can still keep max_dim, but it should have a default value (INT_MAX?).

VincentRouvreau: Hm ok... Maybe more tricky, but doable...
And why not resetting all simplices and then set some values on the vertices ? There is a need ?
Ok for max_dim default value.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1667,6 +1667,41 @@ class Simplex_tree {
     return sh; // None of its faces has the same filtration.
   }
 
+ public:
+  /** \brief This function resets filtration value from a given dimension. Resets all the Simplex_tree when
+   * `min_dim = 0`.
+   * @param[in] filt_value The new filtration value.
+   * @param[in] min_dim The minimal dimension.
+   */
+  void reset_filtration(Filtration_value filt_value, int min_dim) {
```
mglisse, line 1676: Default value of 0 for min_dim?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1667,6 +1667,41 @@ class Simplex_tree {
     return sh; // None of its faces has the same filtration.
   }
 
+ public:
+  /** \brief This function resets filtration value from a given dimension. Resets all the Simplex_tree when
+   * `min_dim = 0`.
```
mglisse, line 1672: Maybe we should comment that with `min_dim>0` this may break the filtration property, and it is the user's responsibility to make it a filtration (using a large enough `filt_value`, or calling `make_filtration_non_decreasing` afterwards for instance).

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1667,6 +1667,41 @@ class Simplex_tree {
...
+      }
+    }
+    clear_filtration(); // Drop the cache.
+  }
+
+ private:
+  /** \brief Recursively resets filtration value from a given dimension.
+   * @param[in] sib Siblings to be parsed.
+   * @param[in] filt_value The new filtration value.
+   * @param[in] min_dim The maximal dimension.
```
mglisse, line 1692: min vs max

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1667,6 +1667,41 @@ class Simplex_tree {
...
+   */
+  void rec_reset_filtration(Siblings * sib, Filtration_value filt_value, int min_dim) {
+    for (auto sh = sib->members().begin(); sh != sib->members().end(); ++sh) {
+      if (min_dim <= dimension(sh)) {
```
mglisse, line 1696: We shouldn't need to compute the dimension of each simplex, the dimension is the same for all the elements of this loop. I think the easiest would be to rename min_dim to min_depth (in the helper only), call recursively with `min_depth-1`, and compare min_depth to 0.

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Done on 12371aa8
Do you think we should default `filt_value` (to `NaN` for instance) ?

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Done on 12371aa8

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Done on 12371aa8

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Optimized on 723252b3

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: > Do you think we should default filt_value (to NaN for instance) ?

As you like.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: "This function resets the filtration value of all the simplices of dimension at least min_dim" might be slightly more explicit.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1667,6 +1667,44 @@ class Simplex_tree {
...
+  /** \brief This function resets filtration value from a given dimension. Resets all the Simplex_tree when
+   * `min_dim = 0`.
+   * `reset_filtration` may break the filtration property with `min_dim > 0`, and it is the user's responsibility to
+   * make it a valid filtration (using a large enough `filt_value`, or calling `make_filtration_non_decreasing`
+   * afterwards for instance).
+   * @param[in] filt_value The new filtration value.
+   * @param[in] min_dim The minimal dimension. Default value is 0.
+   */
+  void reset_filtration(Filtration_value filt_value, int min_dim = 0) {
+    for (auto& simplex : root_.members()) {
```
mglisse, line 1680: Can't we call the recursive function on `&root` to avoid repeating this loop?

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Documentation changed (c++ and python) on e64dd4039e6dff35322ff01a8aa82d79c77e9c8e

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Oh yes you are right ! Simplified on 0c843fd0

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: ok let it like that.

mglisse (approved):

PR #365, opened 2020-07-06
Essential part in wasserstein

tlacombe:
Handle essential parts (points with infinite coordinates) in `gudhi.wasserstein.wasserstein_distance`. 

Allow to automatically ignore essential parts by setting `keep_essential_parts=False` (`True` by default). 

Currently, `enable_autodiff=True` and `keep_essential_parts=True` are incompatible. In order to keep previous implementation correct (ones with `enable_autodiff=True`), we raise a Warning and set `keep_essential_parts=False` implicitly. 
Note: it might be rather simple to extend autodiff to essential parts (at least if there is no pbm of cardinality).

- Convention : if cost is `+infty` (essential parts have different cardinalities, thus any matching can be considered to be optimal), then if `matching=True`, then returns `+np.inf, None` for the tuple `(cost, matching)`. 

TODO: 

- [ ] Check the doc quality.
- [ ] Add more test.


src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+
+    .. note::
+        Assume idX, idY come when calling _handle_essential_parts, thus have same length.
+    '''
+    u = X[idX, axis]
+    v = Y[idY, axis]
+
+    cost = np.sum(np.abs(np.sort(u) - np.sort(v))**(order))  # OT cost in 1D
+
+    sortidX = idX[np.argsort(u)]
```
mglisse, line 137: It feels a bit wasteful to sort u twice, although that's probably negligible.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+    :param idY: indices to consider for this one dimensional OT problem (in Y)
+    :param order: exponent for Wasserstein distanc ecomputation
+    :param axis: must be 0 or 1, correspond to the coordinate which is finite.
+    :returns: cost (float) and match for points with *one* infinite coordinate.
+
+    .. note::
+        Assume idX, idY come when calling _handle_essential_parts, thus have same length.
+    '''
+    u = X[idX, axis]
+    v = Y[idY, axis]
```
mglisse, line 133: I would probably have done that in the caller, but we'll see when we implement enable_autodiff where the most convenient place it for the split, this is fine for now.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+        if len(u) != len(v):
+            return np.inf, None
+
+    # Now we know each essential part has the same number of points in both diagrams.
+    # Handle type 0 and type 1 essential parts (those with one finite coordinates)
+    c1, m1 = _cost_and_match_essential_parts(X, Y, ess_parts_X[0], ess_parts_Y[0], axis=1, order=order)
+    c2, m2 = _cost_and_match_essential_parts(X, Y, ess_parts_X[1], ess_parts_Y[1], axis=0, order=order)
+
+    c += c1 + c2
+    m += m1 + m2
```
mglisse, line 169: It is a bit surprising to use `+=` here when c and m have barely been initialized. Why not `c=c1+c2` and remove the earlier `c=0`?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+        m += list(zip(u, v))  # cost is 0
+
+    return c, np.array(m)
+
+
+def _offdiag(X, enable_autodiff):
+    '''
+    :param X: (n x 2) numpy array encoding a persistence diagram.
+    :param enable_autodiff: boolean, to handle the case where X is a eagerpy tensor.
+    :returns: The off-diagonal part of a diagram `X` (points with finite coordinates).
```
mglisse, line 182: The name "off-diagonal" for finite points is a bit confusing.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+    else:
+        return X[np.where(np.isfinite(X[:,0]) & np.isfinite(X[:,1]))]
+
+
+def wasserstein_distance(X, Y, matching=False, order=1., internal_p=np.inf, enable_autodiff=False,
+                         keep_essential_parts=True):
     '''
-    :param X: (n x 2) numpy.array encoding the (finite points of the) first diagram. Must not contain essential points
-                (i.e. with infinite coordinate).
+    :param X: (n x 2) numpy.array encoding the first diagram. Can now contain essential parts (points with infinite
```
mglisse, line 195: I think we should avoid "now" which belongs in release notes and would soon be outdated in the main doc.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -125,20 +234,49 @@ def wasserstein_distance(X, Y, matching=False, order=1., internal_p=np.inf, enab
...
-            return _perstot(X, order, internal_p, enable_autodiff), np.array([[i, -1] for i in range(n)])
+            cost = _perstot(X, order, internal_p, enable_autodiff)
+            if cost == np.inf:
+                return cost, None
+            else:
+                return cost, np.array([[i, -1] for i in range(n)])
 
+
+    # Second step: handle essential parts
+    if keep_essential_parts:
```
mglisse, line 254: The PR mentions a warning when keep_essential_parts&enable_autodiff, where is it?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -79,6 +80,9 @@ def _perstot(X, order, internal_p, enable_autodiff):
         transparent to automatic differentiation.
     :type enable_autodiff: bool
     :returns: float, the total persistence of the diagram (that is, its distance to the empty diagram).
+
+    .. note::
+        Can be +infty if the diagram has an essential part (points with infinite coordinates).
```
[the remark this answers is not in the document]
tlacombe: - Doc will be more explicit, we aim at handling all (*, *) points, including (+/-inf, +/-inf), etc. 
- From hera's tests ( https://bitbucket.org/grey_narn/hera/src/master/wasserstein/tests/test_hera_wasserstein.cpp ), it seems hera also handles (-inf, *) and (*, inf).

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+        type3 : (-inf, -inf)
+        type4 : (+inf, +inf)
+    .. note::
+        For instance, a[_get_essential_parts(a)[0]] returns the points in a of coordinates (-inf, x) for some finite x.
+    '''
+    if len(a):
+        ess_first_type  = np.where(np.isfinite(a[:,1]) & (a[:,0] == -np.inf))[0] # coord (-inf, x)
+        ess_second_type = np.where(np.isfinite(a[:,0]) & (a[:,1] == np.inf))[0]  # coord (x, +inf)
+        ess_third_type  = np.where((a[:,0] == -np.inf) & (a[:,1] == np.inf))[0]  # coord (-inf, +inf)
+        ess_fourth_type = np.where((a[:,0] == -np.inf) & (a[:,1] == -np.inf))[0] # coord (-inf, -inf)
```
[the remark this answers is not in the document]
tlacombe: Fixed

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Fixed

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Changed it to "finite".

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Done

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +92,137 @@ def _perstot(X, order, internal_p, enable_autodiff):
         return np.linalg.norm(_dist_to_diag(X, internal_p), ord=order)
 
 
-def wasserstein_distance(X, Y, matching=False, order=1., internal_p=np.inf, enable_autodiff=False):
+def _get_essential_parts(a):
+    '''
+    :param a: (n x 2) numpy.array (point of a diagram)
+    :retuns: five lists of indices (between 0 and len(a)) accounting for the five types of points with infinite
```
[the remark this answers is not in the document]
tlacombe: Fixed

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +93,150 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+def _get_essential_parts(a):
+    '''
+    :param a: (n x 2) numpy.array (point of a diagram)
+    :returns: five lists of indices (between 0 and len(a)) accounting for the five types of points with infinite
+    coordinates that can occur in a diagram, namely:
+        type0 : (-inf, finite)
+        type1 : (finite, +inf)
+        type2 : (-inf, +inf)
+        type3 : (-inf, -inf)
+        type4 : (+inf, +inf)
```
mglisse, line 105: Do you count those as being on the diagonal or infinitely far from the diagonal? It isn't well defined what to do with them, I think it would be ok not to support them, but if we support them (maybe we set some filtration values to infinity to "remove" some simplices but still computed the whole diagram), ignoring them (they are on the diagonal) seems more likely to be useful than considering them as really essential. Although returning inf or NaN may advertise that something strange is going on...

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +93,150 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+def _warn_infty(matching):
+    '''
+    Handle essential parts with different cardinalities. Warn the user about cost being infinite and (if
+    `matching=True`) about the returned matching being `None`.
+    '''
+    if matching:
+        warnings.warn('Cardinality of essential parts differs. Distance (cost) is +infty, and the returned matching is None.')
+        return np.inf, None
+    else:
+        warnings.warn('Cardinality of essential parts diffes. Distance (cost) is +infty.')
```
mglisse, line 201: I wonder if we want to warn. I understand it can be useful to a user to understand what is going on. On the other hand, it is just returning the distance, which happens to be infinite, a perfectly valid value. @MathieuCarriere do you have an opinion?

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -88,32 +93,150 @@ def _perstot(X, order, internal_p, enable_autodiff):
...
+    :param X: (n x 2) numpy.array encoding the first diagram. Can contain essential parts (points with infinite
+                        coordinates).
     :param Y: (m x 2) numpy.array encoding the second diagram.
     :param matching: if True, computes and returns the optimal matching between X and Y, encoded as
                      a (n x 2) np.array  [...[i,j]...], meaning the i-th point in X is matched to
                      the j-th point in Y, with the convention (-1) represents the diagonal.
-    :param order: exponent for Wasserstein; Default value is 1.
-    :param internal_p: Ground metric on the (upper-half) plane (i.e. norm L^p in R^2);
+                     Note that if the cost is +inf (essential parts have different number of points,
+                     then the optimal matching will be set to `None`.
```
mglisse, line 216: We could also return an arbitrary matching, say everyone to the diagonal, if that helps callers have a more uniform behavior. But it is probably better indeed if callers are forced to check for inf/None (as in this PR) and don't try to use an arbitrary matching.

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: I get your point. Right now I count them as essential part, i.e. infinitely away from the diagonal. 

One option would be to ignore them but still raising a warning, in the vein of `Warning: points with (+inf,+inf) and (-inf, -inf) are considered as belonging to the diagonal and have been removed.` ?

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Indeed. From my perspective, I like when code is "verbose" and help me to understand what is happening in some critical cases, but I agree that it may be improper to raise a warning when everything is going "as expected".

PS: there is a typo as well, will be corrected.

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: I don't know what is best. 
Returning an arbitrary matching may lead to quite unexpected behavior when it comes to debug as the code is likely to run flawless; thus I personally prefer returning `None` so that the code is likely to crash and the user will first have to figure out how to handle this (e.g. setting `keep_essential_parts=False`).

src/python/gudhi/wasserstein/wasserstein.py
mglisse: https://github.com/grey-narn/hera/blob/b528c4067a8aac346eb307d3c23b82d5953cfe2d/wasserstein/include/wasserstein.h#L286
For reference, Hera now considers (+inf, +inf) as a diagonal point.

src/python/test/test_wasserstein_distance.py
```diff
@@ -5,25 +5,88 @@
...
 
+
 def test_proj_on_diag():
     dgm = np.array([[1., 1.], [1., 2.], [3., 5.]])
     assert np.array_equal(_proj_on_diag(dgm), [[1., 1.], [1.5, 1.5], [4., 4.]])
     empty = np.empty((0, 2))
     assert np.array_equal(_proj_on_diag(empty), empty)
 
+
+def test_finite_part():
```
VincentRouvreau, line 31: [Test coverage](https://8527-174304137-gh.circle-artifacts.com/0/htmlcov/gudhi_wasserstein_wasserstein_py.html) is really, maybe testing `_warn_infty` would help.

VincentRouvreau (commented):
There is also a typo in src/python/gudhi/wasserstein/wasserstein.py on l.337:
```python
    # Comptuation of the otcost using the ot.emd2 library.
```
should be:
```python
    # Computation of the ot cost using the ot.emd2 library.
```

src/python/gudhi/wasserstein/wasserstein.py
tlacombe: Ok I will modify my code to behave in a consistent way (ie considering (+inf,+inf) and (-inf, -inf) as belonging to the diagonal).

src/python/test/test_wasserstein_distance.py
tlacombe: Nice, didn't know that tool. I will write a corresponding test.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -79,6 +81,9 @@ def _perstot(X, order, internal_p, enable_autodiff):
         transparent to automatic differentiation.
     :type enable_autodiff: bool
     :returns: float, the total persistence of the diagram (that is, its distance to the empty diagram).
+
+    .. note::
+        Can be +infty if the diagram has an essential part (points with infinite coordinates).
```
mglisse, line 86: Is the abbreviation "infty" standard in python? "inf" seems more common, if you don't want to spell the whole word.

tlacombe: Changed `infty` for `inf` in the doc. Should I also changed functions' names?

mglisse: I don't care much what the private functions are called.

mglisse (approved):

VincentRouvreau (approved):

PR #368, opened 2020-07-28
Add mechanism to deactivate LaTeX matplotlib rendering

VincentRouvreau:
As #351 is partially fixed, add and document mechanism to deactivate LaTeX matplotlib rendering

### DIFF
--- a/src/python/doc/persistence_graphical_tools_user.rst
+++ b/src/python/doc/persistence_graphical_tools_user.rst
@@ -94,25 +94,8 @@ If you want more information on a specific dimension, for instance:
 LaTeX support
 -------------
 
-By default, persistence graphical tools are using LaTeX support for matplotlib if available (cf.
-`matplotlib text rendering with LaTeX <https://matplotlib.org/3.3.0/tutorials/text/usetex.html>`_).
-It also requires `type1cm` LaTeX package (not detected by matplotlib).
-
-If you are facing issues with LaTeX rendering, like this one:
-
-.. code-block:: none
-
-    Traceback (most recent call last):
-      File "/usr/lib/python3/dist-packages/matplotlib/texmanager.py", line 302, in _run_checked_subprocess
-        report = subprocess.check_output(command,
-    ...
-    ! LaTeX Error: File `type1cm.sty' not found.
-    ...
-
-This is because the LaTeX package is not installed on your system. On Ubuntu systems you can install texlive-full
-(for all LaTeX packages), or more specific packages like texlive-latex-extra, cm-super.
-
-You can still deactivate LaTeX rendering by saying:
+If you are facing issues with `LaTeX <installation.html#latex>`_ rendering, you can still deactivate LaTeX rendering by
+saying:
 
 .. code-block:: python
 
--- a/src/python/doc/persistence_graphical_tools_user.rst
+++ b/src/python/doc/persistence_graphical_tools_user.rst
@@ -90,3 +90,17 @@ If you want more information on a specific dimension, for instance:
     gudhi.plot_persistence_density(persistence=pers_diag,
         dimension=1, legend=True, axes=axes[1])
     plt.show()
+
+LaTeX support
+-------------
+
+By default, persistence graphical tools are using LaTeX support for matplotlib if available (cf.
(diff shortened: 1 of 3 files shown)


src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -20,6 +20,20 @@
 __copyright__ = "Copyright (C) 2016 Inria"
 __license__ = "MIT"
 
+@lru_cache(maxsize=1)
```
mglisse, line 23: Is this still useful in this version?

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -20,6 +20,20 @@
...
+    The returned value is cached.
+    """
+    try:
+        from matplotlib import checkdep_usetex
+        return checkdep_usetex(True)
+    except ImportError:
+        print("This function is not available, you may be missing matplotlib.")
+
+
+_gudhi_matplotlib_use_tex = _matplotlib_can_use_tex()
```
mglisse, line 35: The message "This function is not available, you may be missing matplotlib." will appear as soon as someone does `import gudhi`, since the module is imported by default.

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: You are right, `import matplotlib` will be catched before this call. I remove it.

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Ok, so let's do it differently on 37f80627

src/python/gudhi/persistence_graphical_tools.py
mglisse: Maybe you can add it back with the more recent change?

src/python/doc/persistence_graphical_tools_user.rst
```diff
@@ -90,3 +90,17 @@ If you want more information on a specific dimension, for instance:
...
     plt.show()
+
+LaTeX support
+-------------
+
+By default, persistence graphical tools are using LaTeX support for matplotlib if available (cf.
+`matplotlib text rendering with LaTeX <https://matplotlib.org/3.3.0/tutorials/text/usetex.html>`_).
+It also requires `type1cm` LaTeX package (not detected by matplotlib).
+
+If you are facing issues with LaTeX rendering, you can still deactivate LaTeX rendering by saying:
```
mglisse, line 101: I think it would be nice to show here what the error message looks like, when type1cm is not available (skipping some parts since it is super long, but including at least the beginning and the end). Note that one of the issue reports was about type1ec.sty instead of type1cm.sty.
We could also mention that on Ubuntu, texlive-full (or more specific packages like texlive-latex-extra, cm-super) installs the dependencies.
Do we mention latex in the optional runtime dependencies on the installation page?

mglisse:
I wonder if setting the matplotlib font during a call to those functions is really a good idea. If the user likes some specific font, couldn't they set it themselves once at the beginning of their program? And if they do, it looks like the current code not only overrides their choice for the gudhi functions, it actually overwrites their choice also for future direct calls to matplotlib by the user. So one option would be to remove those settings from the gudhi code, and instead advertise in the doc that we recommend users call matplotlib.pyplot.rc(...) (use latex + CM font) to get prettier plots.
I haven't thought much about it, just putting the idea out there to see what people think. Let me also ping @tlacombe .

src/python/doc/persistence_graphical_tools_user.rst
VincentRouvreau: Done on 9b52376e

> Do we mention latex in the optional runtime dependencies on the installation page?

The dependency is matplotlib, and LaTeX is a matplotlib dependency. I think it does not worth it to mention LaTeX in optional runtime dependencies.

mglisse: > The dependency is matplotlib, and LaTeX is a matplotlib dependency.

LaTeX is an optional dependency of matplotlib, which we explicitly use and thus kind of make a hard dependency. `apt install python3-matplotlib` does not install latex. That's why I was considering it. But ok.

mglisse:
ok, and we can rediscuss the default in september.

PR #371, opened 2020-08-03
Wasserstein autodiff

MathieuCarriere:
I had to do a very small fix in order to make wasserstein_distance (with enable_autodiff=True) work with Tensorflow 2.0. I don't know if this modif would break the code for PyTorch though (I hope not).

### DIFF
--- a/src/python/gudhi/point_cloud/knn.py
+++ b/src/python/gudhi/point_cloud/knn.py
@@ -46,7 +46,7 @@ class KNearestNeighbors:
             sort_results (bool): if True, then distances and indices of each point are
                 sorted on return, so that the first column contains the closest points.
                 Otherwise, neighbors are returned in an arbitrary order. Defaults to True.
-            enable_autodiff (bool): if the input is a torch.tensor, jax.numpy.ndarray or tensorflow.Tensor, this
+            enable_autodiff (bool): if the input is a torch.tensor or tensorflow.Tensor, this
                 instructs the function to compute distances in a way that works with automatic differentiation.
                 This is experimental, not supported for all metrics, and requires the package EagerPy.
                 Defaults to False.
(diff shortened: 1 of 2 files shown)


mglisse:
(please don't commit stuff related to the hera submodule)
Hmm, nonzero returns a tuple, and we want the first element of the tuple, so this patch looks good and should be necessary even with pytorch. I am confused why the current code ever worked... I'll have to investigate.

MathieuCarriere:
I think the bug doesn't show if you never match points to the diagonal, maybe that's why the current code still worked?

mglisse:
And after a quick test, jax.grad doesn't like our function at all. Looks like I oversold it in the doc :-(

mglisse:
I cleaned the branch a bit (the way you merge master without fast-forward leaves some useless merges) and removed jax from the doc since it doesn't work (I am confused because, contrary to tensorflow, I did test with jax at some point, but now it looks hopeless, maybe they changed things in new versions). At least you confirmed that it works with tensorflow and not just pytorch...
I'll merge the branch as soon as circleci is done.

mglisse:
I think the issue appeared because originally I had `[0]` (or equivalent) in the definition of pairs_X_diag, and then changed it because wrapping the indices in a tuple works around a misfeature of pytorch, and forgot to update those tests. And some tensor implementations don't seem to mind the empty case.

mglisse:
By the way, Mathieu, could you write a very small test of wasserstein differentiation with tensorflow (possibly even smaller than what I added for pytorch at the end of test_wasserstein_distance.py), to check for obvious breaks in the future? We can handle the cmake/CI part if you don't want to bother, but the test is probably much easier for you to write...

MathieuCarriere:
Sure I'll add a small test!

PR #377, opened 2020-08-04
Added test for wasserstein distance with tensorflow 2

MathieuCarriere:


mglisse:
Thank you!
This will wait until after the release. We need to add tensorflow detection to cmake, condition the test, add tensorflow to the CI that don't have it, etc.
For the next patches, could you start from a clean `master` from upstream instead of just merging into your version of master? As you can see on github, the history of this branch looks messy. https://github.com/GUDHI/gudhi-devel/blob/master/.github/how_to_use_github_to_contribute_to_gudhi.md#create-a-branch-based-on-the-current-master suggests one workflow (you don't have to follow it, but it can give ideas).

mglisse:
@VincentRouvreau Could you look into the build system part of this PR? I assume some mention of tensorflow is needed for cmake, possibly the test could be split, I don't know if all CI already have tensorflow, etc.

VincentRouvreau:
> @VincentRouvreau Could you look into the build system part of this PR? I assume some mention of tensorflow is needed for cmake, possibly the test could be split, I don't know if all CI already have tensorflow, etc.

I did it on another PR #396

PR #382, opened 2020-08-19
Documents for weighted Rips

yuichi-ike:
I forgot to add my name to documents and was asked to do that :( I remember that Vincent said
> The only problem will be the list of authors that becomes extremely long, but I think we could separate authors by sections, or find another nice solution (this is not a real problem).

and I haven't addressed it yet. The list actually seems extremely long. How should I change it?


VincentRouvreau:
I am OK with the [resulting documentation](https://6125-174304137-gh.circle-artifacts.com/0/tmp/sphinx/index.html#rips-complex).
We will see with @mglisse if we split authors on section or not.

mglisse:
Fine with me. Yes, some reorganization so the name list doesn't grow too much could be nice at some point, but it doesn't look so bad now, and the priority is to have a list that credits all the people who deserve it.

Issue #257, opened 2020-03-18
[Bitmap_cubical_complex] compute_incidence_between_cells has inconsistent const


mglisse:
> Bitmap_cubical_complex_base.h:  virtual int compute_incidence_between_cells(std::size_t coface, std::size_t face) **const** {

> Bitmap_cubical_complex_periodic_boundary_conditions_base.h:  virtual int compute_incidence_between_cells(std::size_t coface, std::size_t face) {

but virtual functions need signatures that exactly match, or they might as well not be virtual.

VincentRouvreau:
Closed by #385

PR #385, opened 2020-08-21
Virtual functions need signatures that exactly match

VincentRouvreau:
A fix proposal for #257 
I was less sure, but I also did it for `inline` virtual functions.


VincentRouvreau:
> I don't think we need `inline`. Adding some `override` would be a nice way to show that the issue is fixed.

I removed `inline` and add some `override` on 023c4cc

mglisse (approved):

PR #405, opened 2020-10-17
Bottleneck distance for one point or negative coordinates

### LINKED ISSUE #399: bottleneck_distance([],[[.99,1.]]) returns .5 ???
Actually, `gudhi.bottleneck_distance([],[[x,1.]])` returns .5 whatever the value of x smaller than 1 (I tried -10, 0, .9)...
Forcing epsilon=0 avoids the issue.
`gudhi.hera.bottleneck_distance` is fine.

mglisse:
Fix #399.
@VincentRouvreau you may want to tweak the tests so they involve negative numbers at some point.

### DIFF
--- a/src/python/doc/bottleneck_distance_user.rst
+++ b/src/python/doc/bottleneck_distance_user.rst
@@ -72,6 +72,6 @@ The output is:
 
 .. testoutput::
 
-    Bottleneck distance approximation = 0.81
+    Bottleneck distance approximation = 0.72
     Bottleneck distance value = 0.75
 
--- a/src/python/doc/bottleneck_distance_user.rst
+++ b/src/python/doc/bottleneck_distance_user.rst
@@ -35,19 +35,19 @@ The following example explains how the distance is computed:
 
     import gudhi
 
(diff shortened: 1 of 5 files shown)


src/python/doc/bottleneck_distance_user.rst
```diff
@@ -35,19 +35,19 @@ The following example explains how the distance is computed:
 
     import gudhi
 
-    message = "Bottleneck distance = " + '%.1f' % gudhi.bottleneck_distance([[0., 0.]], [[0., 13.]])
+    message = "Bottleneck distance = " + '%.1f' % gudhi.bottleneck_distance([[0., 2.]], [[1., 13.]])
```
VincentRouvreau, line 38: If you change the example, the [figure](https://gudhi.inria.fr/python/latest/bottleneck_distance_user.html#distance-computation) is no more correct

VincentRouvreau (commented):
It would be nice to have tests with negative coordinates that guarantees #399 is fixed

src/python/doc/bottleneck_distance_user.rst
mglisse: Indeed. I've reverted that.

VincentRouvreau (approved):

PR #408, opened 2020-11-01
Clean-ups in subsampling

mglisse:
This could improve performance a bit, but shouldn't change the functionality.

Note that we do not test gudhi without -std=gnu++14, so the C++17 code isn't exercised (I compiled and ran the unit test manually).

### DIFF
--- a/src/Subsampling/include/gudhi/sparsify_point_set.h
+++ b/src/Subsampling/include/gudhi/sparsify_point_set.h
@@ -11,7 +11,12 @@
 #ifndef SPARSIFY_POINT_SET_H_
 #define SPARSIFY_POINT_SET_H_
 
-#include <boost/iterator/function_output_iterator.hpp>
+#include <boost/version.hpp>
+#if BOOST_VERSION < 106600
+# include <boost/function_output_iterator.hpp>
+#else
+# include <boost/iterator/function_output_iterator.hpp>
+#endif
 
 #include <gudhi/Kd_tree_search.h>
 #ifdef GUDHI_SUBSAMPLING_PROFILING
@@ -53,7 +58,6 @@ sparsify_point_set(
                    OutputIterator output_it) {
   typedef typename Gudhi::spatial_searching::Kd_tree_search<
       Kernel, Point_range> Points_ds;
-  using std::sqrt;
 
 #ifdef GUDHI_SUBSAMPLING_PROFILING
   Gudhi::Clock t;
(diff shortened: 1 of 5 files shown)


VincentRouvreau:
`test_subsampling_py_test` is failing on azure and appveyor on `assert gudhi.sparsify_point_set(points=point_set, min_squared_dist=1.0) == point_set`

mglisse:
At least in python where we compute the distance with double (so we usually only get an approximation of the distance), what happens when the distance is exactly something is not relevant. In C++, where we can instantiate with an exact number type, this could make sense, but I am not sure which way is more desirable, so I am tempted to pick the one that is easiest with cgal.

mglisse:
Ah, but there is something much more important: I hadn't noticed, but CGAL's Fuzzy_sphere takes a radius (and immediately squares it), contrary to most places that would just deal with a squared radius directly. Various options
* breaking change, sparsify_point_set now takes a minimum distance (not squared)
* use `sqrt`. This doesn't work for Epeck_d (maybe `approximate_sqrt`?).
* define our own Ball type (deriving from Fuzzy_sphere? Maybe not) to use in Kd_tree queries.

VincentRouvreau:
> Ah, but there is something much more important: I hadn't noticed, but CGAL's Fuzzy_sphere takes a radius (and immediately squares it), contrary to most places that would just deal with a squared radius directly. Various options
> 
>     * breaking change, sparsify_point_set now takes a minimum distance (not squared)
> 
>     * use `sqrt`. This doesn't work for Epeck_d (maybe `approximate_sqrt`?).
> 
>     * define our own Ball type (deriving from Fuzzy_sphere? Maybe not) to use in Kd_tree queries.
> 
> 
>     * [ ]  Handle (non-)squared radius

It seems that you choose the 2nd solution and it is fine to me. Maybe we should add a warning in the documentation about Epeck_d.
The cpp test is quite minimalist and couldn't detect this kind of issue.

mglisse:
I used sqrt as the smallest change that let the tests pass while we were discussing. I am not particularly happy with it...

src/Subsampling/include/gudhi/sparsify_point_set.h
```diff
@@ -11,6 +11,8 @@
 #ifndef SPARSIFY_POINT_SET_H_
 #define SPARSIFY_POINT_SET_H_
 
+#include <boost/iterator/function_output_iterator.hpp>
```
VincentRouvreau, line 14: It seems this has been introduced in Boost 1.66 and gudhi requires Boost ≥ 1.56.0.
This modification requires some documentation change.

VincentRouvreau:
> I used sqrt as the smallest change that let the tests pass while we were discussing. I am not particularly happy with it...

It seems to me the best solution would be the 'breaking change, sparsify_point_set now takes a minimum distance (not squared)', no ?
Maybe we should have a quick look on tests that will fail if we do such a change.

src/Subsampling/include/gudhi/sparsify_point_set.h
mglisse: It was `boost/function_output_iterator.hpp` before that, I should probably `#if` the boost version unless we consider that 1.66 is recent enough that we can require it (end of 2017, that may be a bit too recent).

VincentRouvreau (approved):

PR #401, opened 2020-10-15
Zigzag filtration simplex tree

ClementMaria:

PR #409, opened 2020-11-01
Pass a distance function instead of a kernel to choose_n_farthest_points

mglisse:
This is motivated by https://github.com/GUDHI/gudhi-devel/pull/401 by @ClementMaria , but I have been tempted before, see how it simplifies the sparse Rips. This is split from Clement's PR so we can discuss it without it disappearing in the noise.
I think the main question is whether we should keep the old version in addition to the new one, for compatibility, and if so how we should make them coexist (different function name? weird changes to the list of arguments?). Updating some old code to call the new function is rather trivial. In my opinion, the main reason to keep the old version would be if we could foresee a more extensive use of the traits class (aka kernel) to speed things up, for instance if we started using a kd-tree in this function, but I am not aware of clever algorithms for this function, so that doesn't seem likely. So I am suggesting this breaking change.
The use of Euclidean_distance in the python plugin could be replaced with some Squared_euclidean_distance to save the computation of square roots, but I doubt the performance difference is very significant.


src/Subsampling/include/gudhi/choose_n_farthest_points.h
```diff
@@ -86,9 +88,9 @@ void choose_n_farthest_points(Kernel const &k,
     starting_point = dis(gen);
   }
 
-  typename Kernel::Squared_distance_d sqdist = k.squared_distance_d_object();
-
   std::size_t current_number_of_landmarks = 0;  // counter for landmarks
+  static_assert(std::numeric_limits<double>::has_infinity);
```
VincentRouvreau, line 92: `warning: static_assert without a message only available with -std=c++1z or -std=gnu++1z`

src/python/include/Subsampling_interface.h
```diff
@@ -27,23 +28,22 @@ namespace subsampling {
...
-using Subsampling_ft = Subsampling_dynamic_kernel::FT;
 
 // ------ choose_n_farthest_points ------
 std::vector<std::vector<double>> subsampling_n_farthest_points(const std::vector<std::vector<double>>& points,
                                                                unsigned nb_points) {
   std::vector<std::vector<double>> landmarks;
-  Subsampling_dynamic_kernel k;
-  choose_n_farthest_points(k, points, nb_points, random_starting_point, std::back_inserter(landmarks));
+  choose_n_farthest_points(Euclidean_distance(), points, nb_points,
+                           random_starting_point, std::back_inserter(landmarks));
```
VincentRouvreau, line 37: Did you benchmark this modification ? I don't know if CGAL made something smart in `squared_distance_d_object` compared to our naive `Euclidean_distance()` implementation.

mglisse: I didn't benchmark. CGAL doesn't do anything smart there, except for not computing the square root. Well it also uses `CGAL::square` instead of doing `tmp*tmp`, but at least for `double` that doesn't make any difference.

VincentRouvreau (approved):

PR #411, opened 2020-11-02
PD gradient

MathieuCarriere:
A pull request for computing gradient of various persistence diagram constructions with Tensorflow


src/python/gudhi/differentiation/tensorflow.py
```diff
@@ -0,0 +1,251 @@
+import numpy               as np
+import tensorflow          as tf
+import tensorflow_addons   as tfa
```
mglisse, line 3: tensorflow_addons needs do be added to the cmake and CI files.

mglisse: I am adding it to CI in #425, but after that you will still need to add the cmake stuff.

MathieuCarriere: great thanks!!

src/python/gudhi/differentiation/tensorflow.py
```diff
@@ -0,0 +1,251 @@
...
+    perm = np.argsort(pers)
+    indices = list(np.reshape(indices, [-1,2])[perm][::-1,:].flatten())
+    
+    # Pad vertex pairs
+    indices = indices[:2*card] + [0 for _ in range(0,max(0,2*card-len(indices)))]
+    return list(np.array(indices, dtype=np.int32))
+
+class SimplexTreeModel(tf.keras.Model):
+    """
+    TensorFlow model for computing lower-star persistence out of a simplex tree.
```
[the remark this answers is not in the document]
MathieuCarriere: Very good idea, thanks!

src/python/gudhi/differentiation/tensorflow.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Attributes:
+        F (TensorFlow variable): filter function values over the vertices of the simplex tree
+        stbase (string): path to the file containing the simplex tree
```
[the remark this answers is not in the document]
MathieuCarriere: Yep, done

MathieuCarriere:
Pour le coup je ne sais pas trop comment réparer le fail du build windows...

mglisse:
You can ignore the windows mpfr issue, we'll need to modify or disable that CI.

src/python/doc/differentiation.rst
```diff
@@ -0,0 +1,18 @@
...
+
+.. To get rid of WARNING: document isn't included in any toctree
+
+======================
+Differentiation manual
+======================
+
+.. include:: differentiation_sum.inc
+
+In this module, we provide neural network models for computing persistent homology. In particular, we provide TensorFlow 2 models that allow to compute persistence diagrams from complexes available in the Gudhi library, including simplex trees, cubical complexes and Vietoris-Rips complexes. These models can be incorporated at each step of a given neural network architecture, and can be used in addition to `PersLay <https://github.com/MathieuCarriere/gudhi/blob/perslay/src/python/gudhi/representations/perslay.py>`_ to produce topological features.
```
tlacombe, line 11: > we provide neural network models for computing persistent homology.

Wouldn't something like "a model supporting automatic differentiation" be more precise? In sense that it could be use in a more general setting than the one of neural networks, right?
 I am not very familiar with this project so feel free to ignore this comment.

src/python/gudhi/differentiation/tensorflow.py
```diff
@@ -0,0 +1,258 @@
...
+    perm = np.argsort(pers)
+    indices = list(np.reshape(indices, [-1,2])[perm][::-1,:].flatten())
+    
+    # Pad vertex pairs
+    indices = indices[:2*card] + [0 for _ in range(0,max(0,2*card-len(indices)))]
+    return list(np.array(indices, dtype=np.int32))
+
+class LowerStarSimplexTreeModel(tf.keras.Model):
+    """
+    TensorFlow model for computing lower-star persistence out of a simplex tree. Since simplex trees cannot be easily encoded as TensorFlow variables, the model takes as input a path to a file containing the simplex tree simplices, and read it each time the simplex tree is required for computations.
```
tlacombe, line 69: This line is a bit long. Don't know if we really care about it though.

src/python/gudhi/differentiation/tensorflow.py
```diff
@@ -0,0 +1,258 @@
...
+        
+    def call(self):
+        m, d, c = self.mel, self.dim, self.card
+        
+        # Compute distance matrix
+        DX = tf.math.sqrt(tf.reduce_sum((tf.expand_dims(self.X, 1)-tf.expand_dims(self.X, 0))**2, 2))
+        DXX = tf.reshape(DX, [1, DX.shape[0], DX.shape[1]])
+        
+        # Turn numpy function into tensorflow function
+        RipsTF = lambda DX: tf.numpy_function(_Rips, [DX, m, d, c], [tf.int32 for _ in range(4*c)])
```
tlacombe, line 172: Not sure if this is crucial, but `[tf.int32 for _ in range(4*c)]` can be replaced by `[tf.int32]*(4*c)`, which is about 30x faster with c=1000 for instance (same comment holds in other places as well).

tlacombe: To reproduce my claim:
```python
~ ➜  ipython                                                                                                                                                                               
In [1]: import tensorflow as tf
In [2]: c = 1000
In [3]: %timeit [tf.int32 for _ in range(4*c)]
201 µs ± 5.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
In [4]: %timeit [tf.int32]*(4*c)
6.92 µs ± 16.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
In [5]: [tf.int32]*(4*c) == [tf.int32 for _ in range(4*c)]
Out[5]: True
```

src/python/doc/cubical_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,35 @@
...
+    import numpy as np
+    import tensorflow as tf
+
+    Xinit = np.array([[0.,2.,2.],[2.,2.,2.],[2.,2.,1.]], dtype=np.float32)
+    X = tf.Variable(initial_value=Xinit, trainable=True)
+    cl = CubicalLayer(dimension=0)
+
+    with tf.GradientTape() as tape:
+        dgm = cl.call(X)
+        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
```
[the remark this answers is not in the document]
MathieuCarriere: Done!

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
```diff
@@ -0,0 +1,59 @@
...
+    st.insert([8, 9]) 
+    st.insert([9, 10]) 
+
+    Finit = np.array([6.,4.,3.,4.,5.,4.,3.,2.,3.,4.,5.], dtype=np.float32)
+    F = tf.Variable(initial_value=Finit, trainable=True)
+    sl = LowerStarSimplexTreeLayer(simplextree=st, dimension=0)
+
+    with tf.GradientTape() as tape:
+        dgm = sl.call(F)
+        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
```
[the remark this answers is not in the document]
MathieuCarriere: Done!

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,35 @@
...
+    import numpy as np
+    import tensorflow as tf
+
+    Xinit = np.array([[1.,1.],[2.,2.]], dtype=np.float32)
+    X = tf.Variable(initial_value=Xinit, trainable=True)
+    rl = RipsLayer(maximum_edge_length=2., dimension=0)
+
+    with tf.GradientTape() as tape:
+        dgm = rl.call(X)
+        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
```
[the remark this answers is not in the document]
MathieuCarriere: Done!

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
```diff
@@ -0,0 +1,59 @@
...
+    grads = tape.gradient(loss, [F])
+    print(grads[0].numpy())
```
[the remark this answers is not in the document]
MathieuCarriere: Indeed, thanks for spotting this. Just changed the doc.

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,35 @@
...
+
+TensorFlow layer for Vietoris-Rips persistence
+##############################################
+
+.. include:: differentiation_sum.inc
+
+Example of gradient computed from Vietoris-Rips persistence
+-----------------------------------------------------------
+
+.. code-block:: python
```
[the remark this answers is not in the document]
MathieuCarriere: Yep, all three are deterministic, so I updated the doc accordingly.

src/python/gudhi/differentiation/tensorflow.py
mglisse: Actually, now it looks like tensorflow_addons is not used anymore. Should we remove it from gudhi-deploy?

src/python/gudhi/tensorflow/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .cubical_layer import CubicalLayer
+from .lower_star_simplex_tree_layer import LowerStarSimplexTreeLayer
+from .rips_layer import RipsLayer
```
mglisse, line 3: So it is impossible to import say just CubicalLayer without also importing all the other tensorflow layers?
That's a possible choice, I am just wondering if that's deliberate.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,66 @@
...
+    return np.array(ocof, dtype=np.int32)
+
+class CubicalLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing cubical persistence out of a cubical complex
+
+    Attributes:
+        dimension (int): homology dimension
+    """
+    def __init__(self, dimension=1, **kwargs):
```
mglisse, line 47: For some reason sphinx shows
> class gudhi.tensorflow.CubicalLayer(*args, **kwargs)

so it isn't clear at all what arguments I may pass. Or is it conventional that if there is an attribute `dimension` then there is probably also an argument with the same name in `__init__`?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__(dynamic=True, **kwargs)
+        self.dimension = dimension
+
+    def build(self):
+        super.build()
+        
+    def call(self, X):
+        """
+        Compute persistence diagram associated to a cubical complex filtered by some pixel values
```
mglisse, line 56: What does the output look like? (type, shape)

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,66 @@
...
+        ocof[D*(idx+1):D*(idx+2)] = np.unravel_index(cof[count,1], Xs)
+        count += 1
+    return np.array(ocof, dtype=np.int32)
+
+class CubicalLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing cubical persistence out of a cubical complex
+
+    Attributes:
+        dimension (int): homology dimension
```
mglisse, line 45: So there is one fixed dimension. What is the plan if my loss depends on the diagrams in dimensions 0 and 1? Should I make several independent layers, and hope :crossed_fingers: that the cost of the highest dimension dominates so it isn't too bad?

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    def build(self):
+        super.build()
+    
+    def call(self, filtration):
+        """
+        Compute lower-star persistence diagram associated to a function defined on the vertices of the simplex tree
+
+        Parameters:
+            F (TensorFlow variable): filter function values over the vertices of the simplex tree
```
mglisse, line 70: Does it assume that the vertices of the simplex tree are numbered contiguously 0 ... n? More precisely, does F[i] correspond to vertex i even if there is no vertex 0 in the complex?

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,77 @@
...
+    indices = np.reshape(indices, [-1,2])[perm][::-1,:].flatten()
+    
+    return np.array(indices, dtype=np.int32)
+
+class LowerStarSimplexTreeLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing lower-star persistence out of a simplex tree
+
+    Attributes:
+        simplextree (gudhi.SimplexTree()): underlying simplex tree
```
mglisse, line 54:
```suggestion
        simplextree (gudhi.SimplexTree): underlying simplex tree 
```
I guess if this is supposed to be the type?

src/python/doc/rips_complex_tflow_itf_ref.rst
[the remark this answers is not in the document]
mglisse: Probably at some point it would be nice to have a function for that, to get the more readable `total_persistence(dgm, 2)`, but not right now.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,77 @@
...
+    for s,_ in simplextree.get_filtration():
+        simplextree.assign_filtration(s, -1e10)
+
+    # Assign new filtration values
+    for i in range(simplextree.num_vertices()):
+        simplextree.assign_filtration([i], filtration[i])
+    simplextree.make_filtration_non_decreasing()
+    
+    # Compute persistence diagram
+    dgm = simplextree.persistence()
```
mglisse, line 24: It doesn't look like you use `dgm` later, so just `simplextree.compute_persistence()`.

src/python/gudhi/tensorflow/rips_layer.py
```diff
@@ -0,0 +1,75 @@
...
+    #             dimension (homology dimension)
+
+    # Compute the persistence pairs with Gudhi
+    rc = RipsComplex(distance_matrix=DX, max_edge_length=max_edge)
+    st = rc.create_simplex_tree(max_dimension=dimension+1)
+    dgm = st.persistence()
+    pairs = st.persistence_pairs()
+
+    # Retrieve vertices v_a and v_b by picking the ones achieving the maximal
+    # distance among all pairwise distances between the simplex vertices
```
mglisse, line 23: SimplexTree.flag_persistence_generators already does something very similar to this function, and uses edge_with_same_filtration to find the edge. What is the main goal in reimplementing it (and lower_star_persistence_generators)? Do the existing functions have an inconvenient interface? Do they do unnecessary work that makes them slower? The whole point of those 2 functions was that we could call them for pytorch/tensorflow differentiation. If they don't fit this purpose, maybe we should modify them?

src/python/gudhi/tensorflow/rips_layer.py
```diff
@@ -0,0 +1,75 @@
...
+    for s1, s2 in pairs:
+        if len(s1) == dimension+1 and len(s2) > 0:
+            l1, l2 = np.array(s1), np.array(s2)
+            i1 = [l1[v] for v in np.unravel_index(np.argmax(DX[l1,:][:,l1]),[len(l1), len(l1)])]
+            i2 = [l2[v] for v in np.unravel_index(np.argmax(DX[l2,:][:,l2]),[len(l2), len(l2)])]
+            indices.append(i1)
+            indices.append(i2)
+            pers.append(st.filtration(s2)-st.filtration(s1))
+    
+    # Sort points with distance-to-diagonal
```
mglisse, line 34: Is it documented that the output is sorted this way? Is it useful?

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,77 @@
...
+    
+    # Get vertex pairs for optimization. First, get all simplex pairs
+    pairs = simplextree.persistence_pairs()
+    
+    # Then, loop over all simplex pairs
+    indices, pers = [], []
+    for s1, s2 in pairs:
+        # Select pairs with good homological dimension and finite lifetime
```
mglisse, line 32: Is it never useful to optimize the birth time of infinite bars?

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,41 @@
...
+Example of gradient computed from Vietoris-Rips persistence
+-----------------------------------------------------------
+
+.. testcode::
+
+    from gudhi.tensorflow import *
+    import numpy as np
+    import tensorflow as tf
+
+    Xinit = np.array([[1.,1.],[2.,2.]], dtype=np.float32)
```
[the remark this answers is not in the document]
MathieuCarriere: You are right, it was not intended so I updated the doc.

src/python/gudhi/tensorflow/__init__.py
MathieuCarriere: That is not really deliberate, I just don't know how to avoid this -_-'

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Ah I was not putting the doc at the right place! I updated it.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: I updated the doc.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: Yep, so I added a comment about this in the doc. Do you think we should emphasize this somewhere else?

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: That's more or less the plan for now... but indeed this feels really suboptimal

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: you are right, thanks, I modified the doc

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: thanks! done.

src/python/gudhi/tensorflow/rips_layer.py
MathieuCarriere: My bad, I just forgot that function. I am using it now.

src/python/gudhi/tensorflow/rips_layer.py
MathieuCarriere: That is probably from some time ago, when I needed the PDs to be sorted for some reason (that I forgot). It's actually unnecessary so I removed that part.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: It probably is, but forgetting the infinite points makes life with tensorflow easier as the output shape is fixed.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
mglisse: Also skip `dgm=` then? Setting `dgm` to `None` is confusing.

src/python/gudhi/tensorflow/__init__.py
mglisse: Uh? You mean what happens if you remove those 3 lines? :confused:
Or are you asking how to have CubicalLayer directly in gudhi.tensorflow but still require some `import gudhi.tensorflow.cubical_layer` or something?
(again, I am not asking you to change that, just discussing what we want in the end)

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
mglisse: Looks clear enough now that it is mentioned in both places.

src/python/gudhi/tensorflow/rips_layer.py
mglisse: There is a similar function for lower-star filtration that could also be nice to use.
Again, if those functions don't quite fit, or if they perform worse than your code, don't hesitate to say something.

src/python/gudhi/tensorflow/cubical_layer.py
mglisse: At least with this interface it remains possible to add later the possibility to pass something like `dimension=(1,2)` to handle multiple dimensions. I think it is ok for a tensorflow layer to take or return a tuple of tensors instead of just one tensor.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
mglisse: By fixed you mean always of the form `(?,2)` while the infinite features would give something 1-dimensional? We wouldn't connect them to the same later layers anyway. This seems related to the question about several dimensions.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,65 @@
...
+        
+    def call(self, X):
+        """
+        Compute persistence diagram associated to a cubical complex filtered by some pixel values 
+
+        Parameters:
+            X (TensorFlow variable): pixel values of the cubical complex
+
+        Returns:
+            dgm (TensorFlow variable): cubical persistence diagram with shape [num_points, 2]
```
mglisse, line 58: (picking one randomly)
Do we get the complete persistence diagram, including points on the diagonal?
One reason to ask is that when computing H2 for the Rips (ok, I picked the wrong one to comment), collapsing some edges makes the persistence computation faster, but it changes the points on the diagonal. So if we promise to output the true diagonal points, then we cannot do it. If we say there may or may not be diagonal points... I don't know if being vague like that would be ok.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,72 @@
...
+            count += 1
+        L_cofs.append(np.array(ocof, dtype=np.int32))
+
+    return L_cofs
+
+class CubicalLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing cubical persistence out of a cubical complex
+    """
+    def __init__(self, dimensions=[1], **kwargs):
```
mglisse, line 44: Is there a reason to believe that dimension 1 is a good default?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,72 @@
...
+# The parameters of the model are the pixel values.
+
+def _Cubical(X, dimensions):
+    # Parameters: X (image),
+    #             dimensions (homology dimensions)
+
+    # Compute the persistence pairs with Gudhi
+    Xs = X.shape
+    cc = CubicalComplex(dimensions=Xs, top_dimensional_cells=X.flatten())
+    cc.persistence()
```
mglisse, line 18:
```suggestion
    cc.compute_persistence()
```
since we don't use the result?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,72 @@
...
+
+    L_cofs = []
+    for dim in dimensions:
+
+        try:
+            cof = cc.cofaces_of_persistence_pairs()[0][dim]
```
mglisse, line 24: The call to `cofaces_of_persistence_pairs` is inside the loop on dimensions? Why not make it just once before the loop?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,72 @@
...
+        except IndexError:
+            cof = np.array([])
+
+        # Retrieve and ouput image indices/pixels corresponding to positive and negative simplices
+        D = len(Xs) if len(cof) > 0 else 1
+        ocof = np.array([0 for _ in range(D*2*cof.shape[0])])
```
mglisse, line 30: There is `np.zeros(n)` for that.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,72 @@
...
+
+# The parameters of the model are the pixel values.
+
+def _Cubical(X, dimensions):
+    # Parameters: X (image),
+    #             dimensions (homology dimensions)
+
+    # Compute the persistence pairs with Gudhi
+    Xs = X.shape
+    cc = CubicalComplex(dimensions=Xs, top_dimensional_cells=X.flatten())
```
mglisse, line 17: I think you don't need to flatten and pass explicit dimensions anymore, it should just do the right thing with a multidimensional numpy array.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,72 @@
...
+            cof = cc.cofaces_of_persistence_pairs()[0][dim]
+        except IndexError:
+            cof = np.array([])
+
+        # Retrieve and ouput image indices/pixels corresponding to positive and negative simplices
+        D = len(Xs) if len(cof) > 0 else 1
+        ocof = np.array([0 for _ in range(D*2*cof.shape[0])])
+        count = 0
+        for idx in range(0,2*cof.shape[0],2):
+            ocof[D*idx:D*(idx+1)]     = np.unravel_index(cof[count,0], Xs)
```
mglisse, line 33: If I understand correctly, cofaces_of_persistence_pairs returns a flat index, which you unravel, so you can index in the original tensor. Would it be possible instead to flatten the tensorflow index and use directly the flat index in it?

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Yep, you are right, it was quite easy to do so I added the possbility to ask for multiple dimensions at once.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: Thanks! Done.

src/python/gudhi/tensorflow/rips_layer.py
MathieuCarriere: Thanks! I don't think my code is more optimized than this function, so let's keep consistent and use it as you suggest.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: You are right. Again, it was easy to modify so now the code outputs both finite and essential features.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: No reason ;-) I changed to 0.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Thanks! Done.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: You are right, I put it outside the loop.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Thanks! Done.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Thanks! Done.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Good catch! Indeed there was an issue when I did a test with a non-square array, so I made sure the order is right with the option `order='F'` in `np.unravel_index`

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
mglisse: And the Rips?

src/python/gudhi/tensorflow/cubical_layer.py
mglisse: I thought it would be: if we pass an int, return the diagram in that dim, if we pass a tuple/list of dimensions, return a tuple/list of diagrams, but maybe that's overkill, always using tuple/list is indeed probably sufficient, and it is simpler :smiley:

src/python/gudhi/tensorflow/cubical_layer.py
mglisse: Actually, I was more thinking of **not** having a default (same for the other layers), so the user has to think at list 1 second what dimension they want, and so it is clearer when reading the code what dimension is used. Unless there is a good reason to have a default? For cubical specifically, [0, ..., dim-1] could be a sensible default, but that would encourage users not to prune that list when they don't use everything.

src/python/gudhi/tensorflow/cubical_layer.py
mglisse: Could you add that non-square test, to ensure we don't break it later?
What do you think of the option of not unraveling?

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,80 @@
...
+    
+    def call(self, filtration):
+        """
+        Compute lower-star persistence diagram associated to a function defined on the vertices of the simplex tree
+
+        Parameters:
+            F (TensorFlow variable): filter function values over the vertices of the simplex tree. The ith entry of F corresponds to vertex i in self.simplextree
+
+        Returns:
+            dgms (list of tuple of TensorFlow variables): list of lower-star persistence diagrams of length self.dimensions, where each element of the list is a tuple that contains the finite and essential persistence diagrams of shapes [num_finite_points, 2] and [num_essential_points, 1] respectively
```
mglisse, line 69: Ah, you prefer a 2d array of shape `[num_essential_points, 1]` to a 1d array? Ok, I'll trust you that you found it more convenient. Should we change the generator functions that sometimes return 1d arrays to match?

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: Done too!

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Ah I see, this is a good point. Do we already use collapses in the Python version of Rips?

src/python/gudhi/tensorflow/__init__.py
MathieuCarriere: Actually I wanted to be able to do `from gudhi.tensorflow import *`, but I don't know if it is possible to do `from gudhi.tensorflow import CubicalLayer` without also importing the other two. That being said, I do not really care about this `__init__` file, if there is a better, more consistent way of writing it, I'll be happy to use it.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: I agree with you. I've seen a few times people applying differentiation blindly so it makes sense to ask explicitly for dimensions. I modified the code accordingly.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Yep, I added the non-square test. I tried flattening, but since tensorflow does not use Fortran ordering, I had to transpose the matrix first. All in all, I think the code is simpler that way, so I think we should keep this version (with flat indices). Thanks Marc!

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: Mmmh, I prefer the `[num, 1]` shape for essential points mostly for tensorflow reasons (where you often have to use num by dimensions arrays, even with only one dimension). So I don't know if this is really better when not using tensorflow.

MathieuCarriere: Said otherwise, I am not against keeping generators as they are.

mglisse: ok.

src/python/gudhi/tensorflow/cubical_layer.py
mglisse: Note that there is an alternative to transposing the matrix: mirroring the dimensions. If the dimensions are `(a,b,c)` and you give CubicalComplex `(c,b,a)` and `matrix.flatten()`, it *should* also work :crossed_fingers: .

MathieuCarriere: Even better! Thanks Marc, now the only operation we do is a flatten at the beginning for initializing the CubicalComplex.

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
```diff
@@ -0,0 +1,64 @@
...
+############################################################
+
+.. include:: differentiation_sum.inc
+
+Example of gradient computed from lower-star filtration of a simplex tree
+-------------------------------------------------------------------------
+
+.. testcode::
+
+    from gudhi.tensorflow import *
```
VincentRouvreau, line 15: Maybe to be reworked according to [this discussion](https://github.com/GUDHI/gudhi-devel/pull/411#discussion_r742337781)
```suggestion
    from gudhi.tensorflow import LowerStarSimplexTreeLayer
```

src/python/doc/cubical_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,40 @@
...
+########################################
+
+.. include:: differentiation_sum.inc
+
+Example of gradient computed from cubical persistence
+-----------------------------------------------------
+
+.. testcode::
+
+    from gudhi.tensorflow import *
```
VincentRouvreau, line 15: Maybe to be reworked according to [this discussion](https://github.com/GUDHI/gudhi-devel/pull/411#discussion_r742337781)
```suggestion
    from gudhi.tensorflow import CubicalLayer
```

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,39 @@
...
+##############################################
+
+.. include:: differentiation_sum.inc
+
+Example of gradient computed from Vietoris-Rips persistence
+-----------------------------------------------------------
+
+.. testcode::
+
+    from gudhi.tensorflow import *
```
VincentRouvreau, line 15: Maybe to be reworked according to [this discussion](https://github.com/GUDHI/gudhi-devel/pull/411#discussion_r742337781)
```suggestion
    from gudhi.tensorflow import RipsLayer
```

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
MathieuCarriere: Thanks! Done.

src/python/doc/cubical_complex_tflow_itf_ref.rst
MathieuCarriere: Thanks! Done.

src/python/doc/rips_complex_tflow_itf_ref.rst
MathieuCarriere: Thanks! Done.

src/python/gudhi/tensorflow/__init__.py
mglisse: I think what you describe forces to load everything. To be able to load only some, I think it would need to be `from gudhi.tensorflow.cubical_layer import CubicalLayer` (there could still be a short `from gudhi.tensorflow.all import *` for those who want everything). I don't know if doing that would be useful, or if we should stick with loading everything, which is simpler, if a bit wasteful.

src/python/doc/cubical_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,40 @@
...
+Example of gradient computed from cubical persistence
+-----------------------------------------------------
+
+.. testcode::
+
+    from gudhi.tensorflow import CubicalLayer
+    import tensorflow as tf
+
+    X = tf.Variable([[0.,2.,2.],[2.,2.,2.],[2.,2.,1.]], dtype=tf.float32, trainable=True)
+    cl = CubicalLayer(dimension=0)
```
mglisse, line 19: The doc says `dimensions=list`, you may have forgotten to update it. I am surprised the testsuite didn't notice it, I thought it executed the testcode in the doc...

src/python/gudhi/tensorflow/rips_layer.py
```diff
@@ -0,0 +1,86 @@
...
+        Parameters:
+            maximum_edge_length (float): maximum edge length for the Rips complex 
+            dimensions (int): homology dimensions
+        """
+        super().__init__(dynamic=True, **kwargs)
+        self.max_edge = maximum_edge_length
+        self.dimensions = dimensions
+
+    def build(self):
+        super.build()
```
mglisse, line 56: Is that useful? If we remove those 2 lines, doesn't calling `build()` still work and already call the function from the base class?

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
```diff
@@ -0,0 +1,64 @@
...
+    st.insert([1]) 
+    st.insert([2]) 
+    st.insert([3]) 
+    st.insert([4]) 
+    st.insert([5]) 
+    st.insert([6]) 
+    st.insert([7]) 
+    st.insert([8]) 
+    st.insert([9]) 
+    st.insert([10])
```
mglisse, line 30: We could skip the vertex insertions, they are implicitly added by the edges below.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,80 @@
...
+    """
+    TensorFlow layer for computing lower-star persistence out of a simplex tree
+    """
+    def __init__(self, simplextree, dimensions, **kwargs):
+        """
+        Constructor for the LowerStarSimplexTreeLayer class
+  
+        Parameters:
+            simplextree (gudhi.SimplexTree): underlying simplex tree. Its vertices MUST be named with integers from 0 to n = number of vertices
+            dimensions (int): homology dimensions
```
mglisse, line 52:
```suggestion
            dimensions (List[int]): homology dimensions
```
?

src/python/doc/cubical_complex_tflow_itf_ref.rst
MathieuCarriere: Good catch, thanks. I fixed it.

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
MathieuCarriere: Thanks! Done.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: Thanks! Done.

src/python/doc/differentiation_sum.inc
```diff
@@ -0,0 +1,11 @@
...
+   :widths: 40 30 30
+   :header-rows: 0
+
+   * - :Since: GUDHI 3.5.0
+     - :License: MIT
+     - :Requires: `TensorFlow <installation.html#tensorflow>`_
+
+We provide TensorFlow 2 models that can handle automatic differentiation for the computation of persistence diagrams from complexes available in the Gudhi library. 
+This includes simplex trees, cubical complexes and Vietoris-Rips complexes. Detailed example on how to use these layers in practice are available
+in the following `notebook <https://github.com/GUDHI/TDA-tutorial/blob/master/Tuto-GUDHI-optimization.ipynb>`_.
```
mglisse, line 11: Should we mention somewhere that the computations are done on the CPU even if tensorflow is configured to use a GPU?

MathieuCarriere: Good idea! I updated the doc.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: OK I see. I've just added a `min_persistence` parameter (default zero) for gathering points on the diagonal if needed

src/python/gudhi/tensorflow/rips_layer.py
MathieuCarriere: I'm not sure how useful these lines are so I removed them, and it didn't break the code :-)

mglisse:
The CI shows several errors, and in particular while building the doc

```
Exception raised:
      File "/usr/local/miniconda/envs/gudhi_build_env/lib/python3.6/doctest.py", line 1330, in __run
        compileflags, 1), test.globs)
      File "<doctest default[0]>", line 9, in <module>
        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
      File "/Users/runner/.local/lib/python3.6/site-packages/tensorflow/python/training/tracking/data_structures.py", line 441, in __getitem__
        return self._storage[key]
    TypeError: list indices must be integers or slices, not tuple
```

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,76 @@
...
+        Returns:
+            dgms (list of TensorFlow variables): list of cubical persistence diagrams of length self.dimensions, where each element contains a finite persistence diagram of shape [num_finite_points, 2]
+        """
+        # Compute pixels associated to positive and negative simplices 
+        # Don't compute gradient for this operation
+        Xflat = tf.reshape(X, [-1])
+        Xdim = X.shape
+        indices = _Cubical(Xflat.numpy(), Xdim, self.dimensions)
+        # Get persistence diagram by simply picking the corresponding entries in the image
+        self.dgms = [tf.reshape(tf.gather(Xflat, indice), [-1,2]) for indice in indices]
```
mglisse, line 69: The names seem a bit misleading. `indice` (French for `index`?) is still a table of many indices, while `indices` is a list for all dimensions?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,76 @@
...
+        for idx_dim in range(len(self.min_persistence)):
+            min_pers = self.min_persistence[idx_dim]
+            if min_pers >= 0:
+                finite_dgm = self.dgms[idx_dim]
+                persistent_indices = np.argwhere(np.abs(finite_dgm[:,1]-finite_dgm[:,0]) > min_pers).ravel()
```
mglisse, line 74: If finite_dgm is a tensorflow array, should we use `tf.where`, etc to avoid mixing with numpy too much?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Constructor for the CubicalLayer class
+
+        Parameters:
+            dimensions (List[int]): homology dimensions
+            min_persistence (List[float]): minimum distance-to-diagonal of the points in the output persistence diagrams (default None, in which case 0. is used for all dimensions)
+        """
+        super().__init__(dynamic=True, **kwargs)
+        self.dimensions = dimensions
+        self.min_persistence = min_persistence if min_persistence != None else [0. for _ in range(len(self.dimensions))]
```
mglisse, line 50: As Theo says, `[0.] * len(self.dimensions)` is shorter and seems quite readable. Although if you don't find it readable, I don't mind if you keep it as you wrote it here.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,85 @@
...
+
+# The parameters of the model are the vertex function values of the simplex tree.
+
+def _LowerStarSimplexTree(simplextree, filtration, dimensions):
+    # Parameters: simplextree (simplex tree on which to compute persistence)
+    #             filtration (function values on the vertices of st),
+    #             dimensions (homology dimensions),
+    
+    for s,_ in simplextree.get_filtration():
+        simplextree.assign_filtration(s, -1e10)
```
mglisse, line 16: I don't like this kind of magical value. Use -infinity or NaN?

mglisse, line 16: We have reset_filtration if you want to assign the same value to all simplices. get_simplices is cheaper than get_filtration if you don't care about the order.

src/python/gudhi/tensorflow/rips_layer.py
```diff
@@ -0,0 +1,91 @@
...
+        Compute Rips persistence diagram associated to a point cloud
+
+        Parameters:   
+            X (TensorFlow variable): point cloud of shape [number of points, number of dimensions]
+
+        Returns:
+            dgms (list of tuple of TensorFlow variables): list of Rips persistence diagrams of length self.dimensions, where each element of the list is a tuple that contains the finite and essential persistence diagrams of shapes [num_finite_points, 2] and [num_essential_points, 1] respectively
+        """    
+        # Compute distance matrix
+        DX = tf.math.sqrt(tf.reduce_sum((tf.expand_dims(X, 1)-tf.expand_dims(X, 0))**2, 2))
```
mglisse, line 69: I think we could use `tf.norm` or `reduce_euclidean_norm` with a suitable `axis` argument, which would be slightly cheaper and might eventually behave better when 2 points coincide. Sadly, it looks like tensorflow explicitly rejected the idea of providing a function like cdist.

mglisse (commented):
There seems to be some mess with the submodule, it doesn't matter for the review, but we will need to fix it before merging (can be done by Vincent or me).

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,39 @@
...
+.. testcode::
+
+    from gudhi.tensorflow import RipsLayer
+    import tensorflow as tf
+
+    X = tf.Variable([[1.,1.],[2.,2.]], dtype=tf.float32, trainable=True)
+    rl = RipsLayer(maximum_edge_length=2., dimensions=[0])
+
+    with tf.GradientTape() as tape:
+        dgm = rl.call(X)
```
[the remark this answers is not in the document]
MathieuCarriere: Thanks for catching this. It is actually `rl.call(X)[0][0]` since the output is a list of list (first list is dimension, second list is [finite, essential]). I've fixed it.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: You are right, I changed it to `for indices in list_indices`.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: I agree, it is better to make things more consistent. I've modified the code.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: I'm perfectly fine with this. I've modified the code.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
MathieuCarriere: Just updated to `-np.infty`.

MathieuCarriere: Cool, I didn't even know about `reset_filtration`! I've modified the code.

src/python/gudhi/tensorflow/rips_layer.py
MathieuCarriere: Yep, just switched to `tf.norm`, which is also a bit cleaner. Thanks.

src/python/gudhi/tensorflow/rips_layer.py
```diff
@@ -0,0 +1,91 @@
...
+
+        L_indices.append((finite_indices, essential_indices))
+
+    return L_indices
+
+class RipsLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing Rips persistence out of a point cloud
+    """
+    def __init__(self, dimensions, maximum_edge_length=12, min_persistence=None, **kwargs):
```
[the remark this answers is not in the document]
MathieuCarriere: Just switched to `np.infty`.

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,39 @@
...
+        dgm = rl.call(X)[0][0]
+        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
+
+    grads = tape.gradient(loss, [X])
+    print(grads[0].numpy())
+
+.. testoutput::
+
+    [[-0.5 -0.5]
+     [ 0.5  0.5]]
```
mglisse, line 31: Running it on my computer (tensorflow 2.8.0 if it matters), I get a test failure

    Expected:
        [[-0.5 -0.5]
         [ 0.5  0.5]]
    Got:
        [[-0.49999997 -0.49999997]
         [ 0.49999997  0.49999997]]

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
```diff
@@ -0,0 +1,53 @@
...
+        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
+
+    grads = tape.gradient(loss, [F])
+    print(grads[0].indices.numpy())
+    print(grads[0].values.numpy())
+
+.. testoutput::
+
+    [2 4]
+    [-1. 1.]
```
mglisse, line 45: On my computer

    Expected:
        [2 4]
        [-1. 1.]
    Got:
        [2 4]
        [-1.  1.]

(Is it complaining about the number of spaces between `-1.` and `1.`?)

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def call(self, X):
+        """
+        Compute persistence diagram associated to a cubical complex filtered by some pixel values 
+
+        Parameters:
+            X (TensorFlow variable): pixel values of the cubical complex
+
+        Returns:
+            dgms (list of TensorFlow variables): list of cubical persistence diagrams of length self.dimensions, where each element contains a finite persistence diagram of shape [num_finite_points, 2]
```
mglisse, line 61: I am getting a shape `(n,1,2)`, not the `(n,2)` that the doc claims.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,76 @@
...
+        except IndexError:
+            cof = np.array([])
+
+        L_cofs.append(np.array(cof, dtype=np.int32))
+
+    return L_cofs
+
+class CubicalLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing cubical persistence out of a cubical complex
```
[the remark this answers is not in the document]
MathieuCarriere: You are right, thanks. I updated the doc.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,84 @@
...
+        # Don't try to compute gradients for the vertex pairs
+        indices = _LowerStarSimplexTree(self.simplextree, filtration.numpy(), self.dimensions)
+        # Get persistence diagrams
+        self.dgms = []
+        for idx_dim, dimension in enumerate(self.dimensions):
+            finite_dgm = tf.reshape(tf.gather(filtration, indices[idx_dim][0]), [-1,2])
+            essential_dgm = tf.reshape(tf.gather(filtration, indices[idx_dim][1]), [-1,1])
+            min_pers = self.min_persistence[idx_dim]
+            if min_pers >= 0:
+                persistent_indices = np.argwhere(np.abs(finite_dgm[:,1]-finite_dgm[:,0]) > min_pers).ravel()
```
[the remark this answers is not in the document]
MathieuCarriere: My bad, I don't remember why, but I did it only for cubical. I added it to simplex tree and rips as well.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,84 @@
...
+class LowerStarSimplexTreeLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing lower-star persistence out of a simplex tree
+    """
+    def __init__(self, simplextree, dimensions, min_persistence=None, **kwargs):
+        """
+        Constructor for the LowerStarSimplexTreeLayer class
+  
+        Parameters:
+            simplextree (gudhi.SimplexTree): underlying simplex tree. Its vertices MUST be named with integers from 0 to n = number of vertices
```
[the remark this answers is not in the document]
MathieuCarriere: Yep, just updated the doc. As for the parallel computation, it will indeed be a problem, even though I never had to process 2 inputs in parallel. Maybe we should mention it in the doc?

src/python/doc/rips_complex_tflow_itf_ref.rst
MathieuCarriere: Argh, I updated to TF 2.8.0 but still get
    [[-0.5 -0.5]
     [ 0.5  0.5]]
on my machine. I don't know where this could come from...

src/python/doc/ls_simplex_tree_tflow_itf_ref.rst
MathieuCarriere: I added the space just in case ;-)

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: Yep, there was a small bug in the code, the shape should be OK now.

MathieuCarriere: Thanks for catching this.

src/python/doc/rips_complex_tflow_itf_ref.rst
mglisse: Maybe we could add, as hidden code so it doesn't show in the generated html, `import numpy` and `numpy.set_printoptions(precision=4)` (or something equivalent).

MathieuCarriere: Do you mean something like:

`.. testcode::`
`    :hide:`
`    import numpy`
`    numpy.set_printoptions(precision=4)`

`.. testcode::`

`    from gudhi.tensorflow import RipsLayer`
`    import tensorflow as tf`
`[...]`

?

mglisse: Either that or `testsetup`, yes. But I haven't tried it...

MathieuCarriere: OK, just made a push with `testsetup`. I hope this will work.

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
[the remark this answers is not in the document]
mglisse: I was mostly wondering if this is something that tensorflow may decide to do automatically, to make the computation faster, or if it can only happen if a user explicitly asks for it (in that case the warning you already added may be sufficient). There is also what happens say with clone_model, or when tensorflow tries to serialize a model (does it require to make the simplex tree pickle-compatible?), etc.
Maybe that's all fine, or at least can be ignored for now, I am not very familiar with tensorflow so I wanted to point it out in case you later find out that it is an issue.

src/python/doc/rips_complex_tflow_itf_ref.rst
```diff
@@ -0,0 +1,44 @@
...
+TensorFlow layer for Vietoris-Rips persistence
+##############################################
+
+.. include:: differentiation_sum.inc
+
+Example of gradient computed from Vietoris-Rips persistence
+-----------------------------------------------------------
+
+.. testsetup::
+    :hide:
```
mglisse, line 14: `:hide:` is invalid inside `testsetup`, and unnecessary because the code in `testsetup` is not shown anyway.

src/python/doc/rips_complex_tflow_itf_ref.rst
mglisse: This seems to cause random (they don't happen at every run :confused: ) errors in other test code, for instance
```
File "representations.rst", line 45, in default
Failed example:
    import numpy as np
    from gudhi.representations import Landscape
    # A single diagram with 4 points
    D = np.array([[0.,4.],[1.,2.],[3.,8.],[6.,8.]])
    diags = [D]
    l=Landscape(num_landscapes=2,resolution=10).fit_transform(diags)
    print(l)
Expected:
    [[1.02851895 2.05703791 2.57129739 1.54277843 0.89995409 1.92847304
      2.95699199 3.08555686 2.05703791 1.02851895 0.         0.64282435
      0.         0.         0.51425948 0.         0.         0.
      0.77138922 1.02851895]]
Got:
    [[1.0285 2.057  2.5713 1.5428 0.9    1.9285 2.957  3.0856 2.057  1.0285
      0.     0.6428 0.     0.     0.5143 0.     0.     0.     0.7714 1.0285]]
```
Maybe it also requires a `testcleanup` to reset the precision to its default value (8?) after the tensorflow testcode, so it doesn't interfere with the rest.

MathieuCarriere: Like this:
`[...]`
`     print(grads[0].numpy())`
 
`.. testcleanup::`

` .. testoutput::`
 
`     [[-0.5 -0.5]`
`[...]`
?

mglisse: With `numpy.set_printoptions(precision=8)` (or a better way to reset things if there is) in the testcleanup block, yes.
It feels like a hack, but I don't have many other ideas...

MathieuCarriere: OK thanks, just made a new push.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def call(self, X):
+        """
+        Compute persistence diagram associated to a cubical complex filtered by some pixel values 
+
+        Parameters:
+            X (TensorFlow variable): pixel values of the cubical complex
+
+        Returns:
+            dgms (list of TensorFlow variables): list of cubical persistence diagrams of length self.dimensions, where each element contains a finite persistence diagram of shape [num_finite_points, 2]
```
mglisse, line 61: I was going to say that this doesn't match what the other functions return, but actually the code outputs a list of arrays of shape (1,n,2). Is that meant as more or less equivalent to a list of tuples of arrays of shape (n,2) where the tuples have a single element? Anyway, that output is not obvious from the doc.

mglisse: @MathieuCarriere IIRC this is the only point blocking the merge of this branch, could you take a look?

MathieuCarriere:
My bad, this completely slipped off my mind. Do you still have the same issue after the last commit?

mglisse:
> My bad, this completely slipped off my mind. Do you still have the same issue after the last commit?

After the last commit, test_cubical_diff and test_nonsquare_cubical_diff don't pass, as you can see.

```log
______________________________ test_cubical_diff _______________________________

    def test_cubical_diff():
    
        Xinit = np.array([[0.,2.,2.],[2.,2.,2.],[2.,2.,1.]], dtype=np.float32)
        X = tf.Variable(initial_value=Xinit, trainable=True)
        cl = CubicalLayer(dimensions=[0])
    
        with tf.GradientTape() as tape:
            dgm = cl.call(X)[0][0]
>           loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))

../../python/test/test_diff.py:26: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/glisse/.local/lib/python3.9/site-packages/tensorflow/python/util/traceback_utils.py:153: in error_handler
    raise e.with_traceback(filtered_tb) from None
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

e = _NotOkStatusException(), name = 'strided_slice/'

    def raise_from_not_ok_status(e, name):
      e.message += (" name: " + name if name is not None else "")
>     raise core._status_to_exception(e) from None  # pylint: disable=protected-access
E     tensorflow.python.framework.errors_impl.InvalidArgumentError: Index out of range using input dim 1; input has only 1 dims [Op:StridedSlice] name: strided_slice/

/home/glisse/.local/lib/python3.9/site-packages/tensorflow/python/framework/ops.py:7186: InvalidArgumentError
----------------------------- Captured stderr call -----------------------------
2022-04-15 18:50:22.206541: W tensorflow/core/framework/op_kernel.cc:1745] OP_REQUIRES failed at strided_slice_op.cc:108 : INVALID_ARGUMENT: Index out of range using input dim 1; input has only 1 dims
```

Also, there is something wrong with the way you merge master, you downgraded the submodule again. Maybe you forget to run `git submodule update` after pulling a new version of master? I'll fix it just before the merge, but in case you are using some script to merge master into your branches, you may want to add that command somewhere in there.

src/python/gudhi/tensorflow/cubical_layer.py
mglisse: Now the documentation matches the code, that's good. On the other hand, the output for cubical is different from the output for Rips or lower-star, it requires a different number of `[0]` to extract the finite diagram.
* We can consider that this is not a problem, people won't often reuse the same code for the diagrams coming from Rips and Cubical, and even if they do the adaptation is not that bad. And since we don't output essential features for CubicalLayer it would be strange to leave an empty space for it.
* Or we can wrap the diagrams in a tuple of size 1.

That's something for you, @tlacombe and @hensel-f to decide, I am ok with both, I just want to make sure the choice is made consciously.

src/python/test/test_diff.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Xinit = np.array([[1.,1.],[2.,2.]], dtype=np.float32)
+    X = tf.Variable(initial_value=Xinit, trainable=True)
+    rl = RipsLayer(maximum_edge_length=2., dimensions=[0])
+
+    with tf.GradientTape() as tape:
+        dgm = rl.call(X)[0][0]
+        loss = tf.math.reduce_sum(tf.square(.5*(dgm[:,1]-dgm[:,0])))
+    grads = tape.gradient(loss, [X])
+    assert np.abs(grads[0].numpy()-np.array([[-.5,-.5],[.5,.5]])).sum() <= 1e-6
```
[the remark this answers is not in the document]
MathieuCarriere: Yes, I agree, I updated the tests accordingly.

src/python/gudhi/tensorflow/cubical_layer.py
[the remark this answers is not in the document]
MathieuCarriere: You are right, I modified the doc and tried to be as explicit as possible.

MathieuCarriere:
Yes I agree with you about the difference between the Cubical layer output and the two other layer outputs. I think providing the essential diagram for Cubical makes sense since it is actually not empty in dimension 0, but since the only information it contains is the minimal value of the complex, I chose to not output it. However, if anyone thinks it is a better option, I'll be happy to change my code.

tlacombe:
Hello, 
I would personally tip the scale in favor of outputing object with the same shape, as it may happen that a user copy-paste some code s·he wrote for Rips dgm and use it for cubical complexes without reading the doc (thinking "this should be the same") ; having something not working may be frustrating.

However I agree with Felix, the most important thing is to document it accordingly.

MathieuCarriere:
OK I'll quickly change the output then. Thanks for the comments!

MathieuCarriere:
All right, I changed the output, now it should be consistent with the two other functions

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,80 @@
...
+class CubicalLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing the persistent homology of a cubical complex
+    """
+    def __init__(self, dimensions, min_persistence=None, **kwargs):
+        """
+        Constructor for the CubicalLayer class
+
+        Parameters:
+            dimensions (List[int]): homology dimensions
```
mglisse, line 45: I now notice that `CubicalComplex.__init__` also has a parameter called `dimensions` but with a completely different meaning (the shape of the array). Should we use something more explicit like `homology_dimensions`, or do we consider that it is ok with just `dimensions` and hopefully people won't get too confused?

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def call(self, X):
+        """
+        Compute persistence diagram associated to a cubical complex filtered by some pixel values 
+
+        Parameters:
+            X (TensorFlow variable): pixel values of the cubical complex
+
+        Returns:
+            dgms (list of tuple of TensorFlow variables): list of cubical persistence diagrams. The length of this list is the same than that of dimensions, i.e., there is one persistence diagram per homology dimension provided in the input list dimensions. Moreover, the finite and essential parts of the persistence diagrams are provided separately: each element of this list is a tuple of size two that contains the finite and essential parts of the corresponding persistence diagram, of shapes [num_finite_points, 2] and [num_essential_points, 1] respectively. Note that the essential part is always empty in cubical persistence diagrams, except in homology dimension zero, where the essential part always contains a single point, with abscissa equal to the smallest value in the complex, and infinite ordinate
```
mglisse, line 61: The doc currently shows
> Return type: dgms (list of tuple of TensorFlow variables)

so I think it should be
```suggestion
            list of tuple of TensorFlow variables: list of cubical persistence diagrams. The length of this list is the same than that of dimensions, i.e., there is one persistence diagram per homology dimension provided in the input list dimensions. Moreover, the finite and essential parts of the persistence diagrams are provided separately: each element of this list is a tuple of size two that contains the finite and essential parts of the corresponding persistence diagram, of shapes [num_finite_points, 2] and [num_essential_points, 1] respectively. Note that the essential part is always empty in cubical persistence diagrams, except in homology dimension zero, where the essential part always contains a single point, with abscissa equal to the smallest value in the complex, and infinite ordinate
```
(or something like `List[Tuple[tf.Tensor,tf.Tensor]]`)

MathieuCarriere: Good catch, thank you!!

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: That makes sense. I am a bit more in favor of `homology_dimensions` though. If @VincentRouvreau is OK with it, I'll update the code.

mglisse: `homology_dimensions` works for me.
In the sklearn interface, @VincentRouvreau also accepts an int when we want to specify a single dimension, so that would read `homology_dimensions=2`, I hope it is clear enough that we are not asking for 2 unspecified homology groups in this case but we only want homology in dimension 2 :crossed_fingers:

src/python/gudhi/tensorflow/lower_star_simplex_tree_layer.py
```diff
@@ -0,0 +1,84 @@
...
+class LowerStarSimplexTreeLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing lower-star persistence out of a simplex tree
+    """
+    def __init__(self, simplextree, dimensions, min_persistence=None, **kwargs):
+        """
+        Constructor for the LowerStarSimplexTreeLayer class
+  
+        Parameters:
+            simplextree (gudhi.SimplexTree): underlying simplex tree. Its vertices MUST be named with integers from 0 to n = number of vertices. Note that its filtration values are modified in each call of the class.
```
mglisse, line 50: From 0 to n, there are n+1 numbers :wink:

mglisse:
Once the name for dimensions/persistence_dimension/homology_dimensions is settled, this is ok to merge, thanks.

src/python/gudhi/tensorflow/cubical_layer.py
```diff
@@ -0,0 +1,80 @@
...
+        except IndexError:
+            cof = np.array([])
+
+        L_cofs.append(np.array(cof, dtype=np.int32))
+
+    return L_cofs
+
+class CubicalLayer(tf.keras.layers.Layer):
+    """
+    TensorFlow layer for computing the persistent homology of a cubical complex
```
mglisse, line 38: We could mention somewhere what field we are working on, possibly even let users specify it, but that can be done later.

MathieuCarriere: I added it. Since this parameter has a default value, it should not break examples and tests.

src/python/gudhi/tensorflow/rips_layer.py
```diff
@@ -8,15 +8,15 @@
 
 # The parameters of the model are the point coordinates.
 
-def _Rips(DX, max_edge, dimensions):
+def _Rips(DX, max_edge, dimensions, homology_coeff_field=11):
```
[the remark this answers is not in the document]
MathieuCarriere: yes you are right, I removed it.

src/python/gudhi/tensorflow/cubical_layer.py
MathieuCarriere: We can by even more explicit with `homology_dimension_list`, but maybe this is too much?

VincentRouvreau: @MathieuCarriere `homology_dimension_list` may be too restrictive as my [interface](https://github.com/VincentRouvreau/gudhi-devel/blob/8d7244b510eee0d7927c521117198ef286028acf/src/python/gudhi/sklearn/cubical_persistence.py#L50-L52) can accept an `int or a list of int`. When only one dimension is required, it is quite useful and simplifies the code.

MathieuCarriere: OK I'll stick with `homology_dimensions` then

PR #418, opened 2020-11-03
Do not install compiled examples

### LINKED ISSUE #415: make install installs compiled examples in bin/
If I use cmake with `-DWITH_GUDHI_EXAMPLE=ON`, `make install` installs in bin/ not just the utilities, but also the examples. I don't think that makes sense. If we really want to install the examples, it should be the source files that we install, since the whole point of the examples is to read their code.

### LINKED ISSUE #416: make install fails on 'introduction.rst'
With a user version, if I run cmake+make+make install, I get

<pre>
-- Installing: /tmp/j/bin/GudhUI
/usr/lib/python3.8/distutils/extension.py:131: UserWarning: Unknown Extension options: 'cython_directives'
  warnings.warn(msg)
  File "/home/glisse/repos/gudhi3/build/2/build/python/setup.py", line 68, in <module>
    with open("introduction.rst", "r") as fh:
FileNotFoundError: [Errno 2] No such file or directory: 'introduction.rst'
</pre>

VincentRouvreau:
* Fix #415 by removing all install cmake targets from examples


mglisse (approved):
I tried again with `make install` but couldn't make it work. First it failed with "permission denied" (it doesn't obey CMAKE_INSTALL_PREFIX or DESTDIR and tries to install to the system and not the user). Then I temporarily gave it permission to write to /usr/local/lib/python3.8/dist-packages, after which it decided to rebuild the whole plugin (does it want to hardcode the location or something?), and it installed something that looks ok, although it creates stub .py to load the .so, which we don't have for instance in the pip package. And it still finally fails
<pre>
Processing dependencies for gudhi==3.4.0a0
Searching for numpy==1.19.3
Best match: numpy 1.19.3
Adding numpy 1.19.3 to easy-install.pth file
Installing f2py script to /usr/local/bin
error: [Errno 13] Permission denied: '/usr/local/bin/f2py'
</pre>
(I already have numpy 1.19.3 installed, and /usr/bin/f2py3)

Maybe doing a general `make install` shouldn't recurse in python/ but simply print a comment (if the plugin was built) giving a hint that the plugin was not installed, and to install it you should do this or that.

VincentRouvreau:
`make install` only works in the classic way to install it (sometimes requires `--user` but I am not sure - maybe we should add it).
I also tried as explained [here](https://docs.python.org/3/install/index.html#alternate-installation-the-home-scheme) with `--home=` and  `--prefix=` but you need to install it in a python path (must be in PYTHONPATH, must support .pth files, ...)
I can remove this change.

mglisse:
> `make install` only works in the classic way to install it (sometimes requires `--user` but I am not sure - maybe we should add it).
> I also tried as explained [here](https://docs.python.org/3/install/index.html#alternate-installation-the-home-scheme) with `--home=` and `--prefix=` but you need to install it in a python path (must be in PYTHONPATH, must support .pth files, ...)

If `make install` is going to fail on the python plugin for too many users, maybe it shouldn't even try. It could stick to installing headers / utilities, and print a link to the doc explaining how to install the python plugin. I guess when we added the plugin to the installation target, I didn't realize it was so hard to know where to install it :cry:
I welcome other opinions on the subject...

VincentRouvreau:
I reverted #416 fix.

mglisse:
> I didn't realize it was so hard to know where to install it :cry:

Or I guess I should say: I thought that trying to install it in some random place would give you something that python cannot find without PYTHONPATH, but that it would at least succeed to copy the files to that random location. Python works differently from what I am used to...

PR #424, opened 2020-11-05
Test insertion of simplex with NaN filtration value

mglisse:
This patch documents (and tests) that inserting simplices with filtration value NaN does not modify the value for existing simplices.
Now to discuss whether that's actually the behavior we want :wink:

### DIFF
--- a/src/python/gudhi/simplex_tree.pyx
+++ b/src/python/gudhi/simplex_tree.pyx
@@ -184,6 +184,11 @@ cdef class SimplexTree:
         simplices are already present with a higher filtration value, their
         filtration value is lowered.
 
+        .. note::
+
+            Inserting a simplex with filtration value `math.nan` does not
+            modify the filtration value of any simplex already present.
+
         :param simplex: The N-simplex to insert, represented by a list of
             vertex.
         :type simplex: list of int
@@ -367,6 +372,11 @@ cdef class SimplexTree:
     def make_filtration_non_decreasing(self):
         """This function ensures that each simplex has a higher filtration
         value than its faces by increasing the filtration values.
+
+        .. note::
+
+            For this function, a value of NaN for a simplex of dimension 1 or more
+            is considered smaller than any other value.
 
         :returns: True if any filtration value was modified,
             False if the filtration was already non-decreasing.
--- a/src/python/test/test_simplex_tree.py
+++ b/src/python/test/test_simplex_tree.py
@@ -10,6 +10,7 @@
 
 from gudhi import SimplexTree
 import pytest
+import math
 
 __author__ = "Vincent Rouvreau"
 __copyright__ = "Copyright (C) 2016 Inria"
@@ -399,3 +400,17 @@ def test_boundaries_iterator():
 
     with pytest.raises(RuntimeError):
         list(st.get_boundaries([6])) # (6) does not exist
+
+# insert with NaN must not modify filtration values
+# make_filtration_non_decreasing considers NaN small
+def test_nan():
+    st = SimplexTree()
+    st.insert([1], 2.)
+    st.insert([4], 5.)
+    st.insert([2], -1.)
+    st.insert([1, 2, 4], math.nan)
+    assert math.isnan(st.filtration([2, 4]))
+    st.make_filtration_non_decreasing()
+    assert st.filtration([1, 2, 4]) == 5.
+    assert st.filtration([1, 2]) == 2.
+    assert st.filtration([2]) == -1.
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -763,8 +763,10 @@ class Simplex_tree {
     return insert_vertex_vector(copy, filtration);
   }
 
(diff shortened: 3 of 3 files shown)


src/python/test/test_simplex_tree.py
```diff
@@ -399,3 +400,17 @@ def test_boundaries_iterator():
...
+
+# insert with NaN must not modify filtration values
+# make_filtration_non_decreasing considers NaN small
+def test_nan():
+    st = SimplexTree()
+    st.insert([1], 2.)
+    st.insert([4], 5.)
+    st.insert([2], -1.)
+    st.insert([1, 2, 4], math.nan)
+    assert math.isnan(st.filtration([2, 4]))
```
tlacombe, line 412:
```suggestion
    assert math.isnan(st.filtration([2, 4]))
    assert st.filtration([4]) == 5.
    assert st.filtration([2]) == -1.
    assert st.filtration([1]) == 2.
    assert st.filtration([1,2]) == 2.
```
Perhaps we may want to make sure that we did not modify any of the previous values even before calling `st.make_persistence_non_decreasing()`.

src/python/test/test_simplex_tree.py
```diff
@@ -399,3 +400,17 @@ def test_boundaries_iterator():
...
     with pytest.raises(RuntimeError):
         list(st.get_boundaries([6])) # (6) does not exist
+
+# insert with NaN must not modify filtration values
+# make_filtration_non_decreasing considers NaN small
+def test_nan():
+    st = SimplexTree()
+    st.insert([1], 2.)
+    st.insert([4], 5.)
+    st.insert([2], -1.)
```
tlacombe, line 410:
```suggestion
    st.insert([2], -1.)
    st.insert([1,2], 2.)
```
Perhaps already add a single edge (with an admissible value) to make sure that both inserting `nan` and `make_filtration_non_decreasing()` do not play any role there

src/python/gudhi/simplex_tree.pyx
```diff
@@ -184,6 +184,11 @@ cdef class SimplexTree:
         simplices are already present with a higher filtration value, their
         filtration value is lowered.
 
+        .. note::
+
+            Inserting a simplex with filtration value `math.nan` does not
+            modify the filtration value of any simplex already present.
```
tlacombe, line 190: Perhaps complete this note by adding a warning in the vein of what is done for `assign_filtration()` saying (e.g.) "In this particular case, the result will not be a valid filtration anymore. Callers are responsible for fixing this (using `assign_filtration()` or `make_filtration_non_decreasing()` for instance) before calling any function that relies on the filtration property, like `persistence()`.

tlacombe (approved):
From my (biased) perspective, it seems to be a nice addition. Perhaps it is almost useless (I don't really know how people use gudhi in general), but when you typically want to build a lower start filtration, starting from a complex + values on its vertices, this might simplify things a bit.

src/python/test/test_simplex_tree.py
mglisse: I thought that checking after make_persistence_non_decreasing would be sufficient (if we overwrite a value with insert, make_persistence_non_decreasing won't be able to reinvent that value), but why not, done.

src/python/test/test_simplex_tree.py
mglisse: I inserted an extra edge.

src/python/gudhi/simplex_tree.pyx
mglisse: I hadn't realized it, but having some NaN in the complex can have some weird effects. In particular, if I have vertex 0 with value NaN, inserting edge [0, 1] with value 42 does not change the value of vertex 0. I added some confusing text to the doc. We could change that behavior, but would need to check that it doesn't slow things down for normal operations.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -184,6 +184,15 @@ cdef class SimplexTree:
...
+            However, it produces an invalid filtration, which needs to be fixed
+            with :meth:`make_filtration_non_decreasing` or
+            :meth:`assign_filtration`. Inserting simplices with real filtration
+            value may or may not update pre-existing NaN.
```
VincentRouvreau, line 194: It seems to me that inserting a simplices will never update pre-existing NaN.
```python
st.insert([0,1], math.nan)  # return True, all is NaN
st.insert([0,1], 1.)  # return False
list(st.get_filtration())  # returns:  [([0, 1], nan), ([0], nan), ([1], nan)]
st.insert([1,2], 1.)  # return True
list(st.get_filtration())  # returns: [([0, 1], nan), ([0], nan), ([1], nan), ([2], 1.0), ([1, 2], 1.0)]
```

mglisse: Yes, but I don't think that's desirable behavior, so I don't want to document a promise for that behavior if we are likely to break it later.

VincentRouvreau:
It would be nice to test it at C++ level. Here is a [gist](https://gist.github.com/VincentRouvreau/abd4b02f10fc1f33047de2d1019cb25c) that translates the python test in C++.

PR #427, opened 2020-11-17
Support for irregularly sampled Betti curves

gspr:
I recently noticed that GUDHI seems to lack support for computing Betti curves without choosing a sampling grid. I therefore adapted the code I personally use, to (hopefully) fit into the GUDHI way of doing things.

The commit messages explain more of the rationale. However, be aware that the current `BettiCurve` class and the new `IrregularBettiCurve` one do have different semantics for intervals not contained in the range in question. But for intervals that are contained in the range, the old class' `transform` method can be seen as performing that of the new one followed by the appropriate previous-neighbor interpolation on a fixed grid (see tests added in 9fb88b5b3edd6235a166ca2aa20bd691f3b72229). Preliminary benchmarks suggest that the new class is faster than the old, so it may be possible to simplify the old as interpolation of the output of the new.

### DIFF
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -1,11 +1,12 @@
 # This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
 # See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
-# Author(s):       Mathieu Carrière, Martin Royer
+# Author(s):       Mathieu Carrière, Martin Royer, Gard Spreemann
 #
 # Copyright (C) 2018-2020 Inria
 #
 # Modification(s):
 #   - 2020/06 Martin: ATOL integration
+#   - 2020/11 Gard: Irregular (non-evenly sampled) Betti curves.
 
 import numpy as np
 from sklearn.base          import BaseEstimator, TransformerMixin
@@ -349,6 +350,92 @@ class BettiCurve(BaseEstimator, TransformerMixin):
             numpy array with shape (**resolution**): output Betti curve.
         """
         return self.fit_transform([diag])[0,:]
+
+
+class IrregularBettiCurve(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing Betti curves from a list of persistence diagrams. Unlike the BettiCurve class, the curves are not sampled at fixed, regular points. Each curve has all its data intact.
+    """
+
+    def __init__(self, filtration_range = [np.nan, np.nan]):
+        """
+        Constructor for the IrregularBettiCurve class.
+
+        Parameters:
+            filtration_range ([float, float]): minimum and maximum filtration range to consider, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). The behavior is not the same as for sample_range in BettiCurve. In particular, intervals born before x_min are not considered. Intervals born in [x_min, x_max) contribute to the Betti curves as if they were infinite.
+        """
+        self.filtration_range = np.array(filtration_range)
+
+    def fit(self, X, y = None):
+        """
+        See the corresponding function in BettiCurve.
+        """
+        if np.isnan(self.filtration_range).any():
+            pre = DiagramScaler(use=True, scalers=[([0], MinMaxScaler()), ([1], MinMaxScaler())]).fit(X,y)
+            [mx,my],[Mx,My] = [pre.scalers[0][1].data_min_[0], pre.scalers[1][1].data_min_[0]], [pre.scalers[0][1].data_max_[0], pre.scalers[1][1].data_max_[0]]
+            self.filtration_range = np.where(np.isnan(self.filtration_range), np.array([mx, My]), np.array(self.filtration_range))
+
+        return self
+
+    def transform(self, X):
+        """
+        Compute the Betti curve for each persistence diagram.
+
+        Parameters:
+            X (list of n x 2 NumPy arrays): input persistence diagrams.
+        
+        Returns:
+            Pair (xs, bettis), where xs is a length-k NumPy array with all filtration values for which any of the Betti curves change value, and bettis is an (number of diagrams) x k NumPy integer array with the values of the k Betti curves at the corresponding filtration values. The value bettis[i, k] is the value of the Betti curve of the k'th diagram on the interval [xs[i], xs[i+1]).
+        """
+        assert(not np.isnan(self.filtration_range).any())
+
+        N = len(X)
+        cropped_pds = [pd[pd[:, 0] >= self.filtration_range[0]] for pd in X]
+
+        events = np.concatenate([pd.flatten(order="F") for pd in cropped_pds], axis=0)
+        sorting = np.argsort(events)
+        offsets = np.zeros(1 + N, dtype=int)
+        for i in range(0, N):
+            offsets[i+1] = offsets[i] + 2*cropped_pds[i].shape[0]
+        starts = offsets[0:N]
+        ends = offsets[1:N + 1] - 1
+
+        xs = [self.filtration_range[0]]
+        bettis = [[0] for i in range(0, N)]
+
+        for i in sorting:
+            if events[i] >= self.filtration_range[1]:
+                break
+            j = np.searchsorted(ends, i)
+            delta = 1 if i - starts[j] < len(cropped_pds[j]) else -1
+            if events[i] == xs[-1]:
+                bettis[j][-1] += delta
+            else:
+                xs.append(events[i])
+                for k in range(0, j):
+                    bettis[k].append(bettis[k][-1])
+                bettis[j].append(bettis[j][-1] + delta)
+                for k in range(j+1, N):
+                    bettis[k].append(bettis[k][-1])
+
+        xs = np.array(xs)
+        bettis = np.array(bettis, dtype=int)
+
+        return(xs, bettis)
+
+    def __call__(self, diag):
+        """
+        Apply IrregularBettiCurve on a single persistence diagram and outputs the result.
+
+        Parameters:
+            diag (n x 2 numpy array): input persistence diagram.
+
+        Returns:
+            Pair (xs, b) of 1D NumPy arrays, xs containing the filtration values and b containing the Betti curve values. The Betti curve takes the value b[i] on the interval [xs[i], xs[i+1]).
+        """
+        res = self.fit_transform([diag])
+        return (res[0], res[1][0, :])
+
 
 class Entropy(BaseEstimator, TransformerMixin):
     """
--- a/src/python/test/test_representations.py
+++ b/src/python/test/test_representations.py
@@ -2,6 +2,7 @@ import os
 import sys
 import matplotlib.pyplot as plt
 import numpy as np
+import scipy.interpolate
 import pytest
 
 from sklearn.cluster import KMeans
@@ -63,7 +64,7 @@ def test_dummy_atol():
             atol_vectoriser.transform(X=[a, b, c])
 
 
-from gudhi.representations.vector_methods import BettiCurve
+from gudhi.representations.vector_methods import BettiCurve, IrregularBettiCurve
 
 
 def test_infinity():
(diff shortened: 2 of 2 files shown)


gspr:
The CI failures on Windows look unrelated to me.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
...
+        Parameters:
+            filtration_range ([float, float]): minimum and maximum filtration range to consider, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). The behavior is not the same as for sample_range in BettiCurve. In particular, intervals born before x_min are not considered. Intervals born in [x_min, x_max) contribute to the Betti curves as if they were infinite.
+        """
+        self.filtration_range = np.array(filtration_range)
+
+    def fit(self, X, y = None):
+        """
+        See the corresponding function in BettiCurve.
+        """
+        if np.isnan(self.filtration_range).any():
```
VincentRouvreau, line 373: If any NaN, `fit` will do nothing, but `transform` will assert. This is an open question also for @MathieuCarriere , but I am not sure if we should homogenize the behaviour (always assert or always do nothing, or even an exception).

gspr:
Vincent Rouvreau <notifications@github.com> writes:

> @VincentRouvreau commented on this pull request.
>
>
>
>> @@ -2,6 +2,7 @@
>  import sys
>  import matplotlib.pyplot as plt
>  import numpy as np
> +import scipy.interpolate
>
> As you import a new module that is not mandatory, it would be nice to separate your test in a new fle (named `test_betti_curve_representations.py` for instance), and then in `src/python/CMakeLists.txt` you can launch the test iif all the required modules are installed.
> ```cmake
>     if(SCIPY_FOUND AND MATPLOTLIB_FOUND)
>       add_gudhi_py_test(test_betti_curve_representations)
>     endif()
> ```
> The test will not fail (because not launched) if scipy is not
> installed.

Good point. I can make that change after waiting to see if there are any
other comments.

>
>> +    """
> +
> +    def __init__(self, filtration_range = [np.nan, np.nan]):
> +        """
> +        Constructor for the IrregularBettiCurve class.
> +
> +        Parameters:
> +            filtration_range ([float, float]): minimum and maximum filtration range to consider, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). The behavior is not the same as for sample_range in BettiCurve. In particular, intervals born before x_min are not considered. Intervals born in [x_min, x_max) contribute to the Betti curves as if they were infinite.
> +        """
> +        self.filtration_range = np.array(filtration_range)
> +
> +    def fit(self, X, y = None):
> +        """
> +        See the corresponding function in BettiCurve.
> +        """
> +        if np.isnan(self.filtration_range).any():
>
> If any NaN, `fit` will do nothing, but `transform` will assert. This is an open question also for @MathieuCarriere , but I am not sure if we should homogenize the behaviour (always assert or always do nothing, or even an exception).

Right.

 -- Gard

gspr:
Ah, I see that me replying by email kind of breaks GitHub's PR review. I'll try to remember to refrain from doing that.

Also: I'm not really used to GH's PR review process, so don't hesitate to let me know if there's something I need to do.

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: I am more in favour of homogeneous behavior, so I would suggest no assert, but I am ok to keep it this way

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
...
+
+    def transform(self, X):
+        """
+        Compute the Betti curve for each persistence diagram.
+
+        Parameters:
+            X (list of n x 2 NumPy arrays): input persistence diagrams.
+        
+        Returns:
+            Pair (xs, bettis), where xs is a length-k NumPy array with all filtration values for which any of the Betti curves change value, and bettis is an (number of diagrams) x k NumPy integer array with the values of the k Betti curves at the corresponding filtration values. The value bettis[i, k] is the value of the Betti curve of the k'th diagram on the interval [xs[i], xs[i+1]).
```
MathieuCarriere, line 388: The doc is a bit confusing to me: didn't you mean "bettis is an (number of diagrams) x k NumPy integer array with the values of the **(number of diagrams)** Betti curves at the corresponding filtration values." and "The value bettis[i, k] is the value of the Betti curve of the **i**'th diagram on the interval [xs[**k**], xs[**k+1**])."?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
...
+        Apply IrregularBettiCurve on a single persistence diagram and outputs the result.
+
+        Parameters:
+            diag (n x 2 numpy array): input persistence diagram.
+
+        Returns:
+            Pair (xs, b) of 1D NumPy arrays, xs containing the filtration values and b containing the Betti curve values. The Betti curve takes the value b[i] on the interval [xs[i], xs[i+1]).
+        """
+        res = self.fit_transform([diag])
+        return (res[0], res[1][0, :])
```
MathieuCarriere, line 437: I wonder if we should leave the `xs` as an attribute of the class and just output `bettis`. I am saying this because the class nicely fits into scikit-learn framework, but using it in an sklearn Pipeline will probably fail due to the fact that there are two outputs (I might be wrong though).

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
         """
         return self.fit_transform([diag])[0,:]
 
+
+class IrregularBettiCurve(BaseEstimator, TransformerMixin):
```
MathieuCarriere, line 355: I would be in favor of having just one `BettiCurve` class, with a potential parameter `even` or `regular` that would specify whether we sample uniformly or not.

MathieuCarriere:
@gspr Our usual way of doing things is to discuss the comments on this chat, and let the authors do the changes (if some are needed). Then we'll approve and merge once everyone is happy. Thanks a lot for contributing BTW!! Very nice class.

mglisse (commented):
A first few comments.
For the speed compared to BettiCurve, I guess it depends what the diagrams look like. If most intervals are very short (compared to the step of the grid on which we sample), the current BettiCurve should take constant time per interval. If intervals are longer, sorting is indeed faster. It could be that the second one is a better default...

gspr:
> A first few comments.
> For the speed compared to BettiCurve, I guess it depends what the diagrams look like. If most intervals are very short (compared to the step of the grid on which we sample), the current BettiCurve should take constant time per interval. If intervals are longer, sorting is indeed faster. It could be that the second one is a better default...

You're right, of course. That was a careless comment on my part. The running time of `IrregularBettiCurve` should be dominated by whatever sorting algorithm NumPy picks to sort the births and deaths of all intervals in the disjoint union of all the persistence diagrams (unless there is a very large number of empty or almost-empty persistence diagrams, but that seems like a corner case).

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
...
+            filtration_range ([float, float]): minimum and maximum filtration range to consider, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). The behavior is not the same as for sample_range in BettiCurve. In particular, intervals born before x_min are not considered. Intervals born in [x_min, x_max) contribute to the Betti curves as if they were infinite.
+        """
+        self.filtration_range = np.array(filtration_range)
+
+    def fit(self, X, y = None):
+        """
+        See the corresponding function in BettiCurve.
+        """
+        if np.isnan(self.filtration_range).any():
+            pre = DiagramScaler(use=True, scalers=[([0], MinMaxScaler()), ([1], MinMaxScaler())]).fit(X,y)
```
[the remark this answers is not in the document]
gspr: I'm happy to change it. I did it like that just to stay as close as possible to the behavior of `BettiCurve`. I'll make a note of it and change it when I make the other changes.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
...
+                    bettis[k].append(bettis[k][-1])
+
+        xs = np.array(xs)
+        bettis = np.array(bettis, dtype=int)
+
+        return(xs, bettis)
+
+    def __call__(self, diag):
+        """
+        Apply IrregularBettiCurve on a single persistence diagram and outputs the result.
```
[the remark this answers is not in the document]
gspr: You have discovered my shameless copy-pasting from `BettiCurve` ;-)

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
gspr: Sure, I'm happy to change that!

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
gspr: Indeed. How about `ExactBettiCurve`?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +351,92 @@ def __call__(self, diag):
...
+    """
+    This is a class for computing Betti curves from a list of persistence diagrams. Unlike the BettiCurve class, the curves are not sampled at fixed, regular points. Each curve has all its data intact.
+    """
+
+    def __init__(self, filtration_range = [np.nan, np.nan]):
+        """
+        Constructor for the IrregularBettiCurve class.
+
+        Parameters:
+            filtration_range ([float, float]): minimum and maximum filtration range to consider, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). The behavior is not the same as for sample_range in BettiCurve. In particular, intervals born before x_min are not considered. Intervals born in [x_min, x_max) contribute to the Betti curves as if they were infinite.
```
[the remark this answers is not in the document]
gspr: Yeah, it was a silly and arbitrary choice now that I think about it. Personally I am preferential to removing the range from this class altogether, and just write another preprocessing class that deals with restricting a persistence diagram to a given filtration range. What do you think about that?

src/python/gudhi/representations/vector_methods.py
gspr: I most certainly did. Thanks.

src/python/gudhi/representations/vector_methods.py
gspr: Sure thing! I'll make a note and make that change.

src/python/gudhi/representations/vector_methods.py
gspr: That's a good idea.

gspr:
> @gspr Our usual way of doing things is to discuss the comments on this chat, and let the authors do the changes (if some are needed). Then we'll approve and merge once everyone is happy. Thanks a lot for contributing BTW!! Very nice class.

By authors do you mean you the GUDHI authors, or me the contribution author? I'm fine with either, just making sure we don't do double work :-)

MathieuCarriere:
> By authors do you mean you the GUDHI authors, or me the contribution author? I'm fine with either, just making sure we don't do double work :-)

I meant the contributing author ;-)

gspr:
Mathieu Carrière <notifications@github.com> writes:

>> By authors do you mean you the GUDHI authors, or me the contribution author? I'm fine with either, just making sure we don't do double work :-)
>
> I meant the contributing author ;-)

Got it! I should be able to make the changes discussed soon if the
remaining questions are resolved.

VincentRouvreau:
> The CI failures on Windows look unrelated to me.

Indeed, it is unrelated to your dev. I fixed it on #433 and it is merged in master, so you can update your branch and it shall be fine. Sorry about that.

gspr:
The constructive feedback I received here led me to essentially rewrite the class. I have therefore opened a new PR, [PR #441](https://github.com/GUDHI/gudhi-devel/pull/441), and will close this one.

PR #428, opened 2020-11-25
Pybind11 cubical complex

VincentRouvreau:
Cubical and periodic cubical complex user documentations are now merged into one documentation as they have a common parent class to avoid code duplication.
Non conform constructors throw an exception instead of an error message.


src/python/gudhi/cubical/_cubical_complex.cc
```diff
@@ -0,0 +1,91 @@
...
+#include <vector>
+#include <string>
+#include <memory>       // for std::unique_ptr
+
+namespace py = pybind11;
+using namespace Gudhi::cubical_complex;
+
+
+template<typename CubicalComplexOptions = Bitmap_cubical_complex_base<double>>
+class Cubical_complex_interface_ : public Bitmap_cubical_complex<CubicalComplexOptions> {
```
mglisse, line 31: Is this class useful, or could we use directly Bitmap_cubical_complex?

VincentRouvreau: Remove useless class on 81c4df53

src/python/gudhi/cubical/_cubical_complex.cc
```diff
@@ -0,0 +1,91 @@
...
+#include <iostream>
+#include <vector>
+#include <string>
+#include <memory>       // for std::unique_ptr
+
+namespace py = pybind11;
+using namespace Gudhi::cubical_complex;
+
+
+template<typename CubicalComplexOptions = Bitmap_cubical_complex_base<double>>
```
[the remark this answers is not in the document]
VincentRouvreau: Remove useless class on 81c4df53

src/python/gudhi/cubical/_cubical_complex.cc
```diff
@@ -0,0 +1,91 @@
...
+#include <gudhi/Bitmap_cubical_complex_base.h>
+#include <gudhi/Bitmap_cubical_complex_periodic_boundary_conditions_base.h>
+
+#include <Persistent_cohomology_interface.h>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/numpy.h>
+#include <pybind11/stl.h>
+
+#include <iostream>
```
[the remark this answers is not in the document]
VincentRouvreau: Include what you use on 81c4df53 and 169654be

src/python/CMakeLists.txt
```diff
@@ -41,18 +41,11 @@ function( add_gudhi_debug_info DEBUG_INFO )
...
 
 if(PYTHONINTERP_FOUND)
-  if(PYBIND11_FOUND)
-    add_gudhi_debug_info("Pybind11 version ${PYBIND11_VERSION}")
-    set(GUDHI_PYTHON_MODULES "${GUDHI_PYTHON_MODULES}'bottleneck', ")
-    set(GUDHI_PYTHON_MODULES_EXTRA "${GUDHI_PYTHON_MODULES_EXTRA}'hera', ")
-    set(GUDHI_PYTHON_MODULES_EXTRA "${GUDHI_PYTHON_MODULES_EXTRA}'clustering', ")
-  endif()
-  if(CYTHON_FOUND)
+  if(CYTHON_FOUND AND PYBIND11_FOUND)
```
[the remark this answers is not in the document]
VincentRouvreau: Done on 7cd4794f

src/python/doc/cubical_complex_ref.rst
```diff
@@ -11,3 +11,15 @@ Cubical complex reference manual
...
    .. automethod:: gudhi.CubicalComplex.__init__
+
+.. autoclass:: gudhi.PeriodicCubicalComplex
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
+   .. automethod:: gudhi.PeriodicCubicalComplex.__init__
+
+.. autoclass:: gudhi.CubicalComplexBaseClass
```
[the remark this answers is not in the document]
VincentRouvreau: I didn't know this feature, and it seems clearer to me. Done on e1e27ad3
Tell me if you want me to rollback the 2 documentation pages (one for cubical and one for periodic cubical), as I merged the 2 pages because of the base class.

src/python/gudhi/cubical_complex.py
```diff
@@ -276,8 +158,153 @@ def persistence_intervals_in_dimension(self, dimension):
...
+        :type top_dimensional_cells: anything convertible to a numpy ndarray
+
+        Or
+
+        :param perseus_file: A Perseus-style file name.
+        :type perseus_file: string
+        """
+        if ((dimensions is not None) and (top_dimensional_cells is not None)
+            and (perseus_file == '')):
+            CubicalComplexBaseClass._thisptr = Cubical_complex_interface_(dimensions, top_dimensional_cells)
```
mglisse, line 196: The name `_thisptr` looks a bit strange now, but I guess it doesn't hurt.

src/python/gudhi/cubical/_cubical_complex.cc
```diff
@@ -0,0 +1,71 @@
...
+    .def("dimension", py::overload_cast<>(&Cubical_complex_interface_::dimension, py::const_));
+
+  py::class_<Persistent_cohomology_cubical_interface_>(m, "Persistent_cohomology_cubical_interface_")
+    .def(py::init<Cubical_complex_interface_*, bool>())
+    .def("compute_persistence", &Persistent_cohomology_cubical_interface_::compute_persistence)
+    .def("get_persistence", &Persistent_cohomology_cubical_interface_::get_persistence)
+    .def("cofaces_of_cubical_persistence_pairs", &Persistent_cohomology_cubical_interface_::cofaces_of_cubical_persistence_pairs)
+    .def("betti_numbers", &Persistent_cohomology_cubical_interface_::betti_numbers)
+    .def("persistent_betti_numbers", &Persistent_cohomology_cubical_interface_::persistent_betti_numbers)
+    .def("intervals_in_dimension", &Persistent_cohomology_cubical_interface_::intervals_in_dimension);
```
mglisse, line 52: I was going to say we should bind_vector this, but it returns the dreaded vector of pairs :cry: .
We could easily bind_vector `vector<int>` for betti_numbers (maybe others?), but it will change its return type (currently List[int]), so maybe it is better as a separate patch afterwards.

src/python/gudhi/cubical/_cubical_complex.cc
```diff
@@ -0,0 +1,71 @@
...
+  py::class_<Cubical_complex_interface_>(m, "Cubical_complex_interface_")
+    .def(py::init<const char*>())
+    .def(py::init<const std::vector<unsigned>&, const std::vector<Cubical_complex_interface_::Filtration_value>&>())
+    .def("num_simplices", &Cubical_complex_interface_::num_simplices)
+    // C++14 : https://pybind11.readthedocs.io/en/stable/classes.html#overloaded-methods
+    .def("dimension", py::overload_cast<>(&Cubical_complex_interface_::dimension, py::const_));
+
+  py::class_<Persistent_cohomology_cubical_interface_>(m, "Persistent_cohomology_cubical_interface_")
+    .def(py::init<Cubical_complex_interface_*, bool>())
+    .def("compute_persistence", &Persistent_cohomology_cubical_interface_::compute_persistence)
```
mglisse, line 47: Should we pass an extra `py::call_guard<py::gil_scoped_release>()` argument to some of those `.def()`?
(for something like `num_simplices`, it may not be worth it, the overhead of releasing / reacquiring the GIL is likely much more than the time to read one simple value. Similarly for a function that would just return an existing vector, since the nogil part would be empty and all the time spent later in the conversion to python objects)

src/python/doc/cubical_complex_ref.rst
```diff
@@ -8,6 +8,15 @@ Cubical complex reference manual
 .. autoclass:: gudhi.CubicalComplex
    :members:
    :undoc-members:
+   :inherited-members:
    :show-inheritance:
```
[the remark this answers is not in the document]
VincentRouvreau: You are right I removed it on c769f9c6 as there is no added value for the end user.

src/python/doc/cubical_complex_ref.rst
```diff
@@ -8,6 +8,15 @@ Cubical complex reference manual
 .. autoclass:: gudhi.CubicalComplex
    :members:
    :undoc-members:
```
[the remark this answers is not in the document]
VincentRouvreau: You are right. I also removed this one on c769f9c6

src/python/gudhi/cubical/_cubical_complex.cc
```diff
@@ -0,0 +1,71 @@
...
+
+using Periodic_cubical_complex_interface_ =
+    Bitmap_cubical_complex<Bitmap_cubical_complex_periodic_boundary_conditions_base<double>>;
+
+using Persistent_cohomology_periodic_cubical_interface_ =
+    Gudhi::Persistent_cohomology_interface<Periodic_cubical_complex_interface_>;
+
+PYBIND11_MODULE(_cubical_complex, m) {
+  // Cubical complex
+  py::class_<Cubical_complex_interface_>(m, "Cubical_complex_interface_")
```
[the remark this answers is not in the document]
VincentRouvreau: Cubical_complex_interface_ is a typedef, no more a class.
Maybe the name is not very good

src/python/gudhi/cubical_complex.py
VincentRouvreau: I renamed attributes members as it is no more a pointer on 1b816100

src/python/gudhi/cubical/_cubical_complex.cc
VincentRouvreau: Ok let's keep it for a later PR

src/python/gudhi/cubical/_cubical_complex.cc
VincentRouvreau: Done on 7e28033 and 6481f31
CI shall pass when #447 will be merged

mglisse (approved):
Ok.
The doc of compute_persistence looks duplicated, but it doesn't seem so convenient to dedup, and it is just one function.
So this changes from cython to pybind11. Is the advantage that we are trying to standardize on pybind11 and eventually drop cython? Is it a preparation for other changes that we think will be easier with pybind11? Does it change the performance? (hmm, this class only has costly functions except for the questionable num_simplices, so performance is likely irrelevant)

VincentRouvreau:
> So this changes from cython to pybind11. Is the advantage that we are trying to standardize on pybind11 and eventually drop cython? Is it a preparation for other changes that we think will be easier with pybind11? Does it change the performance? (hmm, this class only has costly functions except for the questionable num_simplices, so performance is likely irrelevant)

```bash
$ time (for i in {1..10}; do python3 example/random_cubical_complex_persistence_example.py 15 15 15 15 > /dev/null; done)
### cython
real    0m10,349s
user    0m19,332s
sys     0m5,359s
### pybind11
real    0m10,499s
user    0m19,494s
sys     0m5,479s
```

It is strange how it is a little bit slower with pybind11...
My first idea was to replace cython with pybind11, it would make the maintenance easier (just one language to know, decrease the probability of a bug, ...)
After we can go full with cython or pybind11, I have no strong opinion on it, but I would like to have only one way to make python interfaces.

mglisse:
I am not sure having just one way for python/C++ interfaces is something we really want. Sure, it would be nice. It seems that using pybind11 is more convenient for some things. On the other hand, it is tempting to use cython on all our current .py files, if it gives some small performance gain.

VincentRouvreau:
Maybe some changes are interesting and should be taken from this PR, but let's close this PR as we keep cython and pybind11.

PR #434, opened 2020-12-04
Pybind11 reader utils

VincentRouvreau:
Merge reader_utils and off_reader.
Fix unitary tests.
API changes:
* `read_persistence_intervals_in_dimension` now returns `array([], shape=(0, 2), dtype=float64)` when no persistence is found instead of `array([])`
* `read_persistence_intervals_grouped_by_dimension` and `read_persistence_intervals_in_dimension` are now raising an exception when file is not found (propagated from C++ code).


src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+      m.def("read_persistence_intervals_in_dimension", [](const std::string& persistence_file, int only_this_dim) {
+            // std::vector<std::pair<double, double>> as in read_persistence_intervals_in_dimension
+            // was not giving the good results because of memory alignment
+            // std::vector is mandatory as we don't know in advance the number of persistence intervals to be read
+            std::vector<std::array<double,2>> ret;
+            Gudhi::read_persistence_intervals_and_dimension(persistence_file,
+              boost::make_function_output_iterator([only_this_dim, &ret](std::tuple<int, double, double> t) {
+                if (only_this_dim == get<0>(t) || only_this_dim == -1) ret.push_back({get<1>(t), get<2>(t)});
+              }));
+            return py::array_t<double>(std::vector<std::ptrdiff_t>{static_cast<std::ptrdiff_t>(ret.size()), 2},
```
mglisse, line 81: This copies the vector. It is probably ok in this case, but a comment saying that there is a copy would be nice.
I tried the following
```c++
#include <pybind11/pybind11.h>
#include <pybind11/stl.h>
#include <pybind11/stl_bind.h>
#include <pybind11/numpy.h>

namespace py=pybind11;
typedef std::vector<std::array<double,2>> V;
PYBIND11_MAKE_OPAQUE(V);
PYBIND11_MODULE(marc, m) {
  py::bind_vector<V>(m, "VectorPair", py::buffer_protocol());
  m.def("fff",[](){
      V a(400000000, {1,2});
      return py::array(py::cast(std::move(a)));
      });
}
```
I limited virtual memory to 7G, and managed to call the function from python. If there was a copy, I expect that at some point it would require 2 buffers of 400000000*2 double, more than 12G.

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+    :returns:  The persistence pairs grouped by dimension.
+    :rtype: Dict[int, List[Tuple[float, float]]]
+        )pbdoc"
+      );
+      m.def("read_persistence_intervals_in_dimension", [](const std::string& persistence_file, int only_this_dim) {
+            // std::vector<std::pair<double, double>> as in read_persistence_intervals_in_dimension
+            // was not giving the good results because of memory alignment
+            // std::vector is mandatory as we don't know in advance the number of persistence intervals to be read
+            std::vector<std::array<double,2>> ret;
+            Gudhi::read_persistence_intervals_and_dimension(persistence_file,
```
mglisse, line 77: We can probably release the GIL during the C++ part of this function (it may need to be active again for the py::array part, but the slow part is the C++ I/O anyway).

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+    :returns:  The point set.
+    :rtype: List[List[float]]
+        )pbdoc"
+      );
+      m.def("read_lower_triangular_matrix_from_csv_file", &Gudhi::read_lower_triangular_matrix_from_csv_file<double>,
+        py::arg("csv_file"), py::arg("separator") = static_cast<const char>(';'),
+        R"pbdoc(
+    Read lower triangular matrix from a CSV style file.
+
+    :param csv_file: A CSV file style name.
```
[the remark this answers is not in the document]
VincentRouvreau: I put `Path to a ... file` with a link to the file formats page for OFF and persistence files

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+    :rtype: Dict[int, List[Tuple[float, float]]]
+        )pbdoc"
+      );
+      m.def("read_persistence_intervals_in_dimension", [](const std::string& persistence_file, int only_this_dim) {
+            // std::vector<std::pair<double, double>> as in read_persistence_intervals_in_dimension
+            // was not giving the good results because of memory alignment
+            // std::vector is mandatory as we don't know in advance the number of persistence intervals to be read
+            std::vector<std::array<double,2>> ret;
+            Gudhi::read_persistence_intervals_and_dimension(persistence_file,
+              boost::make_function_output_iterator([only_this_dim, &ret](std::tuple<int, double, double> t) {
```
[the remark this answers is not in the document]
VincentRouvreau: Include what you use on f0f320dc

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+
+    :param off_file: An OFF file style name.
+    :type off_file: string
+
+    :returns:  The point set.
+    :rtype: List[List[float]]
+        )pbdoc"
+      );
+      m.def("read_lower_triangular_matrix_from_csv_file", &Gudhi::read_lower_triangular_matrix_from_csv_file<double>,
+        py::arg("csv_file"), py::arg("separator") = static_cast<const char>(';'),
```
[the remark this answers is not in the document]
VincentRouvreau: No it doesn't help. I just copied/pasted example from pybind11 documentation. Changed on 45c7a651

src/python/gudhi/reader_utils.cc
VincentRouvreau: Ok, now it is quite clear to me how to do it. I exposed the C++ persistence intervals vector with your technique on f03fadc2
The syntax is also clearer and more C++ like. Thanks !

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,103 @@
...
+    :param persistence_file: A persistence file style name.
+    :type persistence_file: string
+
+    :returns:  The persistence pairs grouped by dimension.
+    :rtype: Dict[int, List[Tuple[float, float]]]
+        )pbdoc"
+      );
+      m.def("read_persistence_intervals_in_dimension", [](const std::string& persistence_file, int only_this_dim) {
+            // std::vector<std::pair<double, double>> as in read_persistence_intervals_in_dimension
+            // was not giving the good results because of memory alignment
```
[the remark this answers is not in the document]
VincentRouvreau: I wanted to replay my example and it worked... You are right, and I went to fast with this conclusion.

If I go back with a `vector<pair>`, I cannot bind the vector as you proposed.
This is a PyBind11 internal error, but when I try with a `vector<pair>` I have the following error:
```
pybind11/stl_bind.h:377:5: error: static assertion failed: There is not an appropriate format descriptor for this vector
     static_assert(vector_has_data_and_format<Vector>::value, "There is not an appropriate format descriptor for this vector");
     ^~~~~~~~~~~~~
pybind11/stl_bind.h:380:33: error: ‘format’ is not a member of ‘pybind11::format_descriptor<std::pair<double, double>, void>’
     format_descriptor<T>::format();
```
Do I have to bind the pair as well ?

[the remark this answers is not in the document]
mglisse: Yes, I had gotten the same error. I didn't experiment, so I have no idea what needs tweaking (maybe binders indeed). I would guess that format_descriptor, which is specialized for `T[N]` and `std::array<T,N>`, should also be specialized for double/tuple (or more generally for anything that supports tuple_element/tuple_size), at least when all types are the same. It may be worth opening an issue in pybind11. But maybe that also requires specializing other classes at the same time, or maybe it has undesirable side effects...
(using `std::pair<T,T>` is often considered a bad idea nowadays, as demonstrated by the fact that the committee refused to abbreviate it as `std::pair<T>`)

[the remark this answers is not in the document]
VincentRouvreau: I changed this wrong comment on 8b374238

src/python/gudhi/reader_utils.cc
VincentRouvreau: Done on 99ac0b4b

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+
+#include <gudhi/Points_off_io.h>
+#include <gudhi/reader_utils.h>
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <pybind11/numpy.h>
+#include <pybind11/stl_bind.h>
+
+#include <boost/function_output_iterator.hpp>  // for boost::make_function_output_iterator
```
mglisse, line 19: I think it prints a deprecated warning with newer versions of boost.

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,108 @@
...
+      py::bind_vector<Persistence_intervals>(m, "PersistenceIntervals", py::buffer_protocol());
+      m.def("read_persistence_intervals_in_dimension", [](const std::string& persistence_file, int only_this_dim) {
+            py::gil_scoped_release release;
+            // std::vector<std::pair<double, double>> is not exposable - cf. vector_has_data_and_format in pybind11/stl_bind.h
+            std::vector<std::array<double,2>> ret;
+            Gudhi::read_persistence_intervals_and_dimension(persistence_file,
+              boost::make_function_output_iterator([only_this_dim, &ret](std::tuple<int, double, double> t) {
+                if (only_this_dim == get<0>(t) || only_this_dim == -1) ret.push_back({get<1>(t), get<2>(t)});
+              }));
+            py::gil_scoped_acquire acquire;
```
mglisse, line 86: The `scope` in the name means that their destructor does the opposite action. So you are not supposed to have both release and acquire, you are just supposed to add appropriate `{` and `}` to ensure that the release object is destructed here.

src/python/gudhi/reader_utils.cc
VincentRouvreau: Oh yes you are right, now I remember on your last PR. Done on b7f4cb43

src/python/gudhi/reader_utils.cc
VincentRouvreau: Sorry about that, I completely missed how it was working. Done on cee6228c

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+    Note: the function does not check that birth <= death.
+
+    :param persistence_file: Path to a `persistence diagram <fileformats.html#persistence-diagram>`_ file.
+    :type persistence_file: string
+
+    :returns:  The persistence pairs grouped by dimension.
+    :rtype: Dict[int, List[Tuple[float, float]]]
+        )pbdoc"
+      );
+      py::bind_vector<Persistence_intervals>(m, "PersistenceIntervals", py::buffer_protocol());
```
mglisse, line 81: I think I'd put that at the beginning of the module, kind of like a typedef. Can we skip a couple lines before the next m.def (same elsewhere), so that it is visible where we switch from one function to the next?

src/python/gudhi/reader_utils.cc
```diff
@@ -0,0 +1,113 @@
...
+      m.def("read_persistence_intervals_in_dimension", [](const std::string& persistence_file, int only_this_dim) {
+            std::vector<std::array<double,2>> ret;
```
mglisse, line 83: I think this should use the Persistence_intervals typedef, to confirm that it is using the same type we made opaque and bound.

src/python/gudhi/reader_utils.cc
VincentRouvreau: Done on 63552a11

src/python/gudhi/reader_utils.cc
VincentRouvreau: Oh yes, my fault. Done on 63552a11

src/python/gudhi/reader_utils.cc
mglisse: Do you disagree with the idea of putting a few blank lines between `.def`s? Maybe because clang-format won't respect it? An alternative would be to define everything first
```c++
static auto read_points_from_off_file(const std::string& off_file){...}
char const*const read_points_from_off_file_doc = "Blablabla";
```
which can more easily be separated by empty lines, and then you would be left with just
```c++
m.def("read_points_from_off_file", read_points_from_off_file, py::arg("off_file"), read_points_from_off_file_doc);
```
although a drawback is that the `py::arg` is far from the function itself.
Or we could them inside the module, probably replacing the first one with `auto read_points_from_off_file=[](const std::string& off_file){...}`, to preserve locality without putting everything inside the `.def`.

I am trying to think what a good, readable presentation is for those files. I think you already looked at what some other projects were doing?

VincentRouvreau: Sorry, I just missed the part where you were suggesting to leave blank lines. No I didn't check on other projects what could be a good syntax. PyBind11 documentation doesn't leave any space, maybe not a good idea. Done on aabf1024

mglisse: I meant between every consecutive `.def`, not just before the first one.

src/python/gudhi/reader_utils.cc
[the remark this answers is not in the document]
mglisse: I reported the issue with std::pair as https://github.com/pybind/pybind11/issues/2991 . I also indicated a way we can make it work, if we really need it.

mglisse (commented):
Uh, I had a couple pending comments, not sure why they weren't published earlier.

VincentRouvreau:
This work was just to play with pybind11. Interesting topics to keep in mind (like numpy interfaces)

PR #436, opened 2020-12-08
Add build and tests wo cgal and eigen and wo cgal

### LINKED ISSUE #388: Eigen is mandatory for Python compilation
With #369 Edge Collapse is included by Python Simplex Tree.
This means no Python Simplex Tree when no Eigen, which means also no Alpha, no Rips, no Witness, no Cover, and no Tangential.

### LINKED ISSUE #431: Testsuite without CGAL
When the python plugin is built without CGAL, some tests are run that try to use bottleneck_distance. We need to disable them in that case (entirely or partially). And test that it works. While test_representations is clearly not protected, alpha_rips_persistence_bottleneck_distance.py and bottleneck_basic_example.py do have tests in CMakeLists.txt but still seem to cause problems for this user.

VincentRouvreau:
Fix #431 
I also fixed #388 using your 3rd proposal @mglisse:
>     3. Add some #ifdef so that only the edge collapse function depends on Eigen (and possibly move test_collapse_edges to its own file so it is easier to disable?)
I kept collapse_edges that does nothing but a copy when no Eigen. This is motivated by the fact that I cannot remove `collapse_edges` from simplex_tree.pxd when no Eigen (or I did not find how).


src/python/include/Simplex_tree_interface.h
```diff
@@ -187,6 +196,11 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
       collapsed_stree_ptr->insert({std::get<0>(remaining_edge), std::get<1>(remaining_edge)}, std::get<2>(remaining_edge));
     }
     return collapsed_stree_ptr;
+#else
+    // If no Eigen3, return a copy, as it will be deleted in pyx
+    Simplex_tree_interface* collapsed_stree_ptr = new Simplex_tree_interface(*this);
+    return collapsed_stree_ptr;
```
mglisse, line 202: If the old one gets destroyed immediately, should we move instead of copy? This feels too quiet, I think we should at least warn, possibly even throw an exception if people try to call this function without Eigen.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -17,6 +17,9 @@ __author__ = "Vincent Rouvreau"
 __copyright__ = "Copyright (C) 2016 Inria"
 __license__ = "MIT"
 
+# For unitary tests purpose
+__GUDHI_USE_EIGEN3 = _GUDHI_USE_EIGEN3
```
mglisse, line 21: We could also put this in `__init__.py`, since it is already generated by cmake.

VincentRouvreau: You are right this is simpler. Done on 957da77

src/python/include/Simplex_tree_interface.h
VincentRouvreau: I was not sure, this was the reason of my comment. I throw an exception on 40e0976

src/python/CMakeLists.txt
```diff
@@ -129,10 +129,13 @@ if(PYTHONINTERP_FOUND)
       set(GUDHI_PYTHON_EXTRA_COMPILE_ARGS "${GUDHI_PYTHON_EXTRA_COMPILE_ARGS}'-DDEBUG_TRACES', ")
     endif()
 
+    set(GUDHI_USE_EIGEN3 "False")
```
mglisse, line 132: why not an `else` branch?

VincentRouvreau: Ok, it may make you think that no Eigen is the default behaviour. Else branch on ba3be8d1

mglisse (approved):

PR #438, opened 2020-12-10
Build python 3.9 pip package and use python 3.9 as default test package

VincentRouvreau:


mglisse:
I didn't check what is changing exactly, but note that some packages like pytorch do not support python3.9 yet, so we still need to have some tests with 3.8 for that to be tested. Otherwise, looks good.

VincentRouvreau:
> I didn't check what is changing exactly, but note that some packages like pytorch do not support python3.9 yet, so we still need to have some tests with 3.8 for that to be tested. Otherwise, looks good.

`pip-build-*.yml` builds the python package with what is mandatory to have a full python package (boost, eigen, cgal, ... + build-requirements.txt). Then, no test is performed. It is just to check that the pip wheels can be created.
I am using python 3.9 to be sure it will work on next pip package.
`pip-packaging-*.yml` is doing the same but with all python versions and sends the wheels using twine on pypi.

I will check later when doing the version by building an alpha version, but it should work.

PR #439, opened 2020-12-14
Release note 3.4.0

VincentRouvreau:


.github/next_release.md
```diff
@@ -9,8 +9,9 @@ Below is a list of changes made since GUDHI 3.3.0:
 - [Alpha complex](https://gudhi.inria.fr/doc/latest/group__alpha__complex.html)
      - the C++ weighted version for alpha complex is now available in dimension D.
```
tlacombe, line 10: What does `available in dimension D` mean? Genuine question, but perhaps it can be more explicit; does it mean "for any D, instead of only D=0,1,2" (from what I understand from the current documentation) ?

PR #440, opened 2020-12-19
Docstring improvements in RipsComplex

ulupo:
- `create_simplex_tree` method refers to the Delaunay triangulation instead of the flag complex
- "rips" was not capitalized
- "double" was used in the docs but only "float" (which has double precision) is a Python type


mglisse (approved):
Thanks.

VincentRouvreau (approved):

PR #441, opened 2020-12-20
A more flexible class for computing Betti curves

gspr:
[PR #427](https://github.com/GUDHI/gudhi-devel/pull/427) received helpful comments. Incorporating these lead me to essentially rewriting the class in question. I therefore took the opportunity to start this new PR based on the current HEAD.

Be aware that this new class does not deal with filtration ranges in any way. It seems to me that cropping a persistence diagram to a certain filtration range is rightly considered a preprocessing step, and best not left to each individual representation class.

### DIFF
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -1,14 +1,16 @@
 # This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
 # See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
-# Author(s):       Mathieu Carrière, Martin Royer
+# Author(s):       Mathieu Carrière, Martin Royer, Gard Spreemann
 #
 # Copyright (C) 2018-2020 Inria
 #
 # Modification(s):
 #   - 2020/06 Martin: ATOL integration
+#   - 2020/12 Gard: A more flexible Betti curve class capable of computing exact curves.
 
 import numpy as np
 from sklearn.base          import BaseEstimator, TransformerMixin
+from sklearn.exceptions    import NotFittedError
 from sklearn.preprocessing import MinMaxScaler, MaxAbsScaler
 from sklearn.neighbors     import DistanceMetric
 from sklearn.metrics       import pairwise
@@ -349,6 +351,154 @@ class BettiCurve(BaseEstimator, TransformerMixin):
             numpy array with shape (**resolution**): output Betti curve.
         """
         return self.fit_transform([diag])[0,:]
+
+
+class BettiCurve2(BaseEstimator, TransformerMixin):
+    """
+    A more flexible replacement for the BettiCurve class.
+
+    Examples
+    --------
+    If pd is a persistence diagram and xs is a grid such that xs[0] >= pd.min(), then the result of
+    >>> bc = BettiCurve2(xs)
+    >>> result = bc(pd)
+    and
+    >>> from scipy.interpolate import interp1d
+    >>> bc = BettiCurve2(None)
+    >>> bettis = bc.fit_transform([pd])
+    >>> interp = interp1d(bc.grid_, bettis[0, :], kind="previous", fill_value="extrapolate")
+    >>> result = np.array(interp(xs), dtype=int)
+    are the same.
+    """
+
+    def __init__(self, grid = None):
+        """
+        Constructor for the BettiCurve class.
+
+        Parameters
+        ----------
+        grid: 1d array or None, default=None
+             Filtration grid points at which to compute the Betti curves. Must be strictly ordered. Infinites are OK. If None (default), a grid will be computed that captures all the filtration value changes.
+
+        Attributes
+        ----------
+        grid_: 1d array
+             Contains the compute grid after fit or fit_transform.
+        """
+
+        self.grid_ = np.array(grid)
+
+
+    def fit(self, X, y = None):
+        """
+        Compute a filtration grid that captures all changes in Betti numbers for all the given persistence diagrams.
+
+        Parameters
+        ----------
+        X: list of 2d arrays
+             Persistence diagrams.
+
+        y: None.
+             Ignored.
+        """
+
+        events = np.unique(np.concatenate([pd.flatten() for pd in X], axis=0))
+
+        if len(events) == 0:
+            self.grid_ = np.array([-np.inf])
+        else:
+            self.grid_ = np.array(events)
+
+        return self
+
+
+    def fit_transform(self, X):
+        """
+        Find a sampling grid that captures all changes in Betti numbers, and compute those Betti numbers. The result is the same as fit(X) followed by transform(X), but potentially faster.
+        """
+
+        N = len(X)
+
+        events = np.concatenate([pd.flatten(order="F") for pd in X], axis=0)
+        sorting = np.argsort(events)
+        offsets = np.zeros(1 + N, dtype=int)
+        for i in range(0, N):
+            offsets[i+1] = offsets[i] + 2*X[i].shape[0]
+        starts = offsets[0:N]
+        ends = offsets[1:N + 1] - 1
+
+        bettis = [[0] for i in range(0, N)]
+        if len(sorting) == 0:
+            xs = [-np.inf]
+        else:
+            xs = [events[sorting[0]]]
+
+        for i in sorting:
+            j = np.searchsorted(ends, i)
+            delta = 1 if i - starts[j] < len(X[j]) else -1
+            if events[i] == xs[-1]:
+                bettis[j][-1] += delta
+            else:
+                xs.append(events[i])
+                for k in range(0, j):
+                    bettis[k].append(bettis[k][-1])
+                bettis[j].append(bettis[j][-1] + delta)
+                for k in range(j+1, N):
+                    bettis[k].append(bettis[k][-1])
+
+        self.grid_ = np.array(xs)
+        return np.array(bettis, dtype=int)
+
+
+    def transform(self, X):
+        """
+        Compute Betti curves.
+
+        Parameters
+        ----------
+        X: list of 2d arrays
+            Persistence diagrams.
+
+        Returns
+        -------
+        (len(X))x(len(self.grid_)) array of ints
+             Betti numbers of the given persistence diagrams at the grid points given in self.grid_.
+        """
+
+        if self.grid_ is None:
+            raise NotFittedError("Not fitted. You need to call fit or construct with a chosen sampling grid.")
+
+        N = len(X)
+
+        events = np.concatenate([pd.flatten(order="F") for pd in X], axis=0)
+        sorting = np.argsort(events)
+        offsets = np.zeros(1 + N, dtype=int)
+        for i in range(0, N):
+            offsets[i+1] = offsets[i] + 2*X[i].shape[0]
+        starts = offsets[0:N]
+        ends = offsets[1:N + 1] - 1
+
+        bettis = [[0] for i in range(0, N)]
+
+        i = 0
+        for x in self.grid_:
+            while i < len(sorting) and events[sorting[i]] <= x:
+                j = np.searchsorted(ends, sorting[i])
+                delta = 1 if sorting[i] - starts[j] < len(X[j]) else -1
+                bettis[j][-1] += delta
+                i += 1
+            for k in range(0, N):
+                bettis[k].append(bettis[k][-1])
+
+        return np.array(bettis, dtype=int)[:, 0:-1]
+
+
+    def __call__(self, diag):
+        """
+        Shorthand for transform on a single persistence diagram.
+        """
+        return self.transform([diag])[0, :]
+
 
 class Entropy(BaseEstimator, TransformerMixin):
     """
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -290,7 +290,7 @@ class Silhouette(BaseEstimator, TransformerMixin):
 
 class BettiCurve(BaseEstimator, TransformerMixin):
     """
-    Compute Betti curves from persistence diagrams. There are two modes of operation: with a predefined grid, and without. With a predefined grid, the class computes the Betti numbers at those grid points. Without a predefined grid, it can be fit to a list of persistence diagrams and produce a grid that consists of (at least) the filtration values at which at least one of those persistence diagrams chance Betti numbers, and then compute the Betti numbers at those grid points. In the latter mode, the exact Betti curve is computed for the entire real line.
+    Compute Betti curves from persistence diagrams. There are two modes of operation: with a predefined grid, and without. With a predefined grid, the class computes the Betti numbers at those grid points. Without a predefined grid, it can be fit to a list of persistence diagrams and produce a grid that consists of (at least) the filtration values at which at least one of those persistence diagrams changes Betti numbers, and then compute the Betti numbers at those grid points. In the latter mode, the exact Betti curve is computed for the entire real line.
 
     Parameters
     ----------
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -355,7 +355,7 @@ class BettiCurve(BaseEstimator, TransformerMixin):
 
 class BettiCurve2(BaseEstimator, TransformerMixin):
     """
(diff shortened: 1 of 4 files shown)


src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+        starts = offsets[0:N]
+        ends = offsets[1:N + 1] - 1
+
+        bettis = [[0] for i in range(0, N)]
+        if len(sorting) == 0:
+            xs = [-np.inf]
+        else:
+            xs = [events[sorting[0]]]
+
+        for i in sorting:
```
mglisse, line 436: If we wanted to, I think it would be possible to avoid looping on all the elements in python, but it isn't clear it would really be better (it may use more memory and replace some `y=x` with `y=x+0`). For instance: create an array with ±1 on the k-th line for the k-th diagram (+1 for birth, -1 for death), reorder it according to `sorting`, `cumsum`, and finally extract only the columns whose event is not the same as the previous column. There may be better ways, this just to say that it seems ok to have a loop in python, unless some benchmark proves it to be horrible, and your previous PR seems to indicate that it is fast enough.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
         """
         return self.fit_transform([diag])[0,:]
 
+
+class BettiCurve2(BaseEstimator, TransformerMixin):
+    """
+    A more flexible replacement for the BettiCurve class.
```
mglisse, line 358: Doc is a bit short, but the exact doc we want depends on if we merge this with the other BettiCurve or not. Are infinite values supported? (I think so)
It took me a bit of time to understand what the example was saying, maybe a sentence or two that says the same thing in English instead of code would help understand it (the example itself is good).

src/python/test/test_betti_curve_representations.py
```diff
@@ -0,0 +1,41 @@
...
+        pd[np.random.uniform(0, 1, n) < pinf, 1] = np.inf
+        pds.append(pd)
+    
+    bc = BettiCurve2(None)
+    bc.fit(pds)
+    bettis = bc.transform(pds)
+
+    bc2 = BettiCurve2(None)
+    bettis2 = bc2.fit_transform(pds)
+    success = success and (bc2.grid_ == bc.grid_).all()
```
mglisse, line 29: Why not assert on each line instead of once at the very end of the test, so it is more visible which step failed if one does?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+        events = np.concatenate([pd.flatten(order="F") for pd in X], axis=0)
+        sorting = np.argsort(events)
+        offsets = np.zeros(1 + N, dtype=int)
+        for i in range(0, N):
+            offsets[i+1] = offsets[i] + 2*X[i].shape[0]
+        starts = offsets[0:N]
+        ends = offsets[1:N + 1] - 1
+
+        bettis = [[0] for i in range(0, N)]
+        if len(sorting) == 0:
```
mglisse, line 431: Could you add a very small test for this case? What output do we even want in that case, `np.zeros((N,1))`?

mglisse (commented):
@MathieuCarriere Looking at this new version, does it make more sense to have 2 versions of BettiCurve, or keep only the new one (possibly adding some options to it for compatibility with the older one)?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+        y: None.
+             Ignored.
+        """
+
+        events = np.unique(np.concatenate([pd.flatten() for pd in X], axis=0))
+
+        if len(events) == 0:
+            self.grid_ = np.array([-np.inf])
+        else:
+            self.grid_ = np.array(events)
```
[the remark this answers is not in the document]
gspr: I guess it comes from my general fear and loathing of duck typing, and reflexive wish to know the types of things. Who among us haven't written `x + y` while thinking of `x` and `y` as NumPy arrays when they were lists all along.

I'm happy to change it if you prefer.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+        Constructor for the BettiCurve class.
+
+        Parameters
+        ----------
+        grid: 1d array or None, default=None
+             Filtration grid points at which to compute the Betti curves. Must be strictly ordered. Infinites are OK. If None (default), a grid will be computed that captures all the filtration value changes.
+
+        Attributes
+        ----------
+        grid_: 1d array
```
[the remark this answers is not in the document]
gspr: Good point. I'll move it.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+             Contains the compute grid after fit or fit_transform.
+        """
+
+        self.grid_ = np.array(grid)
```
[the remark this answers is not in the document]
gspr: That sounds like a good idea!

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+        X: list of 2d arrays
+            Persistence diagrams.
+
+        Returns
+        -------
+        (len(X))x(len(self.grid_)) array of ints
+             Betti numbers of the given persistence diagrams at the grid points given in self.grid_.
+        """
+
+        if self.grid_ is None:
```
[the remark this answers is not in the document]
gspr: Indeed, thanks!

src/python/gudhi/representations/vector_methods.py
```diff
@@ -350,6 +352,154 @@ def __call__(self, diag):
...
+
+        events = np.concatenate([pd.flatten(order="F") for pd in X], axis=0)
+        sorting = np.argsort(events)
+        offsets = np.zeros(1 + N, dtype=int)
+        for i in range(0, N):
+            offsets[i+1] = offsets[i] + 2*X[i].shape[0]
+        starts = offsets[0:N]
+        ends = offsets[1:N + 1] - 1
+
+        bettis = [[0] for i in range(0, N)]
```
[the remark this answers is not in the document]
gspr: You are right, I'll fix that. Thanks!

src/python/gudhi/representations/vector_methods.py
gspr: I agree, that's a better way. I would suggest maybe that that can be kicked down the road a little bit though, unless someone feels the urge. I suspect the current implementation is good enough for a lot of the cases where it's useful.

src/python/gudhi/representations/vector_methods.py
gspr: Indeed. I'll expand the docs a bit.

Re infinites: Infinite deaths are supported. Infinite Betti numbers are not. I should check that the implementation doesn't go nuts if fed infinite births.

src/python/test/test_betti_curve_representations.py
gspr: Indeed, I'll fix this. Thanks.

src/python/gudhi/representations/vector_methods.py
gspr: That's a good question. I oscillated back and forth over it more than I'll admit. What's your preference?

src/python/gudhi/representations/vector_methods.py
mglisse: Yes, that was my point, a different implementation can wait (and if we really end up needing more speed, which would be surprising, just rewriting it in C++ may be the best solution).

src/python/gudhi/representations/vector_methods.py
mglisse: I don't really have one right now, maybe @MathieuCarriere does? I'd be tempted to return an array of shape (N,0), but if that breaks your interpolation identity and zeros((N,1)) doesn't, then that one may be better. This corner case really shouldn't matter much, I asked for a test because you seemed to be going out of your way to support this case (and it is good for the coverage statistics).

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
mglisse: I understand, I wrote something because here `events` is defined on the line just above as the result of some `np.*`. Also, IIRC the default for `np.array.__init__` is `copy=True` which seems wasteful, even if completely negligible. But it is ok if you want to keep it like this.

src/python/gudhi/representations/vector_methods.py
gspr: Yeah, `(N, 0)` sounds like a good option, too. I can just specify that the interpolation identity only holds for non-empty diagrams. I'll wait and see if anyone else has any preferences.

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
gspr: After fddeb5724fe2e7f1f37476c5e3cfade992a4edec, I think the code is now much sklearn-compliant in its behavior.

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
gspr: Done.

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
gspr: This is no longer relevant after fddeb5724fe2e7f1f37476c5e3cfade992a4edec.

src/python/test/test_betti_curve_representations.py
gspr: Done.

src/python/gudhi/representations/vector_methods.py
gspr: I ended up going for the convention that even an empty diagram will return a Betti curve having value `0` from `-np.inf`.

gspr:
I think the latest commit deal with most of the comments. I also realized the class wasn't dealing with several [scikit-learn conventions](https://scikit-learn.org/stable/developers/develop.html), such as not setting properties that can be fitted to data (those ending in underscore) elsewhere (like in `__init__`).

I also ended up going for the convention that an empty persistence diagram gets assigned the grid `np.array([-np.inf])` and the zero Betti curve.

MathieuCarriere:
Hi @gspr, sorry about the long time I took to respond. I think it's best to come up with a single implementation to avoid confusion. A quick way for this is to provide a third option for `grid`: in addition to `None` or a predefined grid, you could specify a tuple `(N, min, max)` which basically creates a grid of `N` evenly spaced values between `min` and `max` (which could themselves be estimated from the data). Providing a grid that way would then internally call the old version. That way we could avoid having two classes that compute the same objects with different methods. What do you think?

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: I feel like the best would be to return an array whose second dimension is the same as `len(grid)` (if provided). So a list of empty diagrams would return an array full of 0s. If `grid` is not provided, then I personally think that an array of shape `[N,0]` is the most consistent.

gspr:
Mathieu Carrière ***@***.***> writes:

> Hi @gspr, sorry about the long time I took to respond.

No problem.

> I think it's best to come up with a single implementation to avoid confusion. A quick way for this is to provide a third option for `grid`: in addition to `None` or a predefined grid, you could specify a tuple `(N, min, max)` which basically creates a grid of `N` evenly spaced values between `min` and `max` (which could themselves be estimated from the data). Providing a grid that way would then internally call the old version. That way we could avoid having two classes that compute the same objects with different methods. What do you think?

I like the idea of a third grid option of the form you suggest, but do
you really want it to cold the old version? Wouldn't it be better that
the new version, called with `(N, min, max)` calls the new version again
with `np.linspace(min, max, N)` as grid – for compatibility with the old
version?

gspr:
Mathieu Carrière ***@***.***> writes:

> I feel like the best would be to return an array whose second dimension is the same as `len(grid)` (if provided). So a list of empty diagrams would return an array full of 0s. If `grid` is not provided, then I personally think that an array of shape `[N,0]` is the most consistent.

That is the current behavior. And I agree.

MathieuCarriere:
> I like the idea of a third grid option of the form you suggest, but do
you really want it to call the old version? Wouldn't it be better that
the new version, called with `(N, min, max)`, calls the new version again
with `np.linspace(min, max, N)` as grid – for compatibility with the old
version?

Actually, I was suggesting this to avoid having two classes, but I am fine with what you propose too, as long as we get rid of the old class.

src/python/gudhi/representations/vector_methods.py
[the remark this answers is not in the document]
MathieuCarriere: maybe `bettis=np.zeros([N,1])`?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -285,70 +287,167 @@ def __call__(self, diag):
         """
         return self.fit_transform([diag])[0,:]
 
+
 class BettiCurve(BaseEstimator, TransformerMixin):
     """
-    This is a class for computing Betti curves from a list of persistence diagrams. A Betti curve is a 1D piecewise-constant function obtained from the rank function. It is sampled evenly on a given range and the vector of samples is returned. See https://www.researchgate.net/publication/316604237_Time_Series_Classification_via_Topological_Data_Analysis for more details.
+    A more flexible replacement for the BettiCurve class. There are two modes of operation: with a predefined grid, and without. With a predefined grid, the class computes the Betti numbers at those grid points. Without a predefined grid, it can be fit to a list of persistence diagrams and produce a grid that consists of (at least) the filtration values at which at least one of those persistence diagrams chance Betti numbers, and then compute the Betti numbers at those grid points. In the latter mode, the exact Betti curve is computed for the entire real line.
```
mglisse, line 293: The text may need updating, it states that this class is a replacement for itself :wink:

src/python/test/test_betti_curve_representations.py
```diff
@@ -0,0 +1,54 @@
...
+    pds = []
+    for i in range(0, m):
+        pd = np.zeros((n, 2))
+        pd[:, 0] = np.random.uniform(0, 10, n)
+        pd[:, 1] = np.random.uniform(pd[:, 0], 10, n)
+        pd[np.random.uniform(0, 1, n) < pzero, 0] = 0
+        pd[np.random.uniform(0, 1, n) < pinf, 1] = np.inf
+        pds.append(pd)
+    
+    bc = BettiCurve2(None)
```
mglisse, line 22: I think you need to rename it in the tests as well.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -285,70 +287,167 @@ def __call__(self, diag):
...
+    >>> result = bc(pd)
+    and
+    >>> from scipy.interpolate import interp1d
+    >>> bc = BettiCurve(None)
+    >>> bettis = bc.fit_transform([pd])
+    >>> interp = interp1d(bc.grid_, bettis[0, :], kind="previous", fill_value="extrapolate")
+    >>> result = np.array(interp(xs), dtype=int)
+    are the same.
     """
-    def __init__(self, resolution=100, sample_range=[np.nan, np.nan]):
```
mglisse, line 292: I don't know if we want that, but it would be possible to avoid breaking users' code if we enabled the new exact mode with resolution=None or resolution="exact", instead of introducing a new parameter predefined_grid.

src/python/test/test_betti_curve_representations.py
gspr: Fixed, thanks.

src/python/gudhi/representations/vector_methods.py
mglisse: Would it be ok with you if we changed the interface in such a way? Not breaking existing user code would make it easier for us to merge this branch.
We can do the change, or you can do it, as you prefer.

gspr: Hi. Sorry for forgetting about this discussion. Feel free to change the interface to whatever you prefer – I have no strong feelings about it (just that there should be something à la the exact construction that I proposed available :-)).

Issue #443, opened 2020-12-29
Tangential complex does not work with CGAL 5.2


gspr:
This is just to document the discussion from [1] as an issue. I am preparing a PR.

[1] https://lists.gforge.inria.fr/pipermail/gudhi-users/2020-December/000421.html

gspr:
PR #444 fixes this.

PR #444, opened 2020-12-29
Fix building with CGAL 5.2

gspr:
The fix is based on a similar fix for the alpha complex code, as suggested by Marc Glisse in [1], and fixes issue #443.

[1] https://lists.gforge.inria.fr/pipermail/gudhi-users/2020-December/000421.html


mglisse (approved):
Thanks!
@VincentRouvreau I think we should do a .1 release with this change, after checking that everything works fine with cgal-5.2 (and take the chance to rebuild the pip/conda packages with cgal-5.2).

VincentRouvreau (approved):

PR #406, opened 2020-10-18
Handle duplicated points in choose_n_farthest_points

mglisse:
Fix #404.

PR #457, opened 2021-03-03
Sparse Rips memory corruption

### LINKED ISSUE #456: Potential Bug in Sparse Rips Complex
The calculation of the sparse rips complex (using Python) gives an error for my specific point cloud - which is quite symmetrical. I attach the corresponding (60x60) distance matrix as a txt file. GitHub won't let me pick at better file format -sorry. 

I have not yet found any other examples which give the same error and the unweighted rips complex works well. Furthermore if I perturb all points by a set of very small random numbers the calculation works as desired. The unpredictable nature of the error makes it hard for me to deal with systematically, however. 

When using my mac OS (10.14.6) I get a python runtime error - 'Unknown Error'.

When using my Ubuntu machine (16.04.6) I get a different error that has apparently leaked out of the python wrapper:
"_terminate called after throwing an instance of 'char const*'_"

Thanks

[distance_matrix_that_causes_error.txt](https://github.com/GUDHI/gudhi-devel/files/6075290/distance_matrix_that_causes_error.txt)

mglisse:
Fix #456.
This completes #406, where I missed that this vector needs to be sized based on the total number of points, not on the number that remains after removing duplicates.
(draft because this is still not enough, the redundant points still appear as isolated vertices in the graph)


mglisse:
See #476 instead.

PR #463, opened 2021-03-25
Add equality operator for python Simplex tree

VincentRouvreau:


src/python/gudhi/simplex_tree.pyx
```diff
@@ -639,3 +639,10 @@ cdef class SimplexTree:
             self.thisptr = <intptr_t>(ptr.collapse_edges(nb_iter))
             # Delete old pointer
             del ptr
+
+    def __eq__(self, other:SimplexTree):
```
mglisse, line 643: I wonder if that changes the minimum version of [pc]ython we require. (I think it is fine to require it)

src/python/gudhi/simplex_tree.pyx
```diff
@@ -639,3 +639,10 @@ cdef class SimplexTree:
...
+        """Simplex tree equality operator using C++ depth first search operator==
```
mglisse, line 644: Is it really relevant to say it uses C++ and DFS? All that matters is that it is a structural equality.

src/python/gudhi/simplex_tree.pyx
VincentRouvreau: Python typing involves python >= 3.5 ([source](https://docs.python.org/3/library/typing.html)) and cython >= 0.27 ([source](https://github.com/cython/cython/pull/3799/files)).

src/python/gudhi/simplex_tree.pyx
mglisse:
```suggestion
        """Test for structural equality
```

src/python/gudhi/simplex_tree.pyx
mglisse: Those requirements seem ok to me, should we document them?

mglisse (approved):

VincentRouvreau:
@mglisse As you approved this PR, I will it at the same time as #552 will be approved as there is a clear dependency

PR #472, opened 2021-04-15
Fix contradictory color

mglisse:


VincentRouvreau (approved):
I am ok as "ro" was not part of the [markers](https://matplotlib.org/stable/api/markers_api.html#module-matplotlib.markers).
I could not find the default color value (seems to be the 'default blue') that seems to be the root cause of your fix.
Documentation is ok, so the change is fine to me.

mglisse:
"ro" is a format argument, where you can put more than just markers. In particular, `r` stands for `red`, so the second call was specifying the color as both red and green, hence the warning.

VincentRouvreau:
> "ro" is a format argument, where you can put more than just markers. In particular, `r` stands for `red`, so the second call was specifying the color as both red and green, hence the warning.

Thanks for the clarification.

PR #474, opened 2021-04-15
Compare lists with ==

mglisse:
Otherwise we get this warning

> numpy/core/_asarray.py:102: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.

And indead trying to build an array from `[[1,2],[3]]` is weird.


VincentRouvreau (approved):
I am ok it is weird to create the array `[[0, 1, 2], [0, 1, 2], [0, 1, 2], [3, 4, 5, 6], [3, 4, 5], [3, 4, 5], [3, 6]]` where the elements do not have the same size.

PR #475, opened 2021-04-15
Name the argument n_neighbors of NearestNeighbors

mglisse:
It seems that it has had the same name for a while, so this shouldn't break anything.

> sklearn/utils/validation.py:70: FutureWarning: Pass n_neighbors=3 as keyword args. From version 1.0 (renaming of 0.25) passing these as positional arguments will result in an error


VincentRouvreau (approved):
It was at least there on [scikit learn 0.15 NearestNeighbors](https://scikit-learn.org/0.15/modules/generated/sklearn.neighbors.NearestNeighbors.html) in July 2014.

Issue #413, opened 2020-11-02
choose_n_farthest_points number type


mglisse:
`choose_n_farthest_points` currently hardcodes a number type of `double`. I didn't try, but I assume it would fail if we tried to run in on Epeck_d points.

`Euclidean_distance` manages to be generic, and I think we should do the same for `choose_n_farthest_points`.

VincentRouvreau:
If I have a look to https://github.com/GUDHI/gudhi-devel/blob/7e05e915adc1be285e04eb00d3ab7ba1b797f38d/src/Subsampling/include/gudhi/choose_n_farthest_points.h#L65-L74
`DistanceOutputIterator` is an output iterator, which means `DistanceOutputIterator::value_type` is not known (cf. [std::back_insert_iterator](https://en.cppreference.com/w/cpp/iterator/back_insert_iterator)).
`Point_range` is a range over CGAL points (Epick_d or Epeck_d), so we can have access to `Point_range::value_type` (will be a Point_d), but not to `Point_range::value_type::FT` (`Point_range::value_type::FT_` is private).
`dist_to_L` should be a `std::vector< FT >`, but I don't know how to get this type, maybe by using `decltype` on a "fake" distance computation.
https://github.com/GUDHI/gudhi-devel/blob/7e05e915adc1be285e04eb00d3ab7ba1b797f38d/src/Subsampling/include/gudhi/choose_n_farthest_points.h#L93
Finally, as this comment claims, `Epeck_d::FT` has no infinity value, but this is quite convenient in our case to prevent the user the distance was not computed. We have to find how do so differently.

VincentRouvreau:
A first attempt, that is not satisfying enough in [this branch](https://github.com/VincentRouvreau/gudhi-devel/tree/epeck_choose_n_farthest_points)

PR #476, opened 2021-04-17
Sparse Rips and choose_n_farthest_points, fix #456

### LINKED ISSUE #456: Potential Bug in Sparse Rips Complex

mglisse:
That was way more work than I hoped...
You can see gradual improvements with the commits. The first one is probably not necessary after the whole series, but it still looks better to me. First I made choose_n_farthest_points return all the points, which fixes the bug. But we build a complete graph on equal points, so I updated the sparse Rips so redundant vertices are connected through just one edge. And finally I actually dropped the redundant points.
One small missing optimization is a way to say: stop choose_n_farthest_points once we reach a small enough distance. Currently, that would require passing a special output iterator that throws an exception when that happens...

Fix #456

### DIFF
--- a/src/Rips_complex/include/gudhi/Sparse_rips_complex.h
+++ b/src/Rips_complex/include/gudhi/Sparse_rips_complex.h
@@ -15,7 +15,7 @@
 #include <gudhi/graph_simplicial_complex.h>
 #include <gudhi/choose_n_farthest_points.h>
 
-#include <boost/graph/adjacency_list.hpp>
+#include <boost/graph/graph_traits.hpp>
 #include <boost/range/metafunctions.hpp>
 #include <boost/iterator/counting_iterator.hpp>
 
@@ -118,7 +118,7 @@ class Sparse_rips_complex {
    *
    */
   template <typename RandomAccessPointRange, typename Distance>
-  Sparse_rips_complex(const RandomAccessPointRange& points, Distance distance, double epsilon, Filtration_value mini=-std::numeric_limits<Filtration_value>::infinity(), Filtration_value maxi=std::numeric_limits<Filtration_value>::infinity())
+  Sparse_rips_complex(const RandomAccessPointRange& points, Distance distance, double const epsilon, Filtration_value const mini=-std::numeric_limits<Filtration_value>::infinity(), Filtration_value const maxi=std::numeric_limits<Filtration_value>::infinity())
       : epsilon_(epsilon) {
     GUDHI_CHECK(epsilon > 0, "epsilon must be positive");
     auto dist_fun = [&](Vertex_handle i, Vertex_handle j) { return distance(points[i], points[j]); };
@@ -139,7 +139,7 @@ class Sparse_rips_complex {
    * @param[in] maxi Maximal filtration value. Ignore anything above this scale.
    */
   template <typename DistanceMatrix>
-  Sparse_rips_complex(const DistanceMatrix& distance_matrix, double epsilon, Filtration_value mini=-std::numeric_limits<Filtration_value>::infinity(), Filtration_value maxi=std::numeric_limits<Filtration_value>::infinity())
+  Sparse_rips_complex(const DistanceMatrix& distance_matrix, double const epsilon, Filtration_value const mini=-std::numeric_limits<Filtration_value>::infinity(), Filtration_value const maxi=std::numeric_limits<Filtration_value>::infinity())
       : Sparse_rips_complex(boost::irange<Vertex_handle>(0, boost::size(distance_matrix)),
                             [&](Vertex_handle i, Vertex_handle j) { return (i==j) ? 0 : (i<j) ? distance_matrix[j][i] : distance_matrix[i][j]; },
                             epsilon, mini, maxi) {}
@@ -155,7 +155,7 @@ class Sparse_rips_complex {
    *
    */
   template <typename SimplicialComplexForRips>
-  void create_complex(SimplicialComplexForRips& complex, int dim_max) {
+  void create_complex(SimplicialComplexForRips& complex, int const dim_max) {
     GUDHI_CHECK(complex.num_vertices() == 0,
                 std::invalid_argument("Sparse_rips_complex::create_complex - simplicial complex is not empty"));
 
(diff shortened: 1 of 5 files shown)


src/Rips_complex/include/gudhi/Sparse_rips_complex.h
```diff
@@ -108,17 +164,17 @@ class Sparse_rips_complex {
       complex.expansion(dim_max);
       return;
     }
-    const int n = boost::size(params);
-    std::vector<Filtration_value> lambda(n);
+    const Vertex_handle n = num_vertices(graph_);
```
VincentRouvreau, line 167: `Vertex_handle` is an int, whereas `num_vertices` returns a `std::size_t`

src/Rips_complex/include/gudhi/Sparse_rips_complex.h
```diff
@@ -131,30 +187,32 @@ class Sparse_rips_complex {
...
       auto&& pi = points[i];
       auto li = params[i];
-      if (li < mini) break;
-      for (int j = i + 1; j < n; ++j) {
+      // If we inserted all the points, points with multiplicity would get connected to their first representative,
+      // no need to handle the redundant ones in the outer loop.
+      // if (li <= 0 && i != 0) break;
+      for (Vertex_handle j = i + 1; j < n; ++j) {
         auto&& pj = points[j];
         auto d = dist(pi, pj);
```
VincentRouvreau, line 214: In src/Rips_complex/example/sparse_rips_complex.cpp, we have:
```cpp
  using Point = std::vector<double>;
  using Simplex_tree = Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_fast_persistence>;
  using Filtration_value = Simplex_tree::Filtration_value;
```
As points are a vector of doubles, the distance function will return a `double` value, whereas `Gudhi::Simplex_tree_options_fast_persistence::Filtration_value` is a `float`.
The fix should be fixing the example.

src/Subsampling/include/gudhi/choose_n_farthest_points.h
```diff
@@ -88,34 +88,57 @@ void choose_n_farthest_points(Distance dist,
     starting_point = dis(gen);
   }
 
-  std::size_t current_number_of_landmarks = 0;  // counter for landmarks
-  static_assert(std::numeric_limits<double>::has_infinity, "the number type needs to support infinity()");
   // FIXME: don't hard-code the type as double. For Epeck_d, we also want to handle types that do not have an infinity.
```
VincentRouvreau, line 91: Ok regarding #413 
Maybe we should define a type instead of using double to identify the future fix.

src/Rips_complex/include/gudhi/Sparse_rips_complex.h
```diff
@@ -17,9 +17,68 @@
 
 #include <boost/graph/adjacency_list.hpp>
```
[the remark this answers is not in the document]
mglisse: It was useful in that it includes graph_traits.hpp :wink:
Changed.

src/Rips_complex/include/gudhi/Sparse_rips_complex.h
mglisse: Well, yes, we use int for vertices, and vector uses size_t for its size, so there has to be a conversion somewhere...

src/Rips_complex/include/gudhi/Sparse_rips_complex.h
mglisse: Can you explain what the problem is? We have a distance function that gives a precise result, we have a filtration where we are happy with a lower precision, so we drop some precision in the middle, I think that works...

src/Subsampling/include/gudhi/choose_n_farthest_points.h
mglisse: I think that can be done in the future PR that will fix #413. I believe I made such a fix much simpler, but I'd rather put that in a separate patch.

src/Rips_complex/include/gudhi/Sparse_rips_complex.h
mglisse: Hmm, the way I use i and n, they could actually just be size_t, I'll do that.

mglisse: Done. Note that it is broken that we arbitrarily define Vertex_handle as int in Sparse_rips, but the 2-function design makes that a bit complicated to avoid.

src/Rips_complex/include/gudhi/Sparse_rips_complex.h
VincentRouvreau: It is just a matter of warnings fix, but here, we use points as vector of double. The distance from 2 points will be saved in a double, whereas the filtration values in the simplex tree will be saved as a float.
I think we should use vector of float for a point, or use a double for simplex tree filtration value type.

mglisse: That clearly predates this patch, so we can handle it separately.
I believe it is a perfectly valid use case. I understand why some compilers warn when converting from double to float, but in some cases we may want to do that conversion.
We could add some explicit casts to remove the warning, but it isn't really our place to decide if dropping precision is wanted, it is up to the user. Which tends to favor disabling the warning in the example with a `#pragma`.
We could choose to have an example that does not convert so we don't have to see the warning in the testsuite. That's a cop out, but well...

VincentRouvreau: We can keep it like that, I understand your point, no problem.

VincentRouvreau (approved):

PR #478, opened 2021-04-21
Update Hera, fix #477

### LINKED ISSUE #477: Hera infinite loop with point on the diagonal
```python
import gudhi.hera
import numpy as np

dgm1 = [[0, 0], [0, 1.0]]
dgm2 = [[0, 1.0]]
print(gudhi.hera.wasserstein_distance(dgm1, dgm2, order=2, internal_p=np.inf))
```
does not terminate here.

mglisse:


VincentRouvreau (commented):
I think we should add the failing test case (cf. the issue) in the test suite.

mglisse:
I thought about it, but the test is already in hera's testsuite. We can still be redundant...

VincentRouvreau (approved):
No need to be redundant, I just wanted to be sure it was tested.

PR #482, opened 2021-04-22
Remove make install target from python and rewrite documentation accordingly

### LINKED ISSUE #468: Python gudhi package shall not be installed with 'make install'
gudhi C++ library requires the user to have the proper privileges to be able to install header files. On UNIX systems, as files are installed in `/usr/local` ([source](https://cmake.org/cmake/help/v3.0/variable/CMAKE_INSTALL_PREFIX.html)), this results to install the library by doing:
```bash
sudo make install
```
The user can tweak the installation directory by calling:
```bash
make DESTDIR=/my/directory install
```

The issue with the gudhi python package is that `sudo make install` will lead to do `sudo python setup.py install`.
This is not working if some dependencies are not in the sudoers PATH (cython, for instance).
Moreover, it does not exist already, it will create a directory `/usr/local/lib/python3.X/dist-packages/` with the super user ownership, which is not a good idea as all other future installations will require administrator privileges thereafter (cf. `sudo pip install` vs `pip install --user` posts).
Finally, it is not possible to retrieve `DESTDIR` (or I didn't find how) and I had to document how to install in a different directory by doying `python setup.py install --prefix /my/directory`.

After a quick analysis, I think we should remove python installation through the `make install` command cf. https://github.com/GUDHI/gudhi-devel/blob/7e05e915adc1be285e04eb00d3ab7ba1b797f38d/src/python/CMakeLists.txt#L277
And we should only document to install the python library by doing `python setup.py install`.

VincentRouvreau:
Fix #468

### DIFF
--- a/src/common/doc/installation.h
+++ b/src/common/doc/installation.h
@@ -30,6 +30,10 @@ make \endverbatim
  * This action may require to be in the sudoer or administrator of the machine in function of the operating system and
  * of <a href="https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html">CMAKE_INSTALL_PREFIX</a>.
  *
+ * \note Python module will be compiled by the `make` command, but `make install` will not install it. Please refer to
+ * the <a href="https://gudhi.inria.fr/python/latest/installation.html#gudhi-python-module-installation">Python
+ * module installation documentation</a>.
+ * 
  * \subsection testsuites Test suites
  * To test your build, run the following command in a terminal:
  * \verbatim make test \endverbatim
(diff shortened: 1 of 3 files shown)


mglisse:
We may want to add a word to the C++ installation doc, if `make` builds the python plugin but `make install` doesn't install it.
Other than that, ok, thanks.

VincentRouvreau:
@mglisse And I forgot to say it works with `pip install .`. Do you want this to be mentioned in the documentation ?

mglisse:
> @mglisse And I forgot to say it works with `pip install .`.

Nice :smiley: 
> Do you want this to be mentioned in the documentation ?

Your choice. I think it is nice to mention it, but if you are not very confident about that option and prefer to recommend the other one, we don't have to mention it right now.

mglisse:
ok if you don't touch hera.

Note that I think `pip install .` rebuilds everything and doesn't just install, though I could be wrong.

PR #488, opened 2021-04-29
numpy.take_along_axis used in knn requires numpy>=1.15.0

VincentRouvreau:


mglisse (approved):
If those requirements files are now unnecessary, sure, ok.

PR #497, opened 2021-05-28
pykeops fails on CI and authors are rewritting dependencies

VincentRouvreau:
Tests on linux are ok


azure-pipelines.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         sudo conda install --yes --quiet --name gudhi_build_env python=$(pythonVersion)
         python -m pip install --user -r ext/gudhi-deploy/build-requirements.txt
         python -m pip install --user -r ext/gudhi-deploy/test-requirements.txt
+        python -m pip uninstall pykeops
```
mglisse, line 26: `-y` ?

VincentRouvreau: Ooh yes you are right. I just noticed it was taking a long time.

PR #504, opened 2021-06-17
Persistence graphical tools improvements

### LINKED ISSUE #453: [Python - persistence_graphical_tools] warn when plot drops some intervals
The plotting functions have a parameter `max_intervals=1000` which means that some intervals are not shown. This can be confusing for users, and printing a warning when that happens could help. Of course, that same warning could annoy users who specifically asked for this behavior, maybe it could be restricted to the default case or have a way to disable it.

### LINKED ISSUE #461: [Python - persistence_graphical_tools] Exception or nothing to display on empty diagrams
```python
import gudhi as gd
gd.plot_persistence_diagram([[], [], [], []])
  got: IndexError: list index out of range
  expected: An empty persistence diagram
```

The same for `plot_persistence_density` and `plot_persistence_barcode`

VincentRouvreau:
Fix #453
Fix #461
Review error management (no more prints, warnings and exceptions instead) and add unitary tests
No more plots in loop (optimization)


src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -44,38 +48,64 @@ def __min_birth_max_death(persistence, band=0.0):
...
 def _array_handler(a):
-    '''
+    """
     :param a: if array, assumes it is a (n x 2) np.array and return a
                 persistence-compatible list (padding with 0), so that the
                 plot can be performed seamlessly.
-    '''
-    if isinstance(a[0][1], np.float64) or isinstance(a[0][1], float):
+    """
+    if isinstance(a[0][1], np.floating) or isinstance(a[0][1], float):
```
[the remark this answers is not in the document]
VincentRouvreau: `np.floating` includes `np.float64` and `np.float32`.
I don't know if it can happen, but I find nice we can plot float64 and float32.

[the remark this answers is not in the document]
Hind-M: Ah okay, the type to check is different even if it could be the same precision.

Hind-M (approved):

src/python/gudhi/persistence_graphical_tools.py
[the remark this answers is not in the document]
Hind-M: Ah okay, the type to check is different even if it could be the same precision.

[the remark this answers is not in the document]
mglisse: `np.issubdtype(type(a[0][1]), np.floating)`?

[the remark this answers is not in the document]
mglisse: (for the syntax, `isinstance(x,t1) or isinstance(x,t2)` can be written `isinstance(x,(t1,t2))` I believe)

mglisse:
The comment "fix #453" is a bit short, and the changes are a bit hard to find with the reformatting.
What is the new behavior? By default it prints a warning once, the first time the plot function is called with a list that is too long?

src/python/gudhi/persistence_graphical_tools.py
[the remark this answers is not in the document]
VincentRouvreau: Ok thanks, I didn't know this nicer syntax. Done on 87da488d

VincentRouvreau:
> The comment "fix #453" is a bit short, and the changes are a bit hard to find with the reformatting.
> What is the new behavior? By default it prints a warning once, the first time the plot function is called with a list that is too long?

I let the [warning filter](https://docs.python.org/3/library/warnings.html#the-warnings-filter) to default (which means "print the first occurrence of matching warnings for each location (module + line number) where the warning is issued").

I could set it to `"always"`, but maybe we should define a class like a `PlotWarning` for instance to not impact other warnings.

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -44,38 +48,64 @@ def __min_birth_max_death(persistence, band=0.0):
...
 @lru_cache(maxsize=1)
 def _matplotlib_can_use_tex():
     """This function returns True if matplotlib can deal with LaTeX, False otherwise.
     The returned value is cached.
     """
     try:
         from matplotlib import checkdep_usetex
+
         return checkdep_usetex(True)
     except ImportError:
```
Hind-M, line 100: Shouldn't it be more explicit if we print the import error instead?
For example:

```suggestion
    except ImportError as error:
```

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -44,38 +48,64 @@ def __min_birth_max_death(persistence, band=0.0):
...
-        print("This function is not available, you may be missing matplotlib.")
+        warnings.warn("This function is not available, you may be missing matplotlib.")
```
Hind-M, line 101:
```suggestion
        print(error)
```

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -119,99 +149,92 @@ def plot_persistence_barcode(
...
 
         # Ends plot on infinity value and starts a little bit before min_birth
-        axes.axis([axis_start, infinity, 0, ind])
+        if ind != 0:
+            axes.axis([axis_start, infinity, 0, ind])
         return axes
 
     except ImportError:
-        print("This function is not available, you may be missing matplotlib.")
+        warnings.warn("This function is not available, you may be missing matplotlib.")
```
Hind-M, line 237: Same here with the import error print ?

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -268,109 +290,102 @@ def plot_persistence_diagram(
...
         axes.set_ylabel("Death", fontsize=fontsize)
         axes.set_title("Persistence diagram", fontsize=fontsize)
         # Ends plot on infinity value and starts a little bit before min_birth
-        axes.axis([axis_start, axis_end, axis_start, infinity + delta/2])
+        axes.axis([axis_start, axis_end, axis_start, infinity + delta / 2])
         return axes
 
     except ImportError:
-        print("This function is not available, you may be missing matplotlib.")
+        warnings.warn("This function is not available, you may be missing matplotlib.")
```
Hind-M, line 388: Same here with the import error print ?

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -522,6 +550,4 @@ def plot_persistence_density(
         return axes
 
     except ImportError:
-        print(
-            "This function is not available, you may be missing matplotlib and/or scipy."
-        )
+        warnings.warn("This function is not available, you may be missing matplotlib and/or scipy.")
```
Hind-M, line 553: Same here with the import error print ?

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Done on 55b15a7

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: I kept the use of warning and modify the message on 55b15a7
`error` is not directly printable, dev only have access to its `name` and `path` (cf. [ImportError](https://docs.python.org/3/library/exceptions.html#ImportError))

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Same here

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Same here

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Same here

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -119,114 +149,81 @@ def plot_persistence_barcode(
...
         # Ends plot on infinity value and starts a little bit before min_birth
-        axes.axis([axis_start, infinity, 0, ind])
+        if len(x) != 0:
+            axes.axis([axis_start, infinity, 0, len(x)])
         return axes
 
-    except ImportError:
-        print("This function is not available, you may be missing matplotlib.")
+    except ImportError as import_error:
+        warnings.warn(f"This function is not available.\nModuleNotFoundError: No module named '{import_error.name}'.")
```
mglisse, line 212: I don't really see how that's better than not catching the exception, but it was already like that before the PR, so ok.

mglisse (approved):
I didn't really check the details, but if you think it is good, go ahead and merge it.

PR #506, opened 2021-06-25
Test another sphinx themes

### LINKED ISSUE #42: [Python - documentation] body_max_width is not a valid option for 'classic' theme
Sphinx Python documentation fails on some system because of:
```
   html_theme = 'classic'
   html_theme_options = {
   ...
      "body_max_width": "1200px",
```

Regarding
https://github.com/sphinx-doc/sphinx/blob/master/doc/theming.rst body_max_width option
is only available for 'basic' theme.

### LINKED ISSUE #486: sphinx_paramlinks dependency
The dependency on sphinx_paramlinks is not very clean: we don't document this dependency, we don't check if it is present, we just fail at the point where we try to use it, possibly during `make test` (if we didn't build the doc separately before?), which hides the error message
> Could not import extension sphinx_paramlinks (exception: No module named 'sphinx_paramlinks')

VincentRouvreau:
and some beautifying
Fix #486 
I think we can close #42 as it works fine with this theme, and no warnings/errors about it in [logs](https://circleci.com/api/v1.1/project/github/GUDHI/gudhi-devel/10101/output/102/0?file=true&allocation-id=60d5f02acc82275fa58fd3b8-0-build%2F45903BF6).

[Python documentation](https://10101-174304137-gh.circle-artifacts.com/0/sphinx/index.html)


tlacombe (approved):
Seems nice overall ! :+1:

PR #512, opened 2021-07-22
Remove bad attribute maybe_unused

mglisse:
`CGAL_UNUSED` or `[[maybe_unused]]` is useful if we name the argument but don't use it. If we don't name it, the attribute is wrong as it would apply to the type and not the variable.


VincentRouvreau (approved):
It fixes Azure pipeline issue, on purpose !

PR #529, opened 2021-09-13
make_filtration_non_decreasing no more always called

### LINKED ISSUE #57: make_filtration_non_decreasing in Alpha_complex_3d
https://github.com/GUDHI/gudhi-devel/blob/a33d17ebb1a5f7c830ee155ec2733a0f48211010/src/Alpha_complex/include/gudhi/Alpha_complex_3d.h#L568

I wonder if this is always necessary. At least in the _exact_ case I don't think it is, I am not sure about the _safe_ and _fast_ cases. Anyway the current code is safe, and I doubt the cost matters, this is more of a remark.

VincentRouvreau:
No more called in:
 * alpha complex dD exact version
 * alpha complex 3D exact and safe versions

Fix #57 
(cf. benchmark in this issue)

### DIFF
--- a/src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
+++ b/src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
@@ -554,8 +554,10 @@ Weighted_alpha_complex_3d::Weighted_point_3 wp0(Weighted_alpha_complex_3d::Bare_
     std::clog << "cells \t\t" << count_cells << std::endl;
 #endif  // DEBUG_TRACES
     // --------------------------------------------------------------------------------------------
-    // As Alpha value is an approximation, we have to make filtration non decreasing while increasing the dimension
-    complex.make_filtration_non_decreasing();
+    if (Complexity == complexity::FAST)
+      // As Alpha value is an approximation, we have to make filtration non decreasing while increasing the dimension
+      // Only in FAST version, cf. https://github.com/GUDHI/gudhi-devel/issues/57
+      complex.make_filtration_non_decreasing();
     // Remove all simplices that have a filtration value greater than max_alpha_square
     complex.prune_above_filtration(max_alpha_square);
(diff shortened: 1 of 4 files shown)


src/Alpha_complex/include/gudhi/Alpha_complex_3d.h
```diff
@@ -554,8 +554,10 @@ Weighted_alpha_complex_3d::Weighted_point_3 wp0(Weighted_alpha_complex_3d::Bare_
     std::clog << "cells \t\t" << count_cells << std::endl;
 #endif  // DEBUG_TRACES
     // --------------------------------------------------------------------------------------------
-    // As Alpha value is an approximation, we have to make filtration non decreasing while increasing the dimension
-    complex.make_filtration_non_decreasing();
+    if (Complexity == complexity::FAST)
```
Hind-M, line 557: According to the benchmark referred to in #57, the results with 'safe' mode in 3D are different with some tested data. Shouldn't it be just the 'exact' mode where the make_filtration_non_decreasing may be removed safely like in dD?

VincentRouvreau: @Hind-M Which one ? For 3D, you have to look algo `alpha 3D *`, but not for `alpha dD *`. Maybe this big table is confusing

Hind-M: Oh ok I see, I didn't notice that the algo can be different in 3D table aside from the mode. Maybe the two given tables should be split by algo and not by type of data, or the second table may have two sub-tables (algo alpha3D and algo alpha dD ?). Anyway that's how I read it considering that 4D data was an example used with alpha dD and 3D data with alpha 3D, I don't know if people read it that way too ...

Hind-M (approved):

PR #536, opened 2021-10-15
Online quantization algorithm for gudhi

tlacombe:
Provide a quantization algorithm to "summarize" a collection of persistence diagrams. 

(At least) One thing that may be discussed : 
- I put the code in the `python/gudhi/wasserstein/` repo, because it is of a "Wasserstein metric" flavor (we minimize something in terms of Wasserstein distance between persistence diagrams). However, it does not rely on POT as other functions in this repo do ; we actually never need to explicitly compute a Wasserstein distance/matching explicitly. Perhaps would it belong directly to the `gudhi/` repo ? 

Also TODO :
- [x] Check for `quantization.py` : is the copyright correct?

### DIFF
--- a/src/python/gudhi/wasserstein/quantization.py
+++ b/src/python/gudhi/wasserstein/quantization.py
@@ -9,7 +9,7 @@
 
 
 import numpy as np
-from gudhi import _utils
+from gudhi.wasserstein import _utils
 import warnings
 
 
--- a/src/python/gudhi/wasserstein/wasserstein.py
+++ b/src/python/gudhi/wasserstein/wasserstein.py
@@ -9,7 +9,7 @@
 
 import numpy as np
 import warnings
-from gudhi import _utils
+from gudhi.wasserstein import _utils
 
 try:
     import ot


biblio/bibliography.bib
```diff
@@ -78,6 +78,13 @@ @article{boissonnatmariasimplextreealgorithmica
 language={English}
 }
 
+@article{divolLacombe2021quantization,
+  title={Estimation and Quantization of Expected Persistence Diagrams},
+  author={Divol, Vincent and Lacombe, Th{\'e}o},
+  journal={ICML 2021},
+  year={2021}
```
VincentRouvreau, line 85: A link to the arxiv is welcome:
```suggestion
  year={2021},
  url       = {https://arxiv.org/abs/2105.04852}
```

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,71 @@ The output is:
...
+Doing so, the final codebook is able to properly retrieve the two 1-dimensional features of the underlying points clouds
+(the two loops generating the tori).
+
+.. figure::
+     ./img/quantiz.gif
+     :figclass: align-center
+
+     Online-quantization of a family of persistence diagrams built from random tori.
+
+.. autofunction:: gudhi.wasserstein.quantization
```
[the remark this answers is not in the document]
VincentRouvreau: @tlacombe I think you need to add in `src/python/gudhi/wasserstein/__init__.py`:
```python
from .quantization import quantization
```

Or use `gudhi.wasserstein.quantization.quantization`, but that would not make sense.

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,218 @@
...
+def quantization(pdiagset, k=2, c0=None, batch_size=1, order=2., internal_p=2.):
+    """
+    This quantization algorithm takes a list of diagrams ``pdiagset``, an integer ``k``
+    (or an initial codebook guess ``c0``)
+    and returns a codebook of ``k`` centroids that aims at minimizing the distance between the codebook and the
+    expected persistence diagram underneath the diagrams in ``pdiagset``.
+    The codebook is iteratively updated by going through (batches of) persistence diagrams.
+
+    :param pdiagset: A list of ``numpy.array`` of shape (n x 2) (``n`` can variate) representing a
+                     set of persistence diagrams that we want to quantize.
```
VincentRouvreau, line 144:
```suggestion
    :param pdiagset: a set of persistence diagrams that we want to quantize.
    :type pdiagset: list of n x 2 numpy arrays
```

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,71 @@ The output is:
      [0.7375     0.7625    ]
      [0.2375     0.2625    ]]
 
-Tutorial
-********
+Quantization
+------------
+
```
VincentRouvreau, line 198:
```suggestion

:Since: GUDHI 3.5.0
:Requires: `SciPy <installation.html#scipy>`_

```

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,71 @@ The output is:
...
+    [[0.      1.00025]
+     [0.      3.00125]]
+
+
+Tutorials
+---------
 
 This
 `notebook <https://github.com/GUDHI/TDA-tutorial/blob/master/Tuto-GUDHI-Barycenters-of-persistence-diagrams.ipynb>`_
 presents the concept of barycenter, or Fréchet mean, of a family of persistence diagrams.
```
VincentRouvreau, line 260: Maybe this notebook shall in its own section

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,71 @@ The output is:
...
+If one has access to a sample of measures instead, one may perform online quantization: given a sample
+:math:`\mu_1,\dots,\mu_n` and an initial codebook :math:`c`, we progressively update :math:`c` by going through
+(possibly batches of) the :math:`(\mu_i)_i`. Theoretical properties (including convergence rates and guarantees) of the
+final codebook are provided in :cite:`divolLacombe2021quantization`.
+
+The Figure below showcases the use of the online-quantization approach provided by the
+``gudhi.wasserstein.quantization`` method.
+In this experiment, Rips persistence diagrams (in homology dimension 1) are built from random point clouds supported on
+different tori with some additional noise.
+Starting from an initial codebook ``c0``, centroids are iteratively updated as new diagrams are provided.
```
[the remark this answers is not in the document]
tlacombe: Indeed, here `c0` was meant to be "c at time t=0" ; but I also use the notation c_1,...,c_k to denote the k "centroids" of the codebook. Perhaps I could chance `c0` to `c_init` in the doc, would it be clearer ?

src/python/doc/wasserstein_distance_user.rst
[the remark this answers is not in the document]
tlacombe: Ah, weird, it works on my local fork using `make sphinx`. I will add the line suggested by Vincent.

src/python/doc/wasserstein_distance_user.rst
tlacombe: Indeed (though putting all the notebooks at the end of the doc may be an option as well).

src/python/doc/wasserstein_distance_user.rst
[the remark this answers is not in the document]
tlacombe: Done.

VincentRouvreau:
IMHO the copyright is correct.

tlacombe:
> IMHO the copyright is correct.

Great ! Though, shouldn't it be 2021, now that I read it?

VincentRouvreau:
> > IMHO the copyright is correct.
> 
> Great ! Though, shouldn't it be 2021, now that I read it?

Yes !

tlacombe:
Corrected. 
Any thought on if this code belong to the `/wasserstein` module? From a theoretical perspective, the goal is to solve some minimization with respect to the Wasserstein distance between persistence diagrams (so it makes sense to put it there), but from a practical perspective (code), it does not rely on `pot` contrarily to other functions in the `/wasserstein` module (because we can solve our optimization problem without explicitly computing such distances/matchings). 
My feeling is that it can stay there, doing `gudhi.wasserstein.quantization` makes clear that we are quantizing something with respect to the Wasserstein distance ; but I am open to discussion of course.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,82 @@ The output is:
...
+Quantization
+------------
+
+:Since: GUDHI 3.5.0
+:Requires: `SciPy <installation.html#scipy>`_
+
+The quantization problem consists, given a measure :math:`\mu` (e.g. a persistence diagram) and a budget :math:`k`,
+in finding a measure :math:`\nu` supported on :math:`k` points that is the best approximation possible of :math:`\mu` ;
+the :math:`k`-means problem is a particular example of quantization problem.
+The output of a quantization algorithm is often referred to a codebook; it consists of :math:`k` points (centroids)
```
mglisse, line 215:
```suggestion
The output of a quantization algorithm is often referred to as a codebook; it consists of :math:`k` points (centroids)
```

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,82 @@ The output is:
...
+:math:`\mu_1,\dots,\mu_n` and an initial codebook :math:`c`, we progressively update :math:`c` by going through
+(possibly batches of) the :math:`(\mu_i)_i`. Theoretical properties (including convergence rates and guarantees) of the
+final codebook are provided in :cite:`divolLacombe2021quantization`.
+
+The Figure below showcases the use of the online-quantization approach provided by the
+``gudhi.wasserstein.quantization`` method.
+In this experiment, Rips persistence diagrams (in homology dimension 1) are built from random point clouds supported on
+different tori with some additional noise.
+Starting from an initial codebook ``c0``, centroids are iteratively updated as new diagrams are provided.
+As we use the standard metrics between persistence diagrams (denoted here by :math:`\mathrm{OT}_2`), points in the
```
mglisse, line 228: Not sure about calling that "the standard metrics".

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,82 @@ The output is:
...
+diagrams that are close to the diagonal do not interfere in the codebook update process.
```
mglisse, line 229: So it is the same as having an implicit point on the diagonal in the codebook?

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,82 @@ The output is:
...
+.. autofunction:: gudhi.wasserstein.quantization
+
+Basic example:
+**************
+
+This example outputs a codebook for a set of diagrams with two proeminent modes around :math:`(0,1)` and :math:`(0,3)`.
+Diagrams, as often, also contain points near the diagonal. As we compare the codebook and the input diagrams
+using the standard :math:`2`-Wasserstein metric between persistence diagram (which maps points to the diagonal rather
+than any centroids if it reduces the matching cost), those points do not influence the centroids positions,
+an improvment over using vanilla :math:`k`-mean (Lloyd algorithm) to quantize persistence diagrams.
```
mglisse, line 248:
```suggestion
an improvement over using vanilla :math:`k`-mean (Lloyd algorithm) to quantize persistence diagrams.
```

src/python/doc/wasserstein_distance_user.rst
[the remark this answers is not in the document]
mglisse: Note that c0 is not just a doc thing, it is also the name of a parameter. c0 looks like an ok name for that parameter. It looks like sklearn calls it `init` in [kmeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.KMeans.html).

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,218 @@
...
+import numpy as np
+import scipy.spatial.distance as sc
+import warnings
+
+
+def _dist_to_diag(X, internal_p):
+    return ((X[:, 1] - X[:, 0]) * 2 ** (1. / internal_p - 1))
+
+
+def _build_dist_matrix(X, Y, order=2., internal_p=2):
```
mglisse, line 20: Could we move those 2 functions into a separate file (with a name that starts with `_` unless we want to document them) so we can avoid having several copies?

mglisse (commented):
I guess keeping it in wasserstein/ is ok.

src/python/doc/wasserstein_distance_user.rst
tlacombe: More precisely, having a point in the codebook that represents "all the points on the diagonal" (or, formally, looking at the quotient space where you identify the points on the diagonal).

src/python/gudhi/wasserstein/quantization.py
tlacombe: I totally agree with this idea. How should it be done though? having a file `utils.py` in `wasserstein/` ?

mglisse: > I totally agree with this idea. How should it be done though? having a file `utils.py` in `wasserstein/` ?

If it is an internal file that users are not supposed to know about, I would say `_utils.py`, but yes.

src/python/doc/wasserstein_distance_user.rst
tlacombe: Changed for Wasserstein

src/python/gudhi/wasserstein/quantization.py
tlacombe: Done

tlacombe:
I just realized that I never managed to do the last requested modifications (my local build was broken for some reason at that time). 
I finally did it. 
As I'm working on a new machine, I hope I managed correctly the fork/branching/etc.

src/python/doc/wasserstein_distance_user.rst
```diff
@@ -193,9 +193,82 @@ The output is:
...
+In this experiment, Rips persistence diagrams (in homology dimension 1) are built from random point clouds supported on
+different tori with some additional noise.
+Starting from an initial codebook ``c0``, centroids are iteratively updated as new diagrams are provided.
+As we use the Wasserstein metrics between persistence diagrams (denoted here by :math:`\mathrm{OT}_2`), points in the
+diagrams that are close to the diagonal do not interfere in the codebook update process.
+Doing so, the final codebook is able to properly retrieve the two 1-dimensional features of the underlying points clouds
+(the two loops generating the tori).
+
+.. figure::
+     ./img/quantiz.gif
```
mglisse, line 234: On the one hand, the GIF is cool. On the other hand, I have trouble reading the doc with that thing moving on my screen...

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,189 @@
...
+    if X.shape[1] != 2:
+        raise ValueError("Input batch must be of shape (n x 2), not (%s,%s)" %(X.shape))
+
+    M = _utils._build_dist_matrix(X, c, order=2, internal_p=internal_p)  # Note: Order is useless here
+                                                                  # we just need nearest neighbors.
+
+    if withdiag:
+        a = np.argmin(M[:-1, :], axis=1)
+    else:
+        a = np.argmin(M[:-1, :-1], axis=1)
```
mglisse, line 55: It feels a bit strange to call `_build_dist_matrix`, whose main difference with `cdist` is that it adds the diagonal, just to drop the diagonal immediately... But I don't think it really matters.

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,189 @@
...
+    (in practice, concatenates) on the corresponding batch of diagrams.
+    Note that we may encounter empty diagrams, hence the specific check .ndim==2.
+        If the resulting list is empty (all diagrams are empty), returns empty array.
+    """
+    list_of_non_empty_diags = [pdiagset[i] for i in batches_indices if pdiagset[i].ndim==2]
+    if list_of_non_empty_diags: # There are some non-empty diagrams
+        X_batch = np.concatenate(list_of_non_empty_diags)
+        return X_batch
+    else:
+        return np.array([])
```
mglisse, line 77: It is sometimes useful to force the shape of empty arrays, to `(0,2)` for instance. I don't know if that's the case here.

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,189 @@
...
+    :param batch_size: Size of batches used during the online exploration of the ``pdiagset``.
+                        Default is ``1``.
+    :type batch_size: ``int``
+    :param order: Order of the Wasserstein distance we want to minimize.
+                  Only 2 is implemented for now. Default is ``2.``.
+    :type order: ``float``
+    :param internal_p: Ground metric to assess centroid affectation. Default is ``2.``.
+    :type internal_p: ``float``
+
+    :returns: The final codebook obtained after going through the all pdiagset.
```
mglisse, line 131: `:rtype: kx2 numpy array`?

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,189 @@
...
+    if list_of_non_empty_diags: # There are some non-empty diagrams
+        X_batch = np.concatenate(list_of_non_empty_diags)
+        return X_batch
+    else:
+        return np.array([])
+
+
+def _init_c(pdiagset, k, internal_p=2):
+    """
+    A naive heuristic to initialize a codebook: we take the k points with largest distances to the diagonal
```
mglisse, line 82: What if the first diagram has fewer than k points?

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,189 @@
...
+    :param pdiagset: a set of persistence diagrams that we want to quantize.
+    :type pdiagset: list of n x 2 numpy arrays
+    :param k: number of centroids. Default is ``2``. A naive heuristic is used to initialize the codebook.
+              Not used if an initial codebook ``c0`` has been provided.
+    :type k: ``int``
+    :param init: If provided, overwrite the provided ``k`` (which is now ``c0.shape[0]``), and is used as an
+                initialization for the algorithm. Default is ``None``.
+    :type init: (n x 2) ``numpy.array``, or ``None``.
+    :param batch_size: Size of batches used during the online exploration of the ``pdiagset``.
+                        Default is ``1``.
```
mglisse, line 123: As a user, should I stick to the default value of 1? If I already have all the diagrams, I may think that I don't need an online algorithm, which is for when data appears progressively, and consider using one huge batch under the impression that it disables the "online" stuff and gets the best result.

src/python/gudhi/wasserstein/quantization.py
```diff
@@ -0,0 +1,189 @@
...
+                if lc1 > 0:
+                    # We compute the "gradient", i.e. the optimal descent direction for c_current[j]
+                    # When order == 2, it just consists of pushing c_current[j] towards the barycenter
+                    # of the cells[j] it defines.
+                    grad = np.mean(c_current[j] - cells[j], axis=0)
+                    # We then apply the gradient, with a 1/(1+t) factor to guarantee the convergence of this
+                    # stochastic-gradient-descent like approach (decreasing learning rate).
+                    c_current[j] = c_current[j] - grad / (t + 1)
+        else:
+            raise NotImplemented('Order = %s is not available yet. Only order=2. is valid' %order)
```
mglisse, line 187: I think you could error out earlier (or not provide this option at all and just say that it is W2).

mglisse (commented):
The algorithm is presented as on online algorithm. So it should be normal to give it some data, look at the codebook at that point, pass it more data, look at the updated codebook, etc. The `init` parameter could be used towards that goal, but the number of diagrams (or batches) already processed is forgotten, and indeed `t` (the learning rate) is reset to 0 at every call.

PR #537, opened 2021-10-15
Documentation experiments for learning

VincentRouvreau:
A proposal to organize documentation for learning interfaces and layers


mglisse (commented):
IIUC, the issue is that we have a set A of features (persistence of Rips, persistence of cubical, etc) and a set B of interface types ("old", sklearn, tensorflow, etc), that are essentially orthogonal, although not 100%. From a documentation perspective, we can have one page per element of AxB, which in this example are linked from index.html, which has one section for each element of A, and in each section one link per B. We could also have a single page for an element of A, with the concatenated doc for the various interfaces B that provide this functionality. Or a single page per element of B, with all the features available for this interface. Of course one page can reference other pages (the tensorflow page for alpha-complex can tell users to read the user manual of the old interface for details, and that page can link to the various interfaces for this feature). And we can have mixes, say pages for AxB, but also an introductory page for each B with a longer example (a whole sklearn pipeline for instance).

I think I am ok with having many pages (AxB). It may be good to have more cross-links (the tensorflow version of cubical could have the same links that the cubical user manual has at the top (where a link to the tensorflow version is missing by the way)). Then we might want an introductory page for tensorflow and one for sklearn, a bit like the "user manual" pages, although I am not sure exactly where we would link it from (from cubical-tensorflow, but not from the main page?), although it would be redundant with a well-written tutorial. As mentioned elsewhere, I am not convinced we should put a whole sklearn pipeline example on each page (as opposed to just in an introductory sklearn page), although if others like it, ok.

VincentRouvreau:
It was just a branch for experimentation

PR #549, opened 2021-11-30
Fix this boost deprecated include

VincentRouvreau:


mglisse (approved):
(assuming it works)
Note that CGAL now requires Boost 1.66+ (https://doc.cgal.org/latest/Manual/thirdparty.html), so we could also simplify things and require the same.

VincentRouvreau:
> (assuming it works) Note that CGAL now requires Boost 1.66+ (https://doc.cgal.org/latest/Manual/thirdparty.html), so we could also simplify things and require the same.

Ok, thanks. The dependency modification has been done on CGAL 5.1, let's wait a little bit.
We should also update our documentation about it. I open an issue.

PR #571, opened 2022-01-14
Try this parallelization of appveyor builds

VincentRouvreau:


VincentRouvreau:
Does not work. Appveyor free plan is only one concurrent job per account. We should try another one.

PR #582, opened 2022-02-10
Add SimplexTree copy method and its test

### LINKED ISSUE #131: [python - SimplexTree] SimplexTree.copy()
It could be useful to have a way to clone a simplex tree, more efficiently than a loop on get_filtration doing inserts (we have a copy constructor in C++).

VincentRouvreau:
Fix #131 

This PR proposes this interface:
```python
    from gudhi import SimplexTree
    st = SimplexTree()
    st .insert([1,2,3], 0.)
    a = st.copy()
```

As suggested on this [StackOverflow](https://stackoverflow.com/a/37264151/5092461), this interface is also available:
```python
    from gudhi import SimplexTree
    st = SimplexTree()
    st .insert([1,2,3], 0.)
    a = SimplexTree(st)
```

Also good to know:
* Shallow copy is not available
* Persistence copy is not performed.


src/python/gudhi/simplex_tree.pyx
```diff
@@ -38,17 +39,36 @@ cdef class SimplexTree:
...
+    def __cinit__(self, other = None, copy = True):
+        cdef SimplexTree ostr
+        if other and type(other) is SimplexTree:
+            ostr = <SimplexTree> other
+            if copy:
+                self.thisptr = <intptr_t>(new Simplex_tree_interface_full_featured(dereference(ostr.get_ptr())))
+            else:
+                self.thisptr = ostr.thisptr
+                # Avoid double free - The original is in charge of deletion
+                self.__thisptr_to_be_deleted = False
```
mglisse, line 63: What if the original disappears before the copy?

mglisse (commented):
I see now, the shallow copy thing looks messy and rather useless, I think you were right not to do it in the beginning.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -38,13 +38,27 @@ cdef class SimplexTree:
...
     # The real cython constructor
-    def __cinit__(self):
-        self.thisptr = <intptr_t>(new Simplex_tree_interface_full_featured())
+    def __cinit__(self, other = None):
+        cdef SimplexTree ostr
+        if other and type(other) is SimplexTree:
+            ostr = <SimplexTree> other
+            self.thisptr = <intptr_t>(new Simplex_tree_interface_full_featured(dereference(ostr.get_ptr())))
+        else:
+            self.thisptr = <intptr_t>(new Simplex_tree_interface_full_featured())
```
mglisse, line 61: If I try to construct a SimplexTree from something random, say a string, it will construct an empty SimplexTree without complaining?

VincentRouvreau: You are right that it can lead to misunderstanding like `SimplexTree([1,2,3])`, ...
I moodified it to raise an exception on d5ac245

src/python/gudhi/simplex_tree.pyx
```diff
@@ -63,6 +81,25 @@ cdef class SimplexTree:
          """
         return self.pcohptr != NULL
 
+    def copy(self):
+        """ 
+        :returns: A simplex tree that is a deep copy of itself.
+        :rtype: SimplexTree
+
+        :note: copy requires :func:`compute_persistence` to be launched again as the persistence result is not copied.
```
mglisse, line 89: The formulation looks a bit too strong, people may use the SimplexTree for other purposes and never care about persistence. Maybe "the persistence information is not copied. If you need it in the clone, you have to call `compute_persistence` on it even if you had already computed it in the original."?

src/python/gudhi/simplex_tree.pyx
```diff
@@ -38,13 +38,31 @@ cdef class SimplexTree:
...
+            persistence result is not copied.
         """
 
     # The real cython constructor
-    def __cinit__(self):
-        self.thisptr = <intptr_t>(new Simplex_tree_interface_full_featured())
+    def __cinit__(self, other = None):
+        cdef SimplexTree ostr
+        if other:
+            if type(other) is SimplexTree:
```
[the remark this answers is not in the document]
VincentRouvreau: Replace with `isinstance` on 4e63e52

src/python/gudhi/simplex_tree.pyx
VincentRouvreau: Re-formulated on e7ba896

src/python/gudhi/simplex_tree.pyx
```diff
@@ -38,13 +38,31 @@ cdef class SimplexTree:
...
+                ostr = <SimplexTree> other
+                self.thisptr = <intptr_t>(new Simplex_tree_interface_full_featured(dereference(ostr.get_ptr())))
```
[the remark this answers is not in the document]
VincentRouvreau: Copy simplex tree c++ pointer in a factorized `nogil` function on 41a976c

mglisse (approved):

PR #583, opened 2022-02-11
Debug traces for python module

VincentRouvreau:
`cmake -DDEBUG_TRACES=ON .` activates debug traces for C++, but not for python.
This modification fixes this behaviour by compiling python module with `-DDEBUG_TRACES`


mglisse (approved):
if you've checked that the flag appears in the right places in the log.

PR #584, opened 2022-02-14
A bug in the Entropy function has been solved

Soriano-Trigueros:
The MaxAbsScaler is no longer used to calculate PE. Now, the diagram is divided by the total sum L. 

The function test_entropy_miscalculation have been added to src/python/test/test_representations.py to check if the calculations of entropy are right.

### DIFF
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -532,7 +532,7 @@ class Entropy(BaseEstimator, TransformerMixin):
                     for k in range(min_idx, max_idx):
                         ent[k] += (-1) * p[j] * np.log(p[j])
                 if self.normalized:
-                    ent = ent / (np.linalg.norm(ent, ord=1))
+                    ent = ent / np.linalg.norm(ent, ord=1)
                 Xfit.append(np.reshape(ent,[1,-1]))
 
         Xfit = np.concatenate(Xfit, axis=0)
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -510,16 +510,19 @@ class Entropy(BaseEstimator, TransformerMixin):
         for i in range(num_diag):
             orig_diagram, diagram, num_pts_in_diag = X[i], new_X[i], X[i].shape[0]
             try:
-                new_diagram = DiagramScaler(use=True, scalers=[([1], MaxAbsScaler())]).fit_transform([diagram])[0]
+                #new_diagram = DiagramScaler(use=True, scalers=[([1], MaxAbsScaler())]).fit_transform([diagram])[0]
+                new_diagram = DiagramScaler().fit_transform([diagram])[0]
             except ValueError:
                 # Empty persistence diagram case - https://github.com/GUDHI/gudhi-devel/issues/507
                 assert len(diagram) == 0
                 new_diagram = np.empty(shape = [0, 2])
-
+                
+            p = new_diagram[:,1]
+            L = sum(p)
+            p = p/L
             if self.mode == "scalar":
-                ent = - np.sum( np.multiply(new_diagram[:,1], np.log(new_diagram[:,1])) )
+                ent = -np.dot(p, np.log(p))
                 Xfit.append(np.array([[ent]]))
-
             else:
                 ent = np.zeros(self.resolution)
                 for j in range(num_pts_in_diag):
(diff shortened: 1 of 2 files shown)


mglisse:
As can be seen in the continuous integration, the test fails

    >       assert [[pe_max(diag_ex)]] == sce.fit_transform([diag_ex])
    E       NameError: name 'pe_max' is not defined

Soriano-Trigueros:
Sorry about that. It should work now. 
I have also added the cases Entropy(mode="vector", normalized=False) and Entropy(mode="vector",  normalized=True) to the test.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -510,24 +510,27 @@ def transform(self, X):
...
+                new_diagram = DiagramScaler().fit_transform([diagram])[0]
             except ValueError:
                 # Empty persistence diagram case - https://github.com/GUDHI/gudhi-devel/issues/507
                 assert len(diagram) == 0
                 new_diagram = np.empty(shape = [0, 2])
-
+                
+            p = new_diagram[:,1]
+            L = sum(p)
+            p = p/L
```
mglisse, line 522: I'd rather avoid the generic `sum` and use numpy's version, so `p=p/np.sum(p)` (no need to name `L` if it is only used once), or a version with `np.linalg.norm`.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -510,24 +510,27 @@ def transform(self, X):
...
             if self.mode == "scalar":
-                ent = - np.sum( np.multiply(new_diagram[:,1], np.log(new_diagram[:,1])) )
+                ent = -np.dot(p, np.log(p))
```
mglisse, line 524: I like this change, it is more readable. We could even use scipy.stats.entropy here, but not a few lines below, and the inconsistency would be bad, so let's stick to `dot`.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -510,24 +510,27 @@ def transform(self, X):
...
-
             else:
                 ent = np.zeros(self.resolution)
                 for j in range(num_pts_in_diag):
                     [px,py] = orig_diagram[j,:2]
                     min_idx = np.clip(np.ceil((px - self.sample_range[0]) / step_x).astype(int), 0, self.resolution)
                     max_idx = np.clip(np.ceil((py - self.sample_range[0]) / step_x).astype(int), 0, self.resolution)
                     for k in range(min_idx, max_idx):
-                        ent[k] += (-1) * new_diagram[j,1] * np.log(new_diagram[j,1])
+                        ent[k] += (-1) * p[j] * np.log(p[j])
```
mglisse, line 533: `ent[min_idx:max_idx]-=p[j]*np.log(p[j])` maybe?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -510,24 +510,27 @@ def transform(self, X):
         for i in range(num_diag):
             orig_diagram, diagram, num_pts_in_diag = X[i], new_X[i], X[i].shape[0]
             try:
-                new_diagram = DiagramScaler(use=True, scalers=[([1], MaxAbsScaler())]).fit_transform([diagram])[0]
+                #new_diagram = DiagramScaler(use=True, scalers=[([1], MaxAbsScaler())]).fit_transform([diagram])[0]
+                new_diagram = DiagramScaler().fit_transform([diagram])[0]
```
mglisse, line 514: Does DiagramScaler do anything with no scalers?

mglisse (commented):
Mathieu will tell us if he agrees with removing the scaler, but here are a few comments in the mean time.

src/python/gudhi/representations/vector_methods.py
Soriano-Trigueros: I agree p=p/np.sum(p) is better

src/python/gudhi/representations/vector_methods.py
Soriano-Trigueros: Yes, that is more readeable.
I just realized it should be `ent[min_idx:max_idx+1]-=p[j]*np.log(p[j])` right? Otherwise the last point is always going to be zero

src/python/gudhi/representations/vector_methods.py
Soriano-Trigueros: I was not sure of this part, let's see what Mathieu thinks

src/python/gudhi/representations/vector_methods.py
mglisse: Are you sure? `ent` goes from `0` to `resolution-1`, and `clip(*,0,resolution)` may produce `resolution`.

Soriano-Trigueros: right, I was confused

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: Sorry for the very late reply. Yes, this line is useless now, you can remove it.

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: I like it too!

MathieuCarriere (approved):

Soriano-Trigueros:
Great,
I have removed the try-except blocks were the DiagramScaler was called, since it was doing nothing. Tell me if it is ok now

MathieuCarriere:
It looks good to me, thanks!

PR #596, opened 2022-03-31
Float relative precision alpha python

### LINKED ISSUE #361: [AlphaComplex] set_relative_precision_of_to_double
`CGAL::Lazy_exact_nt<NT>::set_relative_precision_of_to_double` can have a significant impact on the performance of AlphaComplex (changing it from 10^-5 to 10^-3 gained a factor 2 on a file provided by Vincent). It may be useful to interface it in python, with a simpler name.

VincentRouvreau:
Fix #361 
Also remove `thread_local double RELATIVE_PRECISION_OF_TO_DOUBLE = 0.00001;` in Alpha complex 3d that was useless.
Requires cython >= 0.21 for `@staticmethod` (cf. [release note](https://cython.readthedocs.io/en/latest/src/changes.html#id186))

### DIFF
--- a/src/python/gudhi/alpha_complex.pyx
+++ b/src/python/gudhi/alpha_complex.pyx
@@ -31,6 +31,10 @@ cdef extern from "Alpha_complex_interface.h" namespace "Gudhi":
         Alpha_complex_interface(vector[vector[double]] points, vector[double] weights, bool fast_version, bool exact_version) nogil except +
         vector[double] get_point(int vertex) nogil except +
         void create_simplex_tree(Simplex_tree_interface_full_featured* simplex_tree, double max_alpha_square, bool default_filtration_value) nogil except +
+        @staticmethod
+        void set_float_relative_precision(double precision) nogil
+        @staticmethod
+        double get_float_relative_precision() nogil
 
 # AlphaComplex python interface
 cdef class AlphaComplex:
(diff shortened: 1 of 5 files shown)


src/python/gudhi/alpha_complex.pyx
```diff
@@ -31,6 +31,10 @@ cdef extern from "Alpha_complex_interface.h" namespace "Gudhi":
         Alpha_complex_interface(vector[vector[double]] points, vector[double] weights, bool fast_version, bool exact_version) nogil except +
         vector[double] get_point(int vertex) nogil except +
         void create_simplex_tree(Simplex_tree_interface_full_featured* simplex_tree, double max_alpha_square, bool default_filtration_value) nogil except +
+        @staticmethod
+        void set_float_relative_precision(double precision) nogil
```
mglisse, line 35: IIRC for super cheap functions, it is better not to write `nogil`, because releasing and taking the lock again costs more than the function itself. I could be wrong though.

src/python/gudhi/alpha_complex.pyx
```diff
@@ -133,3 +137,27 @@ cdef class AlphaComplex:
...
+        :type precision: float
+        """
+        if precision <=0. or precision >= 1.:
+            raise ValueError("Relative precision value must be strictly greater than 0 and strictly lower than 1")
+        Alpha_complex_interface.set_float_relative_precision(precision)
+    
+    @staticmethod
+    def get_float_relative_precision():
+        """
+        :returns: The float relative precision of filtration values computation in
```
mglisse, line 159: Maybe repeat the fact that it is only for `'safe'`?

mglisse (approved):

src/python/gudhi/alpha_complex.pyx
VincentRouvreau: `nogil` removed on f4b543b

src/python/gudhi/alpha_complex.pyx
VincentRouvreau: Explain the float relative precision is taken into account only for `safe` mode also for the getter on f4b543b

src/python/gudhi/alpha_complex.pyx
mglisse: Actually, I was remembering wrong. Writing `nogil` on those declarations only means it is allowed to call them without the GIL, but it doesn't automatically release the GIL, so you were right to write `nogil`. Could you add it back please? Sorry...

VincentRouvreau: I double checked, and it seems releasing the GIL is just like setting a variable and should not take a lot of time. I think we could use `nogil` for all.
Reverted on 25172af

PR #598, opened 2022-04-05
Make Windows compilation fails on error

VincentRouvreau:
Compiles with '-DEIGEN_DEFAULT_DENSE_INDEX_TYPE=int' as a workaround.


mglisse:
Alternative possible workaround: `-DEIGEN_DEFAULT_DENSE_INDEX_TYPE=int` (or `long`, but definitely not `long long`).

PR #604, opened 2022-05-10
Fix limit tests for plot (and warning in test)

VincentRouvreau:


src/python/test/test_persistence_graphical_tools.py
```diff
@@ -98,8 +98,12 @@ def test_limit_to_max_intervals():
 def _limit_plot_persistence(function):
     pplot = function(persistence=[()])
     assert issubclass(type(pplot), plt.axes.SubplotBase)
+    pplot = function(persistence=[()], legend=True)
```
mglisse, line 101: `[()]` looks like an invalid value for `persistence` according to the doc. The empty case would be `[]`. Am I missing something?

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -321,11 +326,16 @@ def plot_persistence_diagram(
         # line display of equation : birth = death
         axes.plot([axis_start, axis_end], [axis_start, axis_end], linewidth=1.0, color="k")
 
-        x=[birth for (dim,(birth,death)) in persistence]
-        y=[death if death != float("inf") else infinity for (dim,(birth,death)) in persistence]
-        c=[colormap[dim] for (dim,(birth,death)) in persistence]
+        non_empty_diagram = len(persistence[0]) > 0
```
mglisse, line 329: Why does it look at `len(persistence[0])`? I could understand `len(persistence)`, although I would be a bit surprised.

src/python/test/test_persistence_graphical_tools.py
VincentRouvreau: You are completely right. I modified code and test on [393216c](https://github.com/GUDHI/gudhi-devel/pull/604/commits/393216cc99af79526b1bb90ae4da21116884bcbd)

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: You are completely right. I modified code and test on [393216c](https://github.com/GUDHI/gudhi-devel/pull/604/commits/393216cc99af79526b1bb90ae4da21116884bcbd)

mglisse (approved):

Issue #575, opened 2022-01-20
When cmake is launched with `-DWITH_GUDHI_PYTHON=OFF`, cmake shall not search python and its modules


VincentRouvreau:
`WITH_GUDHI_PYTHON` is an option set in `GUDHI_modules.cmake`.
`find_package( PythonInterp )`, `find_python_module("xxx")` and `find_program( SPHINX_PATH sphinx-build )` in `GUDHI_third_party_libraries.cmake` could be skipped when the user explicitely launch cmake with the option `-DWITH_GUDHI_PYTHON=OFF`.

PR #611, opened 2022-05-18
Documentation: make it easier to build only the documentation

albert-github:
Introduce the possibility:
```
-Dbuild_documentation_only=ON
```
so that only the doxygen documentation can be build and we don't have to resort to installing all kinds of sub packages like Boost or manipulate the CMake files:w

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-option(build_documentation_only    "Build only the documentation with doxygen." OFF)
+option(WITH_GUDHI_CPP_DOCUMENTATION_ONLY    "Build only the GUDHI C++ documentation (with doxygen)." OFF)
 
 cmake_minimum_required(VERSION 3.5)
 
@@ -16,7 +16,7 @@ set(GUDHI_MISSING_MODULES "" CACHE INTERNAL "GUDHI_MISSING_MODULES")
 # This variable is used by Cython CMakeLists.txt and by GUDHI_third_party_libraries to know its path
 set(GUDHI_PYTHON_PATH "src/python")
 
-if (NOT build_documentation_only)
+if (NOT WITH_GUDHI_CPP_DOCUMENTATION_ONLY)
   # For third parties libraries management - To be done last as CGAL updates CMAKE_MODULE_PATH
   include(GUDHI_third_party_libraries NO_POLICY_SCOPE)
 endif()
@@ -57,7 +57,7 @@ foreach(GUDHI_MODULE ${GUDHI_MODULES})
   endforeach()
 endforeach()
 
-if (NOT build_documentation_only)
+if (NOT WITH_GUDHI_CPP_DOCUMENTATION_ONLY)
   add_subdirectory(src/GudhUI)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,6 @@
+
+option(build_documentation_only    "Build only the documentation with doxygen." OFF)
+
 cmake_minimum_required(VERSION 3.5)
 
 project(GUDHIdev)
@@ -13,8 +16,10 @@ set(GUDHI_MISSING_MODULES "" CACHE INTERNAL "GUDHI_MISSING_MODULES")
 # This variable is used by Cython CMakeLists.txt and by GUDHI_third_party_libraries to know its path
 set(GUDHI_PYTHON_PATH "src/python")
 
-# For third parties libraries management - To be done last as CGAL updates CMAKE_MODULE_PATH
-include(GUDHI_third_party_libraries NO_POLICY_SCOPE)
+if (NOT build_documentation_only)
+  # For third parties libraries management - To be done last as CGAL updates CMAKE_MODULE_PATH
+  include(GUDHI_third_party_libraries NO_POLICY_SCOPE)
+endif()
 
 include(GUDHI_compilation_flags)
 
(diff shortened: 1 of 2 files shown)


mglisse:
Note that we use doxygen to document the C++ code, but we use sphinx for the python code. The name of the option needs to reflect that somehow and not give the impression that it will build the python sphinx doc. We also need to document it (in src/common/doc/installation.h?). @VincentRouvreau I guess it would make sense to update the CI that runs doxygen to use this option at the same time?

albert-github:
From the doxygen side it is a pity that the documentation of the python code is done by another package (resulting in some different look and feel). It will be too much work to change this quickly as for documentation the python code some sort of rst convention is used.

As there are 2 parts of documentation
- the part, C++, with doxygen
- the, python, part with sphinx

It would indeed be goo to make this more explicit:
- in the name of the option, do you have a suggestion?
  my initial name was `build_documentation_only`, suggestions
  - `build_doxygen_documentation_only`
  - `build_cpp_documentation_only`
  - `build_cpp_doxygen_documentation_only`
- adjust the accompanying help text in the `CMakeLists.txt` file
- adjust the part in the `installation.h`, here there is currently also no distinction between the C++ part and the python part (due to my lack of knowledge I won't be able to write something there about the python / sphinx part).

@VincentRouvreau will probably get my comments as well, but to be sure I added him here as well.

Awaiting the the suggestion for the name.

albert-github:
In the comment https://github.com/GUDHI/gudhi-devel/pull/611#issuecomment-1130401420 also the possibility of CI was mentioned. For a CI environment, but in general as well it might be useful to look at the following settings:
- `WARN_LOGFILE= warnings.log`
  so all the warnings are recorded in a file and are not "hidden" between "progress" messages.

- `WARN_AS_ERROR = FAIL_ON_WARNINGS`
  > If the `WARN_AS_ERROR tag` is set to `YES` then doxygen will immediately stop when a warning
is encountered. If the `WARN_AS_ERROR` tag is set to `FAIL_ON_WARNINGS` then doxygen will continue
running as if `WARN_AS_ERROR` tag is set to `NO`, but at the end of the doxygen process doxygen will return
with a non-zero status.
  >
  > Possible values are: `NO`, `YES` and `FAIL_ON_WARNINGS`.
  >
  > The default value is: `NO`.


- `QUIET = YES`
  to suppress the "progress"  messages to the console.

Or of course a combination of these.

VincentRouvreau:
First of all, thanks for the contribution ! I already improved the doxygen documentation generation with #575 (to avoid all the python cmake stuff) but your PR goes a step further.

> From the doxygen side it is a pity that the documentation of the python code is done by another package (resulting in some different look and feel). It will be too much work to change this quickly as for documentation the python code some sort of rst convention is used.

I understand your purpose, but having a different look and feel for python and c++ is something that we like, because it avoids to mix the 2 versions. But maybe we could have something more DRY by using doxygen for python and c++. Anyway.
I confess that I was not aware doxygen could document python code when I did it with sphinx and I was really liking the sphinx doctest snippet that allows to check the examples in the documentation are working. But I regret using rst (even if python recommend it). I really prefer markdown for several reasons.

> As there are 2 parts of documentation
> 
>     * the part, C++, with doxygen
> 
>     * the, python, part with sphinx
> 
> 
> It would indeed be goo to make this more explicit:
> 
>     * in the name of the option, do you have a suggestion?
>       my initial name was `build_documentation_only`, suggestions
>       
>       * `build_doxygen_documentation_only`
>       * `build_cpp_documentation_only`
>       * `build_cpp_doxygen_documentation_only`

We have a kind of convention for CMake options (not written anywhere but anyway), and they should be in [GUDHI_options.cmake](https://github.com/GUDHI/gudhi-devel/blob/master/src/cmake/modules/GUDHI_options.cmake) file to be available from the "user" and the "dev" version of gudhi.
`WITH_GUDHI_CPP_DOCUMENTATION_ONLY` and `WITH_GUDHI_DOXYGEN_DOCUMENTATION_ONLY` are ok to me.

>     * adjust the accompanying help text in the `CMakeLists.txt` file
> 
>     * adjust the part in the `installation.h`, here there is currently also no distinction between the C++ part and the python part (due to my lack of knowledge I won't be able to write something there about the python / sphinx part).

I think we should just mention somewhere around [these lines in installation.h](https://github.com/GUDHI/gudhi-devel/blob/c5ade8141058b9cd1f9a1ce862e67431df001e9a/src/common/doc/installation.h#L44-L48) that adding `-DWITH_GUDHI_CPP_DOCUMENTATION_ONLY=ON` allows the user to build the documentation without having to install any other third parties than cmake and doxygen.

VincentRouvreau:
> @VincentRouvreau I guess it would make sense to update the CI that runs doxygen to use this option at the same time?

Yes I will do it on another PR when this one will be merged.

VincentRouvreau:
> In the comment [#611 (comment)](https://github.com/GUDHI/gudhi-devel/pull/611#issuecomment-1130401420) also the possibility of CI was mentioned. For a CI environment, but in general as well it might be useful to look at the following settings:
> 
>     * `WARN_LOGFILE= warnings.log`
>       so all the warnings are recorded in a file and are not "hidden" between "progress" messages.
> 
>     * `WARN_AS_ERROR = FAIL_ON_WARNINGS`
>       > If the `WARN_AS_ERROR tag` is set to `YES` then doxygen will immediately stop when a warning
>       > is encountered. If the `WARN_AS_ERROR` tag is set to `FAIL_ON_WARNINGS` then doxygen will continue
>       > running as if `WARN_AS_ERROR` tag is set to `NO`, but at the end of the doxygen process doxygen will return
>       > with a non-zero status.
>       > Possible values are: `NO`, `YES` and `FAIL_ON_WARNINGS`.
>       > The default value is: `NO`.
> 
>     * `QUIET = YES`
>       to suppress the "progress"  messages to the console.
> 
> 
> Or of course a combination of these.

Oh thank you for these good hints ! I was doing it with bash command, but your proposals are much more modern way to do it. I will keep it in mind when I will do the CI stuff.

PR #628, opened 2022-05-23
Documentation: Redirecting some links properly to the gudhi website

albert-github:
In the code (mostly the top bar) there are links to links like `../../licensing/`, this works OK on the website but in the user generated local documentation this gives an unknown link.
- properly create the links as `https://gudhi.inria.fr/licensing/`
- redirect the pages from the website to their own tab / iframe by means of `target="gudhi_web"`, so the manual is still properly visible. The `target="-blank"` hasn't been used as in that case each time when a link was clicked from the manual a new tab would be opened. (see also: https://www.w3schools.com/tags/att_a_target.asp)

### DIFF
--- a/src/Persistent_cohomology/doc/Intro_persistent_cohomology.h
+++ b/src/Persistent_cohomology/doc/Intro_persistent_cohomology.h
@@ -142,7 +142,7 @@ diagram.
 3  1 0.104347 inf 
 3  2 0.138335 inf \endcode
 
-More details on the <a href="../../ripscomplex/">Rips complex utilities</a> dedicated page.
+More details on the <a href="https://gudhi.inria.fr/ripscomplex/" target="gudhi_web">Rips complex utilities</a> dedicated page.
 
 \li <a href="rips_multifield_persistence_8cpp-example.html">
 Persistent_cohomology/rips_multifield_persistence.cpp</a> computes the Rips complex of a point cloud and outputs its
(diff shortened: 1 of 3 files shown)


albert-github:
Maybe I'm missing the point but wouldn't those people also not have the problem that `../../licensing/` etc. points to nowhere and with this it would point to a more sensible place?

VincentRouvreau:
> Maybe I'm missing the point but wouldn't those people also not have the problem that `../../licensing/` etc. points to nowhere and with this it would point to a more sensible place?

Yes you are correct, but there is a good chance that  `../../licensing/` would give an error.

Another example that comes to my mind is that the documentation is generated as an artifact on our CI, and we want to be able to browse it.
For example, this  PR generated the doxygen documentation that is [available here](https://output.circle-artifacts.com/output/job/8dda8943-f186-435a-8fa1-492d1dc2cd06/artifacts/0/doxygen/index.html), and we don't want that when we click on a link to be redirected on https://gudhi.inria.fr

albert-github:
Why shouldn't the artifact, when used by a user, be redirected to https://gudhi.inria.fr/ or do you expect to much traffic or is there some automatic testing that would be problematic?

VincentRouvreau:
Redirection is not practical for our use cases (CI, https://gudhi.github.io/ , ...).
We will keep relative links in our documentations.

PR #630, opened 2022-05-23
Documentation: Render formulas by means of MathJax 3

albert-github:
- create entries in Doxyfile.in for the use of MathJax
- adjust formula so it is properly rendered in MathJax 3 (formula will also work with MathJax 2)

(available in doxygen since doxygen 1.9.2)


mglisse:
What happens if a developer builds the doc locally with an older version of doxygen? Is the unknown option ignored and it works fine? I am ok (if Vincent is) with using a recent version of doxygen for the official build of the doc, but it seems helpful not to make things harder for developers, too many already fail to check their doc before filing a PR.

albert-github:
The `MATHJAX_VERSION` will give a warning like:
```
warning: ignoring unsupported tag 'MATHJAX_VERSION' at line 2, file Doxyfile
```
the `MATHJAX_EXTENSIONS` is more problematic as these are not compatible and hence a line like
```
An AMS symbol \f$ \between \f$
```
will not render in MathJax 2 when `MATHJAX_EXTENSIONS = ams`

I've added a small example for testing purposes: [example.tar.gz](https://github.com/GUDHI/gudhi-devel/files/8757012/example.tar.gz)


When it is still to early to integrate this patch it is, probably, a good reference for the future.

mglisse:
> When it is still to early to integrate this patch it is, probably, a good reference for the future.

Indeed, the question is **when** we will merge this, but we eventually will.

albert-github:
I think I found a good solution so that old and new version are working.

src/cmake/modules/GUDHI_doxygen_target.cmake
```diff
@@ -44,6 +44,13 @@ if(DOXYGEN_FOUND)
     set(GUDHI_DOXYGEN_UTILS_PATH "utilities/*")
   endif()
 
+  if (DOXYGEN_VERSION VERSION_LESS 1.9.2)
```
VincentRouvreau, line 47: If you are ok with it @albert-github , I would display doxygen version (for debug purpose)
```suggestion
  message("++ Doxygen version ${DOXYGEN_VERSION}")
  if (DOXYGEN_VERSION VERSION_LESS 1.9.2)
```

VincentRouvreau (approved):

PR #631, opened 2022-05-23
Extending build directories to ignore

albert-github:
Besides the standard `build/` directory it is better to ignore `build*/` as when one has multiple build directories they can have different names (e.g. multi platform environment)


VincentRouvreau:
Obsolete by #648

PR #633, opened 2022-05-24
No need to copy Doxyfile.in

albert-github:
In my opinion there is no need to copy the `Doxyfile.in` file to the build directory (neither in the User nor the development version).
The usage is:
```
configure_file(${GUDHI_DOXYGEN_SOURCE_PREFIX}/Doxyfile.in "${CMAKE_CURRENT_BINARY_DIR}/Doxyfile" @ONLY)
```
in the file `src/cmake/modules/GUDHI_doxygen_target.cmake` and we see that this uses the file from its original source directory.


VincentRouvreau (approved):
As `Doxyfile.in` is copied in user version, there is no need to copy it in build directory

PR #636, opened 2022-05-25
Optimize silhouettes

wreise:
The purpose of this PR is to optimize the implementation of `.gudhi.representations.Silhouette.transform`. To make sure that the logic is still correct, I added a few tests.

In the end, I propose to use pykeops, because it is the fastest and uses the least memory (on the examples I run it on). See comments for a short benchmarking experiment.


mglisse:
Does the pykeops.torch version work, and does it impact performance?

wreise:
> [getkeops/keops#43](https://github.com/getkeops/keops/issues/43) Pykeops support on windows doesn't seem to be there yet, so I think we need at least to keep a fallback.

Ok, then seems like pykeops just makes it overcomplicated and is not such a good idea. I will propose [the numpy version](https://github.com/GUDHI/gudhi-devel/pull/636/files/1a76ecc3e7459e3461e1f182004362dcb663addd) or a C++ implementation with pybind (if that works from the user perspective).

mglisse:
> I used [the following script](https://gist.github.com/wreise/3e1d86c68377668ef4ff989f43caba41).

IIUC this benchmark uses points below the diagonal, which isn't supported, so it would be safer without those. Also, it contains many points far from the diagonal, which may favor the kind of implementation you are going for. It would be nice, for comparison, to test also with a diagram where you generate points only in a very narrow strip along the diagonal.

wreise:
You are right! I incorporated your suggestions in [the benchmarking script](https://gist.github.com/wreise/3e1d86c68377668ef4ff989f43caba41):
- all points are above the diagonal,
- a parameter, c, governs the concentration of points around the diagonal. There is a single point (0,2) and all the other points have persistence at most c.

Here are the results regarding the execution times.
image

image

I need to find a way to measure the memory consumption - even spawning separate processes, I get unexpected values (probably due to imports).

MathieuCarriere:
Thanks for the nice benchmark! Naive question: would your improvements also apply to landscapes?

mglisse:
The CI tests don't pass. Could you have a look first? In particular the one called "build" that runs on CircleCI.

wreise:
I changed the approach back to `numpy`.

I also experimented with a C++ implementation of silhouette. It's faster than numpy or keops, but I don't think I could keep the same api, where the user can specify a weight-function. The only option reasonable option would be to evaluate the weights in python.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -161,53 +161,23 @@ def transform(self, X):
...
-                events[j].sort(reverse=True)
-                for k in range( min(self.num_landscapes, len(events[j])) ):
-                    ls[k,j] = events[j][k]
+        Xfit = []
+        x_values = np.linspace(self.sample_range[0], self.sample_range[1], self.new_resolution)
+        for i, diag in enumerate(X):
+            midpoints, heights = (diag[:, 0] + diag[:, 1]) / 2., (diag[:, 1] - diag[:, 0]) / 2.
+            tent_functions = np.maximum(heights[None, :] - np.abs(x_values[:, None] - midpoints[None, :]), 0)
+            tent_functions.partition(diag.shape[0] - self.num_landscapes, axis=1)
+            landscapes = np.sort(tent_functions, axis=1)[:, -self.num_landscapes:][:, ::-1].T
```
mglisse, line 171: It looks like you don't take advantage of the partition and still sort the whole array?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -161,53 +161,23 @@ def transform(self, X):
...
-                        events[k].append(landscape_value)
-                        landscape_value -= step_x
-
-            for j in range(self.new_resolution):
-                events[j].sort(reverse=True)
-                for k in range( min(self.num_landscapes, len(events[j])) ):
-                    ls[k,j] = events[j][k]
+        Xfit = []
+        x_values = np.linspace(self.sample_range[0], self.sample_range[1], self.new_resolution)
+        for i, diag in enumerate(X):
```
mglisse, line 167: Do you use `i` anywhere?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -161,53 +161,23 @@ def transform(self, X):
...
             if self.nan_in_range[0]:
-                ls = ls[:,1:]
+                landscapes = landscapes[:, 1:]
             if self.nan_in_range[1]:
-                ls = ls[:,:-1]
-            ls = np.sqrt(2)*np.reshape(ls,[1,-1])
-            Xfit.append(ls)
-
-        Xfit = np.concatenate(Xfit,0)
+                landscapes = landscapes[:, :-1]
```
mglisse, line 176: Couldn't we do that when we define `x_values`?
Also, Silhouette doesn't seem to skip those extreme elements. Should it?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -235,6 +205,7 @@ def __init__(self, weight=lambda x: 1, resolution=100, sample_range=[np.nan, np.
             sample_range ([double, double]): minimum and maximum for the weighted average domain, of the form [x_min, x_max] (default [numpy.nan, numpy.nan]). It is the interval on which samples will be drawn evenly. If one of the values is numpy.nan, it can be computed from the persistence diagrams with the fit() method.
         """
         self.weight, self.resolution, self.sample_range = weight, resolution, sample_range
+        self.im_range = None
```
mglisse, line 208: Is that useful?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -257,44 +229,19 @@ def transform(self, X):
...
-        for i in range(num_diag):
-
-            diagram, num_pts_in_diag = X[i], X[i].shape[0]
-
-            sh, weights = np.zeros(self.resolution), np.zeros(num_pts_in_diag)
-            for j in range(num_pts_in_diag):
-                weights[j] = self.weight(diagram[j,:])
+        for i, diag in enumerate(X):
+            midpoints, heights = (diag[:, 0] + diag[:, 1]) / 2., (diag[:, 1] - diag[:, 0]) / 2.
+            weights = np.array([self.weight(point) for point in diag])
```
mglisse, line 237: `np.apply_along_axis`?

mglisse (commented):
For Silhouette, I wonder how an intermediate version (2 for loops) would perform: for each point, compute the subrange (in terms of indices) where the tent is positive, and add the tent (no need to truncate the negative part) in place to the relevant subrange of the accumulator.
(you don't have to try that, I just mention it so I can forget it)

src/python/gudhi/representations/vector_methods.py
wreise: No, my bad (editor refactoring suggestions)

src/python/gudhi/representations/vector_methods.py
wreise: Also possible! I'm changing that.

src/python/gudhi/representations/vector_methods.py
wreise: No, I remove the `enumerate`.

src/python/gudhi/representations/vector_methods.py
wreise: It's also an option. However, that function doesn't work well for empty arrays (error). The current version produces an array of shape `(0,)`.

src/python/gudhi/representations/vector_methods.py
wreise: I extracted this logic and also applied it in Silhouettes.
FYI: the behavior is not the same in BettiCurves.

mglisse: If we change the behavior, we will need a note in the release notes to warn users about it.
This was mostly a question. @MathieuCarriere @gspr do you have an opinion which ones among Lanscape, Silhouette, BettiCurve, etc, should skip initial and final 0 value when they are asked to guess the bounds? And do we need to tweak the resolution so the output has the expected shape in that case?

src/python/gudhi/representations/vector_methods.py
```diff
@@ -123,6 +123,15 @@ def _automatic_sample_range(sample_range, X, y):
                 pass
         return sample_range
 
+
+def trim_on_edges(x, are_endpoints_nan):
```
mglisse, line 127: I think the convention would be to call it `_trim_on_edges` with a leading underscore since it is not public.

src/python/gudhi/representations/vector_methods.py
wreise: Thanks for noticing! I changed that

src/python/gudhi/representations/vector_methods.py
gspr: Without having thought very thoroughly about this, I think I'm preferential to `BettiCurve` *keeping* the initial and final 0. I think it makes the interpretation a bit more self-explanatory, as all values the curve takes are kept.

mglisse: I think it really depends what you want to do with the output. In order to plot it, having the extreme 0s is convenient. If you are going to feed the output to some classifier or other general ML method, having a coordinate that is always 0 is a waste, there is nothing to learn from it.

gspr: > I think it really depends what you want to do with the output. In order to plot it, having the extreme 0s is convenient. If you are going to feed the output to some classifier or other general ML method, having a coordinate that is always 0 is a waste, there is nothing to learn from it.

True. But then again, the way these representations are documented (at least BettiCurve), they don't give any guarantees that for redundant information (such as repeated Betti numbers in a constant region for BettiCurve) is excluded. For BettiCurve, that happens to be the case, but I personally think that anyone feeding the output to some ML method should ask themselves whether it contains redundant information.

But again, I have no strong opinions on the matter, and don't think that removing the zeros is unreasonable either :-)

src/python/gudhi/representations/vector_methods.py
```diff
@@ -161,53 +172,27 @@ def transform(self, X):
...
+        Xfit = []
+        x_values = self.im_range
+        for diag in X:
+            midpoints, heights = (diag[:, 0] + diag[:, 1]) / 2., (diag[:, 1] - diag[:, 0]) / 2.
+            tent_functions = np.maximum(heights[None, :] - np.abs(x_values[:, None] - midpoints[None, :]), 0)
+            n_points = diag.shape[0]
+            # Get indices of largest elements: can't take more than n_points - 1 (the last ones are in the right position)
+            argpartition = np.argpartition(-tent_functions, min(self.num_landscapes, n_points-1), axis=1)
+            landscapes = np.take_along_axis(tent_functions, argpartition, axis=1)
+            landscapes = landscapes[:, :min(self.num_landscapes, n_points)].T
```
mglisse, line 185: I don't understand the new code. Why use argpartition+take_along_axis instead of just partition? Is there no sorting anymore? I understood the previous version (`landscapes = np.sort(tent_functions, axis=1)[:, -self.num_landscapes:][:, ::-1].T`) better, my previous comment was only that the subarray (`[:, -self.num_landscapes:]`) could be taken before sorting instead of after, to save some work.

src/python/gudhi/representations/vector_methods.py
mglisse: > the way these representations are documented (at least BettiCurve), they don't give any guarantees

we can still aim for a good default.

Looking at the behavior in 3.6.0, when we specify only `resolution=10`, we have
* Landscape removes 0 on both sides
* Silhouette keeps 0 on both sides
* BettiCurve keeps the final 0 but not the initial one

I am sure we can at least improve consistency. My opinion is that we should drop 0 on both sides for all three. Note that this still leaves 0 on both sides for BettiCurve with resolution=None (@gspr, that seems to be the case you care about the most?), and may also leave some 0s on the side for the k-th landscape with k>1 (it isn't that bad to have a useless coordinate, I just want to avoid adding one that is too gratuitous). And we can add an option to bring back all the 0s for people who like plotting those curves, if there is demand.

@MathieuCarriere , would you be ok with such a plan? Anyone else?

(one thing I wondered while looking at it is, assuming the smallest interval that contains all the values where at least one function is non-zero is (0,12) and we want to sample 2 points in this interval, does it make more sense to pick 4 and 8 (as in 0, 4, 8, 12 minus the extremes) or 3 and 9 (a longer sequence would be -3, 3, 9, 15, so 0 and 12 fall in the middle of an interval)?)

src/python/gudhi/representations/vector_methods.py
wreise: You are completely right. I did not manage to apply your suggestion immediately, and I thought I noticed a bug in it, hence the overcomplicated version. Anyway, I think the current version is in the spirit of your original proposition.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -161,53 +172,25 @@ def transform(self, X):
...
+            tent_functions = np.maximum(heights[None, :] - np.abs(x_values[:, None] - midpoints[None, :]), 0)
+            n_points = diag.shape[0]
+            tent_functions.partition(n_points-self.num_landscapes, axis=1)
+            landscapes = np.sort(tent_functions[:, -self.num_landscapes:], axis=1)[:, ::-1].T
 
-            events = []
-            for j in range(self.new_resolution):
-                events.append([])
+            # Complete the array with zeros to get the right number of landscapes
+            if self.num_landscapes > n_points:
```
mglisse, line 186: In this case (not enough points), partition may already have failed with `ValueError: kth(=53) out of bounds (10)`.

src/python/gudhi/representations/vector_methods.py
gspr: > BettiCurve with resolution=None (@gspr, that seems to be the case you care about the most?)

Yes, `resolution=None` was the use case that motivated me to write the current `BettiCurve`class.

mglisse (approved):
But we should handle #715 soon.

PR #637, opened 2022-05-26
issue #613 [cpp documentation] Footer needs to be generated with a more recent version of doxygen

### LINKED ISSUE #613: [cpp documentation] Footer needs to be generated with a more recent version of doxygen
Looking at e.g.  https://gudhi.inria.fr/doc/latest/dir_d44c64559bbebec7f509842c48db8b23.html one can see:

image

which is probably due to an old version of the used footer for the HTML Page.

albert-github:
The problem was that the update of the treeview was not called in the footer, though after correcting this some new problems occurred in respect to the treeview (the small pointers ran into the text) and with the in page table of contents (e.g. in installation.html).
- footer.html: call proper update function by means of `<div>`, as base the 1.9.4 version served but also older version work.
- stylesheet.css, Doxyfile.in: implemented as extra stylesheet and for the "small pointer" problem adjusted the  `.arrow` settings
- header.html:
  - placed the `<link rel="stylesheet" type="text/css" href="https://gudhi.inria.fr/assets/css/styles_feeling_responsive.css" />` more as an extra stylesheet as otherwise a second scrollbar would occur
  - placed the `div id="top">` at the right place
  - corrected `` tag so it is properly closed (so XML compliant)
  - corrected setting of `data-topbar` to `data-topbar=true` as HTML5 does not allow attributes without values.

Fix #613


albert-github:
This proposed pull request obsoletes #635.

VincentRouvreau (approved):
Thanks for the PR and the investigation. It works fine for me with doxygen 1.9.3 and 1.9.4

PR #641, opened 2022-06-13
Doxygen only

VincentRouvreau:
Rework #611 for doxygen only build.
* Update gudhi-deploy submodule
* Warns no more on standard output but in a file
* Also apply the feature for user version also (was missing in previous PR)
* Use new option when only build cpp documentation in CI - useful when performing user version also
* Move new cmake option in the correct file. Disable every option when only build cpp documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,9 +95,9 @@ jobs:
             mkdir build
             cd build
             cmake -DWITH_GUDHI_CPP_DOCUMENTATION_ONLY=ON ..
-            make doxygen 2>&1 | tee dox.log
-            grep warning dox.log
-            cp dox.log html/
+            make doxygen 2>&1
+            grep warning doxygen.log
+            cp doxygen.log html/
             cp -R html /tmp/doxygen
 
       - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
 
   python:
     docker:
-      - image: gudhi/ci_for_gudhi:latest
+      - image: gudhi/doxygen_for_gudhi:latest
     steps:
       - checkout
       - run:
@@ -58,7 +58,7 @@ jobs:
             git submodule update
             mkdir build
             cd build
-            cmake -DUSER_VERSION_DIR=version ..
+            cmake -DWITH_GUDHI_CPP_DOCUMENTATION_ONLY=ON -DUSER_VERSION_DIR=version ..
             make user_version
             cd version
             cmake -DCMAKE_BUILD_TYPE=Release -DWITH_GUDHI_EXAMPLE=OFF -DWITH_GUDHI_UTILITIES=OFF -DWITH_GUDHI_PYTHON=ON -DPython_ADDITIONAL_VERSIONS=3 .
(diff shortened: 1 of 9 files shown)


mglisse:
The option is also useful for instance when we use cmake to create the official archive, which has nothing to do with the documentation, so it would be nice to use a name that fits both uses, saying basically that we are not going to compile anything...

albert-github:
@mglisse thanks for the response.

I think that we have here the situations:
- make an official archive, which does not incorporate third party software
- create a cmake "version" that only builds the documentation.

at the moment the case is that in both cases the same part has to be excluded.
Having 2 names that will do the same (for now) isn't nice but wouldn't it be better in this case so so it is explicit known what is done?

mglisse:
@VincentRouvreau , what do you think? Should we rename the option to something like NO_COMPILATION? Should we have several equivalent ONLY_* options?

albert-github:
A small thought:
Currently with the option the directory `src/GudhUI` is excluded from the archive, this looks a bit strange to me as it is the UI for Gudhi and is excluded for the documentation, by head, as it needs in the cmake phase already the Qt distribution.

albert-github:
@mglisse probably a lack in my knowledge I thought that
```
if (WITH_GUDHI_THIRD_PARTY)
  add_subdirectory(src/GudhUI)
endif()
```
excluded it from the archive as well, but as it is in the `src`  tree it might flow in and in that case it might be indeed sufficient to set it as "no compile" (i.e. `WITH_GUDHI_THIRD_PARTY=OFF`).

VincentRouvreau:
The [gudhi cmake options](https://github.com/GUDHI/gudhi-devel/blob/master/src/cmake/modules/GUDHI_options.cmake) are `WITH_GUDHI_BENCHMARK`, `WITH_GUDHI_EXAMPLE`, `WITH_GUDHI_PYTHON`, `WITH_GUDHI_TEST` and `WITH_GUDHI_UTILITIES`.
These options should have no impact on `make user_version`, that is a kind of [CPack](https://cmake.org/cmake/help/latest/module/CPack.html) (I ignored the existence of it when starting gudhi and it should be rewritten with CPack).

I added:
```cmake
option(WITH_GUDHI_THIRD_PARTY "Activate/deactivate third party libraries cmake detection. When set to OFF, it is usefull for doxygen or user_version i.e." ON)
```
The name `WITH_GUDHI_THIRD_PARTY` is not perfect, but `WITH_GUDHI_CPP_DOCUMENTATION_ONLY` neither, because we could also define `WITH_GUDHI_PYTHON_ONLY`, ...

And what if we were saying that if all options (benchmark, example, python, test and utilities) are false, we do not have to detect third parties ? The problem is reverted and maybe clearer. I want to build the `user_version`, so I set all options to OFF. idem for the doxygen documentation. Maybe we should add `WITH_GUDHI_UI`, or consider the GudhUI as an utility.

@mglisse @albert-github any preferences ?

albert-github:
Getting the option names so that they are right and cover the intention is difficult. 
I don't really have a preference other than what I stated before.

A small side not about the line:
> option(WITH_GUDHI_THIRD_PARTY "Activate/deactivate third party libraries cmake detection. When set to OFF, it is usefull for doxygen or user_version i.e." ON)

- the word "usefull" should be spelled "useful"
- the "i.e."  is hanging a bit in thin air. probably the " When set to OFF, it is usefull for doxygen or user_version i.e." should read like "The option should be set to OFF when creating a user version or just the doxygen documentation."

PR #643, opened 2022-06-15
[Alpha] Get extra point using boundary_opposite_vertex_simplex_range

VincentRouvreau:


VincentRouvreau:
Benchmarks show that the timings are the same, but `propagate_alpha_filtration` is much cleaner this way.

(`_xpt` means extra point, this PR)

NaturalImageStat2x2

5-forex

mglisse (approved):

PR #657, opened 2022-07-01
Ignore complete current build directory when building documentation

albert-github:
Found  the file
```
build/CMakeFiles/ShowIncludes/foo.h
```
during the build of the documentation. The build directory should be excluded.


VincentRouvreau (approved):
I tested with and without your fix and all seems correct

PR #673, opened 2022-08-11
Obsolete doxygen settings

albert-github:
In the current master the settings
- `DOT_TRANSPARENT` and `FORMULA_TRANSPARENT` are obsolete.
- `DOT_FONTNAME` / `DOT_FONTSIZE` are incorporated into the new `...ATTR` settings

### DIFF
--- a/src/Doxyfile.in
+++ b/src/Doxyfile.in
@@ -1452,17 +1452,6 @@ EXT_LINKS_IN_WINDOW    = NO
 
 FORMULA_FONTSIZE       = 10
 
-# Use the FORMULA_TRANPARENT tag to determine whether or not the images
-# generated for formulas are transparent PNGs. Transparent PNGs are not
-# supported properly for IE 6.0, but are supported on all modern browsers.
-#
-# Note that when changing this option you need to delete any form_*.png files in
-# the HTML output directory before the changes have effect.
-# The default value is: YES.
-# This tag requires that the tag GENERATE_HTML is set to YES.
-
-FORMULA_TRANSPARENT    = YES
-
 # Enable the USE_MATHJAX option to render LaTeX formulas using MathJax (see
 # http://www.mathjax.org) which uses client side Javascript for the rendering
 # instead of using pre-rendered bitmaps. Use this if you do not have LaTeX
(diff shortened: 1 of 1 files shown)


VincentRouvreau:
@albert-github What is the doxygen version from which this will become obsolete ? I am with doxygen 1.9.3 and I have the following warnings:
```
warning: ignoring unsupported tag 'DOT_COMMON_ATTR' at line 2178, file /home/gailuron/workspace/gudhi/gudhi-devel/build/version/build/Doxyfile
warning: ignoring unsupported tag 'DOT_EDGE_ATTR' at line 2187, file /home/gailuron/workspace/gudhi/gudhi-devel/build/version/build/Doxyfile
warning: ignoring unsupported tag 'DOT_NODE_ATTR' at line 2195, file /home/gailuron/workspace/gudhi/gudhi-devel/build/version/build/Doxyfile
```

albert-github:
This is now only available in the current master (1.9.5 (0bddda57211df25c81b456d255688f251a939a37)) and will be in the forthcoming 1.9.5 when released. So indeed in the 1.9.3 version it will give problems, to support multiple versions of doxygen you will probably need a strategy analogous to the strategy used by CGAL in respect to doxygen.

albert-github:
The problem was intriguing me a bit (also as I was looking at some other settings regarding `@CMAKE_...@`), but probably some setting is needed in `src/cmake/modules/GUDHI_doxygen_target.cmake` analogous as done for the `GUDHI_DOXYGEN_CLASS_DIAGRAMS`.
Actually the settings can be left out as they are anyway the default settings.

Please advise which version has the preference:
- leaving out the settings as they are default
- using a trick like done for `GUDHI_DOXYGEN_CLASS_DIAGRAMS`

VincentRouvreau:
> 

I think we can say that the minimal version of doxygen is 1.9.5, now that it is available as a conda package, so we could also remove `GUDHI_DOXYGEN_CLASS_DIAGRAMS` mechanism once this PR is merged.

However, the rendering is not exactly the same with your PR and doxygen 1.9.5.
cf. [PR 673 with doxygen 1.9.3](https://pages.saclay.inria.fr/vincent.rouvreau/gudhi/doxygen1.9.3/) and [PR 673 with doxygen 1.9.5](https://pages.saclay.inria.fr/vincent.rouvreau/gudhi/doxygen1.9.5/).
I don't know much about css, but would you have any idea of why the foreground became black ?

albert-github:
I see both in "white" / "light".

To support the light and dark mode a new setting has been introduced: `HTML_COLORSTYLE` and the default here is `AUTO_LIGHT` which means "Automatically set the mode according to the user preference, use light mode if no preference is set", so it looks like your preference is the dark mode. Best is to add the setting `HTML_COLORSTYLE=LIGHT" to the settings file.

VincentRouvreau (approved):
Perfect ! I will set HTML_COLORSTYLE=LIGHT in another PR and mention doxygen version in the installation page.

PR #691, opened 2022-08-31
Add RipsPersistence scikit-learn-like interface

VincentRouvreau:


src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,126 @@
...
+                Short circuit the use of :class:`~gudhi.representations.preprocessing.DimensionSelector` when only one
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            max_edge_length (float): Rips value. Default is +Inf.
+            nb_collapse (int): The number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform on the
+                SimplexTree. Edges are automatically collapsed if the maximal persistence diagrams dimension is greater
+                than 1. Default is 1.
+            newshape (tuple of ints): If cells filtration values require to be reshaped
+                (cf. :func:`~gudhi.sklearn.rips_persistence.RipsPersistence.transform`), set `newshape`
+                to perform `numpy.reshape(X, newshape, order='C')` in
+                :func:`~gudhi.sklearn.rips_persistence.RipsPersistence.transform` method.
```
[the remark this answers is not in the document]
VincentRouvreau: Removed on [efb40fd](https://github.com/GUDHI/gudhi-devel/pull/691/commits/efb40fda69748e3a8c6fa77e7b7216c3584490ba)

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,126 @@
...
+        Constructor for the RipsPersistence class.
+
+        Parameters:
+            homology_dimensions (int or list of int): The returned persistence diagrams dimension(s).
+                Short circuit the use of :class:`~gudhi.representations.preprocessing.DimensionSelector` when only one
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            max_edge_length (float): Rips value. Default is +Inf.
+            nb_collapse (int): The number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform on the
+                SimplexTree. Edges are automatically collapsed if the maximal persistence diagrams dimension is greater
+                than 1. Default is 1.
```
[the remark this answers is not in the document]
VincentRouvreau: I removed the sentence as it is an implementation detail on [2d9f152](https://github.com/GUDHI/gudhi-devel/pull/691/commits/2d9f152d01a77d14590457d36251517280369a97) (if the user asks for 50 iterations of edge collapses, but also only homology_dimensions=0, there is no added value to explain edge collapses won't be performed)
`nb_collapse=-1` for "automatic" behaviour as I prefer this to arguments that can be either `nb_collapse='auto'` or `nb_collapse=10`

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+class RipsPersistence(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the persistence diagrams from a Vietoris-Rips complex.
+    """
+
+    def __init__(
+        self,
+        homology_dimensions,
```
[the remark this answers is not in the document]
mglisse: I'd rather avoid the confusing `persistence_dim_max`, `homology_dimensions` should be enough to determine if we need internally to pass True or False.
By "magic", I mean something that isn't the usual list of integers. It could be `homology_dimensions=-1`, or `None`, `'all'`, `'all_but_max'`, etc.

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __transform(self, points):
+        stree = (self.__get_stree_from_points)(points, (max(self.homology_dimensions) + 1))
+        return [
+            stree.persistence_intervals_in_dimension(dim) for dim in self.homology_dimensions
+        ]
+
+    def __transform_only_this_dim(self, points):
+        stree = (self.__get_stree_from_points)(points, (self.homology_dimensions + 1))
+        return stree.persistence_intervals_in_dimension(self.homology_dimensions)
```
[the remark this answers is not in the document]
VincentRouvreau: Code has been factorized on [e9dede6](https://github.com/GUDHI/gudhi-devel/pull/691/commits/e9dede6f6d590e0d1de7fe8a17a32d9712cf6b0b)

src/python/gudhi/sklearn/rips_persistence.py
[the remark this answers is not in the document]
VincentRouvreau: Ok, I tried to summarize your proposal on [af2399d](https://github.com/GUDHI/gudhi-devel/pull/691/commits/af2399db4ce1a0927ae7776dc7db5c9daa75547f) after our discussion.

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+
+        return [
+            stree.persistence_intervals_in_dimension(dim) for dim in self.dim_list_
+        ]
+
+    def transform(self, X, Y=None):
+        """Compute all the Vietoris-Rips complexes and their associated persistence diagrams.
+
+        :param X: List of point clouds.
```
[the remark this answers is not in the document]
VincentRouvreau: Fixed on 558f130

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,147 @@
...
+        """
+        Constructor for the RipsPersistence class.
+
+        Parameters:
+            homology_dimensions (int or list of int): The returned persistence diagrams dimension(s).
+                Short circuit the use of :class:`~gudhi.representations.preprocessing.DimensionSelector` when only one
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            max_edge_length (float): Rips value. Default is +Inf.
+            input_type (str): Can be 'points' when inputs are point clouds, or 'matrices', when inputs are distance
+                matrices (full square or lower triangular). Default is 'points'.
```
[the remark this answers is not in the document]
VincentRouvreau: I was looking for the occurrence of 'matrices', and Mapper is using "point cloud" or "distance matrix", maybe we should be coherent ?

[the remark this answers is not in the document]
mglisse: I don't mind "point cloud" or "points" if you want to use them. It isn't very precise, everything is going to represent a point cloud, this version more precisely uses coordinates in an orthonormal basis of a Euclidean space. But it seems clear enough.

"distance matrix": note that mapper seems to require a full (symmetric) distance matrix in a dense format, under this name, not the same as here, so we should use a different name anyway. If we decide to reserve "distance matrix" for the full dense case, how should we name others? "lower distance matrix" for what you implemented here ("lower triangular distance matrix" is getting long...)? And we could use "lower distance csr_array" if for instance we wanted the lower part of the matrix, but encoded in the sparse format csr_array ("matrix" without precision implicitly refers to a dense format)?

There is also the subtlety of whether we want to allow encoding vertex filtration values on the diagonal. But that's not really a Rips anymore so maybe it doesn't need to be handled here.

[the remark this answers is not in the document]
VincentRouvreau: On [34c75d2](https://github.com/GUDHI/gudhi-devel/pull/691/commits/34c75d26984d703a8aac80918d228c3d328e1f06) I renamed possible `input_type` as "lower distance matrix" and "point cloud".
Maybe we should decide if `input_type` (or any other argument) with multiple word should be separated with an underscore or a space. ToMaTo has undescores and Mapper has spaces

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,147 @@
...
+        """Compute all the Vietoris-Rips complexes and their associated persistence diagrams.
+
+        :param X: List of point clouds.
+        :type X: list of list of float OR list of numpy.ndarray
+
+        :return: Persistence diagrams in the format:
+
+              - If `homology_dimensions` was set to `n`: `[array( Hn(X[0]) ), array( Hn(X[1]) ), ...]` 
+              - If `homology_dimensions` was set to `[i, j]`: `[[array( Hi(X[0]) ), array( Hj(X[0]) )], [array( Hi(X[1]) ), array( Hj(X[1]) )], ...]`
+        :rtype: list of (,2) array_like or list of list of (,2) array_like
```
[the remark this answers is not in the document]
VincentRouvreau: Yes you are right, we know the output format, documentation should be more concise. I modified it with `numpy ndarray` to be clearer on [0306847](https://github.com/GUDHI/gudhi-devel/pull/691/commits/0306847b1252607c31c99ddbdcc06ebbbf8c6a96)

src/python/test/test_sklearn_rips_persistence.py
```diff
@@ -0,0 +1,56 @@
...
+        # H0
+        assert len(diags[idx][0]) == len(point_clouds[idx])
+        # H1
+        assert len(diags[idx][1]) == 1
+
+def test_h1_only_rips_persistence_of_points_on_a_circle():
+    rips = RipsPersistence(homology_dimensions=1, n_jobs=-2)
+    diags = rips.fit_transform([points.sphere(n_samples = 150, ambient_dim = 2)])[0]
+    assert len(diags) == 1
+    assert 0. < diags[0][0] < 0.5
```
mglisse, line 37: Even with 150 points, if I run it 1000 times, I get 0.5047180621185876 once, so I think we want a bit more margin, either `< 0.6` or a few more points.

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+    ):
+        """
+        Constructor for the RipsPersistence class.
+
+        Parameters:
+            homology_dimensions (int or list of int): The returned persistence diagrams dimension(s).
+                Short circuit the use of :class:`~gudhi.representations.preprocessing.DimensionSelector` when only one
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            max_edge_length (float): Rips value. Default is +Inf.
+            input_type (str): Can be 'point cloud' when inputs are point clouds, or 'lower distance matrix', when
```
mglisse, line 54: I think my main remaining hesitation is how much we want to split things into many different classes vs cram everything into a single class with a myriad of options. Trying to list some variants that may appear in the future
1. input is points (with an implicit or explicit distance function) / input is a graph
2. for graph input, it can be a dense or sparse matrix
2. for graph input, allow or not weights for the vertices (while remaining a flag filtration)
3. if input is points: compute the full distance matrix or use a kd-tree to find neighbors and get a sparse distance matrix (when a non-trivial threshold is specified)
4. sparse Rips (not super useful these days, while approximate, it tends to be worse than the exact edge collapse)
5. weighted Rips as in DTM filtrations: different enough to deserve its own class(es)

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+                inputs are lower triangular distance matrix (can be full square, but the upper part of the distance
+                matrix will not be considered). Default is 'point cloud'.
+            nb_collapse (int): The number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform on the
+                SimplexTree. Default is -1, which means "automatic" (a relatively good enough number of iterations is
+                choosen).
+            homology_coeff_field (int): The homology coefficient field. Must be a prime number. Default value is 11.
+            min_persistence (float): The minimum persistence value to take into account (strictly greater than
+                `min_persistence`). Default value is `0.0`. Set `min_persistence` to `-1.0` to see all values.
```
mglisse, line 62: I wonder if we really want this option. It can make sense, for instance if you sample with a known precision. But in some cases we may be more interested in a minimal ratio death/birth (log-scale). If we don't have this parameter, it is easy to filter afterwards to get the same result. And filtering afterwards is likely not much slower than using the parameter min_persistence.
The reason I am thinking about this option is that supporting the case `min_persistence<0` is annoying. Edge-collapse removes a lot of intervals of length 0, so it makes little sense to have `min_persistence<0` and `nb_collapse>0`. Ripser also skips a lot of those intervals. So we would basically have to keep a slow path that does no edge-collapse and does not use ripser just for the not-so-useful case `min_persistence<0`.
Maybe a compromise could be to keep `min_persistence` but only support ≥0 values, or at least not promise much when it is <0.

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+        n_jobs=None,
+    ):
+        """
+        Constructor for the RipsPersistence class.
+
+        Parameters:
+            homology_dimensions (int or list of int): The returned persistence diagrams dimension(s).
+                Short circuit the use of :class:`~gudhi.representations.preprocessing.DimensionSelector` when only one
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            max_edge_length (float): Rips value. Default is +Inf.
```
DavidLapous, line 53: `max_edge_length` seems too complicated to me, is `threshold` (as 'persistence threshold') ok ?

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+            input_type (str): Can be 'point cloud' when inputs are point clouds, or 'lower distance matrix', when
+                inputs are lower triangular distance matrix (can be full square, but the upper part of the distance
+                matrix will not be considered). Default is 'point cloud'.
+            nb_collapse (int): The number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform on the
```
DavidLapous, line 57: `nb` for number is not consistent with gudhi or e.g. numpy (`st.num_simplices`, ...). Can we change `nb_collapse` to `num_collapses` ?

src/python/gudhi/sklearn/rips_persistence.py
DavidLapous, line 54: I'm not sure the `lower` is necessary here (as it works for full matrices), can we just specify the `lower` part in the doc and swap this option with `distance matrix` ?

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+                SimplexTree. Default is -1, which means "automatic" (a relatively good enough number of iterations is
+                choosen).
+            homology_coeff_field (int): The homology coefficient field. Must be a prime number. Default value is 11.
+            min_persistence (float): The minimum persistence value to take into account (strictly greater than
+                `min_persistence`). Default value is `0.0`. Set `min_persistence` to `-1.0` to see all values.
+            expand_extra_dimension (bool): :func:`~gudhi.SimplexTree.expansion` is performed at
+                `max(homology_dimensions) + 1` if true, and `max(homology_dimensions)` otherwise. Default is true.
+            n_jobs (int): cf. https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html
```
DavidLapous, line 65: We could maybe add a comment before the cf, such as (copied from scikitlearn's doc) : 

Number of jobs to run in parallel. None means 1 unless in a joblib.parallel_backend context. -1 means using all processors

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class RipsPersistence(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the persistence diagrams from a Vietoris-Rips complex.
+    """
+
+    def __init__(
+        self,
+        homology_dimensions,
```
DavidLapous, line 37: I prefer to stick with `degree` instead of `dimension` for homology, its clearer as it avoids, e.g., confusion with other dimension (ambient space, simplex dimensions ...), I propose `degrees` or `homology_degrees`.

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+        # Depends on homology_dimensions is an integer or a list of integer (else case)
+        if isinstance(self.homology_dimensions, int):
+            unwrap = True
+            self.dim_list_ = [ self.homology_dimensions ]
+        else:
+            unwrap = False
+            self.dim_list_ = self.homology_dimensions
+
+        # threads is preferred as Rips construction and persistence computation releases the GIL
+        res = Parallel(n_jobs=self.n_jobs, prefer="threads")(delayed(self.__transform)(inputs) for inputs in X)
```
DavidLapous, line 144: I think a `parallel_config` can override the `prefer` to a backend using processes, which will imply calling the `__get_state__` and `__set_state__` of the simplextree, which can have a large performance impact. What do you think about `backend=threading`  instead ?

DavidLapous (commented):
Few propositions to simplificate the `RipsPersistence` members.

src/python/gudhi/sklearn/rips_persistence.py
DavidLapous: Also, I usually like to put a `tqdm` wrapper, when a `progress` flag is set to true in the class. It works decently well with joblib, and gives an estimation of the time to finish, which can be quite useful. But that implies adding a `tqdm` dependency.

src/python/gudhi/sklearn/rips_persistence.py
mglisse: We have https://gudhi.inria.fr/python/latest/cubical_complex_sklearn_itf_ref.html#gudhi.sklearn.cubical_persistence.CubicalPersistence.__init__ that has already been released, and it would be good if those classes used the same names. CubicalPersistence is fairly recent, we could consider changing the name of the option, but that may break code unless we also keep the old name with some extra code to handle mixing the 2.

src/python/gudhi/sklearn/rips_persistence.py
mglisse: > I think a `parallel_config` can override the `prefer` to a backend using processes, which will imply calling the `__get_state__` and `__set_state__` of the simplextree, which can have a large performance impact.

Neither the input nor the output is a SimplexTree here, and I don't think intermediate objects are serialized / copied.

> What do you think about `backend=threading` instead ?

Since we control the function that is called, we could do that, yes.

src/python/gudhi/sklearn/rips_persistence.py
mglisse: Currently we have [nb_iterations](https://gudhi.inria.fr/python/latest/simplex_tree_ref.html#gudhi.SimplexTree.collapse_edges.params.nb_iterations) in the SimplexTree. And `n_jobs` which uses neither `num_` nor `nb_`...
I don't have a strong opinion on this name. Both can be a bit confusing in that they count the number of passes of the collapse algorithm, not the number of edges that are collapsed.

src/python/gudhi/sklearn/rips_persistence.py
mglisse: What if you have an upper distance matrix?
One thing to take into account is that the current implementation of `__transform` is likely to change completely in the future, with the inclusion of a modified Ripser and direct calls to edge-collapse. It could be that in the new implementation it would be convenient to have access to a full distance matrix, and then it would be strange to have 'distance matrix' mean the lower part and need a new 'full distance matrix'.
On the other hand, maybe copying the input to whatever distance matrix we like always takes a negligible amount of time and we could simplify things like you suggest.

src/python/gudhi/sklearn/rips_persistence.py
DavidLapous: > Neither the input nor the output is a SimplexTree here, and I don't think intermediate objects are serialized / copied.
Yes, I read this code too quickly, it's not an issue indeed

mglisse: > Also, I usually like to put a `tqdm` wrapper, when a `progress` flag is set to true in the class. It works decently well with joblib, and gives an estimation of the time to finish, which can be quite useful. But that implies adding a `tqdm` dependency. What do you think about it ?

Since we have this same pattern with joblib.Parallel in many places, I think we should consider this separately from this PR, for all of them at the same time. Could you open an issue showing what your suggestion looks like? Searching a bit, I find https://github.com/joblib/joblib/issues/972, https://stackoverflow.com/questions/37804279/how-can-we-use-tqdm-in-a-parallel-execution-with-joblib, https://stackoverflow.com/q/24983493/1918193, or even just `Parallel(...,verbose=7)`, but nothing seems perfect.

src/python/CMakeLists.txt
```diff
@@ -549,6 +549,21 @@ if(PYTHONINTERP_FOUND)
...
+      if (NOT CGAL_WITH_EIGEN3_VERSION VERSION_LESS 4.11.0)
+        add_gudhi_py_test(test_sklearn_rips_persistence)
+      endif()
+
+      # Datasets are fetched for these tests 
+      if(WITH_GUDHI_REMOTE_TEST)
+        add_test(NAME rips_complex_sklearn_itf_py_test
+                 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+                 COMMAND ${CMAKE_COMMAND} -E env "${GUDHI_PYTHON_PATH_ENV}"
+                 ${PYTHON_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/example/rips_complex_sklearn_itf.py" "--no-plot")
```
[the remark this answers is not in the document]
VincentRouvreau: "itf" stands for "interface". It was already like that for tensor flow interfaces, so I kept this wording

src/python/doc/rips_complex_sklearn_itf_ref.rst
```diff
@@ -0,0 +1,54 @@
...
+.. To get rid of WARNING: document isn't included in any toctree
+
+Rips complex persistence scikit-learn like interface
+####################################################
+
+.. list-table::
+   :width: 100%
+   :header-rows: 0
+
+   * - :Since: GUDHI 3.7.0
```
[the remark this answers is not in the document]
VincentRouvreau: Modified on cda48d548

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,148 @@
...
+#     Note right of RipsPersistence: ...
+#     thread1->>RipsPersistence: [array( Hi(X[0]) ), array( Hj(X[0]) )]
+#     thread2->>RipsPersistence: [array( Hi(X[1]) ), array( Hj(X[1]) )]
+#     Note right of RipsPersistence: ...
+#     RipsPersistence->>USER: [[array( Hi(X[0]) ), array( Hj(X[0]) )],<br/> [array( Hi(X[1]) ), array( Hj(X[1]) )],<br/> ...]
+
+
+class RipsPersistence(BaseEstimator, TransformerMixin):
+    """
+    This is a class for computing the persistence diagrams from a Vietoris-Rips complex.
```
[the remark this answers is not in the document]
VincentRouvreau: `This is a class for constructing Vietoris-Rips complexes and computing the persistence diagrams from them.` on f3b9ec2f3

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: Ok for `threshold`. Done on 4b73f3f06

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: I renamed `nb_collapse` as `num_collapses` that can be an integer or a bool. Default value is True aka. "automatic mode"
cf. f96eccc58

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: > I think a `parallel_config` can override the `prefer` to a backend using processes, which will imply calling the `__get_state__` and `__set_state__` of the simplextree, which can have a large performance impact. What do you think about `backend=threading` instead ?

From a benchmark I performed, it seems to be the same timings. I opened an [issue on joblib](https://github.com/joblib/joblib/issues/1511) to ask for it. We will do it on a separated issue on gudhi if we should use `backend="threading"` as @mglisse was explaining.

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: Yes, let's keep `homology_dimensions`

src/python/test/test_sklearn_rips_persistence.py
VincentRouvreau: Ok with `< 0.6`. I modified it on [007bd9e](https://github.com/GUDHI/gudhi-devel/pull/691/commits/007bd9e073084b12de561d7f74e8b7526c3b6839)

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: Oh, yes you are right that this can be easily done in a [preprocessing like class](https://github.com/GUDHI/gudhi-devel/blob/master/src/python/gudhi/representations/preprocessing.py). I removed min_persistence argument on 07706df8f

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: Ok, I added it on [bc99315](https://github.com/GUDHI/gudhi-devel/pull/691/commits/bc993151ba1fcd15067fe02c09734031ad1ebbcd)

src/python/gudhi/sklearn/rips_persistence.py
DavidLapous: > Since we have this same pattern with joblib.Parallel in many places, I think we should consider this separately from this PR, for all of them at the same time. Could you open an issue showing what your suggestion looks like? Searching a bit, I find [joblib/joblib#972](https://github.com/joblib/joblib/issues/972), https://stackoverflow.com/questions/37804279/how-can-we-use-tqdm-in-a-parallel-execution-with-joblib, https://stackoverflow.com/q/24983493/1918193, or even just `Parallel(...,verbose=7)`, but nothing seems perfect.

Sorry, I forgot to answer this. The code I was thinking about is something like this:
```python
class RipsPersistence:
  def __init__(..., progress=False):
    ...
    self.progress=progress
  def _compute_rips(self,pointcloud|distance_matrix):
    # code to compute rips for this point cloud
    ...
  def _collapse_edges(self,simplextree):
    # code for collapsing 
    ...
  def _some_other_transform(self,stuff):
    # some code
    ...
  def transform(self,point_clouds):
    simplextrees = Parallel(...)(delayed(self._compute_rips)(point_cloud) 
                 for point_cloud in tqdm(point_clouds, disable = not self.progress, desc="Computing Rips Complexes")
    simplextrees = Parallel(...)(delayed(self._compute_rips)(simplextree) 
                 for simplextree in tqdm(simplextrees, disable = not self.progress, desc="Collapsing edges")
    persistence_diagrams = Parallel(...)(delayed(self._some_other_transform)(simplextree) 
                 for simplextree in tqdm(simplextrees, disable = not self.progress, desc="Computing diagrams")
    return persistence_diagrams
```
I don't know if cutting the log into these steps is good or not (or should depend on a parameter) but it can be useful to know if one of these steps is the bottleneck, and change parameters accordingly. We can also consider adding a verbose level to this, or turn these 3 tqdms into 1 or ...

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,141 @@
...
+                Short circuit the use of :class:`~gudhi.representations.preprocessing.DimensionSelector` when only one
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            threshold (float): Rips maximal edge length value. Default is +Inf.
+            input_type (str): Can be 'point cloud' when inputs are point clouds, or 'lower distance matrix', when
+                inputs are lower triangular distance matrix (can be full square, but the upper part of the distance
+                matrix will not be considered). Default is 'point cloud'.
+            num_collapses (bool or int): Specify if :func:`~gudhi.SimplexTree.collapse_edges` is performed. When
+                `num_collapses` is an integer, it specifies the number of :func:`~gudhi.SimplexTree.collapse_edges`
+                iterations to perform on the SimplexTree. Default is True, which means "automatic" (a relatively good
+                enough number of iterations is choosen).
```
mglisse, line 59: Having something that can be int or bool but where True means something different from 1 doesn't sound right to me.

VincentRouvreau: Ok, let's stick with an int (default to 1 is good enough for the moment) on [86b0a00](https://github.com/GUDHI/gudhi-devel/pull/691/commits/86b0a00a1679ddaff6c0f4485a8d4cb2cdb6e8d5).
We could have some magic numbers in the future API like:
* '-1' for automatic number of collapse detection
* '-2' collapse until it is no more collapsable
* ...

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -0,0 +1,138 @@
...
+                dimension matters (in other words, when `homology_dimensions` is an int).
+            threshold (float): Rips maximal edge length value. Default is +Inf.
+            input_type (str): Can be 'point cloud' when inputs are point clouds, or 'lower distance matrix', when
+                inputs are lower triangular distance matrix (can be full square, but the upper part of the distance
+                matrix will not be considered). Default is 'point cloud'.
+            num_collapses (int): Specify the number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform
+                on the SimplexTree. Default value is 1 (a relatively good enough number of iterations).
+            homology_coeff_field (int): The homology coefficient field. Must be a prime number. Default value is 11.
+            expand_extra_dimension (bool): :func:`~gudhi.SimplexTree.expansion` is performed at
+                `max(homology_dimensions) + 1` if true, and `max(homology_dimensions)` otherwise. Default is true.
```
mglisse, line 60: I am thinking of removing this option from RipsPersistence. While the max_dim thing makes sense for a general simplicial complex, I don't think it (expand_extra_dimension=False) makes much sense for a flag complex. In any case, it is incompatible with edge collapse (and likely ripser), which assumes a true flag complex, not a truncated one (I didn't check, but I think already with 3 points (acute triangle), expand_extra_dimension=False and homology_dimensions=1, the diagram should contain 1 point without collapse, and 0 with collapse).

src/python/doc/rips_complex_sklearn_itf_ref.rst
```diff
@@ -0,0 +1,54 @@
...
+.. To get rid of WARNING: document isn't included in any toctree
+
+Rips complex persistence scikit-learn like interface
+####################################################
+
+.. list-table::
+   :width: 100%
+   :header-rows: 0
+
+   * - :Since: GUDHI 3.9.0
```
mglisse, line 12: Sorry, 3.10 now, my fault :bowing_man:

VincentRouvreau: I modified it on [86cfb79](https://github.com/GUDHI/gudhi-devel/pull/691/commits/86cfb7982270e7324803c50f892883ebc5efaa98) and added the feature in the next release note on [31bbd29](https://github.com/GUDHI/gudhi-devel/pull/691/commits/31bbd292bd33cffcc3b06186c935b2e76ad897da)

src/python/gudhi/sklearn/rips_persistence.py
VincentRouvreau: On [81ebe84](https://github.com/GUDHI/gudhi-devel/pull/691/commits/81ebe849565bdc4782be59577e6cf76f3b197bc2), I removed useless expand_extra_dimension argument and modify persistence_dim_max accordingly

mglisse (approved):
Afterwards, we should add something visible in the page on RipsComplex advertising that if people only want the Rips complex in order to compute its persistence, they should consider using RipsPersistence instead.

VincentRouvreau:
> Afterwards, we should add something visible in the page on RipsComplex advertising that if people only want the Rips complex in order to compute its persistence, they should consider using RipsPersistence instead.

Ok, I merge this old PR and will propose a doc PR for this purpose

PR #694, opened 2022-09-12
Compilation warnings fix

hschreiber:
Just very small changes to fix the warnings while compilation. There also were a lot of boost warnings about deprecated headers, but they fixed them-selves when I installed the latest boost version (1.80).

### DIFF
--- a/src/Persistence_representations/test/persistence_heat_maps_test.cpp
+++ b/src/Persistence_representations/test/persistence_heat_maps_test.cpp
@@ -78,7 +78,7 @@ BOOST_AUTO_TEST_CASE(check_compute_percentage_of_active_of_heat_maps) {
   to_compute_percentage_of_active.push_back(&q);
   to_compute_percentage_of_active.push_back(&r);
   Persistence_heat_maps<constant_scaling_function> percentage_of_active;
-  percentage_of_active.compute_percentage_of_active(to_compute_percentage_of_active, 0.1);
+  percentage_of_active.compute_percentage_of_active(to_compute_percentage_of_active, 0);
 
   Persistence_heat_maps<constant_scaling_function> template_percentage_of_active;
   template_percentage_of_active.load_from_file("data/template_percentage_of_active_of_heat_maps");
--- a/src/Skeleton_blocker/include/gudhi/Skeleton_blocker_simplifiable_complex.h
+++ b/src/Skeleton_blocker/include/gudhi/Skeleton_blocker_simplifiable_complex.h
@@ -267,7 +267,7 @@ void Skeleton_blocker_complex<SkeletonBlockerDS>::remove_blocker_include_in_simp
 template<typename SkeletonBlockerDS>
 void Skeleton_blocker_complex<SkeletonBlockerDS>::tip_blockers(Vertex_handle a, Vertex_handle b,
                                                                std::vector<Simplex> & buffer) const {
-  for (auto const & blocker : this->const_blocker_range(a)) {
+  for (auto const blocker : this->const_blocker_range(a)) {
     Simplex beta = (*blocker);
     beta.remove_vertex(a);
     buffer.push_back(beta);
--- a/src/Coxeter_triangulation/include/gudhi/Functions/Function_affine_plane_in_Rd.h
+++ b/src/Coxeter_triangulation/include/gudhi/Functions/Function_affine_plane_in_Rd.h
@@ -57,7 +57,7 @@ struct Function_affine_plane_in_Rd {
    * The dimension of the vector should be the ambient dimension of the manifold.
    */
   Function_affine_plane_in_Rd(const Eigen::MatrixXd& normal_matrix, const Eigen::VectorXd& offset)
-      : normal_matrix_(normal_matrix), d_(normal_matrix.rows()), k_(normal_matrix.cols()), m_(d_ - k_), off_(offset) {
+	  : normal_matrix_(normal_matrix), d_(normal_matrix.rows()), k_(normal_matrix.cols()), off_(offset) {
     normal_matrix_.colwise().normalize();
   }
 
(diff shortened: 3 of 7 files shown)


VincentRouvreau:
> Just very small changes to fix the warnings while compilation. There also were a lot of boost warnings about deprecated headers, but they fixed them-selves when I installed the latest boost version (1.80).

I can confirm boost warnings like:
```txt
include/boost/detail/iterator.hpp:13:1: note: ‘#pragma message: This header is deprecated. Use <iterator> instead.’
   13 | BOOST_HEADER_DEPRECATED("<iterator>")
      | ^~~~~~~~~~~~~~~~~~~~~~~
In file included from include/boost/smart_ptr/detail/sp_thread_sleep.hpp:22,
                 from include/boost/smart_ptr/detail/yield_k.hpp:23,
                 from include/boost/smart_ptr/detail/spinlock_gcc_atomic.hpp:14,
                 from include/boost/smart_ptr/detail/spinlock.hpp:42,
                 from include/boost/smart_ptr/detail/spinlock_pool.hpp:25,
                 from include/boost/smart_ptr/shared_ptr.hpp:29,
                 from include/boost/property_map/vector_property_map.hpp:14,
                 from include/boost/property_map/property_map.hpp:602,
                 from include/boost/graph/adjacency_list.hpp:27,
                 from gudhi-devel/src/common/include/gudhi/graph_simplicial_complex.h:14,
```
are removed from boost 1.78, cf. https://github.com/boostorg/graph/issues/228

VincentRouvreau (approved):

src/Persistence_representations/test/persistence_heat_maps_test.cpp
```diff
@@ -78,7 +78,7 @@ BOOST_AUTO_TEST_CASE(check_compute_percentage_of_active_of_heat_maps) {
   to_compute_percentage_of_active.push_back(&q);
   to_compute_percentage_of_active.push_back(&r);
   Persistence_heat_maps<constant_scaling_function> percentage_of_active;
-  percentage_of_active.compute_percentage_of_active(to_compute_percentage_of_active, 0.1);
```
mglisse, line 81: My first reaction would be to think that this was supposed to be 0.1 times something, but 0 seems consistent with the expected data/template_percentage_of_active_of_heat_maps. I don't really understand why `compute_percentage_of_active` is named this way, I don't see any percentage in it...

src/Skeleton_blocker/include/gudhi/Skeleton_blocker_simplifiable_complex.h
```diff
@@ -267,7 +267,7 @@ void Skeleton_blocker_complex<SkeletonBlockerDS>::remove_blocker_include_in_simp
 template<typename SkeletonBlockerDS>
 void Skeleton_blocker_complex<SkeletonBlockerDS>::tip_blockers(Vertex_handle a, Vertex_handle b,
                                                                std::vector<Simplex> & buffer) const {
-  for (auto const & blocker : this->const_blocker_range(a)) {
+  for (auto const blocker : this->const_blocker_range(a)) {
```
mglisse, line 270: What was the warning for this one?

hschreiber: warning: 
**loop variable 'blocker' is always a copy because the range of type**
'Gudhi::skeleton_blocker::Skeleton_blocker_complex<Gudhi::skeleton_blocker::Skeleton_blocker_simple_traits>::Const_complex_blocker_around_vertex_range'
(aka 'iterator_range<Blocker_iterator_around_vertex_internal<__map_const_iterator<__tree_const_iterator<std::__1::__value_type<Gudhi::skeleton_blocker::Skeleton_blocker_simple_traits::Vertex_handle, Gudhi::skeleton_blocker::Skeleton_blocker_simplex<Gudhi::skeleton_blocker::Skeleton_blocker_simple_traits::Vertex_handle> *>, std::__1::__tree_node<std::__1::__value_type<Gudhi::skeleton_blocker::Skeleton_blocker_simple_traits::Vertex_handle, Gudhi::skeleton_blocker::Skeleton_blocker_simplex<Gudhi::skeleton_blocker::Skeleton_blocker_simple_traits::Vertex_handle> *>, void *> *, long> >, Skeleton_blocker_simplex<Gudhi::skeleton_blocker::Skeleton_blocker_simple_traits::Vertex_handle> *const> >') **does not return a reference** [-Wrange-loop-analysis]

src/Persistence_representations/test/persistence_heat_maps_test.cpp
hschreiber: It seems to indirectly compute the percentage by counting the number of maps with particular values but without dividing by the number of maps. I think it was supposed to be a percentage first (therefore a cutoff of 0.1), but ended up just counting and the test was not corrected.

mglisse (approved):

src/Skeleton_blocker/include/gudhi/Skeleton_blocker_simplifiable_complex.h
mglisse: Note that in my opinion, clang's warning is bogus. While it is true that the code binds a const reference to a temporary, that's a perfectly normal and idiomatic thing to do. In template code, it could very well be that with different template parameters, the iterator would yield a reference, and `auto const&` or `auto&&` are the best way to handle both cases well at the same time.

mglisse: And indeed the warning has changed since then (https://reviews.llvm.org/D73434) and I don't think `auto const&` can trigger it anymore with just `-Wall` (of course Apple's version of clang lags, so it may still warn 2 years later :disappointed: )

hschreiber: In general, I agree with you, but in this particular case, I don't think it will ever return a reference, no? But I have to admit, I am not sure how this kind of `for` loops are really handled, so I could be wrong.

mglisse: First, it could start returning a reference if we modify Blocker_iterator_around_vertex_internal (or pass it different template parameters) so its `dereference` starts returning a reference.
Second, without changing the existing code, nothing prevents a user from adding a specialization of Blocker_iterator_around_vertex_internal that does return a reference. Just because one instantiation of the template has some behavior doesn't mean that all possible specializations have it.

mglisse: Anyway, this change was ok, and I don't really care about the ones in #710 since they are in tests / examples, but if you see more of those warnings, let's stop "fixing" them and update the compiler instead.

PR #697, opened 2022-09-14
c++17 as a new standard to compile the library

VincentRouvreau:
Update also the installation manual as Visual Studio 2017 is also required.
Fix compilation error in tangential benchmark as std::ofstream::streampos seems deprecated with c++17.

### DIFF
--- a/.github/next_release.md
+++ b/.github/next_release.md
@@ -9,8 +9,8 @@ Below is a list of changes made since GUDHI 3.6.0:
 - [Module](link)
      - ...
 
-- [Module](link)
-     - ...
+- Installation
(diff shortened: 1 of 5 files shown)


mglisse (approved):
ok once we merge something that actually needs it.

VincentRouvreau:
> ok once we merge something that actually needs it.

As @hschreiber needs this feature for the persistence matrix PR #669 , I would like to merge it now. Is it ok for you ?

mglisse:
> As @hschreiber needs this feature for the persistence matrix PR #669 , I would like to merge it now. Is it ok for you ?

The cleaner way would be to have Hannah's branch merge this one and still only merge this one into master later, but let's not bother, ok to merge it now to master.

PR #698, opened 2022-09-16
Translate french comment

### LINKED ISSUE #687: French, user intended, comment block
We see here a comment block in the French language, I think this should be translated into English:

https://github.com/GUDHI/gudhi-devel/blob/fc8290b3ff3236ce88a5e6761dee3242ee1376a0/src/Skeleton_blocker/include/gudhi/Skeleton_blocker/Skeleton_blocker_sub_complex.h#L199-L203

VincentRouvreau:
Fix #687


mglisse (approved):
In several places (not just this PR) you use `i.e.` but it looks strange. Are you confusing it with `e.g.`? Otherwise ok.

VincentRouvreau:
> In several places (not just this PR) you use `i.e.` but it looks strange. Are you confusing it with `e.g.`? Otherwise ok.

Oops, you are right... I was thinking `i.e.` was meaning `in example`. Thanks for the report

PR #704, opened 2022-09-26
Cech_persistence utility for safe/exact/fast mode

VincentRouvreau:


src/Cech_complex/utilities/cech_persistence.cpp
```diff
@@ -77,15 +103,20 @@ int main(int argc, char* argv[]) {
...
                        "Name of an OFF file containing a point set.\n");
 
   po::options_description visible("Allowed options", 100);
   visible.add_options()("help,h", "produce help message")(
+      "exact,e", po::bool_switch(&exact),
+      "To activate exact version of Cech complex (default is false, not available if fast is set)")(
+      "fast,f", po::bool_switch(&fast),
+      "To activate fast version of Cech complex (default is false, not available if exact is set)")(
       "output-file,o", po::value<std::string>(&filediag)->default_value(std::string()),
       "Name of file in which the persistence diagram is written. Default print in std::clog")(
```
mglisse, line 121: Is that true? Looking at the code, it looks like we print the extra information to clog, but the diagram itself is going to cout (which makes sense).

src/Cech_complex/utilities/cech_persistence.cpp
```diff
@@ -25,41 +26,66 @@
...
+  }
 
-  // Construct the Cech complex in a Simplex Tree
-  Simplex_tree simplex_tree;
+  Simplex_tree stree;
+  if (fast_version) {
+    // WARNING : CGAL::Epick_d is fast but not safe (unlike CGAL::Epeck_d)
+    // (i.e. when the points are on a grid)
+    using Fast_kernel = CGAL::Epick_d<CGAL::Dynamic_dimension_tag>;
+    stree = create_simplex_tree<Fast_kernel>(off_file_points, exact_version, max_radius, dim_max);
```
mglisse, line 75: It looks a bit strange to use `exact_version` in the fast case, but why not, it documents what this parameter is, and we already ensure that it is false elsewhere.

mglisse (approved):
I didn't check the details, but I assume you made it look like a cross between the Rips and alpha utilities.

src/Cech_complex/utilities/cech_persistence.cpp
VincentRouvreau: I track this issue in #709 as the issue is also in other utilities

PR #706, opened 2022-10-14
Fix badly centered barcode

### LINKED ISSUE #685: plot_persistence_barcode missing half of the last bar
bugbarcode

```python
import gudhi
import matplotlib.pyplot as plt
gudhi.plot_persistence_barcode([(0,(0,1)),(0,(0,1))])
plt.show()
```
It might just be a bad bounding box / shifting, it looks like there is too much empty space at the top, at the same time as we are missing some part at the bottom.

mglisse:
Fix #685.
The main issue was the line
```python
axes.axis([axis_start, infinity, 0, len(x)])
```
which sets bogus bounds for the y axis. It should go roughly from `-.5` to `len(x)-.5` (or the bars should be aligned differently), but it seems easier to let matplotlib handle this axis automatically (with many bars, the automatic setting leaves much more white at the top and bottom, while the manual setting would let the bars touch the border, so it isn't equivalent)
Other changes are suggestions. I removed the labels on the y axis because seeing `0.25` made no sense, but `axes.yaxis.get_major_locator().set_params(integer=True)` would be another way to handle that.
Not tested much.

### DIFF
--- a/src/python/gudhi/persistence_graphical_tools.py
+++ b/src/python/gudhi/persistence_graphical_tools.py
@@ -194,19 +194,21 @@ def plot_persistence_barcode(
         y=[(death - birth) if death != float("inf") else (infinity - birth) for (dim,(birth,death)) in persistence]
         c=[colormap[dim] for (dim,(birth,death)) in persistence]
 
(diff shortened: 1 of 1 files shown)


VincentRouvreau (approved):
I tested and benchmarked it. Seems nice to me !

PR #708, opened 2022-10-15
Split alpha-complex unit test in 2 files

mglisse:
CI on windows has a hard time compiling this as one file.


src/Alpha_complex/test/Alpha_complex_dim3_unit_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+ *    Author(s):       Vincent Rouvreau
+ *
+ *    Copyright (C) 2015 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#define BOOST_TEST_DYN_LINK
+#define BOOST_TEST_MODULE "alpha_complex"
```
VincentRouvreau, line 12: It doesn't seem to bother the execution, nor the test, but I think we should set unique name for `BOOST_TEST_MODULE`.

```suggestion
#define BOOST_TEST_MODULE "alpha_complex_dim3"
```

VincentRouvreau (approved):

PR #714, opened 2022-10-26
Use std::optional to check C++17 support

mglisse:
Check if #697 actually works (it doesn't).


VincentRouvreau (approved):
Thank you, I missed these points when switching to C++ 17.

PR #717, opened 2022-11-03
Translate n_jobs to workers for SciPy

### LINKED ISSUE #662: [python / knn] n_jobs argument is deprecated for kdtree query with scipy 1.9.0
According to [https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.query.html#scipy.spatial.cKDTree.query](kdtree query documentation), `n_jobs` is deprecated starting from scipy>=1.9.0.

With scipy 1.9.0, this leads to an error in `test_tomato.py`:
```
gudhi/point_cloud/knn.py:318: in transform
    distances, neighbors = self.kdtree.query(X, k=self.k, **qargs)
_ckdtree.pyx:786: in scipy.spatial._ckdtree.cKDTree.query
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: Unexpected keyword argument {'n_jobs': -1}
```

The argument should be named `workers` and the documentation should be checked.

mglisse:
Fix #662.
This is incompatible with versions of SciPy older than 1.6.0 (2 years old). We don't document a lot of minimal versions for Python dependencies, so I am tempted to ignore this small issue.

### DIFF
--- a/src/python/doc/installation.rst
+++ b/src/python/doc/installation.rst
@@ -391,7 +391,7 @@ The :doc:`persistence graphical tools </persistence_graphical_tools_user>` and
 mathematics, science, and engineering.
 
 :class:`~gudhi.point_cloud.knn.KNearestNeighbors` can use the Python package
-`SciPy <http://scipy.org>`_ as a backend if explicitly requested.
+`SciPy <http://scipy.org>`_ :math:`\geq` 1.6.0 as a backend if explicitly requested.
 
 TensorFlow
 ----------
--- a/src/python/gudhi/point_cloud/knn.py
+++ b/src/python/gudhi/point_cloud/knn.py
@@ -314,7 +314,9 @@ class KNearestNeighbors:
             return None
 
         if self.params["implementation"] == "ckdtree":
-            qargs = {key: val for key, val in self.params.items() if key in {"p", "eps", "n_jobs"}}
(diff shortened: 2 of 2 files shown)


VincentRouvreau (approved):
With your fix, `gudhi.point_cloud.knn.KNearestNeighbors(implementation="ckdtree"..., workers=N)` is now accepted, but not documented. But let's keep the documentation as is for not to confuse the users.

`ckdtree.query(..., workers=N)` is available from scipy >= 1.6.0 (december 2020), we should say it in the dedicated [scipy installation manual](https://gudhi.inria.fr/python/latest/installation.html#scipy) and in [test_requirements.txt](https://github.com/GUDHI/gudhi-deploy/blob/main/test-requirements.txt). Can be done on another issue.

mglisse:
> `gudhi.point_cloud.knn.KNearestNeighbors(implementation="ckdtree"..., workers=N)` is now accepted, but not documented

"kwargs – additional parameters are forwarded to the backends" so it is documented :wink:

mglisse:
> [test_requirements.txt](https://github.com/GUDHI/gudhi-deploy/blob/main/test-requirements.txt)

If you don't mind, I'll leave that part to you.

PR #720, opened 2022-11-04
New function to insert several vertices in a Simplex_tree

mglisse:
If I replace the new code in Alpha_complex.h with `complex.insert_batch_vertices(vertices_, std::numeric_limits<Filtration_value>::quiet_NaN());`, the testsuite passes. I didn't include that here because it would also require changing the concept SimplicialComplexForAlpha.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -24,6 +24,7 @@
 #include <boost/iterator/transform_iterator.hpp>
 #include <boost/graph/adjacency_list.hpp>
 #include <boost/range/adaptor/reversed.hpp>
+#include <boost/range/adaptor/transformed.hpp>
 #include <boost/container/static_vector.hpp>
 
 #ifdef GUDHI_USE_TBB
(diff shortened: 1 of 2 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1156,6 +1153,21 @@ class Simplex_tree {
...
+   * @param[in] vertices A range of Vertex_handle
+   * @param[in] filt filtration value of the new vertices (the same for all)
+   *
+   * This may be faster than inserting the vertices one by one, especially in a random order.
+   * The complex does not need to be empty before calling this function. However, if a vertex is
+   * already present, its filtration value is not modified, unlike with other insertion functions. */
+  template <class VertexRange>
+  void insert_batch_vertices(VertexRange const& vertices, Filtration_value filt = 0) {
+    auto verts = vertices | boost::adaptors::transformed([&](auto v){
+        return Dit_value_t(v, Node(&root_, filt)); });
```
VincentRouvreau, line 1166: If we do not want to capture everything, here we could do:
```suggestion
    auto verts = vertices(skel_graph) | boost::adaptors::transformed([&root = this->root_, &skel_graph](auto v){
        return Dit_value_t(v, Node(&root, get(vertex_filtration_t(), skel_graph, v))); });
```
Maybe not that interesting...

mglisse: Is this comment meant for the function insert_graph instead?

src/Simplex_tree/test/simplex_tree_unit_test.cpp
```diff
@@ -1038,3 +1038,17 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(simplex_tree_boundaries_and_opposite_vertex_iterat
       BOOST_CHECK(opposite_vertices.size() == 0);
   }
 }
+
+BOOST_AUTO_TEST_CASE(batch_vertices) {
+  typedef Simplex_tree<> typeST;
```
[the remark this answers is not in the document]
mglisse: Contiguity is already broken by `st.insert_simplex_and_subfaces({3}, 1.5);` before that. This is a test that when there is already a vertex, the batch insertion can still insert other vertices before/after it. We could add a different test for the contiguous case though if you want, the testing is far from exhaustive.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: "not capture everything" -> "be explicit about what we capture"?
I checked, and the lambda in insert_graph already has exactly the size of 2 pointers, this wouldn't change it.
Ah, it would capture `this->root_` instead of `this`, is that what you meant?
Just trying to understand the goal before I change something.

VincentRouvreau: Yes, sorry, I mixed your 2 changes with `boost::adaptors::transformed`.
And yes my comment was to know whether we should be explicit about what we capture in the lambda, or capture everything by reference.
Here, if we capture only 2 pointers, I am ok with it.

VincentRouvreau (approved):

PR #721, opened 2022-11-04
Use fetch_spiral_2d in the doc

mglisse:
I don't know if we want to remove data/points/spiral_2d.csv.


VincentRouvreau:
> I don't know if we want to remove data/points/spiral_2d.csv.

As it is not used at any other places, yes, we can remove it.

VincentRouvreau (approved):

Issue #723, opened 2022-11-06
[python] Unexpected behavior with Rips filtration when distance matrix identity entries are non-zero


tuanpham96:
## Issue 

Recently when computing a distance matrix for a dataset I'm working on, I had forgotten to set the identity entries to 0. After I noticed the bug and corrected it, the results (a few steps after Rips filtration) came out differently. 

I initially thought that the fix wouldn't matter as I assume the filtration would possibly be done on the pairwise distances anyway, and self-loop values would be ignored. 

However, the results came out differently, and below is a minimal example with just `sklearn.make_circles`. 

## Minimal example

- First generate example data

  ``` python
  import numpy as np
  from sklearn.datasets import make_circles
  from sklearn import metrics
  import gudhi as gd
  from copy import deepcopy
  import matplotlib.pyplot as plt
  gd.persistence_graphical_tools._gudhi_matplotlib_use_tex=False
  
  # make data
  X, y = make_circles(n_samples=200, shuffle=True, noise=0.05, random_state=123, factor=0.4)
  
  # plot
  plt.scatter(X[:,0], X[:,1], c = y,s = 50)
  plt.axis('square')
  plt.show()
  ```
  ![gudhi-issue-example-data](https://user-images.githubusercontent.com/42875763/200147960-21a3a7dd-2d75-4cc7-8285-67b57ba0a33e.png)

- Construct the distance matrix, as well as the "wrong" ones with non-zero identity entries.

  ``` python
  # dist matrix
  D0 = metrics.euclidean_distances(X)
  
  # make non-zero for identity entries
  D1 = deepcopy(D0)
  D1[np.eye(len(D1)) > 0] = 0.5
  D2 = deepcopy(D0)
  D2[np.eye(len(D2)) > 0] = 1
  ```
- Compute & plot the persistence barcodes and diagrams
  
  ``` python
  # compute diagrams
  def get_dgm(D):
      skel = gd.RipsComplex(D, max_edge_length=3)
      spx_tree = skel.create_simplex_tree(max_dimension=3)
      dgm = spx_tree.persistence()    
      return dgm
  
  dgm_list = [get_dgm(D) for D in [D0, D1, D2]]
  
  # plot barcodes
  ttls = ['original distance', 'identity is 0.5', 'identity is 1']
  plt.figure(figsize=(18,10))
  for i, dgm in enumerate(dgm_list):
      ax1 = plt.subplot(2,3,i+1)
[...]

PR #725, opened 2022-11-08
Make RipsComplex construction less confusing

mglisse:
* Clarify the doc.
* Mandate the use of named arguments: `points=X` or `distance_matrix=X`, since `RipsComplex(X)` was ambiguous.

The modified test was giving me
> E   TypeError: __cinit__() takes exactly 0 positional arguments (1 given)

As discussed in #723, clarifying the doc is obviously good, while the second patch is more questionable.

### DIFF
--- a/src/python/gudhi/rips_complex.pyx
+++ b/src/python/gudhi/rips_complex.pyx
@@ -45,20 +45,19 @@ cdef class RipsComplex:
                  max_edge_length=float('inf'), sparse=None):
         """RipsComplex constructor.
 
-        :param max_edge_length: Rips value.
-        :type max_edge_length: float
-
         :param points: A list of points in d-Dimension.
-        :type points: list of list of float
+        :type points: List[List[float]]
 
         Or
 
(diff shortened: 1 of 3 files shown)


src/python/gudhi/rips_complex.pyx
```diff
@@ -41,31 +41,30 @@ cdef class RipsComplex:
...
+    def __init__(self, *, points=None, distance_matrix=None,
                  max_edge_length=float('inf'), sparse=None):
         """RipsComplex constructor.
 
-        :param max_edge_length: Rips value.
-        :type max_edge_length: float
-
         :param points: A list of points in d-Dimension.
-        :type points: list of list of float
+        :type points: List[List[float]]
```
VincentRouvreau, line 49: At some point, we should be consistent with WeightedRipsComplex (`Sequence[Sequence[float]]`) and DTMRipsComplex (`numpy.ndarray`)

VincentRouvreau (approved):

PR #727, opened 2022-11-09
New write_points_to_off_file

mglisse:
If we are going to remove OFF data files that have an equivalent fetch_ function, this may help C++ users who liked having an OFF file get it back.

I named the parameter `fname` as in the numpy function. I could have used `off_file` like the reader, although I am not fond of the redundant `off_` part of that name.


src/python/gudhi/off_utils.pyx
```diff
@@ -39,3 +41,17 @@ def read_points_from_off_file(off_file=''):
...
+    A simple wrapper for `numpy.savetxt`.
+
+    :param fname: Name of the OFF file.
+    :type fname: str or file handle
+    :param points: Point coordinates.
+    :type points: numpy array of shape (n, dim)
+    """
+    points = np.array(points, copy=False)
+    assert len(points.shape) == 2
+    np.savetxt(fname, points, header='nOFF\n{} {} 0 0'.format(points.shape[1], points.shape[0]), comments='')
```
VincentRouvreau, line 57: There is a specific case for 3d (cf. [OFF file documentation](https://gudhi.inria.fr/python/latest/fileformats.html#off-file-format)) that `write_points_to_off_file` should consider.

src/python/gudhi/off_utils.pyx
```diff
@@ -39,3 +41,17 @@ def read_points_from_off_file(off_file=''):
             raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT),
                                     off_file)
 
+@cython.embedsignature(True)
+def write_points_to_off_file(fname, points):
+    """Write points to an OFF file.
+
+    A simple wrapper for `numpy.savetxt`.
+
+    :param fname: Name of the OFF file.
```
VincentRouvreau, line 50: We should provide a link to [OFF file documentation](https://gudhi.inria.fr/python/latest/fileformats.html#off-file-format).
IIRC the syntax is:

```suggestion
    :param fname: Name of the `OFF file <fileformats.html#off-file-format>`_.
```

src/python/gudhi/off_utils.pyx
mglisse: There is also a special case for 4d, but if I emit that, we won't be able to read it back... (we don't seem to accept OFF BINARY either)

The nOFF case is still valid for 3d, and it is nice to emit something consistent, so it was on purpose that I did it this way. The main advantage of emitting the special 3d case is that there may be more other software that can read it than nOFF. On the other hand, if the goal is to send a point cloud to another software, I wouldn't choose OFF, plain CSV is more widely used.

Anyway, I changed it now...

PR #735, opened 2022-11-15
Update to Hera 2

mglisse:
I will have follow-ups, but let's get this out of the way first.

@gspr since everything is nicely under include/hera/ now, I replaced the 2 variables with a single HERA_INCLUDE_DIR. For Debian, I guess that will be /usr/include (or anything that doesn't exist, since we don't need any -I to find stuff in /usr/include).


mglisse:
I actually use the very latest commit at the moment. There was some shuffling of files based on the discussion in https://github.com/anigmetov/hera/issues/11. I don't know about official release tagging though.

VincentRouvreau (approved):

PR #736, opened 2022-11-16
Provide matching in hera.wasserstein_distance

mglisse:
This comes on top of #735.

Lots of workarounds, I tried to write them in a way that is not too likely to break if Hera is fixed.

### DIFF
--- a/src/python/gudhi/bottleneck.cc
+++ b/src/python/gudhi/bottleneck.cc
@@ -9,18 +9,17 @@
  */
 
 #include <gudhi/Bottleneck.h>
-
+#include <optional>
 #include <pybind11_diagram_utils.h>
+#include <pybind11/stl.h>
 
 // Indices are added internally in bottleneck_distance, they are not needed in the input.
 static auto make_point(double x, double y, py::ssize_t) { return std::pair(x, y); };
 
 // For compatibility with older versions, we want to support e=None.
-// In C++17, the recommended way is std::optional<double>.
-double bottleneck(Dgm d1, Dgm d2, py::object epsilon)
+double bottleneck(Dgm d1, Dgm d2, std::optional<double> epsilon)
 {
-  double e = (std::numeric_limits<double>::min)();
-  if (!epsilon.is_none()) e = epsilon.cast<double>();
+  double e = epsilon.value_or((std::numeric_limits<double>::min)());
   // I *think* the call to request() has to be before releasing the GIL.
   auto diag1 = numpy_to_range_of_pairs(d1, make_point);
   auto diag2 = numpy_to_range_of_pairs(d2, make_point);
--- a/src/python/gudhi/bottleneck.cc
+++ b/src/python/gudhi/bottleneck.cc
@@ -20,7 +20,7 @@ static auto make_point(double x, double y, py::ssize_t) { return std::pair(x, y)
 double bottleneck(Dgm d1, Dgm d2, std::optional<double> epsilon)
 {
   double e = epsilon.value_or((std::numeric_limits<double>::min)());
-  // I *think* the call to request() has to be before releasing the GIL.
+  // I *think* the call to request() in numpy_to_range_of_pairs has to be before releasing the GIL.
   auto diag1 = numpy_to_range_of_pairs(d1, make_point);
   auto diag2 = numpy_to_range_of_pairs(d2, make_point);
 
--- a/src/python/gudhi/bottleneck.cc
+++ b/src/python/gudhi/bottleneck.cc
@@ -12,6 +12,9 @@
 
 #include <pybind11_diagram_utils.h>
 
+// Indices are added internally in bottleneck_distance, they are not needed in the input.
+static auto make_point(double x, double y, py::ssize_t) { return std::pair(x, y); };
+
 // For compatibility with older versions, we want to support e=None.
 // In C++17, the recommended way is std::optional<double>.
 double bottleneck(Dgm d1, Dgm d2, py::object epsilon)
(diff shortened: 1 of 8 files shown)


src/python/gudhi/bottleneck.cc
```diff
@@ -12,15 +12,18 @@
 
 #include <pybind11_diagram_utils.h>
 
+// Indices are added internally in bottleneck_distance, they are not needed in the input.
+static auto make_point(double x, double y, py::ssize_t) { return std::pair(x, y); };
+
 // For compatibility with older versions, we want to support e=None.
 // In C++17, the recommended way is std::optional<double>.
```
VincentRouvreau, line 19: As we are now in C++17, it may be a good opportunity to fix this, no ? (not mandatory on this PR, we can also open an issue)

src/python/gudhi/hera/wasserstein.cc
```diff
@@ -8,29 +8,126 @@
...
+    dist = std::pow(res.cost, 1./params.wasserstein_power);
+  }
+
+  if(!return_matching)
+    return py::cast(dist);
+
+  if(dist == std::numeric_limits<double>::infinity())
+    return py::make_tuple(dist, py::none());
+
+  // bug in Hera, matching_a_to_b_ is empty if one diagram is empty or both diagrams contain the same points
```
VincentRouvreau, line 62: Could you please open an [issue on hera](https://github.com/anigmetov/hera/issues) to track this ?

src/python/gudhi/hera/wasserstein.cc
```diff
@@ -49,8 +147,9 @@ PYBIND11_MODULE(wasserstein, m) {
             order (float): Wasserstein exponent W_q
             internal_p (float): Internal Minkowski norm L^p in R^2
             delta (float): Relative error 1+delta
+            matching (bool): if ``True``, computes and returns the optimal matching between X and Y, encoded as a (n x 2) np.array [...[i,j]...], meaning the i-th point in X is matched to the j-th point in Y, with the convention that (-1) represents the diagonal.
 
         Returns:
-            float: Approximate Wasserstein distance W_q(X,Y)
+            float|Tuple[float,numpy.array]: Approximate Wasserstein distance W_q(X,Y), and optionally the corresponding matching
```
VincentRouvreau, line 153: On [POT wasserstein version](https://gudhi.inria.fr/python/latest/wasserstein_distance_user.html#gudhi.wasserstein.wasserstein_distance), we also specify that it returns `None` when cost is +inf. Do you think it worth it to specify it here ?

VincentRouvreau (commented):
* Should we advertise it somehow in [wasserstein examples](https://gudhi.inria.fr/python/latest/wasserstein_distance_user.html#basic-example)
* This nice and new feature should be added to `.github/next_release.md`

src/python/gudhi/hera/wasserstein.cc
```diff
@@ -8,29 +8,126 @@
...
-#include <wasserstein.h> // Hera
-
 #include <pybind11_diagram_utils.h>
 
-double wasserstein_distance(
+#ifdef _MSC_VER
+// https://github.com/grey-narn/hera/issues/3
+// ssize_t is a non-standard type (well, posix)
+using py::ssize_t;
+#endif
```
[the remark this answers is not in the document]
mglisse: This hack is there because of a bug in Hera, I'd rather have it only in files that use Hera.
As I mentioned, it could make sense to merge the 2 cc files in hera, now that Hera can handle including bottleneck and wasserstein in the same translation unit. Then this part would only appear once...

src/python/gudhi/hera/wasserstein.cc
mglisse: There is already a TODO in Hera's code:
https://github.com/anigmetov/hera/blob/8bfdd4bd32f005c18b5c75c502b987de552d6e48/include/hera/wasserstein.h#L286

src/python/gudhi/bottleneck.cc
mglisse: It isn't broken, optional is just another way to handle this. But ok, done.

src/python/gudhi/hera/bottleneck.cc
```diff
@@ -16,13 +16,16 @@
 using py::ssize_t;
 #endif
 
-#include <bottleneck.h> // Hera
+#include <hera/bottleneck.h>
+
+// Indices are added internally in bottleneck_distance, they are not needed in the input.
+static auto make_point(double x, double y, py::ssize_t) { return std::pair(x, y); };
```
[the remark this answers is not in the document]
mglisse: This is mostly there as documentation. I think it would be less useful hidden in another file.

src/python/gudhi/hera/bottleneck.cc
```diff
@@ -16,13 +16,16 @@
...
 
 double bottleneck_distance(Dgm d1, Dgm d2, double delta)
 {
   // I *think* the call to request() has to be before releasing the GIL.
```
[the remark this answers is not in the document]
mglisse: I added a precision that the call to request() is inside numpy_to_range_of_pairs. The goal of having this comment here is to discourage people (including me) from moving the call to numpy_to_range_of_pairs inside the nogil region (it should be possible to write a nogil-safe numpy_to_range_of_pairs, but I am not convinced the current code would be safe for that).

src/python/gudhi/hera/wasserstein.cc
mglisse: done

VincentRouvreau (approved):

PR #738, opened 2022-11-16
More submodule options in the howto

mglisse:
Less chance of forgetting `git submodule update` after `git checkout` if it is automatic, maybe.

Should we advertise git@github.com:LOGIN/gudhi-devel.git rather than the https address for cloning? (separate issue)


VincentRouvreau (approved):
We could also explain how to `git submodule update --init` in a post-checkout git hook, but we are not going to explain all the git possibilities.

PR #744, opened 2022-11-18
sklearn's DistanceMetric has moved

mglisse:
Not sure if handling older sklearn is really worth it.

### DIFF
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -13,8 +13,13 @@ import numpy as np
 from sklearn.base          import BaseEstimator, TransformerMixin
 from sklearn.exceptions    import NotFittedError
 from sklearn.preprocessing import MinMaxScaler, MaxAbsScaler
-from sklearn.neighbors     import DistanceMetric
 from sklearn.metrics       import pairwise
+try:
+    # New location since 1.0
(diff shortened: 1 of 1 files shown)


VincentRouvreau (approved):
I agree with your fix but it is strange to me.

I have scikit-learn=1.1.1, but when I perform `from sklearn.neighbors import DistanceMetric`, I have no deprecation warning, while [the doc v1.0](https://scikit-learn.org/1.0/modules/generated/sklearn.metrics.DistanceMetric.html) makes like it was acted that DistanceMetric was moved.

mglisse:
I think I saw the warning somewhere in CI logs.
Maybe for a few releases they kept it available through both locations without warning, and they only started warning later when they decided that they should eventually remove the old location? Or there was a bug in the warning code, or...

With version 1.1.2
```
>>> from sklearn.neighbors import DistanceMetric
>>> DistanceMetric.get_metric('euclidean')
/usr/lib/python3/dist-packages/sklearn/neighbors/_distance_metric.py:10: FutureWarning: sklearn.neighbors.DistanceMetric has been moved to sklearn.metrics.DistanceMetric in 1.0. This import path will be removed in 1.3
  warnings.warn(
<sklearn.metrics._dist_metrics.EuclideanDistance object at 0x7f2a008ae510>
```

PR #748, opened 2022-11-18
Split Delaunay test to reduce memory usage

mglisse:
Another way to reduce the memory need for the compiler. It does increase compile time though.
@VincentRouvreau , do you like it better than using several files?


VincentRouvreau:
Fixed by #789 without the cmake loop

Issue #682, opened 2022-08-17
pip & macos arm64


mglisse:
Follow-up to #533. We have a conda package for macos on arm64. With pip, we have a package for macos on x86_64, but not arm64. Arm64 will soon (< 2 years?) become more common than x86_64 as a macos platform, so I believe we should keep an issue open about adding a pip arm64 package (or making the existing one universal, or removing the existing x86_64 one if we think everybody uses conda on mac).

VincentRouvreau:
@astamm Could you test the release candidate with `pip install gudhi==3.7.0rc1`, please ?

VincentRouvreau:
> Import also works. Do you want me test whether I can actually use it ?

That should be ok. If import is working, the rest should be ok. Thank you @astamm 
I am closing the issue. Official release will follow

PR #750, opened 2022-11-20
Test Mac CI

mglisse:
For #682.


mglisse:
Could it be that, if we provided universal binaries for gmp+mpfr, setting a couple environment variables would be enough to get universal binaries on Mac?

mglisse:
Test successful! I downloaded the wheel on a Mac M1 and ran some AlphaComplex test.

mglisse:
Replace by #767.

Issue #718, opened 2022-11-03
Investigate quadratic behaviour


VincentRouvreau:
As for #703 we should investigate if the same phenomena is present in other simplicial complex construction

To be checked for:
* [ ] Tangential complex
* [ ] Cech complex
* [ ] ...

VincentRouvreau:
```python
import gudhi as gd
from gudhi.datasets.generators import points
data = points.sphere(n_samples = 30000, ambient_dim = 2, radius = 1000., sample = "random")
r = gd.RipsComplex(points=data)
from timeit import default_timer as timer
from datetime import timedelta
start = timer(); st = r.create_simplex_tree(max_dimension=1); print(timedelta(seconds=timer()-start))
```
0:00:16.2 (master) vs 0:00:17.5 (master without [reserve](https://github.com/GUDHI/gudhi-devel/blob/55155afc5e9c9201476b9a2d9096a7ae874a27f6/src/Simplex_tree/include/gudhi/Simplex_tree.h#L1126))

mglisse:
> 0:00:16.2 (master) vs 0:00:17.5 (master without [reserve](https://github.com/GUDHI/gudhi-devel/blob/55155afc5e9c9201476b9a2d9096a7ae874a27f6/src/Simplex_tree/include/gudhi/Simplex_tree.h#L1126))

That call to reserve is not relevant to this issue. I think I fixed insert_graph in a previous PR, so anything based on a single call of insert_graph should be fine now.
(It always seemed wasteful to me to build an intermediate boost graph instead of inserting directly in the Simplex_tree, but that's unrelated and the possible gains are not huge.)

PR #754, opened 2022-11-21
Insert vertices first

mglisse:
Take advantage of the new helper to ensure these 2 functions don't hit #718.
Well, if we insert many edges (0,i) in decreasing order of i we will still hit the same issue, and there would be ways to solve or mitigate that, but it seems less likely than the vertex case.

### DIFF
--- a/src/python/gudhi/simplex_tree.pyx
+++ b/src/python/gudhi/simplex_tree.pyx
@@ -12,6 +12,7 @@ from libc.stdint cimport intptr_t, int32_t, int64_t
 import numpy as np
 cimport gudhi.simplex_tree
 cimport cython
+from numpy.math cimport INFINITY
 
 __author__ = "Vincent Rouvreau"
 __copyright__ = "Copyright (C) 2016 Inria"
@@ -239,7 +240,7 @@ cdef class SimplexTree:
 
     @staticmethod
     @cython.boundscheck(False)
-    def create_from_array(filtrations, double max_filtration=np.inf):
+    def create_from_array(filtrations, double max_filtration=INFINITY):
         """Creates a new, empty complex and inserts vertices and edges. The vertices are numbered from 0 to n-1, and
         the filtration values are encoded in the array, with the diagonal representing the vertices. It is the
         caller's responsibility to ensure that this defines a filtration, which can be achieved with either::
@@ -281,6 +282,8 @@ cdef class SimplexTree:
 
         .. seealso:: :func:`insert_batch`
         """
+        # Without this, it could be slow if we end up inserting vertices in a bad order (flat_map).
+        self.get_ptr().insert_batch_vertices(np.unique(np.stack((edges.row, edges.col))), INFINITY)
         # TODO: optimize this?
         for edge in zip(edges.row, edges.col, edges.data):
             self.get_ptr().insert((edge[0], edge[1]), edge[2])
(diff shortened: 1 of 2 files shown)


VincentRouvreau:
Benchmark results are interesting:
```python
import gudhi
from scipy.spatial.distance import cdist
import numpy as np

points = np.random.rand(10000,3)
distance_matrix = cdist(points, points)
start = timer(); cplx = gudhi.SimplexTree.create_from_array(distance_matrix); print(timedelta(seconds=timer()-start))
# 0:00:00.677781 (master @ 29cefe8ff7840903fc5850c73300e9f641f0fcf8) vs 0:00:00.481747 (PR #754)

points = np.random.rand(10000,4)
distance_matrix = cdist(points, points)
start = timer(); cplx = gudhi.SimplexTree.create_from_array(distance_matrix); print(timedelta(seconds=timer()-start))
# 0:00:00.742254 (master @ 29cefe8ff7840903fc5850c73300e9f641f0fcf8) vs 0:00:00.445652 (PR #754)


points = np.random.rand(10000,5)
distance_matrix = cdist(points, points)
start = timer(); cplx = gudhi.SimplexTree.create_from_array(distance_matrix); print(timedelta(seconds=timer()-start))
# 0:00:00.586704 (master @ 29cefe8ff7840903fc5850c73300e9f641f0fcf8) vs 0:00:00.450024 (PR #754)

vertices = np.random.randint(1000000, size=(1, 1000000))
filtrations = np.random.rand(1000000)
st = gudhi.SimplexTree()
start = timer(); st.insert_batch(vertices, filtrations); print(timedelta(seconds=timer()-start))
# 0:01:03.524540 (master @ 29cefe8ff7840903fc5850c73300e9f641f0fcf8) vs 0:00:00.243189 (PR #754)

edges = np.random.randint(1000000, size=(2, 1000000))
filtrations = np.random.rand(1000000)
st = gudhi.SimplexTree()
start = timer(); st.insert_batch(edges, filtrations); print(timedelta(seconds=timer()-start))
# 0:02:19.975196 (master @ 29cefe8ff7840903fc5850c73300e9f641f0fcf8) vs 0:00:00.663738 (PR #754)

triangles = np.random.randint(1000000, size=(3, 1000000))
filtrations = np.random.rand(1000000)
st = gudhi.SimplexTree()
start = timer(); st.insert_batch(triangles, filtrations); print(timedelta(seconds=timer()-start))
# 0:02:59.745888 (master @ 29cefe8ff7840903fc5850c73300e9f641f0fcf8) vs 0:00:01.291890 (PR #754)
```

mglisse:
Going straight to the most obvious case where it should help, if I use insert_batch to insert 50000 vertices in an empty SimplexTree on master, it takes:
* increasing order: 3ms
* decreasing order: 800ms

With this PR, both versions take 6ms.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -281,6 +282,8 @@ cdef class SimplexTree:
 
         .. seealso:: :func:`insert_batch`
         """
+        # Without this, it could be slow if we end up inserting vertices in a bad order (flat_map).
+        self.get_ptr().insert_batch_vertices(np.unique(np.stack((edges.row, edges.col))), INFINITY)
```
VincentRouvreau, line 286: Would it change something to use `np.concatenate` instead of `np.stack`, for example:
```python
row = np.array([2, 5, 3])
col = np.array([1, 4, 6])
dat = np.array([1, 2, 3])
edges = coo_matrix((dat, (row, col)))
edges = coo_matrix((dat, (row, col)))
np.stack((edges.row, edges.col))
# array([[2, 5, 3],
#        [1, 4, 6]], dtype=int32)
np.concatenate((edges.row, edges.col), axis=None)
#array([2, 5, 3, 1, 4, 6], dtype=int32)
```
I know that `np.unique` is doing the job after, but is it faster when it is already a flatten array ?

mglisse: I don't think it makes a difference whether we use stack, concatenate (with or without the axis argument), etc. The only thing that could make a (very small) difference is if one of those functions managed to optimize the case where we start from a 2d array (as in insert_batch) and use arr[0,:] as row and arr[1,:] as col. It would be possible for stack/concatenate to return a reference to arr instead of copying everything to a new array, but I doubt any of those functions tests for this case.

If concatenate looks more natural to you, I can use that, I really picked stack arbitrarily. Maybe it was shorter to type?

VincentRouvreau: Ok, let's keep it as is

VincentRouvreau (approved):

Issue #339, opened 2020-06-02
example/diagram_vectorizations_distances_kernels.py fit and transform


mglisse:
`example/diagram_vectorizations_distances_kernels.py` now contains 3 dead calls to `fit` methods, and 0 call to `transform`, or the `pairwise_*` functions. With no tests (just running the example), that's a lot of untested code...

PR #755, opened 2022-11-22
Tests for kernels and distances

wreise:
I started from fixing #339 and I think I noticed two bugs, which I also propose to fix.

1. The normalization in `SlicedWassersteinKernel.__call__` should be applied before exp, instead of after. It is correct in `transform` and in other classes.
2. The `PersistenceFisherKernel` has two bandwidths attributes `bandwidth_fisher` and `bandwidth`. According to the documentation, the first is passed to `PersistenceFisherDistance` or `pairwise_persistence_diagram_distances` (with the keyword `bandwidth`), while `bandwidth` itself is used to normalize the value of the distance, in a way analogue to `SlicedWassersteinKernel`.

As for the tests themselves, they just check the consistency between different ways of calculating the same distances.

Remaining:
- [x] test `PersistenceScaleSpaceKernel` and `PersistenceWeightedGaussianKernel`
- [x] test values for some simple diagrams.


wreise:
I noticed that `src/python/gudhi/representations/metrics.py` contains 2 ways to calculate a sliced wasserstein between a pair of diagrams. First, using `_sliced_wasserstein_distance` directly, and a second one, with `_compute_persistence_diagram_projections` followed by `_sliced_wasserstein_distance_on_projections`.
I understand the need for the second version, which precomputes the projection. However, is there an advantage in using `_sliced_wasserstein_distance`?

MathieuCarriere:
> I started from fixing #339 and I think I noticed two bugs, which I also propose to fix.
> 
>     1. The normalization in `SlicedWassersteinKernel.__call__` should be applied before exp, instead of after. It is correct in `transform` and in other classes.

Yes, you are totally right, thanks for catching this! Can't believe I missed it...

>     2. The `PersistenceFisherKernel` has two bandwidths attributes `bandwidth_fisher` and `bandwidth`. According to the documentation, the first is passed to `PersistenceFisherDistance` or `pairwise_persistence_diagram_distances` (with the keyword `bandwidth`), while `bandwidth` itself is used to normalize the value of the distance, in a way analogue to `SlicedWassersteinKernel`.

Indeed, I used `_fisher` so as to not use `bandwidth` twice, so it is better to ask for more consistency.

MathieuCarriere:
> I noticed that `src/python/gudhi/representations/metrics.py` contains 2 ways to calculate a sliced wasserstein between a pair of diagrams. First, using `_sliced_wasserstein_distance` directly, and a second one, with `_compute_persistence_diagram_projections` followed by `_sliced_wasserstein_distance_on_projections`. I understand the need for the second version, which precomputes the projection. However, is there an advantage in using `_sliced_wasserstein_distance`?

Simplicity I would say... Like, if you just want to compute one SW distance for a pair of persistence diagrams, there is no need to precompute projections.

src/python/gudhi/representations/kernel_methods.py
```diff
@@ -125,7 +125,7 @@ def transform(self, X):
         Returns:
             numpy array of shape (number of diagrams in **diagrams**) x (number of diagrams in X): matrix of pairwise sliced Wasserstein kernel values.
         """
-        return pairwise_persistence_diagram_kernels(X, self.diagrams_, kernel="sliced_wasserstein", bandwidth=self.bandwidth, num_directions=self.num_directions, n_jobs=self.n_jobs)
+        return pairwise_persistence_diagram_kernels(self.diagrams_, X, kernel="sliced_wasserstein", bandwidth=self.bandwidth, num_directions=self.num_directions, n_jobs=self.n_jobs)
```
MathieuCarriere, line 128: There is indeed an issue here, but it is in the doc: in scikit-learn pipelines, when computing new kernel matrices on, say, some test data (usually provided as the `X` argument), the shape of the matrix should be [new_test_data, train_data]. Hence, we should keep `X, self.diagrams_` as first two arguments but change the doc to `(number of diagrams in X) x (number of diagrams in **diagrams**)`

wreise: Oh, thank you for pointing that out! I reverted the transposition and I have changed the docs for all transformers in `src/python/gudhi/representations/kernel_methods`.

MathieuCarriere (approved):

VincentRouvreau:
Thank you for the [code coverage](https://output.circle-artifacts.com/output/job/c2f83b13-1b7e-4bee-9ea7-fd2953391156/artifacts/0/htmlcov/index.html) improvement of representations:
Module | before PR #755 | after PR #755
--- | --- | ---
gudhi/representations/__init__.py       	| 	100%    |   100%
gudhi/representations/kernel_methods.py 	| 	71%     |   99%
gudhi/representations/metrics.py 	        | 	76%     |   89%
gudhi/representations/preprocessing.py  	| 	75%     |   75%
gudhi/representations/vector_methods.py     | 	97%     |   97%

VincentRouvreau (approved):

PR #758, opened 2022-11-30
Fix documentation and mechanism to expose subsampling modules that does not depend on cgal

### LINKED ISSUE #730: Subsampling dependency on CGAL
https://gudhi.inria.fr/python/latest/point_cloud.html#subsampling says that gudhi.subsampling depends on CGAL+Eigen. But looking at the C++ code, only sparsify_point_set uses CGAL. It would be nice if the functions that do not use CGAL could be available when CGAL is missing, and the documentation needs to be more precise, since this has an impact on licensing.

VincentRouvreau:
Fix #730 except [this comment](https://github.com/GUDHI/gudhi-devel/issues/730#issuecomment-1331912661)


src/python/include/Subsampling_interface.h
```diff
@@ -99,6 +110,16 @@ std::vector<std::vector<double>> subsampling_sparsify_points_from_file(const std
   std::vector<std::vector<double>> points = off_reader.get_point_cloud();
   return subsampling_sparsify_points(points, min_squared_dist);
 }
+#else
+// ------ sparsify_point_set ------
+std::vector<std::vector<double>> subsampling_sparsify_points(const std::vector<std::vector<double>>&, double) {
+  return std::vector<std::vector<double>>{{}};
```
mglisse, line 116: I don't think you need double `{{}}`, a single `{}` should work. Actually just `return {};`. We could also have thrown the exception from here, but your version with GUDHI_SUBSAMPLING_USE_CGAL is ok.

src/python/gudhi/subsampling.pyx
```diff
@@ -104,7 +109,9 @@ def pick_n_random_points(points=None, off_file='', nb_points=0):
         return subsampling_n_random_points(points, nb_points)
 
 def sparsify_point_set(points=None, off_file='', min_squared_dist=0.0):
-    """Outputs a subset of the input points so that the squared distance
+    """:Requires: `Eigen <installation.html#eigen>`_ :math:`\geq` 3.1.0 and `CGAL <installation.html#cgal>`_ :math:`\geq` 4.11.0
```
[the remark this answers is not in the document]
VincentRouvreau: I see what you mean :thinking: ...
This case is for users who try to build the module themself, but for some reason they are ending with a gudhi module without `sparsify_point_set` function.
Where should it be documented ? In [installation manual](https://gudhi.inria.fr/python/latest/installation.html#cgal), we just say `cf. Module list`, and each module explains its dependencies.
So I put the information at module `Point cloud utilities` level.
Tell me if it is ok for you, cf. fafc40c5

src/python/include/Subsampling_interface.h
VincentRouvreau: Done on 9a2ed4e7

src/Subsampling/include/gudhi/sparsify_point_set.h
```diff
@@ -14,13 +14,19 @@
...
 #ifdef GUDHI_SUBSAMPLING_PROFILING
 #include <gudhi/Clock.h>
 #endif
 
 #include <cstddef>
 #include <vector>
 
+// Make compilation fail - required for external projects
+#if CGAL_VERSION_NR < 1041101000
+# error subsampling::sparsify_point_set is only available with CGAL >= 4.11
```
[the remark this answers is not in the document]
VincentRouvreau: Even if this error will never be shown, I would prefer to keep it because removing it would imply to know exactly what happens for CGAL, and like that we keep consistency and knowledge between code and CMake conditions inside the gudhi code. Is it ok for you ?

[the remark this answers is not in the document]
mglisse: > Even if this error will never be shown, I would prefer to keep it because removing it would imply to know exactly what happens for CGAL,

DRY and locality both say this error makes more sense in the file that actually directly uses CGAL.

> and like that we keep consistency and knowledge between code and CMake conditions inside the gudhi code.

CMake stuff has much lower priority to me...
Do we still care about CGAL < 4.11 at all? Even Ubuntu 18.04 had 4.11.

> Is it ok for you ?

Yes, it is ok to keep it a bit longer, it has been here for a while anyway.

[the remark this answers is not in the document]
mglisse: > Yes, it is ok to keep it a bit longer, it has been here for a while anyway.

Uh, my brain must be off, this is new code added by this PR... I don't like adding this, but I won't fight against it.

src/python/gudhi/subsampling.pyx
```diff
@@ -27,6 +27,11 @@ cdef extern from "Subsampling_interface.h" namespace "Gudhi::subsampling":
     vector[vector[double]] subsampling_sparsify_points(vector[vector[double]] points, double min_squared_dist)
     vector[vector[double]] subsampling_sparsify_points_from_file(string off_file, double min_squared_dist)
 
+cdef extern from "Subsampling_interface.h":
+    cdef int _GUDHI_SUBSAMPLING_USE_CGAL
+
+GUDHI_SUBSAMPLING_USE_CGAL = _GUDHI_SUBSAMPLING_USE_CGAL
```
mglisse, line 33: Is it useful to have both GUDHI_SUBSAMPLING_USE_CGAL and _GUDHI_SUBSAMPLING_USE_CGAL? (I didn't think long about it)

mglisse (approved):

src/python/gudhi/subsampling.pyx
VincentRouvreau: I could do it this way:
```python
cdef extern from "Subsampling_interface.h":
    cdef enum:
        _GUDHI_SUBSAMPLING_USE_CGAL
```
Then in cython `if not _GUDHI_SUBSAMPLING_USE_CGAL:` would work, but the information wouldn't be available at python level (required for unitary tests purpose)

src/python/include/Subsampling_interface.h
```diff
@@ -14,9 +14,19 @@
...
-#include <gudhi/sparsify_point_set.h>
 #include <gudhi/Points_off_io.h>
+
+// Default value is undefined
+#define _GUDHI_SUBSAMPLING_USE_CGAL 0
+
+#if __has_include(<CGAL/version.h>)
+#include <CGAL/version.h>  // for CGAL_VERSION_NR
+#if CGAL_VERSION_NR >= 1041101000
+#define _GUDHI_SUBSAMPLING_USE_CGAL 1
```
mglisse, line 25: First doing `#undef _GUDHI_SUBSAMPLING_USE_CGAL` would avoid a warning.

PR #762, opened 2022-12-05
Insert sorted vertices in batch for tangential complex

VincentRouvreau:
Fix partially #718 for the tangential complex

### DIFF
--- a/src/Tangential_complex/include/gudhi/Tangential_complex.h
+++ b/src/Tangential_complex/include/gudhi/Tangential_complex.h
@@ -625,8 +625,7 @@ class Tangential_complex {
 
     int max_dim = -1;
 
-    // Vertices to be inserted first by the create_complex method to avoid quadratic complexity.
-    // It isn't just [0, n) if some points have multiplicity (only one copy appears in the complex).
+    // Ordered vertices to be inserted first by the create_complex method to avoid quadratic complexity.
     std::vector<typename Simplex_tree_::Vertex_handle> vertices(m_points.size());
     std::iota(vertices.begin(), vertices.end(), 0);
     tree.insert_batch_vertices(vertices);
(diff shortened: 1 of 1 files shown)


VincentRouvreau:
```python
import gudhi as gd
data = points.sphere(n_samples = 500000, ambient_dim = 2, radius = 1000., sample = "random")
tc = gd.TangentialComplex(intrisic_dim = 1,points=data)
tc.compute_tangential_complex()
result_str = 'Tangential contains ' + repr(tc.num_simplices()) + \
    ' simplices - ' + repr(tc.num_vertices()) + ' vertices.'
print(result_str)
# Tangential contains 229123 simplices - 114562 vertices.
start = timer(); st = tc.create_simplex_tree(); print(timedelta(seconds=timer()-start))
# 0:00:24.924412 (master @ 5d0c294572a9e34a7a70e78806ac3cf805357712) vs 0:00:00.391726 (PR #762)
```

src/Tangential_complex/include/gudhi/Tangential_complex.h
```diff
@@ -624,6 +625,12 @@ class Tangential_complex {
 
     int max_dim = -1;
 
+    // Vertices to be inserted first by the create_complex method to avoid quadratic complexity.
+    // It isn't just [0, n) if some points have multiplicity (only one copy appears in the complex).
+    std::vector<typename Simplex_tree_::Vertex_handle> vertices(m_points.size());
+    std::iota(vertices.begin(), vertices.end(), 0);
```
mglisse, line 631: The comment does not match the code at all.

mglisse (commented):
I had forgotten that the tangential complex required a simplex tree instead of some concept like the others. At least here it means we don't need to modify the concept :-)

src/Tangential_complex/include/gudhi/Tangential_complex.h
VincentRouvreau: Bad copy/paste of a comment. Fixed on [6b8da55](https://github.com/GUDHI/gudhi-devel/pull/762/commits/6b8da55930c5a76aefc6725d3cffb17bab5438d0)

mglisse (approved):

PR #765, opened 2022-12-09
pip package for python 3.11

VincentRouvreau:


mglisse (approved):
We could probably drop 3.6 and 3.7, but I'll let you decide when you want to do that.

VincentRouvreau:
> We could probably drop 3.6 and 3.7, but I'll let you decide when you want to do that.

Python 3.6 is only 6 years old, let's keep these versions for the moment.

mglisse:
As mentioned in #767, I think there may be a problem with the osx packaging action with python 3.11, because the path is not ~/.local/bin anymore. But we can discuss that there.

PR #781, opened 2022-12-20
Speed-up persistence of 1d cubical complex

mglisse:
The gains are not huge, but the amount of work wasn't either.
I think it would be worth modifying get_(co)boundary_of_a_cell a bit more to avoid compute_counter_for_given_cell, but I only went for mechanic changes today.

### DIFF
--- a/src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
+++ b/src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
@@ -635,6 +635,7 @@ class Bitmap_cubical_complex_base {
   std::size_t total_number_of_cells;
 
   void set_up_containers(const std::vector<unsigned>& sizes, bool is_pos_inf) {
+    // The fact that multipliers[0]=1 is relied on by optimizations in other functions
     unsigned multiplier = 1;
     for (std::size_t i = 0; i != sizes.size(); ++i) {
       this->sizes.push_back(sizes[i]);
--- a/src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
+++ b/src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
@@ -659,10 +659,13 @@ class Bitmap_cubical_complex_base {
   std::vector<unsigned> compute_counter_for_given_cell(std::size_t cell) const {
     std::vector<unsigned> counter;
     counter.reserve(this->sizes.size());
-    for (std::size_t dim = this->sizes.size(); dim != 0; --dim) {
-      counter.push_back(cell / this->multipliers[dim - 1]);
+    for (std::size_t dim = this->sizes.size(); dim > 1; --dim) {
+      std::size_t quot = cell / this->multipliers[dim - 1];
       cell = cell % this->multipliers[dim - 1];
+      counter.push_back(quot);
     }
+    // Split out the last iteration to save a costly division by multipliers[0]=1
+    counter.push_back(cell);
     std::reverse(counter.begin(), counter.end());
     return counter;
   }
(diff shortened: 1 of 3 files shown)


src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
```diff
@@ -659,10 +659,13 @@ class Bitmap_cubical_complex_base {
...
     std::vector<unsigned> counter;
     counter.reserve(this->sizes.size());
-    for (std::size_t dim = this->sizes.size(); dim != 0; --dim) {
-      counter.push_back(cell / this->multipliers[dim - 1]);
+    for (std::size_t dim = this->sizes.size(); dim > 1; --dim) {
+      std::size_t quot = cell / this->multipliers[dim - 1];
       cell = cell % this->multipliers[dim - 1];
+      counter.push_back(quot);
     }
+    // Split out the last iteration to save a costly division by multipliers[0]=1
```
VincentRouvreau, line 667: `multipliers[0]=1` is always true ?
Maybe we could `assert(multipliers[0] == 1);` (where it is supposed), just to be sure some future modification will not break this. If `multipliers[0]=1` is true by construction, you can forget this comment.

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -726,7 +727,7 @@ class Persistent_cohomology {
   /*  Dictionary establishing the correspondence between the Simplex_key of
    * the root vertex in the union-find ds and the Simplex_key of the vertex which
    * created the connected component as a 0-dimension homology feature.*/
-  std::map<Simplex_key, Simplex_key> zero_cocycles_;
+  std::unordered_map<Simplex_key, Simplex_key> zero_cocycles_;
```
VincentRouvreau, line 730: I imagine you also benchmarked this modification with the Simplex_tree ?

mglisse: I think so, but that was 3 months ago so I don't remember... I'll try again.

mglisse: In the branch `persline` (so I can use the benchmark line.cpp), I get for replacing map with unordered_map

**Before**
```
Mostly unique values
Cubical complex from top cells: 0.86s

Cubical complex from vertices: 0.746s

Simplex tree: 0.921s

Special case for function on a line: 0.007s

Many repeated values
Cubical complex from top cells: 1.043s

Cubical complex from vertices: 0.982s

Simplex tree: 0.628s

Special case for function on a line: 0.007s
```
**After**
```
Mostly unique values
Cubical complex from top cells: 0.702s

Cubical complex from vertices: 0.633s

Simplex tree: 0.772s

Special case for function on a line: 0.007s

Many repeated values
Cubical complex from top cells: 0.989s

Cubical complex from vertices: 0.934s

Simplex tree: 0.622s

Special case for function on a line: 0.007s
```
So sometimes is helps, sometimes it is neutral. I also tested on a Rips graph with a much larger ratio nb_edges / nb_vertices, and it decreased the time from 240ms to 200ms.

src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
VincentRouvreau: As you prefer, either a `GUDHI_CHECK` everywhere you say `Split out the last iteration to save a costly division by multipliers[0]=1`, or a comment where `multipliers` is initialized to explain that the code has been optimized making the assumption that `multipliers[0]=1` when looping on multipliers values.

mglisse: I took option 2 :wink:

PR #789, opened 2023-01-13
Split Delaunay complex in 4 executables to relieve windows CI

VincentRouvreau:
Replace #748


mglisse:
Having 4 copies of the code is not very satisfactory. If we improve the test, we have to copy the improvement to all 4 files... Couldn't we at least put most of the code in a .h file, and then the .cpp would just have
```c++
#include "test_delaunay.h"
#include <CGAL/Epeck_d.h>
BOOST_AUTO_TEST_CASE(Delaunay_complex_exact_kernel_dynamic_simplices_comparison) {
  my_test<CGAL::Epeck_d<CGAL::Dynamic_dimension_tag>>();
}
```
?

VincentRouvreau:
> Having 4 copies of the code is not very satisfactory. If we improve the test, we have to copy the improvement to all 4 files... Couldn't we at least put most of the code in a .h file, and then the .cpp would just have
> 
> ```c++
> #include "test_delaunay.h"
> #include <CGAL/Epeck_d.h>
> BOOST_AUTO_TEST_CASE(Delaunay_complex_exact_kernel_dynamic_simplices_comparison) {
>   my_test<CGAL::Epeck_d<CGAL::Dynamic_dimension_tag>>();
> }
> ```
> 
> ?

Sorry, you are completely right, it was not DRY at all... Done on [d355813](https://github.com/GUDHI/gudhi-devel/pull/789/commits/d35581357525d02e3126e9fe901904cc764e1f5c)

mglisse (approved):
I would have put `#include <boost/test/unit_test.hpp>` in the .h, but that's a detail, ok.

PR #791, opened 2023-01-13
workaround ssize_t in Hera also in Nerve_GIC

mglisse:


src/Nerve_GIC/include/gudhi/GIC.h
```diff
@@ -22,6 +22,13 @@
 # include <gudhi/Bottleneck.h>
 #elif __has_include(<hera/bottleneck.h>)
 # define GUDHI_GIC_USE_HERA 1
+# ifdef _MSC_VER
+// https://github.com/grey-narn/hera/issues/3
+// ssize_t is a non-standard type (well, posix)
+#  include <type_traits>
+#  include <cstdlib>
+using ssize_t = std::make_signed_t<std::size_t>;
```
VincentRouvreau, line 30: Would `std::ptrdiff_t` be a good candidate too ? As it is a signed integer type and already in `<cstdlib>`

mglisse: I think asking for the signed version of size_t minimizes the risk that some other random header elsewhere defines it to something incompatible (it is in the global namespace). For instance, on a 32 bit platform, one place could define it as int and the other as long :boom: 

Another option (untested) would be to define `ssize_t` inside namespace hera (this is already a hack anyway), and then we can define it to whatever we like with little risk of conflict. Still, even in that case, `make_signed_t<size_t>` seems to match most closely the intent of the original author...

VincentRouvreau (approved):

src/Nerve_GIC/include/gudhi/GIC.h
VincentRouvreau: Ok I understand. My remark was because [std::ssize (C++20 function)](https://en.cppreference.com/w/cpp/iterator/size) returns a std::ptrdiff_t, so I think we could do safely (I hope) `using ssize_t = std::ptrdiff_t`.
But ok, your explanation makes sense.

VincentRouvreau (approved):

src/Nerve_GIC/include/gudhi/GIC.h
mglisse: Yes, I guess that would likely work.
Maybe the next time we touch those workarounds we can both move the typedef inside namespace hera and use the simpler ptrdiff_t.

VincentRouvreau (approved):

PR #797, opened 2023-01-17
Add to_file/from_file methods to the SimplexTree

VincentRouvreau:
This is a first naive python implementation that uses C++ `operator>>` and `operator<<`.

To be discussed:
* should the file format be described in [C++ file format](https://gudhi.inria.fr/doc/latest/fileformats.html) and [python file format](https://gudhi.inria.fr/python/latest/fileformats.html)
* should it be implemented through `__show__`/`__str__`/`__repr__` python magic methods ?
* should it use `pandas get_handle` that can compress files on the fly based on the file name, like:
```
from pandas.io.common import get_handle

handles = get_handle('my_file.xz', 'w', compression = "infer")
handles.handle.write("something")
handles.handle.close() # my_file.xz is saved as a compresssed text file
```

### DIFF
--- a/src/python/include/Simplex_tree_interface.h
+++ b/src/python/include/Simplex_tree_interface.h
@@ -233,6 +233,18 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
     auto boundary_srange = Base::boundary_simplex_range(bd_sh);
     return std::make_pair(boundary_srange.begin(), boundary_srange.end());
   }
+
+  void read(std::string filename) {
+    std::ifstream ifs(filename.c_str());
+    ifs >> *this;
+    ifs.close();
+  }
+
+  void write(std::string filename) {
+    std::ofstream ofs(filename.c_str());
+    ofs << *this;
+    ofs.close();
+  }
 };
 
 }  // namespace Gudhi
--- a/src/python/include/Simplex_tree_interface.h
+++ b/src/python/include/Simplex_tree_interface.h
@@ -235,13 +235,13 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
   }
 
   void read(std::string filename) {
-    std::ifstream ifs(filename.c_str());
+    std::ifstream ifs(filename);
     ifs >> *this;
     ifs.close();
   }
 
   void write(std::string filename) {
-    std::ofstream ofs(filename.c_str());
+    std::ofstream ofs(filename);
     ofs << *this;
     ofs.close();
   }
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -1757,7 +1757,7 @@ class Simplex_tree {
 // Print a Simplex_tree in os.
 template<typename...T>
 std::ostream& operator<<(std::ostream & os, Simplex_tree<T...> & st) {
-  for (auto sh : st.filtration_simplex_range()) {
+  for (auto sh : st.complex_simplex_range()) {
     os << st.dimension(sh) << " ";
     for (auto v : st.simplex_vertex_range(sh)) {
       os << v << " ";
--- a/src/python/gudhi/simplex_tree.pxd
+++ b/src/python/gudhi/simplex_tree.pxd
@@ -80,6 +80,8 @@ cdef extern from "Simplex_tree_interface.h" namespace "Gudhi":
         # Expansion with blockers
         ctypedef bool (*blocker_func_t)(vector[int], void *user_data)
         void expansion_with_blockers_callback(int dimension, blocker_func_t user_func, void *user_data)
+        void read(string file_name) nogil
+        void write(string file_name) nogil
 
 cdef extern from "Persistent_cohomology_interface.h" namespace "Gudhi":
     cdef cppclass Simplex_tree_persistence_interface "Gudhi::Persistent_cohomology_interface<Gudhi::Simplex_tree_interface<Gudhi::Simplex_tree_options_full_featured>>":
--- a/src/python/gudhi/simplex_tree.pyx
+++ b/src/python/gudhi/simplex_tree.pyx
@@ -9,6 +9,7 @@
 
 from cython.operator import dereference, preincrement
 from libc.stdint cimport intptr_t, int32_t, int64_t
+from libcpp.string cimport string
 import numpy as np
 cimport gudhi.simplex_tree
 cimport cython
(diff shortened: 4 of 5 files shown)


VincentRouvreau:
Some benchmark:
```python
import gudhi as gd
off_file = gd.__root_source_dir__ + '/data/points/tore3D_1307.off'
points = gd.read_points_from_off_file(off_file = off_file)
stree = gd.RipsComplex(points = points, max_edge_length = 0.5).create_simplex_tree(max_dimension=2)
stree.num_simplices() # 1262223
start = timer(); stree.to_file('torus.st'); print(timedelta(seconds=timer()-start)) # 0.700751 sec
# torus.st file size is 28 Mo
stree = gd.RipsComplex(points = points, max_edge_length = 0.6).create_simplex_tree(max_dimension=2)
stree.num_simplices() # 3556326
start = timer(); stree.to_file('torus.st'); print(timedelta(seconds=timer()-start)) # 2.104864 sec
# torus.st file size is 79 Mo
```

src/python/include/Simplex_tree_interface.h
```diff
@@ -233,6 +233,18 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
...
   }
+
+  void read(std::string filename) {
+    std::ifstream ifs(filename.c_str());
+    ifs >> *this;
+    ifs.close();
+  }
+
+  void write(std::string filename) {
+    std::ofstream ofs(filename.c_str());
```
mglisse, line 244: You can pass the string directly.

src/python/include/Simplex_tree_interface.h
```diff
@@ -233,6 +233,18 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
...
+    ofs << *this;
```
mglisse, line 245: What precision is used for the coordinates?

mglisse (commented):
(not a review)

DavidLapous:
Thanks Vincent for this implementation ! I've tried with a few examples, and it works without any issues.
I think the  `simplextree.from_file` method could be defined as a static method (with `@staticmethod`), as we shouldn't need an instance of a simplextree to create a new one. An alternative would be to have a constructor from a file:
```python
new_simplextree = gd.SimplexTree.from_file(<path>)  # or
new_simplextree = gd.SimplexTree(path=<path>) 
```
I've noticed that re-constructing a simplex tree from a dump is in general slower than recreating it from its dataset (although, maybe if we're doing edge collapses, it might not be the case). I've tried a favorable case from python (removing the io disk operations), using the code
```python
def DumpedSimplexTreeFileToDumpedSimplexTreePython(path:str)->list[tuple[np.ndarray, np.ndarray]]:
	simplextree_file = open(path, "r")
	def line2simplex(line):
		stripped_line = line.strip().split(" ")
		d, s, f = int(stripped_line[0]), np.array(stripped_line[1:-1], dtype=int), float(stripped_line[-1])
		return d,s,f
	lines = [line2simplex(line) for line in simplextree_file.readlines()]
	max_dimension = max(d for d,s,f in lines)
	simplex_lines = [[[s,f] for d,s,f in lines if d == dim] for dim in range(max_dimension+1)]
	return [(np.array([s for s,f in simplices_of_dim], dtype=int).T, np.array([f for s,f in simplices_of_dim], dtype=float)) for dim, simplices_of_dim in enumerate(simplex_lines)]
python_dump = DumpedSimplexTreeFileToDumpedSimplexTreePython("torus.st")

# %%time
st = gd.SimplexTree()
for simplices_filtration in python_dump:
	simplices, filtration = simplices_filtration
	st.insert_batch(simplices, filtration)
# Takes 1.8s, and the st.from_file() takes 2.5s
```
But it is still ~10 times slower than reading the dataset and recreating the simplextree from it.

Do you think there could be a better (pickle-able) python structure from which to reconstruct a simplextree ?

Another point is that I think it would be nice to have a python dump function that returns a pickle-able python structure, to be able to use parallelism functions, e.g., 
```python
from joblib import Parallel, delayed
Parallel(n_jobs=2)(delayed(function)(simplextree.dump()) for simplextree in simplextree_iterable)
```

VincentRouvreau:
@mglisse benchmark to compare filtration_simplex_range and complex_simplex_range (cf. [ce62b37](https://github.com/GUDHI/gudhi-devel/pull/797/commits/ce62b37f17c548f504707ac56687ee4632292fd2)):
```python
import gudhi as gd
off_file = gd.__root_source_dir__ + '/data/points/tore3D_1307.off'

start = timer(); points = gd.read_points_from_off_file(off_file = off_file); stree = gd.RipsComplex(points = points, max_edge_length = 0.5).create_simplex_tree(max_dimension=2); print(timedelta(seconds=timer()-start)) # 0.043483 sec
stree.num_simplices() # 1262223
start = timer(); stree.to_file('torus_0.5.st'); print(timedelta(seconds=timer()-start)) # 0.700751 sec with filtration_simplex_range
                                                                                        # 0.487543 sec with complex_simplex_range
# torus_0.5.st file size is 28 Mo
start = timer(); copy = gd.SimplexTree.create_from_file('torus_0.5.st'); print(timedelta(seconds=timer()-start)) # 0.884387 sec with filtration_simplex_range
                                                                                                                 # 0.709731 sec with complex_simplex_range

start = timer(); points = gd.read_points_from_off_file(off_file = off_file); stree = gd.RipsComplex(points = points, max_edge_length = 0.6).create_simplex_tree(max_dimension=2); print(timedelta(seconds=timer()-start)) # 0.082825 sec
stree.num_simplices() # 3556326
start = timer(); stree.to_file('torus_0.6.st'); print(timedelta(seconds=timer()-start)) # 2.104864 sec with filtration_simplex_range
                                                                                        # 1.366111 sec with complex_simplex_range
# torus_0.6.st file size is 79 Mo
copy = gd.SimplexTree()
start = timer(); copy = gd.SimplexTree.create_from_file('torus_0.6.st'); print(timedelta(seconds=timer()-start)) # 2.657877 sec with filtration_simplex_range
                                                                                                                 # 2.003719 sec with complex_simplex_range
```

mglisse:
> @mglisse benchmark to compare filtration_simplex_range and complex_simplex_range (cf. [ce62b37](https://github.com/GUDHI/gudhi-devel/pull/797/commits/ce62b37f17c548f504707ac56687ee4632292fd2)):

The speed ratio for create_from_file would probably be even larger for a simplicial complex that contains only vertices, and where vertex i has filtration value -i. Anyway, yes, avoid filtration_simplex_range when it isn't needed.

mglisse:
> But it is still ~10 times slower than reading the dataset and recreating the simplextree from it.
> 
> Do you think there could be a better (pickle-able) python structure from which to reconstruct a simplextree ?

For something easy to do in Python, I think the structure you chose (two arrays per dimension, one for the vertices of each simplex and one for the filtrations) is pretty good. Ordering the simplices as per get_simplices rather than get_filtration would probably help a bit, but no miracle.

The expansion function is very fast, specialized C++ code is required to match it. If your simplex tree always comes from the expansion of a graph, in my opinion it would make sense to arrange so you can pickle the graph (the simplex tree before expansion) and not the expanded complex. Even if we write optimized (un)pickle functions that are slightly faster than expansion, the size of the dump can be orders of magnitude smaller before expansion, and that can affect performance.

If instead the simplex tree comes from an alpha complex, which is costly to compute relatively to its size, this is a different matter and dumping the simplex tree should be worth it even with unoptimized code.

DavidLapous:
> For something easy to do in Python, I think the structure you chose (two arrays per dimension, one for the vertices of each simplex and one for the filtrations) is pretty good. Ordering the simplices as per get_simplices rather than get_filtration would probably help a bit, but no miracle.

Do you think it would be useful then to have a python class (e.g. `DumpedSimplexTree`) that holds this kind of data, to "hide the dumped structure to the user" and which has methods to : 1. write this to a file and 2. unpickle it into a simplextree ?  

> The expansion function is very fast, specialized C++ code is required to match it. If your simplex tree always comes from the expansion of a graph, in my opinion it would make sense to arrange so you can pickle the graph (the simplex tree before expansion) and not the expanded complex. Even if we write optimized (un)pickle functions that are slightly faster than expansion, the size of the dump can be orders of magnitude smaller before expansion, and that can affect performance.

So we need to store a variable that store the dimension on which to expand the (dumped) graph in that case. That implies that we have to track this dimension (in that case) before or during the dump (unless there is a fast way to tell that a simplextree is a clique complex).
1.  If the user has expended the simplextree (before dump), store this dimension in the simplextree python structure, or 
2. When using the dump function, the user use a "clique complex flag".

I prefer the first one because it requires no user input, but it is not stable to `insert`ing simplices, and (may) hide some computation to the user.

VincentRouvreau:
Thanks for your feedback and tips. I will come back with another PR but written differently to take into account your advices

mglisse:
> Do you think it would be useful then to have a python class (e.g. `DumpedSimplexTree`) that holds this kind of data, to "hide the dumped structure to the user" and which has methods to : 1. write this to a file and 2. unpickle it into a simplextree ?

It would be useful to have a function that does what is needed for pickle to work. I don't really see much point using a class instead of a list of (tuple of) arrays, but I didn't think very hard. Anyway, I think Vincent is going to write an optimized version so that the dump is smaller and the reconstruction faster.

> So we need to store a variable that store the dimension on which to expand the (dumped) graph in that case. That implies that we have to track this dimension (in that case) before or during the dump (unless there is a fast way to tell that a simplextree is a clique complex).
> 
>     1. If the user has expanded the simplextree (before dump), store this dimension in the simplextree python structure, or
> 
>     2. When using the dump function, the user use a "clique complex flag".
> 
> 
> I prefer the first one because it requires no user input, but it is not stable to `insert`ing simplices, and (may) hide some computation to the user.

I don't much like the idea of making this magic. Some complexes are flag complexes, but that's just one special case, having to remember this information is not nice. And in my opinion, it is the user's responsibility to choose what / when to dump. Imagine that you have a point set in R^2, i.e. a (n,2) array, and you compute the (n,n) array of pairwise distances. You may choose to dump the distance matrix. Or you may choose to dump the point coordinates (much smaller) and reconstruct the matrix when you need it. That's up to the user, numpy/scipy does **not** remember that the matrix was produced by cdist just so pickle could use that information.

PR #798, opened 2023-01-18
Remove hard-coded double and more explicit on Filtration_value type

### LINKED ISSUE #595: hard-coded double in Alpha_complex
https://github.com/GUDHI/gudhi-devel/blob/bbff86f1218fc7bc9976353901aa94cfa54792f6/src/Alpha_complex/include/gudhi/Alpha_complex.h#L407

Should this be `std::numeric_limits<Filtration_value>::quiet_NaN()`, with `static_assert(std::numeric_limits<Filtration_value>::has_quiet_NaN)`?

To support more general types, we may also need to replace `std::isnan(x)` with `isnan(x)` after `using std::isnan;`.

VincentRouvreau:
Fix #595

### DIFF
--- a/src/Alpha_complex/concept/SimplicialComplexForAlpha.h
+++ b/src/Alpha_complex/concept/SimplicialComplexForAlpha.h
@@ -23,7 +23,7 @@ struct SimplicialComplexForAlpha {
   typedef unspecified Simplex_handle;
   /** Handle to specify a vertex. Must be a non-negative integer. */
   typedef unspecified Vertex_handle;
-  /** Handle to specify the simplex filtration value. */
+  /** Handle to specify the simplex filtration value. Must be capable to represent the special value "Not-A-Number"*/
   typedef unspecified Filtration_value;
 
   /** Returns the number of vertices in the simplicial complex. */
--- a/src/Alpha_complex/concept/SimplicialComplexForAlpha.h
+++ b/src/Alpha_complex/concept/SimplicialComplexForAlpha.h
@@ -23,7 +23,7 @@ struct SimplicialComplexForAlpha {
   typedef unspecified Simplex_handle;
   /** Handle to specify a vertex. Must be a non-negative integer. */
   typedef unspecified Vertex_handle;
-  /** Handle to specify the simplex filtration value. Must be capable to represent the special value "Not-A-Number"*/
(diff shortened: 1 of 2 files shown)


src/Alpha_complex/concept/SimplicialComplexForAlpha.h
```diff
@@ -23,7 +23,7 @@ struct SimplicialComplexForAlpha {
   typedef unspecified Simplex_handle;
   /** Handle to specify a vertex. Must be a non-negative integer. */
   typedef unspecified Vertex_handle;
-  /** Handle to specify the simplex filtration value. */
+  /** Handle to specify the simplex filtration value. Must be capable to represent the special value "Not-A-Number"*/
```
mglisse, line 26: Unrelated, but we should drop "handle" from this line, it looks like a bad copy-paste from the 2 previous lines.

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -491,10 +496,10 @@ class Alpha_complex {
         std::clog << vertex << " ";
       }
       std::clog << "is a face of Sigma\n";
-      std::clog << " | isnan(complex.filtration(Tau)=" << std::isnan(complex.filtration(f_boundary)) << std::endl;
+      std::clog << " | isnan(complex.filtration(Tau)=" << isnan(complex.filtration(f_boundary)) << std::endl;
```
mglisse, line 499: Wasn't your `using std::isnan;` in another function?

mglisse: > Is it ok to put it at `Gudhi::alpha_complex` namespace ?

Yes, I don't think it is likely to cause much trouble there.

> Or I repeat it in the 2 methods that are using it, as it cannot be done at class level.

Or that, as you prefer.

VincentRouvreau: Ok, let's go with repeating it in the 2 methods ([f772fb1](https://github.com/GUDHI/gudhi-devel/pull/798/commits/f772fb15dfe6f4fdaa99f2470b6f960a81aed84b)), it avoids embedding it when doing `using namespace Gudhi::alpha_complex`

mglisse (approved):

PR #799, opened 2023-01-20
Change clamping doc

MathieuCarriere:
The purpose of this PR is to separate the documentation of Clamping from the rest of gudhi.representations.preprocessing.


MathieuCarriere:
The reason Clamping appears between classes that are meant for persistence diagrams is because it is part of preprocessing and the classes appear in alphabetical order in the doc. So I don't know how to change this except if I create a new module (like gudhi.representatiosns.utils) containing only Clamping, but this will likely break old code.

mglisse:
> The reason Clamping appears between classes that are meant for persistence diagrams is because it is part of gudhi.representations.preprocessing and the classes in gudhi.representations.preprocessing appear in alphabetical order in the doc. So I don't know how to change this

See what I did for perslay?
I think you can also use `:exclude-members:` to exclude specific classes from the automodule (and then explicitly generate the doc for this class with autoclass either at the beginning or the end).

> except if I create a new module (like gudhi.representations.utils) containing only Clamping, but this will likely break old code.

I am not sure what happens if you create a new module and import Clamping in preprocessing, maybe automodule still generates its doc... Specifying `__all__` can be a way to specify the order in which the documentation appears.

(there is something strange with your use of git, which results in the gazillion merge commits shown in this PR)

MathieuCarriere:
Great thanks! I used `exclude-members`, it seems to be working.

Concerning the commits, indeed I think I messed up with my fork of gudhi master branch once, and now it records every time I merge, which is definitely annoying.

mglisse:
@astamm , you can see the doc at https://output.circle-artifacts.com/output/job/4dd14aff-a529-4121-b193-aff52e46eb0d/artifacts/0/sphinx/representations.html (please ignore the broken style, and ignore perslay which another PR is moving to a different page). Does Clamping become clear enough this way, or do you have other suggestions to improve it?

astamm:
That's much clearer! Maybe we can add to the description that it is meant to complement the four scalers that scikit-learn already provides?

mglisse:
(I wonder if it is worth suggesting to sklearn that they add this functionality. Historically, our suggestions have had a fairly low degree of success there...)

MathieuCarriere:
I've just added references to sklearn

src/python/gudhi/representations/preprocessing.py
```diff
@@ -12,6 +12,49 @@
...
 from sklearn.preprocessing import StandardScaler
 
+
+#############################################
+# Utils #####################################
+#############################################
+
+class Clamping(BaseEstimator, TransformerMixin):
+    """
+    This is a class for clamping a list of values. It is not meant to be called directly on (a list of) persistence diagrams, but it is rather meant to be used as a parameter for the DiagramScaler class. As such it has the same methods and purpose than common scalers from sklearn.preprocessing such as MinMaxScaler, RobustScaler, StandardScaler, etc. A typical use would be for instance if you want to clamp abscissae or ordinates (or both) of persistence diagrams within a pre-defined interval.
```
mglisse, line 22:
```suggestion
    This is a class for clamping a list of values. It is not meant to be called directly on (a list of) persistence diagrams, but it is rather meant to be used as a parameter for the DiagramScaler class. As such it has the same methods and purpose as common scalers from sklearn.preprocessing such as MinMaxScaler, RobustScaler, StandardScaler, etc. A typical use would be for instance if you want to clamp abscissae or ordinates (or both) of persistence diagrams within a pre-defined interval.
```

astamm:
Seems all good to me, thanks a lot!

PR #805, opened 2023-01-29
Always specify python executable for sphinx

### LINKED ISSUE #435: sphinx not using the requested python version
I specified a non-default python version to CMake with `-DPython_ADDITIONAL_VERSIONS=3.8`. The python plugin gets built for this version. But then we call `sphinx-build` which uses the default version of python, not the one I specified, so it fails. In GUDHI_third_party_libraries.cmake, it looks like we have an alternate path when sphinx-build is not found (`pythonX.Y -m sphinx.cmd.build`) which works better. It says that it only works with python 3.X, but I think we can forget about python 2.X now...

mglisse:
Fix #435.
(my system was recently updated to Python 3.11, but there is no tensorflow for that, so I specify 3.10 for gudhi)


VincentRouvreau (approved):
I was not able to reproduce the error, but as the render is ok using conda and on CI, I approve and merge this PR to unblock your developments

PR #807, opened 2023-02-01
[representations] __call__ and fit changes

### LINKED ISSUE #788: Wrong comment in Padding.fit
https://github.com/GUDHI/gudhi-devel/blob/7b9c272f7dad66ad9b05dcdd5ec43e86fda306c4/src/python/gudhi/representations/preprocessing.py#L185

but it actually does `self.max_pts = max([len(diag) for diag in X])`, and `transform` then uses `self.max_pts`, so the comment saying that it does nothing seems wrong.

### LINKED ISSUE #794: Atol `__call__` method needs explicit `.fit()` call first
Usually you can use the `__call__()` method of a class without calling `.fit()` yourself. But it seems no to work in `Atol`.

mglisse:
Based on our discussions.
For Atol, fit_transform on a single diagram (and immediately forgetting the centers!) makes too little sense to me to allow it. Fix #794. Fix #788.

Doc: https://output.circle-artifacts.com/output/job/8bb9280a-58b4-4c46-a360-b753e743f3b0/artifacts/0/sphinx/representations.html


MathieuCarriere (approved):

MathieuCarriere:
great changes, thanks!

MathieuCarriere:
about Atol, are you sure it does not make sense! I agree this is a bit weird, but running K-means on a single diagram and using the centroids to produce a vector is still well-defined no?

mglisse:
> About Atol, are you sure it does not make sense? I agree this is a bit weird, but running K-means on a single diagram and using the centroids to produce a vector is still well-defined no?

But what are you going to do with this vector, and what does it mean? If at least we kept the centers, I could imagine those being useful, but the case I am talking about is one where the centers are immediately removed. So `np.random.rand` is just as good a way to get a vector... Or am I missing something?

MathieuCarriere:
I see what you mean. Indeed, if you completely forget about the centers, there is no way you can get intuition about the vector values, even if they were generated in a structured way. Anyway, I'm fine with the changes, thanks again!

PR #814, opened 2023-02-13
Correct bracket for exponential

wreise:
This PR aims to fix two issues noticed after #755 :
- Incorrect [bracket](https://github.com/wreise/gudhi-devel/blob/20116fbf0d9608a766c056cd689be7cece6705a2/src/python/test/test_representations.py#L96) in exponential makes tests pass almost always. 
- Many lines are marked as tested only because they are covered by `src/python/example/diagram_vectorizations_distances_kernels.py`.

### DIFF
--- a/src/python/test/test_representations.py
+++ b/src/python/test/test_representations.py
@@ -93,7 +93,7 @@ def test_kernel_from_distance():
         kernelClass, kernelParams, tolerance = kernel_dict[kernelName]
         f1 = kernelClass.fit_transform(l1)
         d1 = pairwise_persistence_diagram_distances(l1, metric=kernelName, **kernelParams)
-        assert np.exp(-d1/kernelClass.bandwidth == pytest.approx(f1, **tolerance))
+        assert np.exp(-d1/kernelClass.bandwidth) == pytest.approx(f1, **tolerance)
 
 def test_kernel_distance_consistency():
     l1, l2 = _n_diags(9), _n_diags(11)


src/python/test/test_representations.py
```diff
@@ -87,13 +91,15 @@ def test_distance_transform_consistency():
                                       dict(bandwidth=4.),
                                       dict(rel=1e-3)),
 }
+
+kernel_approx = RBFSampler(gamma=1./2, n_components=100000)
```
mglisse, line 95: `.5` is simpler than trying to get a division to do the right thing.
Why define it here instead of where you use it?

src/python/test/test_representations.py
```diff
@@ -102,6 +108,10 @@ def test_kernel_distance_consistency():
         f2 = kernelClass.transform(l2)
         f12 = np.array([[kernelClass(l1_, l2_) for l1_ in l1] for l2_ in l2])
         assert f12 == pytest.approx(f2, **tolerance)
+        kernelClass.kernel_approx_ = kernel_approx
```
mglisse, line 111: I don't see any `kernel_approx_` in the code, what is this supposed to achieve?

src/python/test/test_representations.py
wreise: It mindlessly copied the code from `src/python/example/diagram_vectorizations_distances_kernels.py`... Indeed, 0.5 is better

src/python/test/test_representations.py
wreise: It was supposed to be `kernel_approx`!

wreise:
After fixing @mglisse 's comments, now the tests crash (at least locally): the approximation seems to be not that good. I will have a closer look.

wreise:
The approximation with ``RBFSampler`` is consistent if
``PersistenceWeightedGaussianKernel(bandwidth, weight_fct)(dgm1, dgm2)`` is equal to the expectation of ``PersistenceWeightedGaussianKernel(bandwidth, weight_fct, kernel_approx=RBFSampler(gamma=gamma))(dgm1, dgm2).`` It is actually the case only when ``gamma = 0.5*1./(bandwidth**2)``. 
Would it be useful to mention that somewhere (or to even restrict the API to force consistency)?

MathieuCarriere:
Sorry for taking so long to answer. I am not so sure about restricting the API, given that `RBFSampler` is also used to approximate other kernels (I think). But it would definitely be good to add a comment about this in the doc of `PersistenceWeightedGaussianKernel`!

PR #817, opened 2023-02-15
Multiparameter simplextree

DavidLapous:
This is a basic implementation of multiparameter simplex trees (for python), via a new (python) class : SimplexTreeMulti.
This code is a cleaned version of my [multipers](https://github.com/DavidLapous/multipers/blob/main/How%20to%20use.ipynb) repository.
It is basically the same simplextree as the standard python one, but with `Filtration_value` being a vector instead of a real number.
Key points:
 - All (finitely)-critical multi-filtration, with any (finite) number of parameter can be encoded by this class.
 - For non-finitely critical multi-filtration (e.g. Multicover bifiltration, degree rips), we will need another structure.
 - Edge collapses are implemented for 2-parameter 1-critical filtrations, using [filtration_domination](https://github.com/aj-alonso/filtration_domination).
 - Filtration values can be projected to a grid (for, e.g. more edge collases) on 1-dimensional lines (to compute the standard persistence) 
 - A SimplexTreeMulti can be exported as a multi-filtration in a scc file, that can be imported afterward using, e.g., Rivet or mpfree. 
 - This class alone is not really useful (yet), as there is no "persistence related function" in this PR, apart from the Euler characteristic.

The attached  [jupyter notebook](https://github.com/GUDHI/gudhi-devel/files/10744065/test_multi.1.pdf) showcases the main functions (sorry for the format, github doesn't allow .ipynb files).


src/python/include/utils/utilities.h
```diff
@@ -0,0 +1,262 @@
...
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#ifndef UTILITIES_H
+#define UTILITIES_H
+
+#include <vector>
+#include <limits>
+#include <omp.h>
```
mglisse, line 16: For now we have been using TBB for parallelism in Gudhi. Since we now use C++17, there are also some standard parallel algorithms. Mixing many different parallelism frameworks may cause problems at some point. OpenMP is badly supported by visual studio, so it probably shouldn't be our first choice.

src/python/include/utils/utilities.h
```diff
@@ -0,0 +1,262 @@
...
+    }
+};
+} //namespace utils
+
+
+
+
+namespace std{
+    template<typename T>
+    void sort(vector<T> &to_sort, bool(*compare)(const T&, const T&)){
```
mglisse, line 245: Not legal.

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+		return dereference(self.get_ptr()) == dereference(other.get_ptr())
+	
+	def euler_char(self, points:np.ndarray) -> np.ndarray:
+		""" Computes the Euler Characteristic of the filtered complex at given (multiparameter) time
+
+		Parameters
+		----------
+		points: 2-dimensional array.
+			List of filtration values on which to compute the euler characteristic.
+			WARNING FIXME : the points have to have the same dimension as the simplextree.
```
mglisse, line 926: There are several notions of dimension for a simplex tree, I think you mean the one of the filtration values.

src/python/include/Simplex_tree_interface_multi.h
```diff
@@ -0,0 +1,324 @@
...
+  euler_chars_type euler_char(const std::vector<std::vector<double>> &point_list){ // TODO multi-critical 
+		const unsigned int npts = point_list.size();
+		if (npts == 0){
+			return {};
+		}
+		const unsigned int nparameters = point_list[0].size();
+		
+		euler_chars_type out(point_list.size(), 0.);
+
+		auto is_greater = [nparameters](const point_type &a, const point_type &b){ //french greater
```
mglisse, line 288: "French" in the sense is_greater_equal?

src/python/include/Simplex_tree_interface_multi.h
```diff
@@ -0,0 +1,324 @@
...
+			simplex = {*it, *(++it)};
+			auto f = Base::filtration(simplexHandle);
+			simplex_list.push_back({simplex, {f[0], f[1]}});
+		}
+	}
+	simplex_list.shrink_to_fit();
+	return simplex_list;
+  }
+
+  euler_chars_type euler_char(const std::vector<std::vector<double>> &point_list){ // TODO multi-critical
```
mglisse, line 279: IIRC it can be computed more efficiently when the points are on a grid, which seems like a common use case, to be discussed with Vadim & Olympio.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1754,6 +1754,17 @@ class Simplex_tree {
...
+//MULTIPERS STUFF
+public: 
+  void set_number_of_parameters(unsigned int num){
+		number_of_parameters_ = num;
+	}
+	unsigned int get_number_of_parameters() const{
+		return number_of_parameters_;
+	}
+private:
+	unsigned int number_of_parameters_;
```
mglisse, line 1767: One extra integer is acceptable, but I wonder if we could make it exist or not depending on the template parameter?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
```diff
@@ -31,11 +31,14 @@ struct Simplex_tree_node_explicit_storage : SimplexTree::Filtration_simplex_base
   typedef typename SimplexTree::Filtration_value Filtration_value;
   typedef typename SimplexTree::Simplex_key Simplex_key;
 
-  Simplex_tree_node_explicit_storage(Siblings * sib = nullptr,
-                                     Filtration_value filtration = 0)
+  Simplex_tree_node_explicit_storage(Siblings * sib,
+                                     Filtration_value filtration) //MULTIPERS : init to 0 not possible
```
mglisse, line 35: Would `Filtration_value filtration = {}` work?

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+
+cimport numpy as cnp
+import numpy as np
+cnp.import_array()
+
+cimport gudhi.simplex_tree_multi
+cimport cython
+from gudhi import SimplexTree ## Small hack for typing
+
+from filtration_domination import remove_strongly_filtration_dominated, remove_filtration_dominated
```
mglisse, line 31: This dependency needs to be documented. Does the whole file need to depend on it, or could we make it so only one function depends on it?

mglisse (commented):
This is not small. Could you give more overview, to help review this? List the files modified explaining what happened to each. If a file is mostly a copy-paste of another file, say it and explain what differs and why we cannot easily share the code.
Random minor comments while trying to find where to start reading.

src/python/include/Simplex_tree_multi.h
```diff
@@ -0,0 +1,183 @@
...
+
+
+
+namespace Gudhi {
+/** Model of SimplexTreeOptions.
+ *
+ * Maximum number of simplices to compute persistence is <CODE>std::numeric_limits<std::uint32_t>::max()</CODE>
+ * (about 4 billions of simplices). */
+
+using value_type = double;
```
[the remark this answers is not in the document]
DavidLapous: Yes, this could be set in the SimplexTree Options.

src/python/include/Simplex_tree_multi.h
```diff
@@ -0,0 +1,183 @@
...
+struct Simplex_tree_options_multidimensional_filtration {
+public:
+	typedef linear_indexing_tag Indexing_tag;
+	typedef int Vertex_handle;
+	typedef std::vector<value_type> Filtration_value;
+	typedef std::uint32_t Simplex_key;
+	static const bool store_key = true;
```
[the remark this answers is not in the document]
DavidLapous: It is used for simplextree conversions, e.g. in the `SimplexTreeMulti.to_scc` functions. I'm not sure it is helpful apart from that.

src/python/include/Simplex_tree_multi.h
```diff
@@ -0,0 +1,183 @@
...
+	static const bool store_filtration = true;
+	static const bool contiguous_vertices = false;
+};
+
+using option_multi = Simplex_tree_options_multidimensional_filtration;
+using option_std = Simplex_tree_options_full_featured;
+bool operator<(const std::vector<value_type>& v1, const std::vector<value_type>& v2)
```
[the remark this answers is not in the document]
DavidLapous: I'm not sure what illegal means here; but I can do a "multi_filtration" class which would make that legal.
That will also help for non-finitely-critical filtrations (which cannot be encoded in a vector).

src/python/include/Simplex_tree_multi.h
```diff
@@ -0,0 +1,183 @@
...
+{
+	bool isSame = true;
+	if (v1.size() != v2.size()) isSame = false;
```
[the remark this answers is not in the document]
DavidLapous: It may be useful when comparing a projection of a filtration onto a lower-dimensional multi-filtration, with another lower-dimensional filtration, without doing the projection, but that would be very niche, and wouldn't be implemented like that. We can remove this indeed.

src/python/include/Simplex_tree_multi.h
```diff
@@ -0,0 +1,183 @@
...
+	}
+	if (isSame) return false;
+	return true;
+}
+
+void multify(const uintptr_t splxptr, const uintptr_t newsplxptr, const unsigned int dimension){
+	Simplex_tree<option_std> &st = *(Gudhi::Simplex_tree<option_std>*)(splxptr);
+	Simplex_tree<option_multi> &st_multi = *(Gudhi::Simplex_tree<option_multi>*)(newsplxptr);;
+	if (dimension <= 0)
+		{std::cout << "Empty filtration\n"; return ;}
```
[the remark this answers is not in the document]
DavidLapous: That is an error. But we don't want to crash the C++ otherwise it will crash the python kernel. I will change that to a cerr.

src/python/gudhi/simplex_tree_multi.pyx
DavidLapous: Yes, the size of the points have to be of the same size as `self.num_parameters`

src/Simplex_tree/include/gudhi/Simplex_tree.h
DavidLapous: I didn't find anywhere else to put it. Its not possible on the SimplexTree options (static). I can put it in the python class, but that would imply passing this integer to c++ each time a c++ function is called.

src/python/include/Simplex_tree_multi.h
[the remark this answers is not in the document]
mglisse: Throwing an exception does not crash anything (but you need to add `except +` in the .pyx so the exception is converted properly). Or you can return something special and check that in python. Or you can check the dimension in python before calling the C++ function. Or...

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
DavidLapous: I'm not sure it will compile with the standard simplextree (`Filtration_value` is a double)

src/python/gudhi/simplex_tree_multi.pyx
DavidLapous: This dependence is only needed in the `collapse_edge` method. I can put this dependence in this method indeed

DavidLapous:
I also have a question : 
For multi-persistence it is often useful (e.g. to compute the eulerchar, the rank invariant) to project the filtration values on a grid. Thus I am sometime playing with multi-simplextrees with filtration values that are coordinates in this grid (that is what the `grid_squeeze` method does, when `coordinate_values = True`).

Would you think another class for "coordinate simplextrees" would be useful ? i.e. `Filtration_value = vector<unsigned int>`

I have specific functions for these kinds of simplextrees (with the coordinate stored in `double`s)

mglisse:
> I also have a question : For multi-persistence it is often useful (e.g. to compute the eulerchar, the rank invariant) to project the filtration values on a grid. Thus I am sometime playing with multi-simplextrees with filtration values that are coordinates in this grid (that is what the `grid_squeeze` method does, when `coordinate_values = True`).
> 
> Would you think another class for "coordinate simplextrees" would be useful ? i.e. `Filtration_value = vector<unsigned int>`
> 
> I have specific functions for these kinds of simplextrees (with the coordinate stored in `double`s)

I assume the question is mostly for Python, since in C++ using `vector<int>` should already work with your branch? We need to understand what the benefits of this integer version would be, so we can balance them against the drawback of yet another type. 53 bits (in a `double`) is enough to store rather large integers. Is the main advantage to make things cleaner, because some operations only make sense with integers and others only with floats, and this way we could make only the meaningful ones available?

src/python/include/Simplex_tree_multi.h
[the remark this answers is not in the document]
mglisse: If there is at least one use, that's ok then.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: One possibility would be `[[no_unique_address]] std::conditional_t<Options::something, int, SomeEmptyType> number_of_parameters_;` so that depending on the option, it would not take any space. Or have Simplex_tree derive from `Maybe_multi_base<SimplexTreeOptions::plouf>` which depending on plouf can be empty or contain an int (and provide 2 methods).
Again, that's not fundamental, I just wanted to mention this possibility.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
mglisse: `double d = {};` compiles just fine.

src/python/include/Simplex_tree_interface_multi.h
```diff
@@ -0,0 +1,324 @@
...
+			auto f = Base::filtration(simplexHandle);
+			simplex_list.push_back({simplex, {f[0], f[1]}});
+		}
+	}
+	simplex_list.shrink_to_fit();
+	return simplex_list;
+  }
+
+  euler_chars_type euler_char(const std::vector<std::vector<double>> &point_list){ // TODO multi-critical 
+		const unsigned int npts = point_list.size();
```
mglisse, line 280: (picking one place randomly)
Using an unsigned type to encode the fact that a number cannot be negative is [not super popular](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es106-dont-try-to-avoid-negative-values-by-using-unsigned) these days, except to match an existing interface. For `npts` it would be natural to use `std::size_t`, what `size()` returns. For `nparameters` which is known to be small, either `size_t` because that's what size returns, or `int` which is the default type for small integers. `unsigned int` tends to be only used when doing bit operations, both because its arithmetic is more error-prone than the corresponding signed type, and because it is often harder to optimize for compilers. Although if you like `unsigned int` we can live with it, we already have quite a few in Gudhi, I just wanted to make sure you were aware of this.

mglisse:
> For multi-persistence it is often useful (e.g. to compute the eulerchar, the rank invariant) to project the filtration values on a grid. Thus I am sometime playing with multi-simplextrees with filtration values that are coordinates in this grid (that is what the `grid_squeeze` method does, when `coordinate_values = True`).

I don't know about the rank invariant, but for the Euler characteristic surface, I don't think you want to actually store the complex with rounded filtration values, you compute the rounded values on the fly and forget them. One algorithm I imagine is that you have a table of 0s matching the grid, then for each simplex you find the corresponding cell in the array (round the filtration values up) and add ±1 to this cell depending on the dimension, and finally you compute a cumulative sum in each dimension on the array.

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+	void flatten_diag(const uintptr_t, const uintptr_t, const vector[value_type], int) nogil
+	void squeeze_filtration(uintptr_t, const vector[vector[value_type]]&, bool) nogil
+	vector[vector[value_type]] get_filtration_values(uintptr_t) nogil
+
+
+cdef bool callback(vector[int] simplex, void *blocker_func):
+	return (<object>blocker_func)(simplex)
+
+# SimplexTree python interface
+cdef class SimplexTreeMulti:
```
mglisse, line 50: Do we want to encode in the name the fact that it is restricted to 1-critical filtrations, or is it unnecessary?

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+		Parameters
+		----------
+		max_dimension:int
+			Max simplicial dimension of the complex. Unless specified, keeps the same dimension.
+		num:int
+			The number of collapses to do.
+		strong:bool
+			Whether to use strong collapses or collapses (slower, but may remove more edges)
+		full:bool
+			Collapses the maximum number of edges if true, i.e., will do at most 100 strong collapses and 100 non-strong collapses afterward.
```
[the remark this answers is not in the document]
DavidLapous: The `full` flag is meant to do all possible collapses, i.e., doing all possible strong collapses, and then all "usual" collapses (definitions [here](https://arxiv.org/pdf/2211.05574.pdf), definition 2.1). As multiparameter persistence algorithms tend to  be very slow, doing all possible collapses can be worth it.
Another reason is that calling this method calls the method `get_edge_list`, and if the user wants to do strong and weak collapses, `get_edge_list` will get called only once.

src/python/include/Simplex_tree_interface_multi.h
```diff
@@ -0,0 +1,324 @@
...
+	  throw std::runtime_error("simplex not found - cannot find boundaries");
+	// this specific case works because the range is just a pair of iterators - won't work if range was a vector
+	auto boundary_srange = Base::boundary_simplex_range(bd_sh);
+	return std::make_pair(boundary_srange.begin(), boundary_srange.end());
+  }
+
+
+
+// ######################## MULTIPERS STUFF
+  void reset_keys(){
```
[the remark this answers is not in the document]
DavidLapous: This is related to a gudhi bug (if it isn't already fixed); the keys of the simplicies of a python simplextree are all 0, so I have to re-initialize them using this function. 
I didn't find a nice name, but `set_keys_to_enumerate` be OK ?
I will remove this functions once this "gudhi python bug" is fixed.

src/python/gudhi/simplex_tree_multi.pxd
```diff
@@ -0,0 +1,102 @@
...
+		dimension_type dimension() nogil
+		dimension_type upper_bound_dimension() nogil
+		bool find_simplex(vector[int] simplex) nogil
+		bool insert(vector[int] simplex, filtration_type filtration) nogil
+		vector[pair[simplex_type, filtration_type]] get_star(vector[int] simplex) nogil
+		vector[pair[simplex_type, filtration_type]] get_cofaces(vector[int] simplex, int dimension) nogil
+		void expansion(int max_dim) nogil except +
+		void remove_maximal_simplex(simplex_type simplex) nogil
+		bool prune_above_filtration(filtration_type filtration) nogil
+		# bool make_filtration_non_decreasing() nogil
```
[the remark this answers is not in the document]
DavidLapous: Yes, all function that are not related to either persistence, or relying on filtration comparison, usually work without any issue. I'm not sure the `prune_above_filtration` works properly. The `expansion` works for the simplex part, but it calls the `make_filtration_non_decreasing` afterwards, that is rewritten in the `simplex_tree_multi.pyx`.
The code for `make_filtration_non_decreasing` is in python for the moment, so it's super slow.

src/python/include/Simplex_tree_interface_multi.h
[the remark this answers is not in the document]
mglisse: That's not a bug. The name "key" is misleading, it is arbitrary data. The persistence algorithm does start by assigning an index as key, but then changes it during its execution.
Yes, `set_keys_to_enumerate` is much clearer, thanks.

src/python/include/Simplex_tree_interface_multi.h
DavidLapous: OK, I will change that to an int; with all of the other occurrences of unsigned int, as I'm not doing bit operations.

src/python/gudhi/simplex_tree_multi.pyx
DavidLapous: This class supports multi-critical filtrations. Knowing the number of parameter `n`, the multiple "birthtimes" of a simplex filtration `f`, can be recovered using, e.g., `np.array_split(f, len(f) // n)`.  c.f. the attached notebook on the initial message.
It is true, though, that a lot of functions do not support multi-critical filtrations yet; they should all have a comment saying that.

src/python/include/Simplex_tree_interface_multi.h
```diff
@@ -0,0 +1,324 @@
...
+	for (auto &simplexhandle : Base::skeleton_simplex_range(dimension)){
+		if (Base::dimension(simplexhandle) == dimension){
+			std::vector<int> simplex;
+			simplex.reserve(dimension+1);
+			for (int vertex : Base::simplex_vertex_range(simplexhandle))
+				simplex.push_back(vertex);
+			simplex_list.push_back(simplex);
+		}
+	}
+	simplex_list.shrink_to_fit();
```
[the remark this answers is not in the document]
DavidLapous: I supposed that the copy to python was maybe smaller this way. If it's not necessary, I can remove that.

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+
+		:param simplex: The N-simplex, represented by a list of vertex.
+		:type simplex: list of int
+		:returns:  The simplicial complex filtration value.
+		:rtype:  float
+		"""
+		# filtration_vector = self.get_ptr().simplex_filtration(simplex)
+		# filtrations = np.array_split(filtration_vector, len(filtration_vector) // self.num_parameters)
+		# return filtrations.squeeze() # If 1 critical returns only 1 filtration
+		return self.get_ptr().simplex_filtration(simplex) # Same output type, better
```
mglisse, line 145: I like `reshape((-1, num_parameters))`, it makes it clear that there are multiple critical points.
`squeeze` has the problem that you end up having to test the size of the output for every call. If the whole filtration is 1-critical, it may be convenient to avoid the extra dimension, but I don't think it is worth the resulting inconsistency.

src/python/gudhi/simplex_tree_multi.pyx
mglisse: Ah, ok, I misinterpreted those comments.

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+# SimplexTree python interface
+cdef class SimplexTreeMulti:
+	"""The simplex tree is an efficient and flexible data structure for
+	representing general (filtered) simplicial complexes. The data structure
+	is described in Jean-Daniel Boissonnat and Clément Maria. The Simplex
+	Tree: An Efficient Data Structure for General Simplicial Complexes.
+	Algorithmica, pages 1–22, 2014.
+
+	This class is a filtered, with keys, and non contiguous vertices version
+	of the simplex tree.
```
mglisse, line 58: I don't think this doc gets compiled by sphinx currently, and it does not always reflect the **multi**filtration.

src/python/gudhi/simplex_tree_multi.pyx
```diff
@@ -0,0 +1,970 @@
...
+		return self.copy()
+
+	def filtration(self, simplex)->filtration_type:
+		"""This function returns the filtration value for a given N-simplex in
+		this simplicial complex, or +infinity if it is not in the complex.
+
+		:param simplex: The N-simplex, represented by a list of vertex.
+		:type simplex: list of int
+		:returns:  The simplicial complex filtration value.
+		:rtype:  float
```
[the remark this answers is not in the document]
DavidLapous: Yes I have to update the doc :wink:

DavidLapous:
Thanks Marc for the review,
- I've made a class for finitely critical filtrations, I had to do some "jardinage" because I couldn't get cython & c++ to cast a class to it's inherited class (multi_filtrations are vectors of double)
- I removed the (mostly) unused filed in the utils folder and renamed it 
- and some cleanup to comply with your comments.

mglisse:
> * The default values is changed from `0` to a new static `Filtration_value`  in the simplextree structure : `inf_ = std::numeric_limits<Filtration_value>::infinity();`. This default value can be discussed (and changed if `Filtration_value` doesn't have an infinity), but 0 isn't natural for me, aswell as the default behavior of insert, which changes the values of the faces of the simplex inserted.

This kind of change should really be handled as a separate PR, otherwise it just makes everything harder to review.
Are those changes you need for the multi-parameter code, or just preferences? I haven't looked at the code, but we should be careful with changes that are likely to break users' code.

DavidLapous:
These changes should all be behind a ```if constexpr (Options::is_multiparameter)``` flag. I still need to fix this default value to use the `const Filtration_value&` to prevent copies; I can change this default value to `{}` and set the default value in python instead of c++ to have a more consistent behavior.

DavidLapous:
After re-reading the code in the PR, I noticed that my previous statements were false :
 - The default value is always `{}` and only the python interface has different values.
 - The static simplextree variable member `inf_` is only meant to deal with the `filtration` which has to return `inf` on the `null_simplex`.

DavidLapous:
I had to add a static variable `null_value` to the dummy filtration, as filtration values are passed via reference in this pr. 

I'm also not sure that the python interface will be able to run with `python<3.10` and `cython<3`, as I'm using some python 3.10 features (e.g. function typing) and I had to rewrite some stuff for cython3. I guess that's why the `python_without_cgal` build fails.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -970,8 +980,28 @@ class Simplex_tree {
     Simplex_handle simplex_one = insertion_result.first;
     bool one_is_new = insertion_result.second;
     if (!one_is_new) {
-      if (filtration(simplex_one) > filt) {
-        assign_filtration(simplex_one, filt);
+      if (!(filtration(simplex_one) <= filt)) { // TODO : For multipersistence, it's not clear what should be the default, especially for multicritical filtrations
+        if constexpr (SimplexTreeOptions::is_multi_parameter){
```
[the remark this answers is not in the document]
DavidLapous: Not really, as multiparameter filtrations have to be done carefully, I don't want the insertion to change the filtration values of the other simplices. We could define a behavior based on if the filtration value is comparable to the previous filtration value and create a multicritical filtration / change the values of the other simplices (that's `if`s), but I'm not sure that someone would want to deal with this complicated behavior; doing this by hand seems fine. I thus assume here that the user is smart.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1603,12 +1633,22 @@ class Simplex_tree {
       if (dim == 0) return;
       // Find the maximum filtration value in the border
       Boundary_simplex_range&& boundary = boundary_simplex_range(sh);
-      Boundary_simplex_iterator max_border = std::max_element(std::begin(boundary), std::end(boundary),
+      typename SimplexTreeOptions::Filtration_value max_filt_border_value;
+      if constexpr (SimplexTreeOptions::is_multi_parameter){
+        // in that case, we assume that Filtration_value has a `push_to` member to handle this.
```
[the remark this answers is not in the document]
DavidLapous: It's a possibility, but I'm only using python, and I will not maintain / deal with a c++ interface that I don't use.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2290,6 +2342,7 @@ struct Simplex_tree_options_full_featured {
   static const bool contiguous_vertices = false;
   static const bool link_nodes_by_label = false;
   static const bool stable_simplex_handles = false;
+  static const bool is_multi_parameter = false;
```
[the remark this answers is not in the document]
DavidLapous: either `has_multiple_parameters` or `is_multi_parameter`. I prefer the second one.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: It is not about making a C++ interface. Just move the file, include it in `Simplex_tree.h` and add something like:
```C++
struct Simplex_tree_options_multi_parameter {
  typedef linear_indexing_tag Indexing_tag;
  typedef int Vertex_handle;
  typedef Finitely_critical_multi_filtration<float> Filtration_value;
  typedef std::uint32_t Simplex_key;
  static const bool store_filtration = true;
  static const bool contiguous_vertices = false;
  static const bool link_nodes_by_label = false;
  static const bool stable_simplex_handles = false;
  static const bool is_multi_parameter = true;
};
```
at the end of the file with the other options instead of having it in `/python/include/Simplex_tree_multi.h`. That is the only change. Then for the documentation, in the concept of the options (concept/SimplexTreeOptions.h) you mention the need of `push_to` and that you can find an example of implementation as ` Finitely_critical_multi_filtration<T>`.

The difference is that it will make the tree usable for C++ users without you having to change much or to maintain anything.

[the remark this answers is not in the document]
hschreiber: @mglisse Not really, it was just to illustrate how we would use it. If Filtration_value needs a special method like `push_to`, we cannot just let it hang there without example for the user, in particular if code was already written. We can also put this just in the tests or in an example, but at least in the concept, the implementation of `Finitely_critical_multi_filtration<T>` should be mentioned.

PR #828, opened 2023-03-09
Misc cubical simplifications

mglisse:
It may be easier to look at each commit instead of the global diff.


src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex.h
```diff
@@ -471,7 +352,7 @@ class Bitmap_cubical_complex : public T {
    **/
```
VincentRouvreau, line 352: Is it possible to take the opportunity to rewrite the documentation of this range ?
`Class needed for compatibility with Gudhi. Not useful for other purposes.` is not really attractive.

mglisse: I already removed one of those in a previous PR. Let's not delay those incremental fixes by asking for more, I'll handle it in a later PR (there are many possible clean-ups left...).

VincentRouvreau (approved):

PR #829, opened 2023-03-09
Don't crash on bogus assign_filtration

### LINKED ISSUE #756: Crash on assign_filtration
For a given SimplexTree() if I assign a filtration value to an edge that is not in the one skeleton, the program crashes :
```
import gudhi as gd
st = gd.SimplexTree()
st.assign_filtration([0, 1], 3.14)
```
Instead should we:
- raise an exception?
- insert the edge `st.insert([0, 1])` and assign anyway?

mglisse:
Fix #756.

### DIFF
--- a/src/python/include/Simplex_tree_interface.h
+++ b/src/python/include/Simplex_tree_interface.h
@@ -53,7 +53,10 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
   }
 
   void assign_simplex_filtration(const Simplex& simplex, Filtration_value filtration) {
-    Base::assign_filtration(Base::find(simplex), filtration);
+    Simplex_handle sh = Base::find(simplex);
+    if (sh == Base::null_simplex())
+      throw std::invalid_argument("Cannot assign a filtration to a simplex that is not in the complex");
+    Base::assign_filtration(sh, filtration);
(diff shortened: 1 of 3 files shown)


src/python/include/Simplex_tree_interface.h
```diff
@@ -53,7 +53,10 @@ class Simplex_tree_interface : public Simplex_tree<SimplexTreeOptions> {
   }
 
   void assign_simplex_filtration(const Simplex& simplex, Filtration_value filtration) {
-    Base::assign_filtration(Base::find(simplex), filtration);
+    Simplex_handle sh = Base::find(simplex);
+    if (sh == Base::null_simplex())
+      throw std::invalid_argument("Cannot assign a filtration to a simplex that is not in the complex");
```
VincentRouvreau, line 58: Should we explain that the python method will throw an exception in simplex is not found ? `ValueError` in python

VincentRouvreau: "Obvious" exceptions seem to be not documented under python cf.  [numpy.genfromtxt](https://numpy.org/devdocs/reference/generated/numpy.genfromtxt.html#numpy-genfromtxt), `np.genfromtxt("unexisting_file.txt")` throws a `FileNotFoundError`

PR #832, opened 2023-03-10
Avoid crash in Alpha_complex::get_point

mglisse:


VincentRouvreau (commented):
Should we also document/test the python part ? With your fix, `get_point` will throw an `IndexError`

mglisse:
> With your fix, `get_point` will throw an `IndexError`

It already does for an argument that is negative or too large.
I am not convinced it is necessary to document that an erroneous input results in an error, but if you think it is important, I'll add a sentence.

VincentRouvreau:
> > With your fix, `get_point` will throw an `IndexError`
> 
> It already does for an argument that is negative or too large. I am not convinced it is necessary to document that an erroneous input results in an error, but if you think it is important, I'll add a sentence.

We document the c++ exception, so yes I would document the python exception.

VincentRouvreau:
> We document the c++ exception, so yes I would document the python exception.

"Obvious" exceptions seem to be not documented under python cf.  [numpy.genfromtxt](https://numpy.org/devdocs/reference/generated/numpy.genfromtxt.html#numpy-genfromtxt), `np.genfromtxt("unexisting_file.txt")` throws a `FileNotFoundError`

VincentRouvreau (approved):

PR #835, opened 2023-03-13
Modify python documentation third parties

VincentRouvreau:
Add pydata-sphinx-theme as a requirement and that version 0.8.0 is the minimal supported version


mglisse (approved):
Probably some CMake changes are also needed?

mglisse (approved):
Maybe after we change the build system we can see if there is an easy way to specify a minimal version for a package.

PR #836, opened 2023-03-17
Changes for onetbb use instead of deprecated TBB

### LINKED ISSUE #511: Check support for oneTBB
Our code was written for use with TBB, we need to check if it also works with the more recent oneTBB. I think Tangential_complex/benchmark/benchmark_tc.cpp may be problematic. I also reported to Hera their use of parallel_do in https://github.com/anigmetov/hera/issues/7. FindTBB.cmake may be outdated (maybe it should just be removed?). https://software.intel.com/content/www/us/en/develop/documentation/onetbb-documentation/top/onetbb-developer-guide/migrating-from-threading-building-blocks-tbb.html has relevant information.
At some point we could investigate using the standard C++17 API to replace most of our uses of TBB, but currently gcc/clang implement that on top of TBB, so it wouldn't remove the TBB dependency.

VincentRouvreau:
* Use oneTBB cmake target
* As tbb::task_scheduler_init has been deprecated, it has been removed from Tangential complex benchmark. It is best to leave the decision of how many threads to use to the task scheduler

Tested with oneTBB conda release package. Not tested in debug mode or other modes.

Fix #511


src/Tangential_complex/benchmark/benchmark_tc.cpp
```diff
@@ -578,16 +579,16 @@ int main() {
...
 #endif
 #endif
       /*for (Concurrent_mesher_config::get().num_work_items_per_batch = 5 ;
         Concurrent_mesher_config::get().num_work_items_per_batch < 100 ;
         Concurrent_mesher_config::get().num_work_items_per_batch += 5)*/ {
 #ifdef GUDHI_USE_TBB
-      tbb::task_scheduler_init init(
-                                    num_threads > 0 ? num_threads : tbb::task_scheduler_init::automatic);
+      if (num_threads > 0)
+        tbb::global_control global_limit(tbb::global_control::max_allowed_parallelism, num_threads);
```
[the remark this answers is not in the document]
VincentRouvreau: Oh yes you are right ! Now I understand why Clément did it this way. I tried to do so on [72789aa](https://github.com/GUDHI/gudhi-devel/pull/836/commits/72789aad4512d2e54d45f3322b67dce1fdd1ceb8)

src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -52,12 +52,36 @@ option(WITH_GUDHI_USE_TBB "Build with Intel TBB parallelization" ON)
...
 if(WITH_GUDHI_USE_TBB)
-  set(TBB_FIND_QUIETLY ON)
-  find_package(TBB)
-  if (TBB_FOUND)
-    include(${TBB_USE_FILE})
-    message("TBB found in ${TBB_LIBRARY_DIRS}")
-    add_definitions(-DGUDHI_USE_TBB)
+  find_package(TBB QUIET)
+  # cf. https://github.com/oneapi-src/oneTBB/blob/master/cmake/README.md
+  # if (TARGET TBB::tbb) # does not work...
```
mglisse, line 57: Can you describe what was going wrong? The error message seemed to be about get_target_property(TBB::tbbmalloc), but if it was the `if` test that returned the wrong answer, I would have expected an error on the previous line. I am confused how changing just the `if` can have fixed things.

mglisse (commented):
Now ideally we would find a dataset on which to run the benchmark, to see if the performance still increases with the number of threads...

src/cmake/modules/GUDHI_third_party_libraries.cmake
VincentRouvreau: As it is confusing, I tried to explain better what I meant on [55fef38](https://github.com/GUDHI/gudhi-devel/pull/836/commits/55fef38c2fb757edf02ec113fd35b312e4fa03e6)

mglisse: But in what way did it not work? Did it answer yes when TBB was not found, or no when TBB was actually found? Was it bad on all platforms?

mglisse: IIUC, looking at the "pip build linux" CI, the version with if(TBB_FOUND) succeeds because it decides that it has **not** found TBB.

VincentRouvreau: In the case `pip build linux`, TBB is not installed (cf. [Dockerfile](https://github.com/GUDHI/gudhi-deploy/blob/main/Dockerfile_for_pip) - I also checked in `/usr/include` and `/usr/local/include` after a fresh docker pull).
Right after `find_package(TBB)`, `if(TARGET TBB::tbb)` returns true, while `if (TBB_FOUND)` returns false. On this image, cmake is in version 3.12 (July 2018). This may be the root cause. Cmake minimal version is still 3.8 for GUDHI. I would keep the `if (TBB_FOUND)` version. Do you agree ?

VincentRouvreau:
> By the way, what version of TBB do we have on the CI platforms? A mix of old and oneTBB?

There is no TBB for pip packages (maybe we can think about activating it if it worth it).
TBB versions are:
* CircleCI (linux build) - 2021.5.0
* Azure CI Windows - 2021.8.0
* Azure CI OSx - 2021.8.0-1

We have no old TBB installation tested. I can test locally with an old TBB version, to see if it works (only tangential benchmark should fail), but I think we can skip it.

src/cmake/modules/GUDHI_third_party_libraries.cmake
mglisse: Ah, `find_package(TBB)` ends up using the FindTBB.cmake from CGAL, and that file seems to have a strange behavior. `find_package(TBB CONFIG)` would hopefully be immune to that. I don't remember if we ever use the TBB functionalities in CGAL, and if this might cause trouble. I am not against sticking to TBB_FOUND, but I find it worth understanding.

src/python/doc/installation.rst
```diff
@@ -265,12 +267,11 @@ The following examples require `Eigen <http://eigen.tuxfamily.org/>`_ version :m
 Threading Building Blocks
 -------------------------
 
-`Intel® TBB <https://www.threadingbuildingblocks.org/>`_ lets you easily write
-parallel C++ programs that take full advantage of multicore performance, that
-are portable and composable, and that have future-proof scalability.
+`Intel® TBB <https://www.threadingbuildingblocks.org/>`_ lets you easily write parallel C++ programs that take full
```
mglisse, line 270: This URL seems to redirect elsewhere now, we might as well link directly there.

src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -52,14 +52,43 @@ option(WITH_GUDHI_USE_TBB "Build with Intel TBB parallelization" ON)
...
-  set(TBB_FIND_QUIETLY ON)
-  find_package(TBB)
-  if (TBB_FOUND)
-    include(${TBB_USE_FILE})
-    message("TBB found in ${TBB_LIBRARY_DIRS}")
-    add_definitions(-DGUDHI_USE_TBB)
+  find_package(TBB CONFIG)
+  if(TARGET TBB::tbb)
+    # Specific windows case with its debug/release management
+    if(CMAKE_BUILD_TYPE MATCHES Debug)
```
mglisse, line 58: It is considered a Bad Thing™ to test CMAKE_BUILD_TYPE, some cmake generators (like visual studio) ignore it and produce build instructions for both release and debug. The advice is to try and use generator expressions like `$<$<CONFIG:Debug>:DEBUG_MODE>` where possible.
I think we can ignore this issue for now, but it is something to keep in mind.

mglisse (approved):
I'd still like to see a run of the benchmark showing a dependency between the number of threads and the running time, to make sure that the benchmark is still useful...

PR #838, opened 2023-03-26
Use autodoc_class_signature to document __init__

mglisse:
I was trying to avoid the repetition of the arguments
```
class gudhi.representations.preprocessing.Clamping(minimum=-inf, maximum=inf)
__init__(minimum=-inf, maximum=inf)
```
and it came with side benefits.

We should still check that the doc looks ok with this change.

### DIFF
--- a/src/python/doc/conf.py
+++ b/src/python/doc/conf.py
@@ -19,6 +19,8 @@ extensions = [
     "sphinx_paramlinks",
 ]
 
+autodoc_class_signature = "separated"
+
 bibtex_bibfiles = ["../../biblio/bibliography.bib"]
 
 templates_path = ["_templates"]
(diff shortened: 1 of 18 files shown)


src/python/doc/conf.py
```diff
@@ -19,6 +19,8 @@
     "sphinx_paramlinks",
 ]
 
+autodoc_class_signature = "separated"
```
VincentRouvreau, line 22: I didn't know this feature, and I prefer it this way (arguments are described only in constructor) !

VincentRouvreau: Argh... it creates a `__new__` entry for `*Complex` (but not `WeightedRipsComplex`, nor `DTMRipsComplex`) and `SimplexTree` classes, don't know why. :thinking: 

image

VincentRouvreau: Ok, thanks, I did not know this cython behaviour

VincentRouvreau (approved):

PR #842, opened 2023-03-28
Qt vs TBB conflict about `emit`

### LINKED ISSUE #841: Qt defines emit as a macro and TBB as a function
Compiling GudhUI with TBB enabled results in
```
/usr/include/oneapi/tbb/profiling.h:229:15: error: expected unqualified-id before ‘)’ token
  229 |     void emit() { }
      |               ^
```
(for example https://github.com/oneapi-src/oneTBB/issues/547)
Apparently we could tell Qt not to define this macro, with QT_NO_EMIT, or maybe QT_NO_KEYWORDS or QT_NO_META_MACROS or... But that may require changing the code in GudhUI that uses those macros.

mglisse:
Fix #841.
From what I can read, `emit` exists purely for decoration, so a comment should work just as well.

### DIFF
--- a/src/GudhUI/CMakeLists.txt
+++ b/src/GudhUI/CMakeLists.txt
@@ -32,6 +32,7 @@ if (OPENGL_FOUND)
         target_link_libraries( GudhUI ${QGLVIEWER_LIBRARIES} )
         target_link_libraries( GudhUI ${OPENGL_gl_LIBRARY} ${OPENGL_glu_LIBRARY} )
         if(TARGET TBB::tbb)
+          add_compile_definitions( QT_NO_EMIT )
           target_link_libraries( GudhUI TBB::tbb)
         endif()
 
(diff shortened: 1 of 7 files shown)


VincentRouvreau:
> Fix #841. From what I can read, `emit` exists purely for decoration, so a comment should work just as well.

Do you think it helps if it is kept it as a comment ? (I have no strong opinion on it...)

mglisse:
> Do you think it helps if it is kept it as a comment ? (I have no strong opinion on it...)

It pretty much was a comment to begin with (Qt just does `#define emit`). Keeping it seems to preserve the intent of the author to document something. I am trying to change as little as possible since I don't know how this code works.

VincentRouvreau:
> > Do you think it helps if it is kept it as a comment ? (I have no strong opinion on it...)
> 
> It pretty much was a comment to begin with (Qt just does `#define emit`). Keeping it seems to preserve the intent of the author to document something. I am trying to change as little as possible since I don't know how this code works.

By the way, I tested it and it works fine

PR #843, opened 2023-03-29
generate long_description for the pip package from cmake

### LINKED ISSUE #808: introduction.rst not installed properly by pip
`introduction.rst` installation should be reworked considering [this documentation](https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/) as this file is sometimes installed in `~/.local `.
File copy could also be avoided and `long_description` could be filled by CMake reading `introduction.rst`.

VincentRouvreau:
Fix #808


mglisse (approved):
I assume the text still ends up where it was supposed to end up (was it the PyPI page?).

VincentRouvreau:
> I assume the text still ends up where it was supposed to end up (was it the PyPI page?).

Yes, that's it. I will test it when I will build the next release candidate

PR #847, opened 2023-03-30
New code for persistence of 2d cubical from top cells

mglisse:
For now only available through the Python CubicalPersistence, so we have more time to think about the exact C++ interface we want to commit to.
I hesitated between putting the code in Bitmap_cubical_complex/ or Persistent_cohomology/ .

Benchmark (without TBB):
> Construction from top cells: 0.189s
> initialize_filtration: 0.632s
> Compute persistent homology: 1.319s
> Total old code: 2.152s
>
> Total new code: 0.047s
>
> Total new code with index: 0.051s


src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,595 @@
...
+  // The square i + dy is right above i.
+  Index dy;
+
+  // Squares keep their index from the input.
+  // Vertices have the same index as the square at their bottom left (smaller x and y)
+  // Store the filtration value of vertices that could be critical. We could store them
+  // in some map, or recompute them on demand, but this strongly affects performance.
+  std::unique_ptr<T[]> data_v_;
+  T& data_vertex(Index i){ return data_v_[i]; }
+  T data_vertex(Index i) const { return data_v_[i]; }
```
hschreiber, line 83: As the access levels of `data_v_[i]` and `data_vertex(i)` are the same, are those getter function not useless? (see also below for ds_parents_*) Even if later, say, `ds_parent_v_` becomes private, would the method `ds_parent_vertex` really be useful as public method?

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,595 @@
...
+#include <cstddef>
+
+namespace Gudhi::cubical_complex {
+
+// When building a cubical complex from top-dimensional cells, there are
+// normally more vertices than input top cells ((x+1)*(y+1) instead of x*y).
+// However, the top cells along the boundary turn out to be collapsible, and we
+// only need to work with (x-1)*(y-1) vertices.
+template <class Filtration_value, class Index = std::size_t, bool output_index = false>
+struct Persistence_on_rectangle {
```
hschreiber, line 49: I guess, everything is public the time we decide on an interface?

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,595 @@
...
+  }
+  void set_parent_square(Index child, Index parent) {
+    GUDHI_CHECK(child != parent, std::logic_error("Bug in Gudhi: use mark_*_critical instead of set_parent"));
+    ds_parent_square(child) = parent;
+  }
+
+  // Locally pair simplices around each square.
+  // Work implicitly from input, only store the filtration value of critical vertices (squares are already in input).
+  // Store critical edges for later processing.
+  void fill_and_pair() {
```
hschreiber, line 211: Is it possible to decompose this method to have a better overview?

hschreiber:
> I hesitated between putting the code in Bitmap_cubical_complex/ or Persistent_cohomology/ .

Good question. As it does not use cohomology, I am not sure it would belong in Persistent_cohomology/. On the other hand, Cubical Complex is a data structure and not meant to compute persistence by it-self. But I would still keep it in the latter.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
mglisse: The class is really meant as an implementation detail, at least for now. I started with a single function, and the things that are now member functions were lambdas. But this was inconvenient to work with because tools don't handle lambdas so well, and the profile mostly said that most of the time was taken by operator() calling operator(), followed by operator()...
If I had to make something public right now, it would be the free function at the end, while the class would remain hidden.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,595 @@
...
+ * @tparam Index This is used to index the elements of `input`, so it must be large enough to represent the size of `input`.
+ * @param[in] input Pointer to `nrows*n_cols` filtration values for the square cells. Note that the values are assumed to be stored in C order, unlike `Gudhi::cubical_complex::Bitmap_cubical_complex` (you can exchange `n_rows` and `n_cols` for compatibility).
+ * @param[in] n_rows number of rows of `input`.
+ * @param[in] n_cols number of columns of `input`.
+ * @param[out] out0 For each interval (b, d) in the persistence diagram of dimension 0, the function calls `out0(b, d)`.
+ * @param[out] out1 Same as `out0` for persistence in dimension 1.
+ * @returns The global minimum, which is not paired and is thus the birth of an infinite persistence interval of dimension 0.
+ */
+template <bool output_index = false, typename Filtration_value, typename Index, typename Out0, typename Out1>
+auto persistence_on_rectangle_from_top_cells(Filtration_value const* input, Index n_rows, Index n_cols, Out0&&out0, Out1&&out1){
```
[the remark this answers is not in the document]
mglisse: Yes, exactly. Well, for the dual there is an implicit exterior cell to make it really match H1, but that's a detail.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
mglisse: The goal wasn't to have getters for a private member, the goal was to abstract away how this data is stored. At various points it was stored in a map, or accessed as storage[i/2].filt, or...
It is true that it ends up as a really trivial wrapper...

mglisse:
> As it does not use cohomology, I am not sure it would belong in Persistent_cohomology/. On the other hand, Cubical Complex is a data structure and not meant to compute persistence by it-self. But I would still keep it in the latter.

One thing that may look strange is that it ends up in a different place than the 1d code.
(well, as long as it is undocumented we can still move this C++ code around later)

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
mglisse: I am not sure how to do that in a useful way.
The function starts by setting up the vocabulary: `i` is the current square, its upper-left vertex is referred to as `v_ul`, etc. Then it iterates on all the squares. And the scary part, well-delimited in its for loop, is the case analysis for interior squares, which depends on which neighboring squares have a bigger / smaller filtration value.

I could split out the handling of a single interior square into its own function, although I would then have to duplicate the vocabulary lambdas, but I am not sure it would be much clearer.

Do you have some precise suggestion?

mglisse: By "overview", do you mean you would like the content of the for loops to be short enough that we can see them completely on one screen?
A comment
```
// Iterate on all the squares
// For each square in the first row: ...
// For each row:
//     First square
//     For each interior square: ...
//     Last square
// For each square in the last row: ...
```
could play the same role?

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
hschreiber: Ok, makes sense.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
hschreiber: If only the free function is public, there is no possibility to only compute H0 or only H1. Does computing both not make much of a difference than computing only one? I am wondering because in one of your comments you write that `primal` and `dual` can be dominant.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,595 @@
...
+// * try to loosen tight dependency chains, load values several instructions before they are needed. Performing 2 find_set in lock step surprisingly doesn't help.
+
+/**
+ * @private
+ * Compute the persistence diagram of a function on a 2d cubical complex, defined as a lower-star filtration of the values at the top-dimensional cells.
+ *
+ * @tparam output_index If false, each argument of the out functors is a filtration value. If true, it is instead the index of this filtration value in the input.
+ * @tparam Filtration_value Must be comparable with `operator<`.
+ * @tparam Index This is used to index the elements of `input`, so it must be large enough to represent the size of `input`.
+ * @param[in] input Pointer to `nrows*n_cols` filtration values for the square cells. Note that the values are assumed to be stored in C order, unlike `Gudhi::cubical_complex::Bitmap_cubical_complex` (you can exchange `n_rows` and `n_cols` for compatibility).
```
hschreiber, line 557: typo: nrows --> n_rows

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,595 @@
...
+ * @param[out] out1 Same as `out0` for persistence in dimension 1.
+ * @returns The global minimum, which is not paired and is thus the birth of an infinite persistence interval of dimension 0.
+ */
+template <bool output_index = false, typename Filtration_value, typename Index, typename Out0, typename Out1>
+auto persistence_on_rectangle_from_top_cells(Filtration_value const* input, Index n_rows, Index n_cols, Out0&&out0, Out1&&out1){
+#ifdef GUDHI_DETAILED_TIMES
+  Gudhi::Clock clock;
+#endif
+  GUDHI_CHECK(n_rows >= 2 && n_cols >= 2, std::domain_error("The complex must truly be 2d, i.e. at least 2 rows and 2 columns"));
+  Persistence_on_rectangle<Filtration_value, unsigned, output_index> X;
```
hschreiber, line 570: Why `unsigned` and not `Index`?

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
hschreiber: I see what you mean. It is just that there are so many imbricated `if`s and small letters everywhere, that I always get lost at some point. I am wondering if replacing `u`, `d`, `r`, `l` by their respective meanings would not make them easier to read (even though I agree that the meanings are clear).
Thankfully, the collapsing function of the editor is helping a lot already!

hschreiber:
> One thing that may look strange is that it ends up in a different place than the 1d code.
> (well, as long as it is undocumented we can still move this C++ code around later)

As `compute_persistence_of_function_on_line` is already in `Persistent_cohomology`, than we can also put this code there too. It does make sense after thinking about it longer, as persistent cohomology and persistent homology gives the same results anyway, so it is just a matter of name. The only thing is that being in a module named "cohomology" gives the feeling that the algorithms have to use cocyles. But in the end, there are not really using cycles explicitly either, so well...

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
hschreiber: No, the little functions are a very good idea! 
I meant `r` ---> `right`, `l`---> `left` etc.
If I wrote all those imbricated conditions, I wouldn't want to touch them again either. And if it is the fastest way, they should be left like that anyway.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
mglisse: Hmm, I think I had Index, documented Index, then temporarily used unsigned for a benchmark and forgot to put it back. Fixed.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
mglisse: I just pushed a patch that renames to longer names. What do you think?
(I am not convinced that it is more readable, but I am ok with it if you prefer this)

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,596 @@
...
+      ds_parent_square(b) = a;
+      if constexpr (output_index)
+        out(e.f.out(), b);
+      else
+        out(e.f.out(), input(b));
+    }
+  }
+};
+// Ideas for improvement:
+// * for large hard (many intervals) inputs, primal/dual dominate the running time because of the random reads in find_set and input(a/b). The input load would be cheaper if we stored it with parents, but then find_set would be slower.
```
VincentRouvreau, line 546: Lines should have less than 120 characters per line

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,596 @@
...
+// * to increase memory locality, maybe pairing, which is currently arbitrary, could use some heuristic to favor some pairs over others.
+// * try to loosen tight dependency chains, load values several instructions before they are needed. Performing 2 find_set in lock step surprisingly doesn't help.
+// * To handle very big instances, we could remove data_v_ and recompute it on demand as the min of the inputs of i, i+1, i+dy and i+dy+1. On hard instances, it wouldn't save that much memory (and it is a bit slower). On easy instances, if we also remove the calls to reserve(), the saving is less negligible, but we still have ds_parent_*_ that take about as much space as the input. We could, on a subarray, fill a dense ds_parent, then reduce it and export only the critical vertices and boundary to some sparse datastructure, but it doesn't seem worth the trouble for now.
+
+/**
+ * @private
+ * Compute the persistence diagram of a function on a 2d cubical complex, defined as a lower-star filtration of the values at the top-dimensional cells.
+ *
```
VincentRouvreau, line 554: Why not exposing the documentation just like for persistence on a line ?

```suggestion
 *
 * \ingroup persistent_cohomology
 *
```

VincentRouvreau: Sorry, I missed your point `For now only available through the Python CubicalPersistence, so we have more time to think about the exact C++ interface we want to commit to`. Forget this comment.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,596 @@
...
+#ifdef DEBUG_TRACES
+ #include <iostream>
+#endif
+#include <vector>
+#include <memory>
+#include <algorithm>
+#include <stdexcept>
+#include <cstddef>
+
+namespace Gudhi::cubical_complex {
```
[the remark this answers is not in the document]
VincentRouvreau: Sorry, I missed the discussion above

VincentRouvreau:
> > One thing that may look strange is that it ends up in a different place than the 1d code.
> > (well, as long as it is undocumented we can still move this C++ code around later)
> 
> As `compute_persistence_of_function_on_line` is already in `Persistent_cohomology`, than we can also put this code there too. It does make sense after thinking about it longer, as persistent cohomology and persistent homology gives the same results anyway, so it is just a matter of name. The only thing is that being in a module named "cohomology" gives the feeling that the algorithms have to use cocyles. But in the end, there are not really using cycles explicitly either, so well...

My first feeling was that I was not understanding why this persistence computation specific for 2d cubical was in `Bitmap_cubical_complex` and not in `Persistent_cohomology`, but I was missing your explanation about homology and cohomology. I would advise to put this code in `Persistent_cohomology`, but I don't have a strong opinion on the question.

src/python/gudhi/_pers_cub_lowdim.cc
```diff
@@ -28,7 +32,7 @@ PYBIND11_MAKE_OPAQUE(Vf);
 PYBIND11_MAKE_OPAQUE(Vd);
 
 template<class T>
-py::array fun(py::array_t<T> data) {
+py::array fun1(py::array_t<T> data) {
```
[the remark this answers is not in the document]
mglisse: 1/2 represent the dimension, and we already call compute_persistence_of_function_on_line and persistence_on_rectangle_from_top_cells in them so grep already finds them. If I rename fun1 to persistence_on_a_line, then it becomes easy to confuse with compute_persistence_of_function_on_line. Would `wrap1d` and `wrap2d` work for you?

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,596 @@
...
+    return ds_find_set_(v, [this](Index i) -> Index& { return ds_parent_vertex(i); });
+  }
+  Index ds_find_set_square(Index v) {
+    return ds_find_set_(v, [this](Index i) -> Index& { return ds_parent_square(i); });
+  }
+
+  struct Edge {
+    T f;
+    Index v1, v2; // v1 < v2
+    Edge() = default;
```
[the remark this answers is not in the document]
mglisse: Even if not used, the default constructor may be relevant in making the type "trivial" which can have performance implications. I would need to check the impact on codegen before removing it.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
hschreiber: Sorry, I answer two month later, but yes, I prefer the new version.

src/python/gudhi/_pers_cub_lowdim.cc
```diff
@@ -45,9 +49,34 @@ py::array fun(py::array_t<T> data) {
...
+}
+
+PYBIND11_MODULE(_pers_cub_lowdim, m) {
   py::bind_vector<Vf>(m, "VectorPairFloat" , py::buffer_protocol());
   py::bind_vector<Vd>(m, "VectorPairDouble", py::buffer_protocol());
-  m.def("persistence_on_a_line", fun<float>, py::arg().noconvert());
-  m.def("persistence_on_a_line", fun<double>);
+  m.def("_persistence_on_a_line", fun1<float>, py::arg().noconvert());
+  m.def("_persistence_on_a_line", fun1<double>);
+  m.def("_persistence_on_rectangle_from_top_cells", fun2);
```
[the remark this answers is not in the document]
VincentRouvreau: Ok, thanks for the answer, and sorry, my question arrives very late, so ok if you don't remember exactly.
This was more a question than something to be done.

src/Bitmap_cubical_complex/include/gudhi/Persistence_on_rectangle.h
```diff
@@ -0,0 +1,596 @@
...
+    // The unique_ptr could be std::vector, but the initialization is useless.
+    data_v_.reset(new T[input_size - dy - 1]); // 1 row/column less for vertices than squares
+    ds_parent_v_.reset(new Index[input_size - dy - 1]);
+    ds_parent_s_.resize(input_size); // Initializing the boundary squares to 0 is important, it represents the infinite exterior cell.
+    edges.reserve(input_size / 2); // What is a good estimate here? For a random 1000x1000 input, we get ~311k edges. For a checkerboard, ~498k.
+  }
+
+  bool has_larger_input(Index a, Index b, Filtration_value fb) const {
+    // Is passing fb useful, or would the compiler notice that it already has it available?
+    GUDHI_CHECK(a != b, std::logic_error("Bug in Gudhi"));
```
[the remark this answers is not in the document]
VincentRouvreau: I agree, that you will have the line number and that some of them may never happen. What annoys me, is that when you grep `"Bug in Gudhi"` in the code, you will have several occurences.

src/python/gudhi/_pers_cub_lowdim.cc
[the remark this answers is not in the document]
VincentRouvreau: `wrap_persistence_1d` and `wrap_persistence_2d` ?

mglisse:
My reasoning for putting the file in cubical rather than persistence (when it does both) was that cubical seemed more specific. If we put everything related to persistence together (Ripser as well?), it may be a bit messy. At the time, I had put the 1d case in persistence because 1d cubical is also a simplicial complex, but now I think it would have been better in cubical. Anyway, the directory is not visible for users, all that matters is how we document it (not at all for now...), so I put it in persistence like you both wanted.

src/Bitmap_cubical_complex/benchmark/persistence_2d.cpp
```diff
@@ -0,0 +1,92 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
```
VincentRouvreau, line 1: Do we move this file also in Persistent_cohomology/benchmark ?

VincentRouvreau (commented):
If you could also add a line in `.github/next_release.md` for this new feature, and that will be ok on my side

src/Bitmap_cubical_complex/benchmark/persistence_2d.cpp
mglisse: Oops... done.

VincentRouvreau (approved):

PR #848, opened 2023-03-31
use 'is not' instead of '!=' when comparing to None

ajouellette:
This fixes a problem in the weighted alpha complex where the construction fails if the weights are passed as a numpy array. This was not being caught in the tests, since the weights are being passed as a list instead of an array.

### DIFF
--- a/src/python/gudhi/alpha_complex.pyx
+++ b/src/python/gudhi/alpha_complex.pyx
@@ -87,14 +87,14 @@ cdef class AlphaComplex:
             points = read_points_from_off_file(off_file = off_file)
 
         # weights are set but is inconsistent with the number of points
-        if weights != None and len(weights) != len(points):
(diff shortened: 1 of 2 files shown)


mglisse (approved):
Thank you for this PR. To be precise, the issue is that `np.array([1,2]) != None` returns `array([ True,  True])`.

We have a few other `== None` and `!= None` elsewhere, but probably in less dangerous places, so we can change them later.

mglisse:
Note that afterwards we should add a test where AlphaComplex takes the points and weights as numpy arrays, to make sure we don't accidentally reintroduce the bug later. But it is ok not to do it as part of this PR.

mglisse:
By the way, don't worry about the "pip build windows" CI platform that has a red cross: vcpkg is broken as usual, that's not related to your patch.

VincentRouvreau (approved):

PR #856, opened 2023-04-04
Remove newshape cubical

VincentRouvreau:
* Use numpy.reshape in example.
* Externalize sklearn cubical example as a python file.
* Test sklearn cubical example if GUDHI_REMOTE_TEST (as it fetchs some data)


src/python/CMakeLists.txt
```diff
@@ -326,79 +326,103 @@ if(PYTHONINTERP_FOUND)
...
+    add_gudhi_py_test(test_cubical_complex)
+
+    # Datasets are fetched for these tests 
+    if(SKLEARN_FOUND AND WITH_GUDHI_REMOTE_TEST)
+      add_gudhi_py_test(test_sklearn_cubical_persistence)
+
+      add_test(NAME cubical_complex_sklearn_itf_py_test
+               WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+               COMMAND ${CMAKE_COMMAND} -E env "${GUDHI_PYTHON_PATH_ENV}"
+               ${PYTHON_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/example/cubical_complex_sklearn_itf.py")
```
mglisse, line 424: Is it the same code that was in the doc? If so, isn't it a bit too long for a test? Should it include an assertion at the end that the classification score was not too horrible?

VincentRouvreau: Yes, I moved the code from the doc in an example file that is now included in the doc and tested by the test suite only if scikit-learn and `WITH_GUDHI_REMOTE_TEST` (as data are fetched).

The test is quite long (90 sec. on my laptop, 490 sec. on circleCI, 373 sec. on Azure Windows, 230 sec. on azure OSx), but it worth it because this example changed a lot since its introduction in 3.6.0 and has been broken several times. I don't want to add it to sphinx_py_test that already takes a lot of time.

Don't you think this is a good compromise ?

mglisse: For someone running the testsuite in parallel, this doubles the time the whole testsuite takes, that's too long.
Now if you don't use all the images but only `X=X[:10000]` (same for `y`, and it doesn't have to be exactly 10000 I picked that randomly), the running time becomes much more acceptable, and the score is still good.

src/python/gudhi/sklearn/cubical_persistence.py
```diff
@@ -98,7 +92,7 @@ def __transform(self, cells):
     def transform(self, X, Y=None):
         """Compute all the cubical complexes and their associated persistence diagrams.
 
-        :param X: List of cells filtration values (`numpy.reshape(X, newshape, order='C'` if `newshape` is set with a tuple of ints).
+        :param X: List of cells filtration values.
         :type X: list of list of float OR list of numpy.ndarray
```
mglisse, line 96: "list of list of float" can also be list of list of list of ... of float. I think list of array-like would work, if you don't want to limit to list of ndarray.

src/python/gudhi/sklearn/cubical_persistence.py
```diff
@@ -98,7 +92,7 @@ def __transform(self, cells):
     def transform(self, X, Y=None):
         """Compute all the cubical complexes and their associated persistence diagrams.
 
-        :param X: List of cells filtration values (`numpy.reshape(X, newshape, order='C'` if `newshape` is set with a tuple of ints).
+        :param X: List of cells filtration values.
```
mglisse, line 95: Maybe we could be a bit clearer, though when I try I mostly end up repeating the type :-(
Filtration values of the top-dimensional cells or vertices for each complex?

VincentRouvreau: I took your proposal on [c0caed4](https://github.com/GUDHI/gudhi-devel/pull/856/commits/c0caed4de511df2695079eaf93b28bed860d59b7), thanks !

src/python/gudhi/sklearn/cubical_persistence.py
VincentRouvreau: Ok for list of array-like on [c0caed4](https://github.com/GUDHI/gudhi-devel/pull/856/commits/c0caed4de511df2695079eaf93b28bed860d59b7). I don't know if we should precise float ?

src/python/example/cubical_complex_sklearn_itf.py
```diff
@@ -0,0 +1,53 @@
...
+# Learn from the train subset
+pipe.fit(X_train, y_train)
+# Predict from the test subset
+predicted = pipe.predict(X_test)
+
+print(f"Classification report for TDA pipeline {pipe}:\n" f"{metrics.classification_report(y_test, predicted)}\n")
+
+report = metrics.classification_report(y_test, predicted, output_dict=True)
+assert report['0']['f1-score'] > 0.65
+assert report['0']['f1-score'] > 0.65
```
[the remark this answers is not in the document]
VincentRouvreau: Bad copy/paste, fixed on [c0caed4](https://github.com/GUDHI/gudhi-devel/pull/856/commits/c0caed4de511df2695079eaf93b28bed860d59b7)

src/python/example/cubical_complex_sklearn_itf.py
```diff
@@ -0,0 +1,53 @@
...
+# Import TDA pipeline requirements
+from gudhi.sklearn.cubical_persistence import CubicalPersistence
+from gudhi.representations import PersistenceImage, DiagramSelector
+
+X, y = fetch_openml("mnist_784", version=1, return_X_y=True, as_frame=False)
+
+# Take only the 500 images from this 70.000 images dataset
+# Remove these 2 lines if you want to test the full dataset
+X=X[:500]
+y=y[:500]
```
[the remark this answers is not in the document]
VincentRouvreau: File blacked on [c0caed4](https://github.com/GUDHI/gudhi-devel/pull/856/commits/c0caed4de511df2695079eaf93b28bed860d59b7)

src/python/gudhi/sklearn/cubical_persistence.py
```diff
@@ -35,7 +35,6 @@ class CubicalPersistence(BaseEstimator, TransformerMixin):
     def __init__(
         self,
         homology_dimensions,
-        newshape=None,
```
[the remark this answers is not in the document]
VincentRouvreau: Add `newshape` removal in release note on [2c3e7b1](https://github.com/GUDHI/gudhi-devel/pull/856/commits/2c3e7b11df12c2a80c8e5d3d02dd1e9f1da6aa64)

mglisse (approved):

PR #877, opened 2023-05-04
[Simplex_tree] Fast stars access

VincentRouvreau:
These development was taken from @ClementMaria on #401 and bugs were fixed.
Tested with sanitizer and there is no memory leaks.
As it is a development in progress I did not perform benchmarks yet.


src/Simplex_tree/include/gudhi/Simplex_tree/nodes_by_label.h
```diff
@@ -0,0 +1,57 @@
...
+    typename SimplexTree::List_max_vertex* find(typename SimplexTree::Vertex_handle v) {
+      auto it_v = nodes_label_to_list_.find(v);
+      if (it_v != nodes_label_to_list_.end()) {
+        return it_v->second;
+      } else {
+        return nullptr;
+      }
+    }
+    // map Vertex_handle v -> pointer to list of all Nodes with label v.
+    std::map<typename SimplexTree::Vertex_handle, typename SimplexTree::List_max_vertex*> nodes_label_to_list_;
```
hschreiber, line 52: Is there a reason why the value of the map needs to be a pointer ? (`typename SimplexTree::List_max_vertex*`)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1017,9 +1061,16 @@ class Simplex_tree {
...
+    if constexpr (Options::link_nodes_by_label) {  // faster cofaces computation
+      Simplex_vertex_range rg = simplex_vertex_range(simplex);
+      std::vector<Vertex_handle> simp(rg.begin(), rg.end());
+      // must be sorted in decreasing order
+      assert(std::is_sorted(simp.begin(), simp.end(), std::greater<Vertex_handle>()));
+      return Star_simplex_range(Star_simplex_iterator(this, simp), Star_simplex_iterator());
+    } else {
+      return cofaces_simplex_range(simplex, 0);
+    }
   }
```
hschreiber, line 1075: `Cofaces_simplex_range cofaces_simplex_range(const Simplex_handle simplex, int codimension)` could also benefit from the optimization if we are able to track the heights of nodes we are crossing. In the meantime, we can have (eventhough it is not really pretty I admit...):
```C++
Cofaces_simplex_range cofaces_simplex_range(const Simplex_handle simplex, int codimension) {
    if constexpr (Options::link_nodes_by_label) {
       if (codimension == 0) return star_simplex_range(simplex);
       [usual code...]
    } else {
       [usual code...]
    }
}
```

mglisse (commented):
Uh, for some reason github shows all these old comments as "pending", maybe I forgot to "finish" them...

src/Persistent_cohomology/test/persistent_cohomology_unit_test.cpp
```diff
@@ -175,6 +175,7 @@ struct MiniSTOptions {
   static const bool store_key = true;
   static const bool store_filtration = false;
   static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = false;
```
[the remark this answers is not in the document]
VincentRouvreau: Here, I had to set a value true or false to `link_nodes_by_label` as it is now mandatory (as MiniSTOptions does not inherit from another Simplex_tree option).
The aim of this test is to check an exception is raised when `num_simplices` > `max<Simplex_key>` and `link_nodes_by_label` has no effect to this mechanism.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -639,7 +666,12 @@ class Simplex_tree {
     return (sh->second.children()->parent() == sh->first);
   }
 
-    /** \brief Given a range of Vertex_handles, returns the Simplex_handle
+  /** \brief Returns the children of the node in the simplex tree pointed by sh.
+   * Invalid if sh has no children.
```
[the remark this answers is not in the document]
VincentRouvreau: GUDHI_CHECK has been added on [cb19d90](https://github.com/GUDHI/gudhi-devel/pull/877/commits/cb19d90bfe45785711a656b57eb0af11a9f0a954)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1160,8 +1214,11 @@ class Simplex_tree {
         sh->second.assign_children(new Siblings(&root_, sh->first));
       }
 
-      sh->second.children()->members().emplace(v,
-          Node(sh->second.children(), get(edge_filtration_t(), skel_graph, edge)));
+      auto it = (sh->second.children()->members().emplace(
```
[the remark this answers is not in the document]
VincentRouvreau: I removed useless parenthesis on [b620d29](https://github.com/GUDHI/gudhi-devel/pull/877/commits/b620d2956364e2e624925a3b1927a7153dc6d1fa)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1767,6 +1835,83 @@ class Simplex_tree {
...
+  }
+  /* \brief Returns true if the node in the simplex tree pointed by
+   * sh has children. node must have label u*/
+  bool has_children(Node& node, Vertex_handle u) const { return (node.children()->parent() == u); }
+
+ private:
+  // update all extra data structures in the Simplex_tree. Must be called after all
+  // simplex insertions.
+  void update_simplex_tree_after_node_insertion(Simplex_handle sh) {
+    std::cout << "update_simplex_tree_after_node_insertion" << std::endl;
```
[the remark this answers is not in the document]
VincentRouvreau: I moved new debug traces in between `#if DEBUG_TRACES` on [f17bd8f](https://github.com/GUDHI/gudhi-devel/pull/877/commits/f17bd8f56c96aaf0bf6a5a344775215302df89c3)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1160,8 +1214,11 @@ class Simplex_tree {
         sh->second.assign_children(new Siblings(&root_, sh->first));
       }
 
-      sh->second.children()->members().emplace(v,
-          Node(sh->second.children(), get(edge_filtration_t(), skel_graph, edge)));
+      auto it = (sh->second.children()->members().emplace(
+                     v, Node(sh->second.children(), get(edge_filtration_t(), skel_graph, edge))))
+                    .first;
+      // update extra data structures for new simplex
```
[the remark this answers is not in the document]
VincentRouvreau: I removed some useless comments on [a7d0830](https://github.com/GUDHI/gudhi-devel/pull/877/commits/a7d083093720d81256cf86324d3b6ad8fcd950af) for update_simplex_tree_after_node_insertion

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1017,9 +1061,16 @@ class Simplex_tree {
...
-
-  Cofaces_simplex_range star_simplex_range(const Simplex_handle simplex) {
-    return cofaces_simplex_range(simplex, 0);
+  Star_simplex_range star_simplex_range(const Simplex_handle simplex) {
+    if constexpr (Options::link_nodes_by_label) {  // faster cofaces computation
+      Simplex_vertex_range rg = simplex_vertex_range(simplex);
+      std::vector<Vertex_handle> simp(rg.begin(), rg.end());
+      // must be sorted in decreasing order
+      assert(std::is_sorted(simp.begin(), simp.end(), std::greater<Vertex_handle>()));
+      return Star_simplex_range(Star_simplex_iterator(this, simp), Star_simplex_iterator());
```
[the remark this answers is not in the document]
VincentRouvreau: Pass the vector by rvalue reference instead of copying it on 073a62e15

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+template <class Stree>
+void benchmark_stars_computation() {
+  using Vertex_handle  = typename Stree::Vertex_handle;
+  using Simplex_handle = typename Stree::Simplex_handle;
+
+  std::clog << "... Simplex_tree construction" << std::endl;
+
+  Stree st;
+  const Vertex_handle VERTEX_MAX = 5000;
```
mglisse, line 53: Is the test supposed to crash if I reduce VERTEX_MAX to 500?

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
```diff
@@ -0,0 +1,90 @@
...
+    st.insert_simplex_and_subfaces(rand_int_range<Vertex_handle>(2, 5, VERTEX_MAX));
+  std::cout << "... " << st.num_vertices() << " vertices and " << st.num_simplices() << " simplices." << std::endl;
+  
+  Gudhi::Clock benchmark_search("... Looking for random existing simplices");
+  const int SH_SIZE = 10000;
+  std::vector<Simplex_handle> sh_list;
+  sh_list.reserve(SH_SIZE);
+
+  std::uniform_int_distribution<int> dist(0, st.num_simplices());
+  // This method is not equivalent but quite faster than iterating randomly on complex_simplex_range
```
mglisse, line 64: I am not sure what you are comparing to, but I expect copying all Simplex_handle to a vector once then picking random ones in the vector would already be faster. Did you try std::sample?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1017,9 +1065,16 @@ class Simplex_tree {
    * \param simplex represent the simplex of which we search the star
    * \return Vector of Simplex_handle, empty vector if no cofaces found.
```
mglisse, line 1066: Not necessarily a vector.

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
```diff
@@ -0,0 +1,90 @@
...
+    for (int nb_rand = dist(rd); nb_rand > 0; nb_rand--)
+      simplex++;
+    sh_list.push_back(*simplex);
+  }
+  std::clog << benchmark_search << std::endl;
+
+  Gudhi::Clock benchmark_stars("Benchmark the stars search of the random simplices");
+  // Just browse the stars from the list
+  for (auto& sh : sh_list)
+    for ([[maybe_unused]] const auto& simplex : st.star_simplex_range(sh)) { }
```
mglisse, line 77: I think we should store the Simplex_handle in a volatile variable, to prevent the compiler from optimizing things out.
`{ volatile auto _ = simplex; }` for instance, possibly with maybe_unused.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
```diff
@@ -11,6 +11,13 @@
 #ifndef SIMPLEX_TREE_SIMPLEX_TREE_NODE_EXPLICIT_STORAGE_H_
 #define SIMPLEX_TREE_SIMPLEX_TREE_NODE_EXPLICIT_STORAGE_H_
 
+// Empty base optimization for MSVC - https://learn.microsoft.com/en-us/cpp/cpp/empty-bases
+#if _MSC_VER && !__INTEL_COMPILER
+ #define EMPTY_BASE_CLASS_OPTIMIZATION __declspec(empty_bases)
```
mglisse, line 16: Start macro names with `GUDHI_`? We may also want to define it in some common header.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
```diff
@@ -11,6 +11,13 @@
 #ifndef SIMPLEX_TREE_SIMPLEX_TREE_NODE_EXPLICIT_STORAGE_H_
 #define SIMPLEX_TREE_SIMPLEX_TREE_NODE_EXPLICIT_STORAGE_H_
 
+// Empty base optimization for MSVC - https://learn.microsoft.com/en-us/cpp/cpp/empty-bases
+#if _MSC_VER && !__INTEL_COMPILER
```
mglisse, line 15: Any reference for the `!__INTEL_COMPILER`?

VincentRouvreau: It was mentionned in [this stackoverflow](https://stackoverflow.com/questions/5850358/is-there-a-preprocessor-define-that-is-defined-if-the-compiler-is-msvc), but yes this may not be the case for us. I remove it.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
VincentRouvreau: I renamed it on e33b19896
Let's see later in which common header it would be the most convenient

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
VincentRouvreau: I added your proposal on 35c331235

src/Simplex_tree/include/gudhi/Simplex_tree/nodes_by_label.h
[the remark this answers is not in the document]
VincentRouvreau: use unordered_map as we do not iterate on it (only in the destructor), cf. 90d0b0de4

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
VincentRouvreau: I removed VERTEX_MAX and rephrase some comments on c0d24ccbb

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
VincentRouvreau: 468db5e60 : It is quite better to use std::sample to pick 20.000 random simplices from the Simplex_tree

src/Simplex_tree/include/gudhi/Simplex_tree/nodes_by_label.h
```diff
@@ -0,0 +1,57 @@
...
+namespace Gudhi {
+
+  // trivial data structure, in case the option is disallowed.
+  template <typename SimplexTree>
+  struct Nodes_by_label_dummy {};
+
+  // use the Node hooks Hooks_simplex_base to put all Nodes with same label u in an
+  // intrusive list.
+  template <typename SimplexTree>
+  struct Nodes_by_label_intrusive_list {
```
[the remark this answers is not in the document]
VincentRouvreau: You are right and I removed it on 1a348e405

src/Simplex_tree/include/gudhi/Simplex_tree/hooks_simplex_base.h
```diff
@@ -0,0 +1,69 @@
...
+  struct Hooks_simplex_base_link_nodes {
+   private:
+    BOOST_COPYABLE_AND_MOVABLE(Hooks_simplex_base_link_nodes)
+   public:
+    Hooks_simplex_base_link_nodes() {}
+    // the copy constructor, inherited by the Node class, exchanges hooks,
+    // and make the ones of this invalid.
+    // this is used when stored in a map like DS, using copies when
+    // performing insertions and rebalancing of the rbtree
+    Hooks_simplex_base_link_nodes(const Hooks_simplex_base_link_nodes& other) {
```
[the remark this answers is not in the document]
VincentRouvreau: I replaced boost syntax from Hooks_simplex_base_link_nodes with a c++11 syntax on 95475ea94 , but only a move syntax is not sufficient enough for Simplex_tree constructor and move unitary tests

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1132,11 +1183,14 @@ class Simplex_tree {
       dimension_ = 1;
     }
 
-    root_.members_.reserve(num_vertices(skel_graph)); // probably useless in most cases
-    auto verts = vertices(skel_graph) | boost::adaptors::transformed([&](auto v){
-        return Dit_value_t(v, Node(&root_, get(vertex_filtration_t(), skel_graph, v))); });
-    root_.members_.insert(boost::begin(verts), boost::end(verts));
-    // This automatically sorts the vertices, the graph concept doesn't guarantee the order in which we iterate.
```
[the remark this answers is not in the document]
VincentRouvreau: Sorry, I missed this merge. I fixed it on ff4b890c1 (tested with Simplex_tree_example_from_cliques_of_graph)

src/Simplex_tree/include/gudhi/Simplex_tree/hooks_simplex_base.h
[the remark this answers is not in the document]
VincentRouvreau: I fixed documentation and remove comments about movable/copyable part on bdbab2d61

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1339,11 +1400,16 @@ class Simplex_tree {
         for (auto new_sib_member = new_sib->members().begin();
              new_sib_member != new_sib->members().end();
              new_sib_member++) {
+           // update data structures for all new simplices
+           update_simplex_tree_after_node_insertion(new_sib_member);
```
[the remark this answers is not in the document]
VincentRouvreau: Yes, in `siblings_expansion_with_blockers`, we need to `update_simplex_tree_after_node_insertion / update_simplex_tree_before_node_removal`, the same way we create / delete siblings in function of the blocker oracle result

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,289 @@
...
+          }  // we found a coface
+        }
+        curr_sib = curr_sib->oncles();
+      }
+      return false;
+    }
+
+   private:
+    SimplexTree* cpx_;
+    std::vector<Vertex_handle> simp_;  // vertices of simplex, in reverse order
```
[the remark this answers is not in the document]
VincentRouvreau: Yes, in boundary iterators:
 https://github.com/GUDHI/gudhi-devel/blob/23431ea266e6ba6d8a1926707fadfe31bb7db904/src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_iterators.h#L176
Do you think we should use a static vector, too ?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1767,6 +1835,87 @@ class Simplex_tree {
...
+  }
+
+  // update all extra data structures in the Simplex_tree. Must be called before
+  // all simplex removals
+  void update_simplex_tree_before_node_removal(Simplex_handle sh) {
+#ifdef DEBUG_TRACES
+    std::clog << "update_simplex_tree_before_node_removal" << buffer_byte_size << std::endl;
+#endif  // DEBUG_TRACES
+    if constexpr (Options::link_nodes_by_label) {
+      sh->second.unlink_hooks();  // remove from lists of same label Nodes
```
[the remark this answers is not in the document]
hschreiber: If we remove a lot of vertices, removing empty lists could be worth it, as it was the case for the compressed version of cohomology. On the other side, I don't know if it happens often that vertices are removed in practice and then the lists will never be empty. But the zigzag flag iterator seems to remove them.

src/Simplex_tree/include/gudhi/Simplex_tree/nodes_by_label.h
VincentRouvreau: This is a good point and I tried to do so as it was more convenient for the simplex tree destructor and for the use of the second term.
But I think Clément is using a pointer because `boost::intrusive::list` is movable but not copyable, cf.
https://github.com/boostorg/intrusive/blob/060a6398c768c1091400ba304113fccaa64b79f3/include/boost/intrusive/list.hpp#L121
I don't see how to do it differently, any idea ?

mglisse: It is possible to use move-only types with unordered_map, you just have to be a bit careful which exact functions you use.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,289 @@
...
+    bool operator()(typename SimplexTree::Hooks_simplex_base& curr_hooks) {
+      Node& curr_node = static_cast<Node&>(curr_hooks);
+      auto vertex_it = simp_.begin();  // largest label is first
+      // first Node must always have label simp_.begin(); we assume it is true
+      auto curr_sib = cpx_->self_siblings(curr_node, *vertex_it);
+      if (++vertex_it == simp_.end()) {
+        return true;
+      }
+      while (curr_sib->oncles() != nullptr) {
+        if (curr_sib->parent() == *vertex_it) {
```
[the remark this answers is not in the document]
VincentRouvreau: Same timings with  `if (curr_sib->parent() < *vertex_it)`

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Ok, but this implies to cast `star_simplex_range` as a `cofaces_simplex_range`, which is taking some time (0.036 sec. vs 0.017 sec. without casting):
```cpp
      Star_simplex_range star_range(Star_simplex_iterator(this, std::move(simp)), Star_simplex_iterator());
      return Cofaces_simplex_range(star_range.begin(), star_range.end());
```

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
```diff
@@ -0,0 +1,85 @@
...
+  sh_list.reserve(SH_SIZE);
+
+  std::sample(st.complex_simplex_range().begin(), st.complex_simplex_range().end(), std::back_inserter(sh_list),
+              SH_SIZE, std::mt19937 {rd()});
+  std::clog << benchmark_search << std::endl;
+
+  Gudhi::Clock benchmark_stars("Benchmark the stars search of the random simplices");
+  // Just browse the stars from the random simplices list
+  for (auto& sh : sh_list)
+    for ([[maybe_unused]] const auto& simplex : st.star_simplex_range(sh))
```
mglisse, line 71: `simplex` is used now, the attribute looks unnecessary.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Which benchmark? It is true that putting the handles in a vector is not free, but for a sparse complex, with the new option to use linked lists, it should save us from the expensive search on a large part of the complex.

By the way, you could skip `star_range` and pass the iterators directly to the vector (Cofaces_simplex_range).

VincentRouvreau: Benchmark is the one provided in this PR: https://github.com/VincentRouvreau/gudhi-devel/blob/fast_cofaces/src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp

I tried to skip `star_range`, but it is still slower (0.033 sec. vs 0.019 sec.). As you said, putting the handles in a vector is not free, but the return type is much understandable. I am ok to go with this solution as it will simplify the interfaces and the documentation. And you ?

src/Simplex_tree/benchmark/simplex_tree_stars_benchmark.cpp
VincentRouvreau: You are right, it is used just below. I removed `[[maybe_unused]]` on 5e43c7c

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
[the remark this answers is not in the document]
VincentRouvreau: There was a bug in the unitary test. Fixed on [32cca6f](https://github.com/GUDHI/gudhi-devel/pull/877/commits/32cca6fbe4bada53fb2b3b1104e51e62f9b33c71)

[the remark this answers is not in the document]
VincentRouvreau: This code is now passing the fixed test:
```cpp
#include <functional>  // for std::greater
...
    bool operator()(typename SimplexTree::Hooks_simplex_base& curr_hooks) {
      Node& curr_node = static_cast<Node&>(curr_hooks);
      auto vertex_it = simp_.begin();  // largest label is first
      // first Node must always have label simp_.begin(); we assume it is true
      auto sh = cpx_->self_siblings(curr_node, *vertex_it)->find(*vertex_it);
      // is simp_ a face of the simplex defined by sh ?
      return std::includes(cpx_->simplex_vertex_range(sh).begin(), cpx_->simplex_vertex_range(sh).end(),
                           simp_.begin(), simp_.end(), std::greater<Vertex_handle>());
    }
```
Just a little bit slower (0.022 sec. versus 0.017 sec. for the benchark), but much more expressive and maintainable.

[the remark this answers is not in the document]
VincentRouvreau: I merged it on 0e4ba24

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: A vector is a simpler object, but if using the self-made iterators is faster (as it avoids filling a vector), why not make the cofaces an iterator too? I find it still weird that we have to different types "star_simplex_range" and "cofaces_simplex_range", when mathematically speaking there are the same thing. We just use one as special case, but I wouldn't be surprised if it confuses more than one user later. Can we not factorize something into the iterator? For example having two iterators, one for the codim==0, the other for codim>0, inheriting both from the same base class and they act differently depending on the value of Options::link_nodes_by_label. At the end, we would have a single type "cofaces_simplex_range" for both methods. It will be more work, but I think it will be cleaner.

hschreiber: By the way, Clément made de star access an iterator, because I need to be able to iterate two stars at the same time and stop when the smallest star finishes (ie a complexity depending on the smallest star, not on the biggest) for the tower algorithm. Other applications could need something similar too, but for the star of a higher dimensional simplex. This would be in favor of the iterator range instead of the vector.

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: It was no more the case, but thanks to [5104751](https://github.com/GUDHI/gudhi-devel/pull/877/commits/5104751ff686175d30b0089ddb20f6f21a108fa9) I rolled back the return type of star_simplex_range.

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: star_simplex_range and coface_simplex_range now return the same type on [5104751](https://github.com/GUDHI/gudhi-devel/pull/877/commits/5104751ff686175d30b0089ddb20f6f21a108fa9)

src/Simplex_tree/include/gudhi/Simplex_tree/nodes_by_label.h
VincentRouvreau: Thanks @mglisse for the help, done on 77d4d1c

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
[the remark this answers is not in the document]
VincentRouvreau: Use boost static_vector and factorize the type in the Simplex tree on 6e5af55

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1030,19 +1087,41 @@ class Simplex_tree {
...
+
+    if constexpr (Options::link_nodes_by_label) {
+      // faster cofaces computation
+      if (codimension == 0) {
+        // Always true for a star
+        select = [](Const_boost_iterator) { return true; };
+      } else {
+        // Specific coface case
+        select = [this, simplex, codimension](Const_boost_iterator iter) {
+            return ((this->dimension(simplex) + codimension) == this->dimension(iter));
```
mglisse, line 1101:
```suggestion
        select = [this, dim=this->dimension(simplex)+codimension](Const_boost_iterator iter) {
            return dim == this->dimension(iter);
```
Or something like that to avoid recomputing the dimension of `simplex` all the time?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -165,6 +173,22 @@ class Simplex_tree {
...
+   * linked in a list.
+   */
+  using Optimized_star_simplex_iterator = Simplex_tree_optimized_star_simplex_iterator<Simplex_tree>;
+  /** \brief Range for an optimized search for the star of a simplex. */
+  using Optimized_star_simplex_range = boost::iterator_range<Optimized_star_simplex_iterator>;
+
+  using Const_boost_iterator = const boost::container::vec_iterator<std::pair<int, Node >*, false>;
+  using Filtered_boost_predicate = std::function<bool(Const_boost_iterator)>;
+  std::function<bool(Const_boost_iterator)> select;
+  using Optimized_star_simplex_filtered_range = boost::filtered_range<decltype(select), Optimized_star_simplex_range>;
```
mglisse, line 190: I am trying to think how safe this is. filtered_range does not keep a copy of its range argument, so returning a filtered_range of a local range that is going out of scope is dangerous. However, filtered_range only uses the range to call begin() and end() on it, and stores filtered versions of both. Since Optimized_star_simplex_range is a boost::iterator_range (i.e. a pair of iterators begin/end), extracting just its begin and end and dropping the range should be safe.
I think this deserves a comment in the code, since any small change can easily break this.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1767,6 +1846,96 @@ class Simplex_tree {
...
+  typedef boost::intrusive::list<Hooks_simplex_base_link_nodes, List_member_hook_t,
+                                 boost::intrusive::constant_time_size<false>> List_max_vertex;
+  // type of hooks stored in each Node, Node inherits from Hooks_simplex_base
+  typedef typename std::conditional<Options::link_nodes_by_label, Hooks_simplex_base_link_nodes,
+                                    Hooks_simplex_base_dummy>::type Hooks_simplex_base;
+  /** Data structure to access all Nodes with a given label u. Can be used for faster
+   * computation. */
+ private:
+  // if Options::link_nodes_by_label is true, store the lists of Nodes with same label, empty otherwise.
+  // unordered_map Vertex_handle v -> pointer to list of all Nodes with label v.
```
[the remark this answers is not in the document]
VincentRouvreau: I forgot to update this comment when doing [77d4d1c](https://github.com/GUDHI/gudhi-devel/pull/877/commits/77d4d1c484a5a9c0f22718dd6bd4b8116c09a50b). Comment updated on 4876df9

src/Simplex_tree/include/gudhi/Simplex_tree/hooks_simplex_base.h
```diff
@@ -0,0 +1,64 @@
...
+ *    Copyright (C) 2020 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#ifndef SIMPLEX_TREE_HOOKS_SIMPLEX_BASE_H_
+#define SIMPLEX_TREE_HOOKS_SIMPLEX_BASE_H_
+
+#include <boost/intrusive/list.hpp>
```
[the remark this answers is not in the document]
VincentRouvreau: Good advice. I launched the test suite in debug mode and no assert

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -165,6 +173,22 @@ class Simplex_tree {
...
+   * extra pointers in each node of the simplex tree. The Nodes of same label are
+   * linked in a list.
+   */
+  using Optimized_star_simplex_iterator = Simplex_tree_optimized_star_simplex_iterator<Simplex_tree>;
+  /** \brief Range for an optimized search for the star of a simplex. */
+  using Optimized_star_simplex_range = boost::iterator_range<Optimized_star_simplex_iterator>;
+
+  using Const_boost_iterator = const boost::container::vec_iterator<std::pair<int, Node >*, false>;
+  using Filtered_boost_predicate = std::function<bool(Const_boost_iterator)>;
+  std::function<bool(Const_boost_iterator)> select;
```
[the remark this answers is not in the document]
VincentRouvreau: select is a class member because `boost::adaptors::filter` returns a `boost::filtered_range<decltype(select), Optimized_star_simplex_range>;`.

I opened an [issue on boost range documentation](https://github.com/boostorg/range/pull/147) because this was missing.

I am also not a big fan of `std::function`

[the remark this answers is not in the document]
mglisse: I don't understand how that's related. `decltype(select)` is just `Filtered_boost_predicate`.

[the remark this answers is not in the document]
VincentRouvreau: Ok, now I got it. I moved select from class attribute to a variable in `cofaces_simplex_range` method on e9f7c6a

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: I am ok with your proposal, done on 2adeccc

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,289 @@
...
+ * link_nodes_by_label == true) in order to connect all nodes with a fixed label u
+ * into a list.
+ *
+ * Forward iterator, value_type is SimplexTree::Simplex_handle.
+ *
+ * The implementation uses a filter iterator on all Nodes of label u, checking
+ * whether they represent a coface of \f$\sigma\f$ or not.
+ */
+template <class SimplexTree>
+class Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator
```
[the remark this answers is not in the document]
VincentRouvreau: No, and it is not available in the documentation

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1767,6 +1846,96 @@ class Simplex_tree {
...
+        return nullptr;
+      }
+    }
+    return nullptr;
+  }
+
+  // basic methods implemented for Nodes, and not Simplex_handle. The hooks in nodes_label_to_list_ gives access to
+  // Nodes.
+ public:
+  // set of methods taking a node and a vertex_handle as input. For internal use only.
```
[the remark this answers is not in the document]
VincentRouvreau: That's a good idea and the code is clearer on 0757252

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
VincentRouvreau: I replaced `std::function` with a functor on [8c21fdf](https://github.com/GUDHI/gudhi-devel/pull/877/commits/8c21fdf5d37cb65da8e42f5dba64b27ef3218307). Timings are almost the same.

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: I added a comment on [d72926c](https://github.com/GUDHI/gudhi-devel/pull/877/commits/d72926c48a5628f7c72d0deb0fadedfb94776bdd)

src/Simplex_tree/include/gudhi/Simplex_tree/hooks_simplex_base.h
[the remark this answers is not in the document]
mglisse:
```diff
--- a/src/Simplex_tree/test/simplex_tree_unit_test.cpp
+++ b/src/Simplex_tree/test/simplex_tree_unit_test.cpp
@@ -593,6 +593,7 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(NSimplexAndSubfaces_tree_insertion, typeST, list_o
 template<class typeST, class Vertex_handle>
 void test_cofaces(typeST& st, const std::vector<Vertex_handle>& expected, int dim, const std::vector<typename typeST::Simplex_handle>& res) {
   std::size_t nb_res = 0;
+  typeST marc = st;
   if (dim == 0) {
     typename typeST::Cofaces_simplex_range stars = st.star_simplex_range(st.find(expected));
     for (auto simplex = stars.begin(); simplex != stars.end(); ++simplex) {
```
causes the test to fail.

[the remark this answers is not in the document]
VincentRouvreau: You are right. I modified copy constructor to copy hooks instead of exchanging them on [b7322d5](https://github.com/GUDHI/gudhi-devel/pull/877/commits/b7322d5435e6f10cd2a236319b33a4da5f402f0e)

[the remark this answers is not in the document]
mglisse: Now this version fails:
```diff
--- a/src/Simplex_tree/test/simplex_tree_unit_test.cpp
+++ b/src/Simplex_tree/test/simplex_tree_unit_test.cpp
@@ -591,8 +591,9 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(NSimplexAndSubfaces_tree_insertion, typeST, list_o
 }
 
 template<class typeST, class Vertex_handle>
-void test_cofaces(typeST& st, const std::vector<Vertex_handle>& expected, int dim, const std::vector<typename typeST::Simplex_handle>& res) {
+void test_cofaces(typeST& st_, const std::vector<Vertex_handle>& expected, int dim, const std::vector<typename typeST::Simplex_handle>& res) {
   std::size_t nb_res = 0;
+  typeST st = st_;
   if (dim == 0) {
     typename typeST::Cofaces_simplex_range stars = st.star_simplex_range(st.find(expected));
     for (auto simplex = stars.begin(); simplex != stars.end(); ++simplex) {
```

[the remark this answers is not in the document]
VincentRouvreau: Ok. I added some specific code and tests on [95ce710](https://github.com/GUDHI/gudhi-devel/pull/877/commits/95ce710cfbc2697de328b943e35b15becaf0e309) for copy/move constructor/assignement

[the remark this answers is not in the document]
mglisse: Maybe we could make things a bit less misleading now?
```diff
--- a/src/Simplex_tree/include/gudhi/Simplex_tree/hooks_simplex_base.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree/hooks_simplex_base.h
@@ -28,12 +28,15 @@ namespace Gudhi {
   struct Hooks_simplex_base_link_nodes {
    public:
     Hooks_simplex_base_link_nodes() {}
+    // The following 2 functions are conceptually wrong, but when copying a Simplex_tree it is more convenient to have them (and fix things asap after they are called).
     // copy constructor, inherited by the Node class, requires to copy hooks
-    Hooks_simplex_base_link_nodes(const Hooks_simplex_base_link_nodes& other)
-    : list_max_vertex_hook_(other.list_max_vertex_hook_) { }
+    Hooks_simplex_base_link_nodes(const Hooks_simplex_base_link_nodes&)
+    : list_max_vertex_hook_() { }
     // copy assignment
-    Hooks_simplex_base_link_nodes& operator=(const Hooks_simplex_base_link_nodes& other) {
-      list_max_vertex_hook_.swap_nodes(other.list_max_vertex_hook_);
+    Hooks_simplex_base_link_nodes& operator=(const Hooks_simplex_base_link_nodes&) {
+#ifdef GUDHI_DEBUG
+      Member_hook_t().swap_nodes(list_max_vertex_hook_);
+#endif
       return *this;
     }
     // move constructor
@@ -52,7 +55,7 @@ namespace Gudhi {
         boost::intrusive::link_mode<boost::intrusive::auto_unlink>>
         Member_hook_t;
 
-    mutable Member_hook_t list_max_vertex_hook_;
+    Member_hook_t list_max_vertex_hook_;
   };
 
 }  // namespace Gudhi
```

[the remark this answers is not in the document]
VincentRouvreau: I took your suggestion as is on [9c0611f](https://github.com/GUDHI/gudhi-devel/pull/877/commits/9c0611f38a2d851f773357b04ba3e2e75e3bc76d). I didn't get why we cannot say:
```cpp
Hooks_simplex_base_link_nodes& operator=(const Hooks_simplex_base_link_nodes&) = delete;
```
(I have some compilation error)
instead of throwing an exception.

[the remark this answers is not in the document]
mglisse: Something like: internally, copying the array delegates the work to a more general function like insert, which needs operator= when there are already elements in the array. So we need operator= to compile, but the code using it is dead in our case.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,283 @@
...
+  // forest have label max_v.
+  //[it_,end_) == range of Simplex_handles of the roots of the cofaces trees (any dim)
+  Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator<SimplexTree> it_;
+  Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator<SimplexTree> end_;
+  // curr Simplex_handle, returned by operator*, pointing to a coface of s
+  Simplex_handle sh_;
+  // set of siblings containing sh_ in the Simplex_tree
+  Siblings* sib_;  //
+  // use a bfs search to avoid calling sib_->members().find(.)
+  std::queue<Siblings*> bfs_queue_;
```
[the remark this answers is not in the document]
VincentRouvreau: I replaced the queue by a static_vector on [45421f7](https://github.com/GUDHI/gudhi-devel/pull/877/commits/45421f76dca18ac5cd8223ac8180a7ffd1ef11bd) and it is quite faster:
* stars 0.049 sec. -> 0.017 sec.
* cofaces (codim=1) 0.051 sec. -> 0.018 sec.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,285 @@
...
+      : st_(cpx), it_(cpx, std::move(simp)), end_(), sh_(*it_), sib_(st_->self_siblings(sh_)), bfs_queue_() {
+    if (it_ == end_) {
+      st_ = nullptr;
+      return;
+    }  // no coface subtree => end()
+    is_root_ = true;
+    sh_ = *it_;                      // sh_ is the root
+    sib_ = st_->self_siblings(sh_);  // Siblings containing sh_
+    if (st_->has_children(sh_)) {
+      bfs_queue_.insert(bfs_queue_.begin(), st_->children(sh_));
```
mglisse, line 190: You are trying to keep it as a queue, but I am not convinced that the order actually matters. Assuming it doesn't, you can handle it as a stack and just push_back when you want to add an element.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,285 @@
...
+  //[it_,end_) == range of Simplex_handles of the roots of the cofaces trees (any dim)
+  Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator<SimplexTree> it_;
+  Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator<SimplexTree> end_;
+  // curr Simplex_handle, returned by operator*, pointing to a coface of s
+  Simplex_handle sh_;
+  // set of siblings containing sh_ in the Simplex_tree
+  Siblings* sib_;  //
+  // use a bfs search to avoid calling sib_->members().find(.)
+  // A static_vector is quite faster than a queue
+  // 100 seems a conservative bound on the number of children a Sibling can coutain for now.
```
mglisse, line 276: What? The root Siblings can have as many children as there are vertices (minus 1). And even if that one somehow cannot appear in coface computation, the Siblings of the first vertex can have n-2 (you can just add all possible triangles for instance).
If the size of this container is indeed related to Siblings (that would make sense for a BFS), we cannot use a static_vector. If instead it is bounded by the depth of the Simplex_tree (which would probably be the case for a DFS?), we could use 40 like we do elsewhere.

I don't know if there was a good reason for Clément to choose a BFS or if he picked this order arbitrarily.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
VincentRouvreau: Ok, let's use a stack on [ca218a9](https://github.com/GUDHI/gudhi-devel/pull/877/commits/ca218a989b250885e8cd8056f425b566b0df9a64). Order doesn't seem to matter

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
VincentRouvreau: Hum... You are right. From the test I performed, the size was not going up to 20 (with my benchmark), this is why I put 100. I don't know what was Clément intention, but let's go with a std::vector as timings are still good and it will be dynamic.
I don't think it will grow that much as the main code that is used is:
```cpp
      if (!children_stack_.empty()) {
        sib_ = children_stack_.back();
        children_stack_.pop_back();
        sh_ = sib_->members().begin();  // don't track dimensions
        if (st_->has_children(sh_)) {
          children_stack_.push_back(st_->children(sh_));
        }
      }
```
I am using a std::vector from [ca218a9](https://github.com/GUDHI/gudhi-devel/pull/877/commits/ca218a989b250885e8cd8056f425b566b0df9a64)

mglisse: I would say the "main" code is `++sh_` on the first line (iterate on all the nodes of this Siblings) and then the last `else` branch that adds the corresponding children. So it adds all the children to the stack before it starts looking at them (normal for BFS, although it could be done more implicitly) and the structure can contain roughly n_vertices * dimension objects. Of course, this structure will only grow large in cases where one vertex has many neighbors, which is a bad case for this optimization. But at least we want to avoid crashing in that case, even if it will be slow. So yes, vector (or `stack<vector>`).

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: So now again it isn't necessarily a vector?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,282 @@
...
+    is_coface() : cpx_(nullptr) {}
+    is_coface(SimplexTree* cpx, Static_vertex_vector&& simp) : cpx_(cpx), simp_(simp) {}
+
+    // Return true iff traversing the Node upwards to the root reads a
+    // coface of simp_
+    bool operator()(typename SimplexTree::Hooks_simplex_base& curr_hooks) {
+      Node& curr_node = static_cast<Node&>(curr_hooks);
+      auto vertex_it = simp_.begin();  // largest label is first
+      // first Node must always have label simp_.begin(); we assume it is true
+      auto sh = cpx_->self_siblings(cpx_->simplex_handle_from_node(curr_node))->find(*vertex_it);
```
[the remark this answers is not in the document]
VincentRouvreau: Ok, I got it, I removed `find` calls on [5542527](https://github.com/GUDHI/gudhi-devel/pull/877/commits/55425270cef0d9a6cf3f90fbaec4a96eac195a0a). `max_v_` was no more usefull

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1767,6 +1872,75 @@ class Simplex_tree {
...
+      } else {
+        return nullptr;
+      }
+    }
+    return nullptr;
+  }
+
+  /** \brief Helper method that returns the corresponding Simplex_handle from a member element defined by a node. */
+  static Simplex_handle simplex_handle_from_node(Node& node) {
+    return (Simplex_handle)(boost::intrusive::get_parent_from_member<Dit_value_t>(&node, &Dit_value_t::second));
```
mglisse, line 1908: The cast is because we have a pointer whereas we want an iterator, and boost is kind enough to provide an explicit constructor for this conversion? ok.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1767,6 +1872,75 @@ class Simplex_tree {
...
+  void update_simplex_tree_after_node_insertion(Simplex_handle sh) {
+#ifdef DEBUG_TRACES
+    std::clog << "update_simplex_tree_after_node_insertion" << std::endl;
+#endif  // DEBUG_TRACES
+    if constexpr (Options::link_nodes_by_label) {
+      auto it = nodes_label_to_list_.find(sh->first);
+      if (it == nodes_label_to_list_.end()) {  // create a new list
+        it = (nodes_label_to_list_.emplace(sh->first, List_max_vertex())).first;
+      }
+      it->second.push_back(sh->second);  // insert at the end of the list
```
mglisse, line 1927: Aren't those 5 lines equivalent to `nodes_label_to_list_[sh->first].push_back(sh->second);`?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -187,7 +229,16 @@ class Simplex_tree {
...
-  typedef std::vector<Simplex_handle> Cofaces_simplex_range;
+  typedef typename std::conditional<Options::link_nodes_by_label,
+                                    Optimized_star_simplex_filtered_range,  // faster implem
+                                    std::vector<Simplex_handle>>::type Cofaces_simplex_range;
+
+  // 40 seems a conservative bound on the dimension of a Simplex_tree for now, as it would not fit on the biggest
+  // hard-drive.
+  // static_vector still has some overhead compared to a trivial hand-made
+  // version using std::aligned_storage, or compared to making suffix_ static.
+  using Static_vertex_vector = boost::container::static_vector<Vertex_handle, 40>;
```
mglisse, line 240: `/// @private` or something to hide it in the doc?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2001,6 +2175,7 @@ struct Simplex_tree_options_full_featured {
   static const bool store_key = true;
   static const bool store_filtration = true;
   static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = false;
```
mglisse, line 2178: This new option needs to be documented in concept/SimplexTreeOptions.h. Maybe we should mention its existence in the doc of star_simplex_range and cofaces_simplex_range.
From its name, it sounds like Simplex_tree_options_full_featured should have this option (otherwise it gets as confusing as intmax_t being shorter than int128_t). On the other hand, a number of current uses of Simplex_tree_options_full_featured don't need the new option. That's something that requires some thought, so if we merge this branch, I want a blocking issue to force us to settle that before the next release.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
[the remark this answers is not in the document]
VincentRouvreau: Ok, I marked the documentation class as private on [4dde99f](https://github.com/GUDHI/gudhi-devel/pull/877/commits/4dde99fb1cede0482c5fcefb618d425d40318d55)

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Marked as private on [0bb0bd3](https://github.com/GUDHI/gudhi-devel/pull/877/commits/0bb0bd3bdab957c1cf77a2d0649f7397c8ab22b9)

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Nice move ! I took your proposal on [ca8bfc5](https://github.com/GUDHI/gudhi-devel/pull/877/commits/ca8bfc5884d99148110a69f619b82b94ab4657db) as `nodes_label_to_list_[sh->first]` will create a new element in the map if `sh->first` is not there.

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Yes, I had to cast it

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -165,6 +174,39 @@ class Simplex_tree {
...
+    int dim_;
+   public:
+    Fast_cofaces_predicate(Simplex_tree* st, int codim, int dim)
+      : st_(st), codim_(codim), dim_(dim) {}
+    bool operator()( Const_boost_iterator iter ) const {
+      if (codim_ == 0)
+        // Always true for a star
+        return true;
+      // Specific coface case
+      return dim_ + codim_ == st_->dimension(iter);
```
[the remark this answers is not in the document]
VincentRouvreau: Store `dim_ + codim_` on [21f527e](https://github.com/GUDHI/gudhi-devel/pull/877/commits/21f527e72fc7e1d482168e781d731a32a06fdf13), no need to compute it on each iteration

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -187,7 +229,16 @@ class Simplex_tree {
   /** \brief Range over the vertices of a simplex. */
   typedef boost::iterator_range<Simplex_vertex_iterator> Simplex_vertex_range;
   /** \brief Range over the cofaces of a simplex. */
-  typedef std::vector<Simplex_handle> Cofaces_simplex_range;
+  typedef typename std::conditional<Options::link_nodes_by_label,
+                                    Optimized_star_simplex_filtered_range,  // faster implem
+                                    std::vector<Simplex_handle>>::type Cofaces_simplex_range;
```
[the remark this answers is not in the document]
VincentRouvreau: Ok, I added a comment on 3e4303b

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,282 @@
...
+      // is simp_ a face of the simplex defined by sh ?
+      return std::includes(cpx_->simplex_vertex_range(sh).begin(), cpx_->simplex_vertex_range(sh).end(),
+                           simp_.begin(), simp_.end(), std::greater<Vertex_handle>());
```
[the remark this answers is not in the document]
VincentRouvreau: Ok, plus some minor changes on [6bb7718](https://github.com/GUDHI/gudhi-devel/pull/877/commits/6bb7718b9b31c7b269f9f73ab14000450fb02fab)

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Add some documentation on dd26bff

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: I specified cofaces/stars_simplex_range return type in function of link_nodes_by_label on [a27f109](https://github.com/GUDHI/gudhi-devel/pull/877/commits/a27f1097a207f4d3eb7a496227f2d0773195be6b)

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+   private:
+    SimplexTree* cpx_;
+    Static_vertex_vector simp_;  // vertices of simplex, in reverse order
+  };
+
+  typedef boost::filter_iterator<is_coface, typename SimplexTree::List_max_vertex::iterator>
+      Filtered_cofaces_simplex_iterator;
```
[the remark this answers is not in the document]
VincentRouvreau: It would be a `boost::filtered_range<decltype(is_coface()), typename SimplexTree::Hooks_simplex_base>` ?

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: We don't have to be that specific (we don't want users to rely on the fact that it is a vector in some case), but that will do for now.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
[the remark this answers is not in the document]
mglisse: `filtered_range<is_coface, List_max_vertex>`?

PR #885, opened 2023-05-24
Add networkx in optionnal runtime dependencies

### LINKED ISSUE #884: Networkx should be listed in python run time dependencies
In [Python installation manual](https://gudhi.inria.fr/python/latest/installation.html#run-time-dependencies), networkx should be mentioned as it is used in `CoverComplexPy.get_networkx`

VincentRouvreau:
Fix #884


mglisse (approved):
The cmake stuff seems useless for now, but this way it will be ready if we add a test that for get_networkx.

PR #892, opened 2023-06-08
Only periodic cubical shall set persistence_dim_max to True

VincentRouvreau:
Also required unitary tests changes

### DIFF
--- a/src/python/test/test_cubical_complex.py
+++ b/src/python/test/test_cubical_complex.py
@@ -89,8 +89,8 @@ def simple_constructor(cub):
     assert cub.__is_persistence_defined() == False
     assert cub.persistence() == [(0, (1.0, float("inf")))]
     assert cub.__is_persistence_defined() == True
-    assert cub.betti_numbers() == [1, 0, 0]
-    assert cub.persistent_betti_numbers(0, 1000) == [0, 0, 0]
+    assert cub.betti_numbers() == [1, 0]
+    assert cub.persistent_betti_numbers(0, 1000) == [0, 0]
 
 def test_simple_constructor_from_top_cells():
     cub = CubicalComplex(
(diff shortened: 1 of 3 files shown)


src/python/test/test_cubical_complex.py
```diff
@@ -152,8 +152,8 @@ def test_dimension_file_constructor():
     assert cub.__is_persistence_defined() == False
     assert cub.persistence() == [(1, (0.0, 100.0)), (0, (0.0, float("inf")))]
     assert cub.__is_persistence_defined() == True
-    assert cub.betti_numbers() == [1, 0, 0]
-    assert cub.persistent_betti_numbers(0, 1000) == [1, 0, 0]
+    assert cub.betti_numbers() == [1, 0]
+    assert cub.persistent_betti_numbers(0, 1000) == [1, 0]
```
mglisse, line 156: Ah, yes, I forgot it would have this kind of consequence. If we wanted to, we could just append a `0` to keep the same output. I don't know what is better...

mglisse:
Thinking about it some more, all this really changes is whether there is an extra 0 in Betti numbers (there are no positive d-cells, and dim_max is only checked for positive simplices). If we follow Hannah's suggestion to keep the extra 0 (she is the only one who answered), we might as well not apply the patch at all. Sorry for suggesting it in the first place.

Paradoxically, the patch would be a tiny bit more useful for the fully periodic case, where there is one positive d-cell, involved in a huge cycle that contains all d-cells. However, 1) this would require quite a few fixes (adding one infinite interval, etc), 2) we wouldn't benefit much: we would save one allocation for the cocycle, but we would not save the costly reduction to prove that this last square is positive. So let's forget it.

VincentRouvreau:
Ok, forget this suggestion

PR #896, opened 2023-06-19
ignore simplices of infinite filtration values during persistence computation

DavidLapous:
**Changes:** 
Skips the simplices with infinite filtration value in the persistence computations. 

**Usecase:**
When computing persistence of different subcomplexes of the same simplicial complex, it is inefficient to recreate the simplextree for every computation, so setting values to infinity with this patch allows to skip the simplicies that are not in the subcomplex, avoiding some persistence computation.
Typically, computing the homology of the restriction to an horizontal line in 2-parameter persistence can be sped up with such strategy.
As far as I'm aware, homology at infinity is not taken into account, so this will not affect the results.

### DIFF
--- a/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
+++ b/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
@@ -115,6 +115,7 @@ class Persistent_cohomology {
     }
     Simplex_key idx_fil = 0;
     for (auto sh : cpx_->filtration_simplex_range()) {
+      if (cpx_->filtration(sh) == std::numeric_limits<Filtration_value>::infinity()) break;
       cpx_->assign_key(sh, idx_fil);
       ++idx_fil;
       dsets_.make_set(cpx_->key(sh));
@@ -175,6 +176,7 @@ class Persistent_cohomology {
     // Compute all finite intervals
     for (auto sh : cpx_->filtration_simplex_range()) {
       int dim_simplex = cpx_->dimension(sh);
+      if (cpx_->filtration(sh) == std::numeric_limits<Filtration_value>::infinity()) break;
       switch (dim_simplex) {
         case 0:
           break;
@@ -189,6 +191,7 @@ class Persistent_cohomology {
     // Compute infinite intervals of dimension 0
     Simplex_key key;
     for (auto v_sh : cpx_->skeleton_simplex_range(0)) {  // for all 0-dimensional simplices
+    if (cpx_->filtration(v_sh) == std::numeric_limits<Filtration_value>::infinity()) continue;
       key = cpx_->key(v_sh);
 
       if (ds_parent_[key] == key  // root of its tree


mglisse:
I am of two minds about this.
1. On the one hand, this is clearly the wrong place for this test. If we consider that these cells are not in the filtration, we should make it so they are not output by `filtration_simplex_range` (nor by `skeleton_simplex_range`).
2. On the other hand, this patch might not cause too much trouble. I think we should test `std::numeric_limits<Filtration_value>::has_infinity && ...` to avoid trouble with a Filtration type without infinity (infinity() would likely return 0 there...). I can't easily come up with a use case for infinite values that we don't want to ignore. And this test should not be too expensive. I ran the `line.cpp` benchmark to compute 0-dim persistence with no infinite value (a priori the case that should suffer most), and this patch seems to slow down calls to `compute_persistent_cohomology()` by something like 4% (with 1 compiler on 1 platform, not reliable for such a small difference).

@VincentRouvreau, @hschreiber, others, do you have a strong opinion about this?

hschreiber:
> I am of two minds about this.
> 
>     1. On the one hand, this is clearly the wrong place for this test. If we consider that these cells are not in the filtration, we should make it so they are not output by `filtration_simplex_range` (nor by `skeleton_simplex_range`).
> 
>     2. On the other hand, this patch might not cause too much trouble. I think we should test `std::numeric_limits<Filtration_value>::has_infinity && ...` to avoid trouble with a Filtration type without infinity (infinity() would likely return 0 there...). I can't easily come up with a use case for infinite values that we don't want to ignore. And this test should not be too expensive. I ran the `line.cpp` benchmark to compute 0-dim persistence with no infinite value (a priori the case that should suffer most), and this patch seems to slow down calls to `compute_persistent_cohomology()` by something like 4% (with 1 compiler on 1 platform, not reliable for such a small difference).
> 
> 
> @VincentRouvreau, @hschreiber, others, do you have a strong opinion about this?

I see why it would make sense to not output simplices at infinity in `filtration_simplex_range`, but at the same time I don't find it impossible that someone would still want to iterate over them at the end of the filtration. The need of infinity as filtration value is very particular in the first place. On the other side, ignoring them from the beginning would make `initialize_filtration` faster if more then half of the simplices are at infinity. But I am not sure if that is the use case.

VincentRouvreau:
Here is what I have with this modification on `line` benchmark:
```txt
* master                                         * ignore_infinite_persistence

$ ./line 10000000                                $ ./line 10000000                 
Mostly unique values                             Mostly unique values                    
Cubical complex from top cells: 6.134s           Cubical complex from top cells: 6.196s                                      
                                                 
Cubical complex from vertices: 6.49s             Cubical complex from vertices: 6.696s                                    
                                                 
Simplex tree: 7.001s                             Simplex tree: 7.13s                    
                                                 
Special case for function on a line: 0.075s      Special case for function on a line: 0.076s                                           
                                                 
Many repeated values                             Many repeated values                    
Cubical complex from top cells: 2.997s           Cubical complex from top cells: 3.018s                                      
                                                 
Cubical complex from vertices: 2.348s            Cubical complex from vertices: 2.458s                                     
                                                 
Simplex tree: 5.364s                             Simplex tree: 5.339s                    
                                                 
Special case for function on a line: 0.073s      Special case for function on a line: 0.071s                                           
```

DavidLapous:
What about putting the `if` here
```c++
  void initialize_filtration(bool ignore_infinite_values = false) {
    filtration_vect_.clear();
    filtration_vect_.reserve(num_simplices());
    for (Simplex_handle sh : complex_simplex_range()){
      if (ignore_infinite_values && filtration(sh) == std::numeric_limits<Filtration_value>::infinity()) continue;
      filtration_vect_.push_back(sh);
    }
```
in `Simplex_tree.h` line 934 ? and do the same for the skeleton iterator (I'm not familiar with this code though) ?
I don't mind dealing with this flag in C++.
 - the sort in `initialize_filtration` should be way faster in the scenarii where a lot of simplices are at infinity
 - but we do more `if` but less `push_back`s if there is a lot of things at infinity (no more `break`s)

Otherwise, a maybe cleaner way would be to do this`if` in a new simplex iterator (or in `complex_simplex_range` with a flag), which would ignore each simplex **and its cofaces** if this simplex is at infinity (should be possible as the simplextree is a tree).

mglisse:
> which would ignore each simplex **and its cofaces** if this simplex is at infinity (should be possible as the simplextree is a tree).

Ignoring the children of a simplex (a subset of its cofaces) is easy, but for other cofaces, it is easier to check if their filtration value is infinite than to look at their children.


From a C++ viewpoint, I think a natural way to deal with infinite cells would be to define a (placeholder name) filtered_simplex_tree (like we have filtered_range to iterate on a subset of the elements of a range). That would be a class containing a pointer to a real simplex tree, and with methods where some just forward to the simplex tree, and some are reimplemented to pretend that simplices with an infinite filtration value are not part of the complex. And then we would use Persistent_cohomology on this wrapper instead of the original complex.

Without the wrapper, handling it in initialize_filtration looks ok to me. However, pcoh expects skeleton_simplex_range to be consistent with filtration, so we would also need to do that one (without parameter since pcoh calls it without parameter), and either the structure starts looking inconsistent (skeleton(dim=100) != all-simplices), or we do it for all iterators but then we cannot see those infinite cells anymore and it becomes messy to work with this complex.

One possible solution would be to remove the call to skeleton_simplex_range in pcoh (and then handle just initialize_filtration like you propose). It is only used to iterate on vertices at the end (to find the connected components), but we have already iterated on those before, so we could for instance store them in a vector in pcoh. It would waste a bit of space, but it shouldn't be that bad, and for the cubical complex it would likely be faster than the current iterator.
@hschreiber , since you are looking at the persistence code, would this make sense to you?

hschreiber:
> However, pcoh expects skeleton_simplex_range to be consistent with filtration

More precisely, pcoh need that the keys while traversing `skeleton_simplex_range` to be consistent. If `filtration_simplex_range` ignores infinite values, no key is assigned to them. But this is not a problem, if we just add a condition to the `if` in the `skeleton_simplex_range` loop, testing if the filtration value is inf. We just need to be put this condition first to avoid segmentation fault when calling `ds_parent_[key] == key`. I don't think this would add much overhead to the loop. 

Storing the (keys of) vertices in a vector is probably also not a bad idea. In any case, I don't think that `skeleton_simplex_range` should ignore values at infinity, as the range has nothing to do with filtration values.

> From a C++ viewpoint, I think a natural way to deal with infinite cells would be to define a (placeholder name) filtered_simplex_tree

Honestly, I always thought that we should separate the "filtered" from the simplex tree and do exactly that. Those are two different functionalities and from a developer point of view, the simplex tree class would become much more readable. It would also make it easier to add new variants, like the multi-parameter simplex tree for multi-parameter persistence or the Morse simplex tree for zigzag.

mglisse:
My proposed resolution would thus be

- Short term
  1. I do a pull request removing skeleton_simplex_range from pcoh
  2. David does a pull request to add a parameter to initialize_filtration
- Long term
  1. We investigate various refactoring possibilities for the Simplex_tree (I'll open an issue)

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -172,11 +166,17 @@ class Persistent_cohomology {
...
   void compute_persistent_cohomology(Filtration_value min_interval_length = 0) {
     interval_length_policy.set_length(min_interval_length);
+    Simplex_key idx_fil = -1;
+    std::vector<Simplex_key> vertices; // so we can check the connected components at the end
     // Compute all finite intervals
     for (auto sh : cpx_->filtration_simplex_range()) {
+      cpx_->assign_key(sh, ++idx_fil);
+      dsets_.make_set(cpx_->key(sh));
       int dim_simplex = cpx_->dimension(sh);
+      if (cpx_->filtration(sh) == std::numeric_limits<Filtration_value>::infinity()) break;
```
VincentRouvreau, line 176: As @mglisse was suggesting:
```suggestion
      if (std::numeric_limits<Filtration_value>::has_infinity &&
          cpx_->filtration(sh) == std::numeric_limits<Filtration_value>::infinity()) break;
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1012,11 +1012,15 @@ class Simplex_tree {
    *
    * Any insertion, deletion or change of filtration value invalidates this cache,
    * which can be cleared with clear_filtration().  */
-  void initialize_filtration() {
+  void initialize_filtration(bool ignore_infinite_values = true) {
```
mglisse, line 1015: Shouldn't the default be `false`, for consistency with the current behavior?

DavidLapous: Yes indeed, the simplex iterator by filtration values will be affected by this. Furthermore the majority of the filtrations used have no infinite values in general (Rips, alpha, cubical,...), I also think setting default value to `false` is better.

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -173,6 +173,8 @@ class Persistent_cohomology {
       cpx_->assign_key(sh, ++idx_fil);
       dsets_.make_set(cpx_->key(sh));
       int dim_simplex = cpx_->dimension(sh);
+      if (std::numeric_limits<Filtration_value>::has_infinity &&
+          cpx_->filtration(sh) == std::numeric_limits<Filtration_value>::infinity()) break;
```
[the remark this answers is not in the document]
DavidLapous: Indeed, this is no longer needed, thanks Marc.

mglisse:
> Or should we try passing something more general than a boolean for infinity, maybe a threshold (keep only simplices with filtration lower than X), or even a callback (we can have something less generic in the python interface)? Or is that overkill and likely useless?

In my opinion this is the main question that needs to be answered. If we consider that only infinity is worth ignoring, then I am ok with the current PR (although some details in the doc above the function clarifying the consequences of this option for other functions would make it even better).

DavidLapous:
I personally only need this version. 
For the threshold version, we would also need a boolean argument to state if the threshold is included or not (especially for the infinite threshold). I think the callback version is very dangerous; but if someone has a use-case, why not.

VincentRouvreau (approved):

VincentRouvreau:
I think we can merge it as is and see later if there are other needs

PR #903, opened 2023-07-03
Direct Python interface for edge collapse

mglisse:
Edge collapse doesn't use the SimplexTree at all, so it makes sense to have a separate interface. Mostly I wanted a wrapper so we could (at some point in the future, not in the current branch) use it in RipsPersistence, but I thought I might as well make some version public.
The pybind11 version doesn't directly have the coo_matrix interface, because I am not convinced that it is the most convenient. For instance, torch.sparse_coo_tensor wants a single tensor for i+j (what I chose to output).
There didn't seem to be a great place to put the doc, I first thought of appending it to the ref manual of SimplexTree, but eventually put it separately, with a link from the persistent cohomology section of the index.
The name gudhi.collapse_edges may be a bit too generic, but it should be enough to get the discussion started.
The submodule is imported by default, which goes against what I tend to recommend in other reviews :upside_down_face: I would be ok requiring `from gudhi.edge_collapse import collapse_edges` from users if that feels cleaner.


src/python/gudhi/_edge_collapse.cc
```diff
@@ -0,0 +1,66 @@
...
+  if (buff.shape[0] == 0) {
+    py::array_t<Index> indices({{ 2, 0 }}, {{ 0, 0 }});
+    py::array_t<Filtr> filtrs;
+    return py::make_tuple(std::move(indices), std::move(filtrs));
+  }
+  auto& edges = *new Edges();
+  {
+    py::gil_scoped_release release;
+    edges.reserve(bufi.shape[0]);
+    Index n_edges = static_cast<Index>(bufi.shape[0]);
```
VincentRouvreau, line 42: I don't think it costs to call `bufi.shape[0]`, but it could be:
```suggestion
    Index n_edges = static_cast<Index>(bufi.shape[0]);
    edges.reserve(n_edges);
```

src/python/gudhi/_edge_collapse.cc
```diff
@@ -0,0 +1,66 @@
...
+ *
+ *    Copyright (C) 2023 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#include <vector>
+#include <utility>
+
```
VincentRouvreau, line 13: Missing include `<stdexcept>` for exceptions management

VincentRouvreau (commented):
Some code I wrote to test/benchmark the feature. I would have thought the coo_matrix conversion would have been more costly, but no !

```python
import gudhi as gd
import numpy as np
from scipy.sparse import coo_matrix

off_file = gd.__root_source_dir__ + '/data/points/tore3D_1307.off'
points = gd.read_points_from_off_file(off_file = off_file)
stree = gd.RipsComplex(points = points).create_simplex_tree(max_dimension=1)

edges = [ [edge[0][0], edge[0][1], edge[1]] for edge in stree.get_skeleton(1) if len(edge[0]) == 2 ]
edges = list(map(list, zip(*edges)))

mat = coo_matrix((edges[2], (edges[0], edges[1])))

start = timer(); res = gd.collapse_edges(mat); end = timer(); print(timedelta(seconds=end-start))
# 0:00:05.368246

res.shape
# (1306, 1307)
len(res.data)
# 94993
mat.shape
# (1306, 1307)
len(mat.data)
# 853471

print(stree.num_simplices()-stree.num_vertices())
# 853471 -> OK

start = timer(); stree.collapse_edges(); end = timer(); print(timedelta(seconds=end-start))
# 0:00:05.278990

print(stree.num_simplices()-stree.num_vertices())
# 94993 -> OK
```

src/python/gudhi/edge_collapse.py
```diff
@@ -0,0 +1,35 @@
...
+
+    :param input_edges: Input weighted graph.
+    :type input_edges: scipy.sparse.coo_matrix
+    :param nb_iterations: The number of times we apply the algorithm. Default is 1.
+    :type nb_iterations: int
+    :rtype: scipy.sparse.coo_matrix
+    """
+    from scipy.sparse import coo_matrix
+
+    r = _collapse_edges(input_edges.row, input_edges.col, input_edges.data)
```
VincentRouvreau, line 34: `nb_iterations` is not propagated to `_collapse_edges` call

src/python/doc/persistent_cohomology_sum.inc
```diff
@@ -21,4 +21,5 @@
    |                                                                 | * :doc:`simplex_tree_ref`                                                                                             |
    |                                                                 | * :doc:`cubical_complex_ref`                                                                                          |
    |                                                                 | * :doc:`periodic_cubical_complex_ref`                                                                                 |
+   |                                                                 | * :doc:`edge_collapse`                                                                                                |
```
VincentRouvreau, line 24: I am not sure of where we should mention this new module. In C++, we have:

image

To do something equivalent for the python version, you can add in `src/python/doc/index.rst`, between Rips and Witness complex:
```rst

Edge collapse
=============

.. table::
   :widths: 30 40 30

   +-----------------------------------------------------------------+---------------------------------------------------------------------+-----------------------------------------------+
   | .. figure::                                                     | Edge collapse is able to reduce any flag filtration to a smaller    | :Author: Siddharth Pritam, Marc Glisse        |
   |      ../../doc/Collapse/dominated_edge.png                      | flag filtration with the same persistence, using only the           |                                               |
   |      :figclass: align-center                                    | 1-skeletons of a simplicial complex.                                | :Since: GUDHI 3.9.0                           |                                         
   |                                                                 | The reduction is exact and the persistence homology of the reduced  |                                               |
   |                                                                 | sequence is identical to the persistence homology of the input      |                                               |
   |                                                                 | sequence. The resulting method is simple and extremely efficient.   |                                               |
   |                                                                 |                                                                     |                                               |
   |                                                                 | Computation of edge collapse and persistent homology of a filtered  |                                               |
   |                                                                 | flag complex via edge collapse as described in                      |                                               |
   |                                                                 | :cite:`edgecollapsearxiv`.                                          |                                               |
   |                                                                 |                                                                     |                                               |
   +-----------------------------------------------------------------+---------------------------------------------------------------------+-----------------------------------------------+
   |                                                                 |  * :doc:`edge_collapse`                                                                                             |
   +-----------------------------------------------------------------+---------------------------------------------------------------------------------------------------------------------+
```

mglisse: I moved it where you said. But (also in C++) it is a strange place. ("Basic operation: contraction" is also in a strange place)
Bah, we can always move it later...

VincentRouvreau (approved):

PR #904, opened 2023-07-03
Fix address sanitizer for Simplex tree serialization UT

VincentRouvreau:


mglisse (approved):
It would also have been possible to do `char buffer[256];` or use a unique_ptr, but it doesn't matter.

PR #910, opened 2023-07-07
Disable warning for self move test

mglisse:
If I try to disable it for any version of gcc, gcc-12 prints
> warning: unknown option after ‘#pragma GCC diagnostic’ kind [-Wpragmas]

### DIFF
--- a/src/Simplex_tree/test/simplex_tree_ctor_and_move_unit_test.cpp
+++ b/src/Simplex_tree/test/simplex_tree_ctor_and_move_unit_test.cpp
@@ -151,7 +151,14 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(simplex_copy_constructor, Simplex_tree, list_of_te
   BOOST_CHECK(st == st8);
   BOOST_CHECK(st7 == st);
 
+#if __GNUC__ >= 13
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wself-move"
+#endif
   st = std::move(st);
+#if __GNUC__ >= 13
+#pragma GCC diagnostic pop
+#endif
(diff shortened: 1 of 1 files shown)


src/Simplex_tree/test/simplex_tree_ctor_and_move_unit_test.cpp
```diff
@@ -151,7 +151,14 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(simplex_copy_constructor, Simplex_tree, list_of_te
...
   BOOST_CHECK(st7 == st);
 
+#if __GNUC__ >= 13
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wself-move"
+#endif
   st = std::move(st);
+#if __GNUC__ >= 13
+#pragma GCC diagnostic pop
+#endif
```
VincentRouvreau, line 161: Should we consider clang case ?
```cpp
#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wself-move"
#endif
#if __GNUC__ >= 13
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wself-move"
#endif
  st = std::move(st);
#if __GNUC__ >= 13
#pragma GCC diagnostic pop
#endif
#ifdef __clang__
#pragma clang diagnostic pop
#endif
```

VincentRouvreau (approved):

src/Simplex_tree/test/simplex_tree_ctor_and_move_unit_test.cpp
mglisse: IIRC clang was not giving me any warning there. Someone who does get the warning can handle that later.

PR #919, opened 2023-07-23
Minimum enclosing ball filtration values

mglisse:
Another use case for #912. In particular, if you give it a Delaunay triangulation, you get the filtration values of the Delaunay-Cech (squared).
In some cases we may want to compute it only up to some dimension, but I expect we would want to `prune_above_dimension` in that case, so the new function doesn't need to handle it explicitly.

@VincentRouvreau Either after this branch is merged or during the review, it would be good to make some comparison between alpha-complex and Delaunay-Cech (i.e. computing the Delaunay triangulation and calling the new function to assign filtration values).
- the 2 should have the same persistence diagram up to rounding errors (a good test that things are working properly)
- benchmark: IIRC in the alpha-complex most of the time is spent in geometric computations for the filtration values. It would be interesting to see how the running-time of the Delaunay-Cech compares.

Doc: https://output.circle-artifacts.com/output/job/3c348709-ec0c-46c2-81f7-5ef27a5cc918/artifacts/0/doxygen/group__cech__complex.html

### DIFF
--- a/src/Cech_complex/include/gudhi/MEB_filtration.h
+++ b/src/Cech_complex/include/gudhi/MEB_filtration.h
@@ -16,6 +16,7 @@ namespace Gudhi::cech_complex {
 /**
  * \ingroup cech_complex
  *
+ * \brief
  * Given a simplicial complex and an embedding of its vertices, this assigns to
  * each simplex a filtration value equal to the squared radius of its minimal
  * enclosing ball.
@@ -23,11 +24,11 @@ namespace Gudhi::cech_complex {
  * Applied on a Cech complex, it recomputes the same values (squared). Applied on a Delaunay triangulation, it computes the Delaunay-Cech filtration.
  *
  * \tparam Kernel CGAL kernel: either Epick_d or Epeck_d.
- * \tparam SimplicialComplexForMEB Simplex_tree with Simplex_key an integer type large enough to store the number of simplices.
  * \tparam PointRange Random access range of `Kernel::Point_d`.
  *
- * @param[in] points Embedding of the vertices of the complex.
+ * @param[in] k The geometric kernel.
  * @param[in] complex The simplicial complex.
+ * @param[in] points Embedding of the vertices of the complex.
  * @param[in] exact If true and `Kernel` is <a href="https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epeck__d.html">CGAL::Epeck_d</a>, the filtration values are computed exactly. Default is false.
  */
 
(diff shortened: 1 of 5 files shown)


src/Cech_complex/include/gudhi/MEB_filtration.h
```diff
@@ -0,0 +1,112 @@
...
+
+namespace Gudhi::cech_complex {
+
+/**
+ * \ingroup cech_complex
+ *
+ * \brief
+ * Given a simplicial complex and an embedding of its vertices, this assigns to
+ * each simplex a filtration value equal to the squared radius of its minimal
+ * enclosing ball.
```
VincentRouvreau, line 22: Here we could explain what the acrnym MEB stands for:
```suggestion
 * enclosing ball (MEB).
```

src/Cech_complex/include/gudhi/MEB_filtration.h
```diff
@@ -0,0 +1,112 @@
...
+      auto verts = complex.simplex_vertex_range(sh);
+      auto vert_it = verts.begin();
+      Vertex_handle u = *vert_it;
+      Vertex_handle v = *++vert_it;
+      auto&& pu = points[u];
+      Point_d m = k.midpoint_d_object()(pu, points[v]);
+      FT r = k.squared_distance_d_object()(m, pu);
+      if (exact) CGAL::exact(r);
+      complex.assign_key(sh, cache_.size());
+      complex.assign_filtration(sh, std::max(cvt(r), Filtration_value(0)));
```
VincentRouvreau, line 63: IIUC, this means that `squared_distance_d_object` can return negative values ?

VincentRouvreau (commented):
I don't know where, but I think we should mention/explain that one can `prune_above_dimension` to avoid some computations.
As it is more related to the use of the Simplex_tree, I am not sure.

src/Cech_complex/include/gudhi/MEB_filtration.h
mglisse: With exact computations, no, but then we wouldn't need a lot of other code (see [1488c9b](https://github.com/GUDHI/gudhi-devel/pull/919/commits/1488c9bed79a0371cff48f12f51db0ea8a6ac840)).
Without rounding errors... it still seem very unlikely to get a negative result, but better safe than sorry, especially since I expect this to take a negligible amount of time.

src/Cech_complex/include/gudhi/MEB_filtration.h
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * \ingroup cech_complex
+ *
+ * \brief
+ * Given a simplicial complex and an embedding of its vertices, this assigns to
+ * each simplex a filtration value equal to the squared radius of its minimal
+ * enclosing ball.
+ *
+ * Applied on a Cech complex, it recomputes the same values (squared). Applied on a Delaunay triangulation, it computes the Delaunay-Cech filtration.
```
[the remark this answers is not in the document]
mglisse: Delaunay that derives from Alpha sounds strange, if anything I would expect the reverse (although if it is private inheritance, it could be ok).
Sure, you can open an issue about improving the visibility of Delaunay (through better doc or separate API or ...).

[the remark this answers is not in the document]
VincentRouvreau: cf. #967

VincentRouvreau (approved):

PR #920, opened 2023-07-26
Ripser Python integration

mglisse:
This makes RipsPersistence use a modified version of Ripser.
There is no public C++ interface for now, the main goal was to speed up the Python code.
I am providing a utility "ripser" which is on purpose equivalent to upstream's ripser-coeff, to make comparisons easier. I expect packagers prefix all program names with "gudhi-" so it should not conflict with the official ripser, but I now realize that CMake doesn't make this easy, Debian does it by hand for each exe :disappointed: 
I refactored the code a bit (though not quite as much as I wanted) in the process of understanding it and turning it into library code.
I kept some type names like value_t and coefficient_t for compatibility with Ripser (make it easy to import future changes in Ripser, or code from other branches of Ripser).
@VincentRouvreau , could you help add a few tests, in particular to check that Ripser and SimplexTree compute (almost) the same thing?


src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -73,36 +81,90 @@ def fit(self, X, Y=None):
...
+        num_collapses = self.num_collapses
+        input_type = self.input_type
+        threshold = self.threshold
+        n = inp.shape[0] if input_type == 'distance coo_matrix' else len(inp)
+        max_dimension = min(max(self.dim_list_), max(0, n-3))
+        # Ripser needs to encode simplices and coefficients in 128 bits, which may not always fit
+        # Instead of a 256 bit version which may not always suffice either, fall back to SimplexTree
+        use_simplex_tree = math.comb(n, min(n // 2, max_dimension + 2)) >= (1 << (128 - (self.homology_coeff_field - 2).bit_length()))
+        if num_collapses == 'auto':
+            num_collapses = 1 if max_dimension > (not use_simplex_tree) else 0
```
DavidLapous, line 96:
```suggestion
        if num_collapses is not None:
            num_collapses = 1 if max_dimension > (not use_simplex_tree) else 0
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -38,7 +44,7 @@ def __init__(
         homology_dimensions,
         threshold=float('inf'),
         input_type='point cloud',
-        num_collapses=1,
+        num_collapses='auto',
```
DavidLapous, line 47:
```suggestion
        num_collapses:Optional[int]=None,
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -49,12 +55,14 @@ def __init__(
...
-            num_collapses (int): Specify the number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform
-                on the SimplexTree. Default value is 1 (a relatively good enough number of iterations).
+            threshold (float): Rips maximal edge length value. Default is +Inf. Ignored if input_type is 'distance coo_matrix'.
+            input_type (str): Can be 'point cloud' when inputs are point clouds, 'full distance matrix',
+                'lower distance matrix' when inputs are lower triangular distance matrix (can be full square,
+                but the upper part will not be considered), or 'distance coo_matrix' for a distance matrix in SciPy's
+                sparse format, which should contain each edge at most once (avoid the symmetric) and no diagonal entry.
+                Default is 'point cloud'.
+            num_collapses (int|str): Specify the number of iterations of :func:`~gudhi.flag_filtration.edge_collapse.reduce_graph`
+                (edge collapse) to perform on the graph. Default value is 'auto'.
```
DavidLapous, line 65:
```suggestion
                (edge collapse) to perform on the graph. Default value is `None`.
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -38,7 +44,7 @@ def __init__(
         homology_dimensions,
         threshold=float('inf'),
         input_type='point cloud',
```
DavidLapous, line 46:
```suggestion
        input_type:Literal["point cloud", "full distance matrix", "lower distance matrix", "distance coo_matrix"]='point cloud',
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -38,7 +44,7 @@ def __init__(
         homology_dimensions,
         threshold=float('inf'),
```
DavidLapous, line 45:
```suggestion
        threshold:float=float('inf'),
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -38,7 +44,7 @@ def __init__(
         homology_dimensions,
```
DavidLapous, line 44:
```suggestion
        homology_dimensions:Union[int,Iterable[int]],
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -7,8 +7,14 @@
 # Modification(s):
 #   - YYYY/MM Author: Description of the modification
```
DavidLapous, line 9:
```suggestion

from typing import Union,Iterable,Literal,Optional
```

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -49,12 +55,14 @@ def __init__(
...
-                matrix will not be considered). Default is 'point cloud'.
-            num_collapses (int): Specify the number of :func:`~gudhi.SimplexTree.collapse_edges` iterations to perform
-                on the SimplexTree. Default value is 1 (a relatively good enough number of iterations).
+            threshold (float): Rips maximal edge length value. Default is +Inf. Ignored if input_type is 'distance coo_matrix'.
+            input_type (str): Can be 'point cloud' when inputs are point clouds, 'full distance matrix',
+                'lower distance matrix' when inputs are lower triangular distance matrix (can be full square,
+                but the upper part will not be considered), or 'distance coo_matrix' for a distance matrix in SciPy's
+                sparse format, which should contain each edge at most once (avoid the symmetric) and no diagonal entry.
+                Default is 'point cloud'.
+            num_collapses (int|str): Specify the number of iterations of :func:`~gudhi.flag_filtration.edge_collapse.reduce_graph`
```
DavidLapous, line 64:
```suggestion
            num_collapses (int|None): Specify the number of iterations of :func:`~gudhi.flag_filtration.edge_collapse.reduce_graph`
```

DavidLapous (commented):
I didn't check that everything works, but I added some typing to the functions' signature. 
According to [the doc](https://docs.python.org/3.8/library/typing.html) these functions exist in py3.8. I think we should do it in general, as it greatly helps by showing warning + autocompletions (espetially for the `input_type`) in editors.
For `num_collapses`, I morally prefer the 'auto' keyword, but it is easier to use `Optional[int]` in practice. We can replace it by `Union[int,Literal["auto"]]` if we prefer the morale.

mglisse:
> For `num_collapses`, I morally prefer the 'auto' keyword, but it is easier to use `Optional[int]` in practice. We can replace it by `Union[int,Literal["auto"]]` if we prefer the morale.

Normally I'd be all in favor of `None` (the default choice for a magic value in Python), but I find that hearing "how many collapses do you want? None.", and having it mean that I **do** want collapses, is really confusing. But if multiple people vote for `None` (current count: 1), I'll change it. The string also has the advantage that we could add other named strategies later ('max'?), though I am not sure we'll ever really want that.

Thanks for the typing, I'll look at it after the holidays.

mglisse:
I added some type hints.
Before adding more of them in Gudhi, we need to look at the interaction with sphinx. Indicating the type twice is obviously not good, but getting the same output from sphinx after removing the types from the docstring seems non-trivial (some options and plugins claim to do that, but they failed in a very quick test).

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -73,36 +84,90 @@ def fit(self, X, Y=None):
...
+                i, j, f = _lower_to_coo(inp, threshold)
+                inp = coo_matrix((f, (i, j)), shape=(n,) * 2)
+                input_type = 'distance coo_matrix'
+
+        if num_collapses > 0:
+            assert input_type == 'distance coo_matrix'
+            inp = reduce_graph(inp, num_collapses)
+
+        if use_simplex_tree:
+            from gudhi import SimplexTree
```
VincentRouvreau, line 146:
```python
from .. import SimplexTree
```
and at the top of the file seems to be more _pythonic_

src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -73,36 +84,90 @@ def fit(self, X, Y=None):
...
+            #     inp = inp.T
+            dgm = _full(inp, max_dimension=max_dimension, max_edge_length=threshold, homology_coeff_field=self.homology_coeff_field)
+        elif input_type == 'lower distance matrix':
+            dgm = _lower(inp, max_dimension=max_dimension, max_edge_length=threshold, homology_coeff_field=self.homology_coeff_field)
+        elif input_type == 'distance coo_matrix':
+            # Switch to coo_array (danger: row/col seem deprecated)?
+            dgm = _sparse(inp.row, inp.col, inp.data, inp.shape[0], max_dimension=max_dimension, max_edge_length=threshold, homology_coeff_field=self.homology_coeff_field)
         else:
-            raise ValueError("Only 'point cloud' and  'lower distance matrix' are valid input_type")
+            raise ValueError("Only 'point cloud', 'lower distance matrix', 'full distance matrix' and 'distance coo_matrix' are valid input_type") # move to __init__?
```
VincentRouvreau, line 167: From a user point of view, I think it is better to catch exception at construction

src/common/include/gudhi/uint128.h
```diff
@@ -0,0 +1,147 @@
...
+  static constexpr bool is_exact = true;
+  static constexpr bool has_infinity = false;
+  static constexpr bool is_modulo = true;
+  static constexpr int digits = 128;
+  static constexpr int radix = 2;
+  // etc
+};
+#else
+namespace Gudhi::numbers {
+typedef unsigned __int128 uint128_t;
```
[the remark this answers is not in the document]
mglisse: ISTR that it was the first thing I tried but it was significantly worse, but I'll check again.
The current code is supposed to be compatible with 32 bit CPUs: `long long` has to be at least 64 bits in C++ (possibly through emulation).

src/python/gudhi/sklearn/rips_persistence.py
mglisse: For something like SimplexTree that is already imported at the top level anyway, indeed (I thought I had already moved it, but that was probably another import).

src/python/gudhi/sklearn/rips_persistence.py
mglisse: `input_type` could have been changed by `set_params`, so I don't think the constructor is sufficient, but I guess we could have the test in 2 places.

src/common/include/gudhi/uint128.h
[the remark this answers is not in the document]
mglisse: I checked again. On linux, `boost::multiprecision::uint128_t` delegates to `unsigned __int128` for storage and most operations. On some testcases, it has a bit of overhead (10%, mostly because of checks in `operator<<` and `operator>>`). On windows, there is no 128bit type, and boost implements it as a bignum, so it takes 50% more storage (it stores the size), and operations are generally slower. On some testcases it is not noticable, but I have a testcase (a bit pathological: asking for homology in dim 0-25 for a circle with 24 points) where it multiplies the running time by more than 4 compared to linux. Fake_uint128 has less than 10% overhead on the same testcase. Saying that we don't care about windows is tempting, but if we do care, a factor 4 in the running time seems worth it.

src/python/gudhi/sklearn/rips_persistence.py
mglisse: https://scikit-learn.org/stable/developers/develop.html#instantiation says
> To summarize, an `__init__` should look like: [...] There should be no logic, not even input validation,

I am ok with violating this recommendation if you think we should, but I wanted to get confirmation first.

VincentRouvreau: Ah... I was not aware of these recommendations, sorry about that. We should decide if we follow it or not and explain it in the [code_conventions](https://github.com/GUDHI/gudhi-devel/blob/master/.github/code_conventions.md). Not for this PR.

src/common/include/gudhi/uint128.h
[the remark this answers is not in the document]
VincentRouvreau: Ok, let's keep your implementation in this case, but if you could add some comments to explain why we don't go with `boost::multiprecision::uint128_t`, that would be appreciate, and we can remove it later if boost is doing some imrpovements on windows side.

src/python/gudhi/sklearn/rips_persistence.py
mglisse: We did discuss them a few times already :wink:
It led to some changes in the representations module.
I think we should *try* to follow it (if we don't, something may mysteriously fail deep in a pipeline), but not dogmatically:
* If we decide to do some argument validation in `__init__`, I don't really see how that can break anything.
* Currently we don't have a default value for homology_dimensions, and I am ok with that. If it causes problems, the error message is likely to be easy to understand, and we can consider adding a default later.

VincentRouvreau:
> > For `num_collapses`, I morally prefer the 'auto' keyword, but it is easier to use `Optional[int]` in practice. We can replace it by `Union[int,Literal["auto"]]` if we prefer the morale.
> 
> Normally I'd be all in favor of `None` (the default choice for a magic value in Python), but I find that hearing "how many collapses do you want? None.", and having it mean that I **do** want collapses, is really confusing. But if multiple people vote for `None` (current count: 1), I'll change it. The string also has the advantage that we could add other named strategies later ('max'?), though I am not sure we'll ever really want that.
> 
> Thanks for the typing, I'll look at it after the holidays.

And why not keep an int value, where `-1` would mean automatic number of collapses ? It makes me think of [joblib Parallel](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html), where
 > if `n_jobs=-1` is given, joblib tries to use all CPUs.

mglisse:
> And why not keep an int value, where `-1` would mean automatic number of collapses ? It makes me think of [joblib Parallel](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html), where
> 
> > if `n_jobs=-1` is given, joblib tries to use all CPUs.

Because the way I naturally interpret `num_collapses=-1` is: collapse the **maximum** number of times, collapse until it cannot be collapsed anymore. (which by the way is an option we might add in the future, under one name or another, but it requires some work to do it intelligently)

VincentRouvreau:
> @VincentRouvreau , could you help add a few tests, in particular to check that Ripser and SimplexTree compute (almost) the same thing?

I did so on [a branch ripser_test](https://github.com/VincentRouvreau/gudhi-devel/tree/ripser_test). If you need to do some comments on it, I could make a draft PR of it. Just tell me.

mglisse:
> I did so on [a branch ripser_test](https://github.com/VincentRouvreau/gudhi-devel/tree/ripser_test).

I integrated a modified version. This helped me notice a bug in the way I was computing the cone radius (it was safe but made it useless).

src/python/test/test_sklearn_rips_persistence.py
```diff
@@ -10,6 +10,12 @@
 
 from gudhi.datasets.generators import points
 from gudhi.sklearn.rips_persistence import RipsPersistence
+from gudhi import RipsComplex, SimplexTree
+from gudhi._ripser import _lower, _full, _sparse, _lower_to_coo, _lower_cone_radius
+from gudhi import bottleneck_distance
```
VincentRouvreau, line 15: If you import bottleneckdistance (and not the hera version), CGAL is required for this test (but not for the module).
Ah... forget this point, CGAL is already required because of `gudhi.datasets.generators`.

mglisse: Ah, it is good that you thought about it, because it didn't cross my mind at all.
Indeed, CGAL is already required for this particular test. Ideally it would be better if we didn't gratuitously require CGAL to test things that don't actually use CGAL, but I guess it is ok for a test.

VincentRouvreau:
> We could have predefined behaviors for `num_collapses=-1` (would mean auto) and for `num_collapses=sys.maxsize` (would mean collapse until it cannot be collapsed anymore). But maybe not pythonic way of doing it...

In order to take into account `auto` and `max` possibilities, I vote in favour of a string.

VincentRouvreau (approved):

mglisse:
Maybe I should have started from the fork in giotto-ph instead of the original, it is faster (without using threads) by a factor 6+ on the first example I try :disappointed: 
Well, TODO for later...

Issue #395, opened 2020-09-25
[Python - persistence_graphical_tools] numpy (N x 3) a input could be convenient


VincentRouvreau:
User reports they use numpy to load/save persistence diagrams with dimension information.

```python
np.loadtxt('./rips.pers')                                                                                                                                          

array([[2.       , 0.138335 ,       inf],
       [1.       , 0.104347 ,       inf],
       ...,

       [1.       , 0.11169  , 0.111695 ]])

```

A patch to make it work:
```python
def _array_handler(a):
    '''
    :param a: if array, assumes it is a (n x 2) or (n x 3) np.array and return a
                persistence-compatible list (padding with 0 or dimension), so that the
                plot can be performed seamlessly.
    '''
    if len(a[0]) == 3:
        if isinstance(a[0][2], np.float64) or isinstance(a[0][2], float):
            return [[int(x[0]), (x[1], x[2])] for x in a]
    if len(a[0]) == 2:
        if isinstance(a[0][1], np.float64) or isinstance(a[0][1], float):
            return [[0, x] for x in a]
    return a
```
Some documentation to be done.

mglisse:
3 coordinates: dim, birth, death? That seems similar to a format that giotto likes, it is worth checking if that matches.

Another interesting format, produced for instance by the scikit-tda interface to ripser, is a list of nx2 arrays, one array per dimension.

At some point, with too many formats, it may become dangerous to guess, although for now it still seems doable.

Eventually, we may want to deprecate our current List[Tuple[int,Tuple[float,float]]], which is less convenient than the alternatives.

PR #925, opened 2023-07-28
New format for persistence

mglisse:
(not ready: missing tests at least, and doing the same for cubical)
As a step towards using more numpy arrays, this provides an option to SimplexTree.persistence() to output a list of arrays (1 numpy array per dimension), instead of our current list of tuple (all together, with the dimension being part of the tuple).
I open the PR to start the discussion.

### DIFF
--- a/src/python/gudhi/simplex_tree.pyx
+++ b/src/python/gudhi/simplex_tree.pyx
@@ -612,7 +612,7 @@ cdef class SimplexTree:
         """
         self.get_ptr().expansion_with_blockers_callback(max_dim, callback, <void*>blocker_func)
 
-    def persistence(self, homology_coeff_field=11, min_persistence=0, persistence_dim_max = False):
+    def persistence(self, homology_coeff_field=11, min_persistence=0, persistence_dim_max = False, output_type = 'old'):
         """This function computes and returns the persistence of the simplicial complex.
 
         :param homology_coeff_field: The homology coefficient field. Must be a
@@ -627,11 +627,18 @@ cdef class SimplexTree:
             maximal dimension in the complex is computed. If false, it is
             ignored. Default is false.
         :type persistence_dim_max: bool
+        :param output_type: Format of the output. 'old' for the legacy list of (dim,(birth,death)),
+            'array by dimension' for a list of nx2 numpy arrays (one per dimension).
+        :type output_type: str
         :returns: The persistence of the simplicial complex.
         :rtype:  list of pairs(dimension, pair(birth, death))
         """
         self.compute_persistence(homology_coeff_field, min_persistence, persistence_dim_max)
-        return self.pcohptr.get_persistence()
+        if output_type == 'array by dimension':
+            v = self.pcohptr.intervals_by_dimension()
+            return [ np.asarray(dgm) for dgm in v ] # std::move(dgm)?
+        else:
+            return self.pcohptr.get_persistence()
 
     def compute_persistence(self, homology_coeff_field=11, min_persistence=0, persistence_dim_max = False):
         """This function computes the persistence of the simplicial complex, so it can be accessed through
--- a/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
+++ b/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
@@ -690,6 +690,20 @@ class Persistent_cohomology {
     return result;
   }
 
+  /** @brief Returns persistence intervals for each dimension.
+   * @return A vector of diagrams, one per dimension starting from 0, where each diagram is a vector of persistence intervals (birth and death).
+   */
+  std::vector<std::vector<std::pair<Filtration_value, Filtration_value>>>
+  intervals_by_dimension() {
+    std::vector<std::vector<std::pair<Filtration_value, Filtration_value>>> result;
+    result.resize(dim_max_);
+    for (auto && pair : persistent_pairs_) {
+      auto b = get<0>(pair);
+      result[cpx_->dimension(b)].emplace_back(cpx_->filtration(b), cpx_->filtration(get<1>(pair)));
+    }
+    return result;
+  }
+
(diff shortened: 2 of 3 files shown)


src/python/gudhi/simplex_tree.pyx
```diff
@@ -627,11 +627,18 @@ cdef class SimplexTree:
...
+            'array by dimension' for a list of nx2 numpy arrays (one per dimension).
+        :type output_type: str
         :returns: The persistence of the simplicial complex.
         :rtype:  list of pairs(dimension, pair(birth, death))
         """
         self.compute_persistence(homology_coeff_field, min_persistence, persistence_dim_max)
-        return self.pcohptr.get_persistence()
+        if output_type == 'array by dimension':
+            v = self.pcohptr.intervals_by_dimension()
+            return [ np.asarray(dgm) for dgm in v ] # std::move(dgm)?
```
mglisse, line 639: We should probably do the usual workaround for empty dgm to get a shape (0,2).

VincentRouvreau:
(this PR requires to merge master for compilation to pass)
```python
import gudhi
rips_complex = gudhi.RipsComplex(points=[[1, 1], [7, 0], [4, 6], [9, 6], [0, 14], [2, 19], [9, 17]],
                                 max_edge_length=12.0)

simplex_tree = rips_complex.create_simplex_tree(max_dimension=1)
simplex_tree.persistence(output_type='old')
# [(0, (0.0, inf)), (0, (0.0, 8.94427190999916)), (0, (0.0, 7.280109889280518)), (0, (0.0, 6.082762530298219)),
#  (0, (0.0, 5.830951894845301)), (0, (0.0, 5.385164807134504)), (0, (0.0, 5.0))]
simplex_tree.persistence(output_type='array by dimension')
# [array([[0.        , 5.        ],
#        [0.        , 5.38516481],
#        [0.        , 5.83095189],
#        [0.        , 6.08276253],
#        [0.        , 7.28010989],
#        [0.        , 8.94427191],
#        [0.        ,        inf]])]
```
This format is quite interesting and was initiated from discussion on #395

martinroyer:
I don't have much to add except that I am greatly in favor of this and already using the new format using the following simple piece of translator code, if it is any help to anyone / anywhere:
```
def _diag_to_list_by_dim_format(pdiagram, tda_max_dim=None):
    """ transforms list of Tuple(dimension, Tuple(x,y)) into list of ndarray[[x_0, y_0], [x_1, y_1], ...]
     for each dimension, so if Tuple(x,y) has dimension i it will be found in the ith ndarray of
     the resulting list, e.g.:
         [(2, (1.0577792405537423, 1.1003878733068035)),
          (0, (0.0, np.inf)),
          (0, (0.0, 1.0556057201636535)),
          (0, (0.0, 0.8756047102452433))]
     transforms into
         [array([[0.        ,        inf],
                 [0.        , 1.05560572],
                 [0.        , 0.87560471]]),
         array([], shape=(0, 2), dtype=float64),
         array([[1.05777924, 1.10038787]])].
    """
    max_dimension = max(dim for (dim, _) in pdiagram) if tda_max_dim is None else tda_max_dim
    by_dim_pdiagram = [np.array([_ for (dim, _) in pdiagram if dim == i]) for i in range(0, max_dimension + 1)]
    return [diag if len(diag) else np.empty(shape=[0, 2]) for diag in by_dim_pdiagram]
```

I also suggested future tests for diagram vectorizers in the new format there: https://github.com/GUDHI/gudhi-devel/pull/1017/files#diff-509c3dbe85dd6b515d6d3e33d0f6c074686dcb27367baa653eeedeeec534f1c8. As Marc hinted it allows for nice interactions with `sklearn.compose.ColumnTransformer` for instance.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -627,11 +627,18 @@ cdef class SimplexTree:
...
         :type persistence_dim_max: bool
+        :param output_type: Format of the output. 'old' for the legacy list of (dim,(birth,death)),
+            'array by dimension' for a list of nx2 numpy arrays (one per dimension).
+        :type output_type: str
         :returns: The persistence of the simplicial complex.
         :rtype:  list of pairs(dimension, pair(birth, death))
         """
         self.compute_persistence(homology_coeff_field, min_persistence, persistence_dim_max)
-        return self.pcohptr.get_persistence()
+        if output_type == 'array by dimension':
```
DavidLapous, line 637: "array by dimension" seems a bit long to me, is "arrays" ok ? and 'old' isn't very telling :  "tuples" ?
Also, should we raise a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) to change the default to the array one afterward ?

src/python/gudhi/simplex_tree.pyx
DavidLapous, line 639: Tuples are slightly faster than lists in python, but that's negligible here.

src/python/gudhi/simplex_tree.pyx
```diff
@@ -612,7 +612,7 @@ cdef class SimplexTree:
         """
         self.get_ptr().expansion_with_blockers_callback(max_dim, callback, <void*>blocker_func)
 
-    def persistence(self, homology_coeff_field=11, min_persistence=0, persistence_dim_max = False):
+    def persistence(self, homology_coeff_field=11, min_persistence=0, persistence_dim_max = False, output_type = 'old'):
```
DavidLapous, line 615: -- Not completely related to this PR, but I think we should do type annotations here, i.e., 
```python
from typing import Optional,Literal, Union
def persistence(self, homology_coeff_field:int=11, min_persistence:float=0., persistence_dim_max:Optional[int] = None, output_type:Literal['old','array by dimension'] = 'old') -> list:
    pass
```
especially to have auto-completion for long arguments such as "array by dimension". 
I also changed the `False` to a `None`.

PR #927, opened 2023-08-03
Cython3 compatibility

### LINKED ISSUE #779: Compatibility with cython3
(you can easily experiment with cython3 after `pip install -U --pre cython`)
Building the gudhi python plugin with cython 3.0.0a11, I get a number of errors like
> E       AttributeError: 'gudhi.simplex_tree.SimplexTree' object has no attribute '__is_defined'

Apparently cython changed the way it handles double underscores, and it now mangles those names in a way that breaks our use. If that works, I think we should switch to a single underscore, no need for the second one.

VincentRouvreau:
Fix #779 
Fix Cython warning

### DIFF
--- a/src/python/gudhi/simplex_tree.pxd
+++ b/src/python/gudhi/simplex_tree.pxd
@@ -48,7 +48,7 @@ cdef extern from "Simplex_tree_interface.h" namespace "Gudhi":
         Simplex_tree_interface_full_featured() nogil
         Simplex_tree_interface_full_featured(Simplex_tree_interface_full_featured&) nogil
         double simplex_filtration(vector[int] simplex) nogil
-        void assign_simplex_filtration(vector[int] simplex, double filtration) nogil except +
+        void assign_simplex_filtration(vector[int] simplex, double filtration) except + nogil
         void initialize_filtration() nogil
         int num_vertices() nogil
         int num_simplices() nogil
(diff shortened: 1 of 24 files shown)


src/python/gudhi/simplex_tree.pxd
```diff
@@ -81,8 +81,8 @@ cdef extern from "Simplex_tree_interface.h" namespace "Gudhi":
         Simplex_tree_skeleton_iterator get_skeleton_iterator_end(int dimension) nogil
         pair[Simplex_tree_boundary_iterator, Simplex_tree_boundary_iterator] get_boundary_iterators(vector[int] simplex) nogil except +
         # Expansion with blockers
-        ctypedef bool (*blocker_func_t)(vector[int], void *user_data)
-        void expansion_with_blockers_callback(int dimension, blocker_func_t user_func, void *user_data)
+        ctypedef bool (*blocker_func_t)(vector[int], void *user_data) except +
+        void expansion_with_blockers_callback(int dimension, blocker_func_t user_func, void *user_data) except +
```
mglisse, line 85: I wonder about the exceptions. I don't think the C++ code is likely to throw an exception. If an exception is thrown, it probably comes from the python callback. Would `except *` or `except +*` also work? What happens if we throw an exception from the callback?

mglisse (commented):
The exception stuff is the subtle part, the rest is ok.

src/python/gudhi/simplex_tree.pxd
mglisse: I think there would also be possibilities adding noexcept on the other side, but this is ok, this function is slow anyway. I filed #935 for the future.

mglisse (approved):

Issue #296, opened 2020-04-29
[python] warnings


mglisse:
I just discovered (yes, I should learn python properly...) https://docs.python.org/3/library/warnings.html . That seems like a good fit for `print("Warning: computing barycenter of empty diag set. Returns None")` in barycenter.py.

tlacombe:
Noticed. I will do a PR incorporating soon.

VincentRouvreau:
List of warnings candidates:
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/example/rips_complex_diagram_persistence_from_correlation_matrix_file_example.py#L46-L49
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/gudhi/representations/metrics.py#L172-L174
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/gudhi/representations/metrics.py#L179-L181
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/gudhi/representations/metrics.py#L297-L299
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/gudhi/representations/metrics.py#L420-L422
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/gudhi/wasserstein/wasserstein.py#L14-L17
* [ ] https://github.com/GUDHI/gudhi-devel/blob/6b5b1c0f5d4a1652e668f3ee076ec7fb1b220096/src/python/test/test_simplex_tree.py#L548-L553 (TBD if there is some added value to warn in a unitary test)

PR #929, opened 2023-08-04
Use warnings python package to warn user instead of print('warning:...')

VincentRouvreau:
Fix partially #296


mglisse:
Please rephrase the comment so github doesn't automatically close the issue. Most of our uses of `print` are wrong, although I didn't check which ones should be warnings and which ones errors (exceptions).

mglisse (approved):

PR #930, opened 2023-08-04
[Simplex tree] Stable simplex handles

hschreiber:
The computation of oscillating rips (as in PR #401) needs the simplex handles to remain valid after insertion and removal.
This requires another dictionnary type for the nodes. As it is less performant (the coface benchmarks show 0.3s instead of 0.01 for the optimized version for example), the changes are added as an option which is disabled by default.


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1913,9 +1939,22 @@ class Simplex_tree {
...
+   * Not optimal, if Options::simplex_handle_strong_validity == true. Requires that the node is not a copy.
+   */
   static Simplex_handle simplex_handle_from_node(Node& node) {
-    return (Simplex_handle)(boost::intrusive::get_parent_from_member<Dit_value_t>(&node, &Dit_value_t::second));
+    if constexpr (Options::simplex_handle_strong_validity) {
+      Siblings* children = node.children();
+      //verifies if node is a leaf
+      for (auto it = children->members().begin(); it != children->members().end(); ++it){
+        // WARNING: does not work if nodes were copied, but I don't see another way to verify if two leaf nodes are
+        // equal. And I guess that `get_parent_from_member` needs a real pointer to the flat_map element too anyway?
```
mglisse, line 1951: I am trying to understand what the problem is here. We have a container (tree or vector) of pairs. From a reference to the second element of the pair, we want an iterator on this pair. get_parent_from_member gets us a pointer to the pair. For the vector iterator, boost kindly provides an explicit constructor of the iterator from the pointer. For a tree, this constructor is missing? I would need to dig through the boost sources to see if there is a workaround. Ideally (?) we would change the coface intrusive list so it contains Simplex_handle instead of `Node*`, but I don't know how hard that is.
Even if we end up using slow code with find, having a linear search before find seems strange.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2183,6 +2222,41 @@ struct Simplex_tree_options_full_featured {
...
   static const bool link_nodes_by_label = false;
+  static const bool simplex_handle_strong_validity = false;
+};
+
+/** Model of SimplexTreeOptions, same as `Simplex_tree_options_full_featured`
+ *  but the possibility of much bigger keys and in particular of negative keys.
+ *
+ * Maximum number of simplices to compute persistence is <CODE>std::numeric_limits<std::int64_t>::max()</CODE>
+ * (way enough simplices). */
+struct Simplex_tree_options_wide_indexation {
```
mglisse, line 2233: All those new `Simplex_tree_options_*` don't really seem related to this PR. I am not convinced we need a ton of *default* options to choose from, I'd rather users derive from one of very few *defaults* and customize the property they want. Even for the purpose of testing, we can define an options class in the test file.

src/Simplex_tree/concept/SimplexTreeOptions.h
```diff
@@ -29,5 +29,7 @@ struct SimplexTreeOptions {
   static constexpr bool contiguous_vertices;
   /// If true, the lists of `Node` with same label are stored to enhance cofaces and stars access.
   static const bool link_nodes_by_label;
+  /// If true, Simplex_handle will not be invalidated after insertions or removals.
+  static const bool simplex_handle_strong_validity;
```
mglisse, line 33: I was wondering if the name of the option should contain "stable" (or stability), but since this is related to iterator invalidation, validity seems fine.

mglisse (commented):
Yes, this can be useful, even without zigzags.

src/Simplex_tree/concept/SimplexTreeOptions.h
hschreiber: We could simply call it "stable_simplex_handles"? `simplex_handle_strong_validity` is the name Clément chose, so I kept it, but the other version would be shorter and still understandable.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -112,7 +114,13 @@ class Simplex_tree {
   // Note: this wastes space when Vertex_handle is 32 bits and Node is aligned on 64 bits. It would be better to use a
   // flat_set (with our own comparator) where we can control the layout of the struct (put Vertex_handle and
   // Simplex_key next to each other).
-  typedef typename boost::container::flat_map<Vertex_handle, Node> Dictionary;
+  typedef typename boost::container::flat_map<Vertex_handle, Node> flat_map;
+  //Dictionary::iterator remain valid under insertions and deletions,
+  //necessary when computing zigzag filtration
+  typedef typename boost::container::map<Vertex_handle, Node> map;
```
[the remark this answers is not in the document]
hschreiber: I didn't thought about testing the difference between boost::map et std::map, as I somehow thought that Clément tested it. But as nothing was really finished or compiled, I doubt it now, so I will test this! Also with std::unordered_map while I am at it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -185,16 +194,14 @@ class Simplex_tree {
...
-
   class Fast_cofaces_predicate {
     Simplex_tree* st_;
     int codim_;
     int dim_;
    public:
     Fast_cofaces_predicate(Simplex_tree* st, int codim, int dim)
       : st_(st), codim_(codim), dim_(codim + dim) {}
-    bool operator()( Const_boost_iterator iter ) const {
+    bool operator()( const Simplex_handle iter ) const {
```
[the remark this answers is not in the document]
hschreiber: It was already a `const iterator` before, just named `Const_boost_iterator` a bit higher above, which was even more confusing as it was difficult to see that `boost::container::vec_iterator<std::pair<int, Node >*, false>` is simply a simplex handle when the dictionnary is a flat_map. If we introduce one day a `Const_simplex_handle` (together with the `const Simplex_tree`), we could replace it here with.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: unordered_map would likely require a lot of changes, we really rely on the order.
boost::map likely has the advantage over std::map that it also has a constructor that takes boost::container::ordered_unique_range and can thus share more code with flat_map, which is a good enough reason for me.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -614,7 +623,11 @@ class Simplex_tree {
    *
    * One can call filtration(null_simplex()). */
   static Simplex_handle null_simplex() {
-    return Dictionary_it(nullptr);
+    if constexpr (Options::simplex_handle_strong_validity) {
+      return Dictionary_it();
+    } else {
+      return Dictionary_it(nullptr);
+    }
```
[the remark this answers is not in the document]
hschreiber: Yes, I am changing that.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -649,7 +662,12 @@ class Simplex_tree {
   size_t num_simplices(Siblings * sib) {
     auto sib_begin = sib->members().begin();
     auto sib_end = sib->members().end();
-    size_t simplices_number = sib_end - sib_begin;
+    size_t simplices_number;
+    if constexpr (Options::simplex_handle_strong_validity){
+        simplices_number = std::distance(sib_begin, sib_end);
```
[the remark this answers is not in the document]
hschreiber: I kept all those special cases, as I was not sure if the boost containers and iterators where always standard on this aspect. But if it is the case, I will be able to remove a lot of the constexpr for the std::next etc.
And for `sib->members().size()`, that is a very good question... It is the most simple solution...

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -52,6 +54,7 @@
 #include <iterator>  // for std::distance
 #include <type_traits>  // for std::conditional
 #include <unordered_map>
+#include <iterator>
```
mglisse, line 57: It is already included 3 lines above.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -764,7 +774,11 @@ class Simplex_tree {
   Simplex_handle find_vertex(Vertex_handle v) {
     if (Options::contiguous_vertices) {
```
mglisse, line 775: I think **all** tests of Options::contiguous_vertices should also test `!Options::simplex_handle_strong_validity` (or we could have a static_assert forbidding to have both, though that doesn't sound as good). Then you shouldn't need the 2-argument version of `std::next` anywhere.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Indeed, is removed.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I replaced them all. I also converted them to constexpr.
It is probably better to allow them both in case we find some other optimization where both are possible?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: For the map, I was not able to make `get_parent_from_member` work, but I missed perhaps something. The boost documentation is not really helpful... Digging through the sources works perhaps better.
I also don't like having a linear search, but I didn't found another workaround... And replacing Node* with Simplex handles is not really a solution, specifically because they are not always stable. Except if we choose to have two different types of lists, depending on the value of `simplex_handle_strong_validity`, what is clearly possible, just uglier. So, I will look into it.

hschreiber: So, the only difficulty there is to replace the Node* by simplex handles is that we either need a wrapper around the handles to inherit the hook or we have to use something else than a intrusive list. The rest is quit straightforward.
If we replace the intrusive list, we need something easy to iterate over and to remove from.

mglisse: > I also don't like having a linear search, but I didn't found another workaround...

Everywhere we call this function, we also know the Vertex_handle that goes with the Node.

> And replacing Node* with Simplex handles is not really a solution, specifically because they are not always stable.

Simplex_handle and Node* are exactly as (un)stable as each other, since they only differ by a constant offset.

mglisse: Boost's map::iterator is roughly of type `iterator_from_iiterator<tree_iterator>`. I can go back and forth between that and a node_ptr (which is a true pointer `compact_rbtree_node*`):
```c++
  auto it = m.begin();
  auto iit = it.get();
  auto pn = iit.pointed_node();
  decltype(iit) iit2; iit2 = pn;
  decltype(it) it2(iit2);
  assert(it==it2);
```
Now going between that node_ptr and our `Node*` (different kind of node), since both are pointers, we can cast to `const char*` and compute the difference to find the offset, and then we can add/remove the offset (and cast twice) to go from one pointer to the other.
Yeah that's pretty ugly, and a bit fragile using public but undocumented stuff, but Boost.Container isn't a very active library.

hschreiber: > Everywhere we call this function, we also know the Vertex_handle that goes with the Node.
Ok, if we add the vertex handle in the parameters, we could do the verification in log instead of linear time. If that is fine, I would do that until we have something better.

>  Simplex_handle and Node* are exactly as (un)stable as each other, since they only differ by a constant offset.
Sorry, I was thinking about `Node`, not `Node*`. But isn't that still a problem, by the way? If I remember well, boost intrusive lists do not copy their members, which means that the `link_nodes_by_label` optimization does not work if we add or remove too many things in the tree. Or do references somehow not get invalidated?

mglisse: Look at Hooks_simplex_base_link_nodes: it has a move constructor/assignment to adapt the links any time a node is moved.

hschreiber: So, that mean that the list only keeps track of the position of the member `list_max_vertex_hook_` and not of the class `Hooks_simplex_base_link_nodes`? I agree that the hooks are the right ones, but the node it-self and therefore the hook wrapper changed address, no? Or perhaps there is something I am missing about inheritance?

hschreiber: I added the vertex handle in the new version. Is it okay this way?
In the meantime, I will be looking at the offset to see if we can switch from Node to iterator as you suggested.

hschreiber: So, this works on my side for boost::map :
```C++
static Simplex_handle simplex_handle_from_node(Node& node) {
    const char* shift = (const char*)(&node) - 32;
    boost::intrusive::compact_rbtree_node<void*>* it_ptr = (boost::intrusive::compact_rbtree_node<void*>*)shift;
    boost::intrusive::tree_iterator<
        boost::intrusive::bhtraits<
            boost::container::base_node<
                std::pair<const int, Simplex_tree_node_explicit_storage<Simplex_tree>>,
                boost::container::dtl::intrusive_tree_hook<void*, boost::container::red_black_tree, true>, true>,
            boost::intrusive::rbtree_node_traits<void*, true>, boost::intrusive::normal_link, boost::intrusive::dft_tag,
            3>,
        false>
        it_ii;
    it_ii = it_ptr;
    Dictionary_it sh(it_ii);

    return sh;
  }
```
A more "automatic" version:
```C++
static Simplex_handle simplex_handle_from_node(Node& node) {
    Dictionary_it testIt = node.children()->members().begin();
    Node* testNode = &testIt->second;
    auto testIIt = testIt.get();
    auto testPtr = testIIt.pointed_node();
    auto shift = (const char*)(testNode) - (const char*)(testPtr);
    decltype(testPtr) sh_ptr = decltype(testPtr)((const char*)(&node) - shift);
    decltype(testIIt) sh_ii;
    sh_ii = sh_ptr;
    Dictionary_it sh(sh_ii);

    return sh;
  }
```

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -73,7 +73,12 @@ class Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator
     // coface of simp_
     bool operator()(typename SimplexTree::Hooks_simplex_base& curr_hooks) {
       Node& curr_node = static_cast<Node&>(curr_hooks);
-      auto sh = cpx_->simplex_handle_from_node(curr_node);
+      Simplex_handle sh;
+      if constexpr (SimplexTree::Options::simplex_handle_strong_validity){
+        sh = cpx_->simplex_handle_from_node(curr_node, simp_.front());
+      } else {
+        sh = cpx_->simplex_handle_from_node(curr_node);
+      }
```
mglisse, line 81: if we stick to the implementation for map that needs the vertex, I guess we could always pass front as 2nd argument, and simply inside simplex_handle_from_node have two implementations (with `if constexpr[), one of which ignores the second argument. But let's see if the version without vertex works first.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: I know it feels ugly, but it is contained inside this one helper function, so I would be in favor of doing this (say the second "automatic" version). You can add the explicit types as comments next to it if you want.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1919,11 +1928,25 @@ class Simplex_tree {
...
 
+  /** \brief Helper method that returns the corresponding Simplex_handle from a member element defined by a node.
+   */
+  static Simplex_handle simplex_handle_from_node(Node& node, Vertex_handle label) {
+    Siblings* children = node.children();
+    auto it = children->members().find(label);
+    if (it != children->members().end()) {  // verifies if node is a leaf
+      return it;
+    }
+    return children->oncles()->find(label);
```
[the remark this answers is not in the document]
hschreiber: This part is obsolete if we use the shift-recast method, but you are totally right... I completely forgot that a label cannot appear twice the same path down and that one can simply test `node.children()->parent()`... Also that there is already a method `has_children`... So, the correct version would be:
```C++
      if (node.children()->parent() == label) {  // verifies if node is a leaf
        return children->oncles()->find(label);
      } else {
        return children->members().find(label);
      }
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: It is also much faster and avoids the label parameter. The new version is committed. I kept the old version as a comment in case some problem arises in the future.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_star_simplex_iterators.h
```diff
@@ -73,7 +73,8 @@ class Simplex_tree_optimized_cofaces_rooted_subtrees_simplex_iterator
     // coface of simp_
     bool operator()(typename SimplexTree::Hooks_simplex_base& curr_hooks) {
       Node& curr_node = static_cast<Node&>(curr_hooks);
-      auto sh = cpx_->simplex_handle_from_node(curr_node);
+      Simplex_handle sh;
+      sh = cpx_->simplex_handle_from_node(curr_node);
```
mglisse, line 77: Can we put this back in a single line?

mglisse (approved):

src/Simplex_tree/concept/SimplexTreeOptions.h
mglisse: I am ok with either name.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1921,9 +1930,48 @@ class Simplex_tree {
     return nullptr;
   }
 
-  /** \brief Helper method that returns the corresponding Simplex_handle from a member element defined by a node. */
+  /** \brief Helper method that returns the corresponding Simplex_handle from a member element defined by a node.
+   */
   static Simplex_handle simplex_handle_from_node(Node& node) {
-    return (Simplex_handle)(boost::intrusive::get_parent_from_member<Dit_value_t>(&node, &Dit_value_t::second));
+    if constexpr (Options::simplex_handle_strong_validity){
+      //Relies on the Dictionnary type to be boost::container::map<Vertex_handle, Node>.
```
VincentRouvreau, line 1937:
```suggestion
      //Relies on the Dictionary type to be boost::container::map<Vertex_handle, Node>.
```

src/Simplex_tree/concept/SimplexTreeOptions.h
hschreiber: @VincentRouvreau Do you have an opinion on this?

VincentRouvreau: @hschreiber No strong opinion. Choose the one you prefer, as the name and the documentation is quite clear for me

src/Simplex_tree/concept/SimplexTreeOptions.h
```diff
@@ -29,5 +29,7 @@ struct SimplexTreeOptions {
   static constexpr bool contiguous_vertices;
   /// If true, the lists of `Node` with same label are stored to enhance cofaces and stars access.
   static const bool link_nodes_by_label;
+  /// If true, Simplex_handle will not be invalidated after insertions or removals.
```
mglisse, line 32: Maybe we could mention here that it slows down most operations. Or maybe that's obvious, we wouldn't provide the false case if it did not have advantages. We don't mention that contiguous_vertices can speed up some operations, so **not** saying anything here either is consistent.

hschreiber: That is what I thought. If there were no disadvantages to have stable simplex handles, it would not be at false by default. But we could add something if you think it is better, also for `contiguous_vertices` for consistency.

PR #941, opened 2023-08-22
small fix to make cmake find OpenGL when OpenGL_GL_PREFERENCE=GLVND

hschreiber:
Replaced "${OPENGL_gl_LIBRARY} ${OPENGL_glu_LIBRARY}" with "OpenGL::GL OpenGL::GLU" to generalize the usage of OpenGL for GudhUI. If OpenGL_GL_PREFERENCE=GLVND instead of OpenGL_GL_PREFERENCE=LEGACY, the path to the library is not stored in OPENGL_gl_LIBRARY which causes a linking problem.


mglisse (approved):
It is convenient that these imported targets are available since 3.8, exactly the minimum version of cmake we require.

VincentRouvreau (approved):
Works fine for me and target cmake syntax is nicer

Issue #179, opened 2019-12-19
Build directory stored in compiled Python modules


gspr:
The Python modules seem to store knowledge of the directory in which GUDHI was built, for example through `__root_source_dir__ = "@CMAKE_SOURCE_DIR@"`. While this probably doesn't hurt, it seems rather unconventional for the *build* directory to be available as a constant to the end user. It surely shouldn't be useful for anything?

(I'm reporting this as an issue because it causes Debian to mark the GUDHI package as unreproducible).

mglisse:
Thanks for the report. It looks like this is used in documentation examples, so they can find data files on which to run the examples. Using `__data_dir__=__root_source_dir__+'/data'` (or a more descriptive name) instead would clarify that, but wouldn't really help you. I am not sure what would be a convenient other way to achieve this.

gspr:
Marc Glisse <notifications@github.com> writes:

> Thanks for the report. It looks like this is used in documentation
> examples, so they can find data files on which to run the examples.

Couldn't paths relative to the root of the source directory be used?

mglisse:
> Couldn't paths relative to the root of the source directory be used?

I don't understand your suggestion. That's pretty much what we have now, `__root_source_dir__` gives us the root of the source directory so we can specify a path relative to it. Since commands are run from the build directory (at best), we can't easily guess where the sources are.
A drawback of this approach is that it is meant for people who build from source, but that's a minority. Most people will download a conda package, and the examples from the documentation won't run on their machine.
One way around it would be to package those data files in some way, so they can be installed (preferably as a separate package). That's probably a non-trivial amount of work.

VincentRouvreau:
`__root_source_dir__` was a hack I found to make the documentation work for each install.
It was before the Debian and conda package (same issue with conda).
@bertrandmichel was asking me if we could add the data inside the python module like they do on [scikit-learn](https://scikit-learn.org/stable/datasets/index.html) and I think this would be the most satisfying solution.
The problem I have is that I have to copy all data files inside the gudhi python directory with CMake. Maybe I could think of another solution.

gspr:
Marc Glisse <notifications@github.com> writes:

>> Couldn't paths relative to the root of the source directory be used?
>
> I don't understand your suggestion. That's pretty much what we have
> now, `__root_source_dir__` gives us the root of the source directory
> so we can specify a path relative to it. Since commands are run from
> the build directory (at best), we can't easily guess where the sources
> are.

[...]

PR #942, opened 2023-08-23
Remove __root_source_dir__ from cubical documentation

VincentRouvreau:
Fix partially #179 for cubical complex documentation

### DIFF
--- a/src/python/doc/cubical_complex_user.rst
+++ b/src/python/doc/cubical_complex_user.rst
@@ -73,31 +73,22 @@ Knowing the sizes of the bitmap, by a series of modulo operation, we can determi
 present in the product that gives the cube :math:`C`. In a similar way, we can compute boundary and the coboundary of
 each cube. Further details can be found in the literature.
 
-Input Format
+Construction
 ------------
 
 In the current implantation, filtration is given either at the maximal cubes or at the vertices, and it is then
 extended by the lower star filtration to all cubes.
 There are a number of constructors that can be used to construct cubical complex by users
 who want to use the code directly. They can be found in the :doc:`cubical_complex_ref`.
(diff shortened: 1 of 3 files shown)


mglisse (approved):
The main thing this PR does is remove examples with Perseus input from the doc (the removal of root_source_dir is just a by-product). I am ok with that as long as it is a choice, i.e. that we prefer showing an example that takes an array. If we remove it just because we are not sure how to keep it cleanly but would otherwise want to have it in the doc, then I don't like it.

VincentRouvreau:
> The main thing this PR does is remove examples with Perseus input from the doc (the removal of root_source_dir is just a by-product). I am ok with that as long as it is a choice, i.e. that we prefer showing an example that takes an array.

I think we should remove the use of `__root_source_dir__` (cf. #179) as it does not work for pip and conda packages.
Constructing a cubical from a perseus style file is a specific case, that, IMHO, we should not put in cubical user manual, otherwise users may think they have to construct a perseus file from their data to be able to use a cubical.
Showing how to construct a cubical from a numpy array is much more clear example. If users have images, for example, they will find easily how to read an image as a numpy array.

> If we remove it just because we are not sure how to keep it cleanly but would otherwise want to have it in the doc, then I don't like it.

No, that was not the aim of this PR. What I also was disliking was that:
* not easy for the user to find data from `/data/bitmap/cubicalcomplexdoc.txt`
* the note section for perseus files are duplicated in [input-format](https://gudhi.inria.fr/python/latest/cubical_complex_user.html#input-format) and in [persus file format](https://gudhi.inria.fr/python/latest/fileformats.html#perseus)

mglisse:
> Showing how to construct a cubical from a numpy array is much more clear example.

Ok then. The title of the PR misled me to think that this was not your main motivation.

Issue #875, opened 2023-04-24
Maximal dimension in LowerStarSimplexTreeLayer


MathieuCarriere:
Currently, the `LowerStarSimplexTreeLayer` class does not have a `persistence_dim_max` argument, so this argument is set to its default value (`False`) when persistence is computed inside the class. Hence, the layer can never output the essential part of the diagrams in the maximal homology dimension. Thus, it would be nice to add a `persistence_dim_max` parameter to the class.

PR #948, opened 2023-08-24
Fix for #875

MathieuCarriere:


mglisse (approved):
It would have been even better with a new test, but since this is just a wrapper for other code, ok.

VincentRouvreau (approved):
Same remark as Marc, a test would be nice, but ok as is.
UT and notebooks are still working nicely

PR #956, opened 2023-09-08
On OSX, '-mmacosx-version-min=XXX' compile and link options management

### LINKED ISSUE #899: Macos options
Instead of an explicit `-mmacosx-version-min=10.14` in src/python/CMakeLists.txt and `ARCHFLAGS: "-arch arm64 -arch x86_64"` in .github/workflows/pip-packaging-osx.yml, we should investigate using CMAKE_OSX_DEPLOYMENT_TARGET and CMAKE_OSX_ARCHITECTURES. Maybe we already did and ended up with what we have because of the complicated cmake+setuptools mix.

The current github action works if we remove `-mmacosx-version-min=10.14`, but I didn't check the azure one.

This is mostly a reminder that the macos situation is currently messy.

VincentRouvreau:
when asked by MACOSX_DEPLOYMENT_TARGET environment variable.
Fix #899

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -255,7 +255,8 @@ if(PYTHONINTERP_FOUND)
 
     # Specific for Mac
     if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-      # if MACOSX_DEPLOYMENT_TARGET environment variable is set
+      # CMAKE_OSX_DEPLOYMENT_TARGET is always set on OSx
+      # but it can be modified by the user by setting MACOSX_DEPLOYMENT_TARGET environment variable
       if(CMAKE_OSX_DEPLOYMENT_TARGET)
         set(GUDHI_PYTHON_EXTRA_COMPILE_ARGS "${GUDHI_PYTHON_EXTRA_COMPILE_ARGS}'-mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET}', ")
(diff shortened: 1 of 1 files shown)


src/python/CMakeLists.txt
```diff
@@ -255,8 +255,11 @@ if(PYTHONINTERP_FOUND)
 
     # Specific for Mac
     if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
-        set(GUDHI_PYTHON_EXTRA_COMPILE_ARGS "${GUDHI_PYTHON_EXTRA_COMPILE_ARGS}'-mmacosx-version-min=10.14', ")
-        set(GUDHI_PYTHON_EXTRA_LINK_ARGS "${GUDHI_PYTHON_EXTRA_LINK_ARGS}'-mmacosx-version-min=10.14', ")
+      # if MACOSX_DEPLOYMENT_TARGET environment variable is set
```
mglisse, line 258: Isn't CMAKE_OSX_DEPLOYMENT_TARGET always defined? (not an issue, just surprised by the comment)

mglisse (approved):

src/python/CMakeLists.txt
VincentRouvreau: I was not aware of this behavior, but yes you are right, it seems CMAKE_OSX_DEPLOYMENT_TARGET is always set on OSx. I reworked the comment on [ee221bd](https://github.com/GUDHI/gudhi-devel/pull/956/commits/ee221bd58fc5cb4b9eb575660d758f8651d00217)

PR #957, opened 2023-09-08
[skip ci] let's try to skip azure CI

VincentRouvreau:


VincentRouvreau:
None of the tricks explained at https://github.com/microsoft/azure-pipelines-agent/issues/858 work :cry:

PR #962, opened 2023-09-15
Python documentation requires sphinxcontrib-bibtex to be built

VincentRouvreau:


mglisse:
Github won't let me comment on it, but you may want to update the message
```cmake
    else(SPHINX_PATH)
      message("++ Python documentation module will not be compiled because sphinx and sphinxcontrib-bibtex were not found")
```

PR #966, opened 2023-09-18
Missing erase after remove_if

### LINKED ISSUE #964: remove_if unused result
https://github.com/GUDHI/gudhi-devel/blob/911ee9d797a1fb328ee19441d6bf621d7a99ff75/src/Subsampling/include/gudhi/choose_n_farthest_points.h#L323
clang++ with libc++ warns that it is suspicious not to use the return value of remove_if. I need to look at this code again and either actually use the return value, or add a comment explaining why not.

mglisse:
Fix #964.
I don't know how I managed to miss that... It was in a place where the code was still valid, and the slowdown was limited enough not to make the runtime seem strange. Good thing one compiler decided to warn about it.
It does not seem to impact the result (good), but it does improve performance (something like 20%, depending on the input).

### DIFF
--- a/src/Subsampling/benchmark/choose_n_farthest_points.cpp
+++ b/src/Subsampling/benchmark/choose_n_farthest_points.cpp
@@ -76,7 +76,16 @@ int main(int argc, char**argv) {
             << " vs metric "              << std::chrono::duration_cast<std::chrono::milliseconds>((time_stop2 - time_start2)).count()
             << "  (Boost version " << BOOST_VERSION << ")\n";
   if (dists != dists2 || results != results2) {
+    // Note that with many points, it often happens that 2 points with the same distance are swapped in the output.
     std::cerr << "Results differ\n";
+#ifdef LOG_DIFF
+    std::ofstream log_gen("log_gen");
+    std::ofstream log_met("log_met");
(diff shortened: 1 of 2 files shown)


src/Subsampling/benchmark/choose_n_farthest_points.cpp
```diff
@@ -76,7 +76,16 @@ int main(int argc, char**argv) {
             << " vs metric "              << std::chrono::duration_cast<std::chrono::milliseconds>((time_stop2 - time_start2)).count()
             << "  (Boost version " << BOOST_VERSION << ")\n";
   if (dists != dists2 || results != results2) {
+    // Note that with many points, it often happens that 2 points with the same distance are swapped in the output.
     std::cerr << "Results differ\n";
```
VincentRouvreau, line 80: Does it happen that results differ ? I was not able to reach this code.

VincentRouvreau (approved):

src/Subsampling/benchmark/choose_n_farthest_points.cpp
mglisse: With 100k points and seed 0, I get 1 swap. With 50k points, I need to try several seeds before I get a swap.

PR #968, opened 2023-09-20
warns if Hera submodule is not initialzied

### LINKED ISSUE #963: CMake: check for missing Hera
I think CMake should check early whether Hera is present, and if not remind users about `git submodule update --init` (and possibly `HERA_INCLUDE_DIR`).

VincentRouvreau:
Fix #963

### DIFF
--- a/src/cmake/modules/GUDHI_submodules.cmake
+++ b/src/cmake/modules/GUDHI_submodules.cmake
@@ -3,3 +3,9 @@ set(HERA_INTERNAL_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/ext/hera/include)
 set(HERA_INCLUDE_DIR ${HERA_INTERNAL_INCLUDE_DIR} CACHE PATH "Directory where one can find hera/{wasserstein.h,bottleneck.h}")
 # since everything is cleanly under include/hera/, there is no harm always including it
 include_directories(${HERA_INCLUDE_DIR})
+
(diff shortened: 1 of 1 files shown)


src/cmake/modules/GUDHI_submodules.cmake
```diff
@@ -3,3 +3,9 @@ set(HERA_INTERNAL_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/ext/hera/include)
 set(HERA_INCLUDE_DIR ${HERA_INTERNAL_INCLUDE_DIR} CACHE PATH "Directory where one can find hera/{wasserstein.h,bottleneck.h}")
 # since everything is cleanly under include/hera/, there is no harm always including it
 include_directories(${HERA_INCLUDE_DIR})
+
+if (NOT EXISTS ${HERA_INCLUDE_DIR}/hera/wasserstein.h OR NOT EXISTS ${HERA_INCLUDE_DIR}/hera/bottleneck.h)
+  message(WARNING "${HERA_INCLUDE_DIR}/hera/{wasserstein.h,bottleneck.h} are not found.\n\
```
mglisse, line 8: Since the build is going to fail later anyway, shouldn't we have something stronger than a warning? If we error out immediately, the message is more likely to be seen than if it is lost in the middle pages of cmake+make stuff.

VincentRouvreau: I would let it as a warning, as if you say `cmake -DWITH_GUDHI_PYTHON=OFF`, hera is not required and it will compile fine

mglisse (approved):

PR #969, opened 2023-09-21
missing iostream import

DavidLapous:
`output_diagram` method needs `iostream`.

### DIFF
--- a/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
+++ b/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
@@ -19,6 +19,7 @@
 #include <boost/pending/disjoint_sets.hpp>
 #include <boost/intrusive/list.hpp>
 
+#include <iostream>
 #include <map>
 #include <unordered_map>
 #include <utility>


mglisse (approved):
It isn't good practice to have other headers depend on the full iostream, but since we use `cout` in this file, the includes have to reflect it indeed.

DavidLapous:
> It isn't good practice to have other headers depend on the full iostream

I didn't know that, what's the reason ? Isn't this include ignored by some kind of `#pragma once` in general (so adding this line doesn't cost anything) ?

mglisse:
> I didn't know that, what's the reason ? Isn't this include ignored by some kind of `#pragma once` in general (so adding this line doesn't cost anything) ?

If there is another `#include <iostream>`, sure, the subsequent ones are free. The issue is that it prevents from writing a program that uses your header but does not include `<iostream>` at all. And including `<iostream>` doesn't just declare some stuff, it also causes at runtime the initialization of the standard I/O system. That's not a problem for the type of applications we write, but in other areas (say embedded systems) it is something people try to avoid.

VincentRouvreau (approved):

PR #971, opened 2023-09-22
Fix fast cofaces for insert_batch_vertices

### LINKED ISSUE #970: [Simplex_tree] insert_batch_vertices shall call update_simplex_tree_after_node_insertion
On #877 insert_batch_vertices was not considered and this method shall call update_simplex_tree_after_node_insertion for cofaces/stars to work when`link_nodes_by_label` is activated

VincentRouvreau:
Fix #970 
To be discussed if storing information in a `std::vector<Vertex_handle>` is a good idea or not.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -1317,24 +1317,14 @@ class Simplex_tree {
    * already present, its filtration value is not modified, unlike with other insertion functions. */
   template <class VertexRange>
   void insert_batch_vertices(VertexRange const& vertices, Filtration_value filt = 0) {
-    // When Options::link_nodes_by_label, update_simplex_tree_after_node_insertion must be called only if
-    // not already inserted - copy vertex handles from flat_map
-    std::vector<Vertex_handle> old_members(root_.members().size());
-    if constexpr (Options::link_nodes_by_label) {
-      for (auto sh = root_.members().begin(); sh != root_.members().end(); sh++)
-        old_members.emplace_back(sh->first);
-    }
-
     auto verts = vertices | boost::adaptors::transformed([&](auto v){
-      return Dit_value_t(v, Node(&root_, filt)); });
+        return Dit_value_t(v, Node(&root_, filt)); });
     root_.members_.insert(boost::begin(verts), boost::end(verts));
     if (dimension_ < 0 && !root_.members_.empty()) dimension_ = 0;
-    for (auto sh = root_.members().begin(); sh != root_.members().end(); sh++) {
-      if constexpr (Options::link_nodes_by_label) {
-        if (std::find(old_members.begin(), old_members.end(), sh->first) == std::end(old_members))
+    if constexpr (Options::link_nodes_by_label) {
+      for (auto sh = root_.members().begin(); sh != root_.members().end(); sh++) {
+        if (nodes_label_to_list_.find(sh->first) == nodes_label_to_list_.end())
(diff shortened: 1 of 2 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1317,10 +1317,26 @@ class Simplex_tree {
    * already present, its filtration value is not modified, unlike with other insertion functions. */
   template <class VertexRange>
   void insert_batch_vertices(VertexRange const& vertices, Filtration_value filt = 0) {
+    // When Options::link_nodes_by_label, update_simplex_tree_after_node_insertion must be called only if
+    // not already inserted - copy vertex handles from flat_map
+    std::vector<Vertex_handle> old_members(root_.members().size());
+    if constexpr (Options::link_nodes_by_label) {
+      for (auto sh = root_.members().begin(); sh != root_.members().end(); sh++)
+        old_members.emplace_back(sh->first);
```
mglisse, line 1325: You create old_members that contains n copies of 0 then append the rest to it? Did you mean to use reserve instead?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1317,10 +1317,26 @@ class Simplex_tree {
...
+    }
+
     auto verts = vertices | boost::adaptors::transformed([&](auto v){
-        return Dit_value_t(v, Node(&root_, filt)); });
+      return Dit_value_t(v, Node(&root_, filt)); });
     root_.members_.insert(boost::begin(verts), boost::end(verts));
     if (dimension_ < 0 && !root_.members_.empty()) dimension_ = 0;
+    for (auto sh = root_.members().begin(); sh != root_.members().end(); sh++) {
+      if constexpr (Options::link_nodes_by_label) {
+        if (std::find(old_members.begin(), old_members.end(), sh->first) == std::end(old_members))
```
mglisse, line 1334: No, using std::find yields a quadratic cost, whereas both vectors are sorted so it would be possible to walk on both at the same time in linear time.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1317,10 +1317,26 @@ class Simplex_tree {
...
+          update_simplex_tree_after_node_insertion(sh);
+      } else {
+        update_simplex_tree_after_node_insertion(sh);
```
mglisse, line 1337: I don't understand why this appears in both branches, but with a test in one branch?

mglisse (commented):
Try a simple loop with `if(!x.is_linked()) update(x);`, it will be less likely to go wrong.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1321,6 +1321,12 @@ class Simplex_tree {
         return Dit_value_t(v, Node(&root_, filt)); });
     root_.members_.insert(boost::begin(verts), boost::end(verts));
     if (dimension_ < 0 && !root_.members_.empty()) dimension_ = 0;
+    if constexpr (Options::link_nodes_by_label) {
+      for (auto sh = root_.members().begin(); sh != root_.members().end(); sh++) {
+        if (nodes_label_to_list_.find(sh->first) == nodes_label_to_list_.end())
```
mglisse, line 1326: Or something like
```suggestion
        if (!sh->second.list_max_vertex_hook_.is_linked())
```
?

PR #976, opened 2023-09-28
Multiparameter persistence, C++ part.

DavidLapous:
Cpp part of PR #817.
List of changes : 
 - new `is_multi_parameter` flag in the simplextree's options to deal with filtration values that behave as array-like instead of float-like.
 - Filtration values are now as const ref to prevent array-like copies
   - This implies adding static members for trivial filtrations : `null_value` and `inf_`
 - new member `filtration_mutable` to have a non-const access to the filtration.
 - For multiparameter filtrations, deactivated the `assign` behavior of the insert to prevent unwanted behaviors (and stay 1-critical). The filtration property is no longer maintained.
 - New array-like structure as `Filtration_value`, child of `std::vector` with numpy-like pointwise operations, in `Simplex_tree/multi_filtrations/Finitely_critical_multi_filtrations.h`
 - The `Simplex_tree/Simplex_tree_multi.h` contains the options of a simplextree multi, functions that can convert a 1-parameter simplextree to a multiparameter simplextree and vice versa, aswell as useful functions to push filtration values on a discrete grid (for discrete computations.
 - I also copied the unit tests of the simplextree to apply them to this new structure.

Feel free to move files/functions if they don't follow Gudhi's standard, or commit if its moral changes.

**Edit 20/02/26:**
PR dependencies:
- #1121
- #1122 
- #1176
- #1191 
- #1297


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -924,7 +933,7 @@ class Simplex_tree {
    */
   template<class InputVertexRange = std::initializer_list<Vertex_handle>>
   std::pair<Simplex_handle, bool> insert_simplex_and_subfaces(const InputVertexRange& Nsimplex,
-                    Filtration_value filtration = 0) {
```
VincentRouvreau, line 927: github interface doesn't let me comment on the documentation part of this method, but it is said:
```txt
   * If the insertion fails (the simplex is already there), the bool is set to false. If the insertion
   * fails and the simplex already in the complex has a filtration value strictly bigger than 'filtration',
   * we assign this simplex with the new value 'filtration', and set the Simplex_handle field of the
   * output pair to the Simplex_handle of the simplex.
  ```
We should explain the behaviour difference when `SimplexTreeOptions::is_multi_parameter = true`

File: src/Simplex_tree/include/gudhi/Simplex_tree.h [hunk 2]
```diff
@@ -970,8 +979,20 @@ class Simplex_tree {
     Simplex_handle simplex_one = insertion_result.first;
     bool one_is_new = insertion_result.second;
     if (!one_is_new) {
-      if (filtration(simplex_one) > filt) {
-        assign_filtration(simplex_one, filt);
+      if (!(filtration(simplex_one) <= filt)) { 
+        if constexpr (SimplexTreeOptions::is_multi_parameter){
```
VincentRouvreau, line 983: If we do nothing here when `SimplexTreeOptions::is_multi_parameter = true`, we could avoid a lot of code here.
I think we could test `if constexpr (!SimplexTreeOptions::is_multi_parameter)` upper in the method

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2246,6 +2277,18 @@ class Simplex_tree {
   /** \brief Upper bound on the dimension of the simplicial complex.*/
   int dimension_;
   bool dimension_to_be_lowered_ = false;
+
+//MULTIPERS STUFF
+public: 
+  void set_number_of_parameters(int num){
+		number_of_parameters_ = num;
```
VincentRouvreau, line 2284: This is a comment for all your changes, but we are using 2 spaces as indentation for the c++ part, cf. https://github.com/GUDHI/gudhi-devel/blob/master/.github/code_conventions.md#code-style

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1610,12 +1631,22 @@ class Simplex_tree {
       if (dim == 0) return;
       // Find the maximum filtration value in the border
       Boundary_simplex_range&& boundary = boundary_simplex_range(sh);
-      Boundary_simplex_iterator max_border = std::max_element(std::begin(boundary), std::end(boundary),
+      Filtration_value max_filt_border_value;
+      if constexpr (SimplexTreeOptions::is_multi_parameter){
+        // in that case, we assume that Filtration_value has a `push_to` member to handle this.
+        max_filt_border_value = Filtration_value(this->number_of_parameters_);
+        for (auto &face_sh : boundary){
+          max_filt_border_value.push_to(filtration(face_sh)); // pushes the value of max_filt_border_value to reach simplex' filtration
```
VincentRouvreau, line 1639: Put the comment upper the code as lines should not exceed 120 characters per line (when possible), cf. https://github.com/GUDHI/gudhi-devel/blob/master/.github/code_conventions.md#code-style

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/finitely_critical_filtrations.h
```diff
@@ -0,0 +1,161 @@
...
+#define FINITELY_CRITICAL_FILTRATIONS_H_
+
+#include <iostream>
+#include <algorithm>
+#include <limits>
+
+namespace Gudhi::multiparameter::multi_filtrations{
+
+template<typename T=float>
+class Finitely_critical_multi_filtration : public std::vector<T> {
```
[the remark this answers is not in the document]
DavidLapous: Yes, furthermore they are not multi-critical for the moment, so `Multi_filtration_value` seems ok

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,229 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+ *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+ *    Author(s):       David Loiseaux, Hannah Schreiber
+ *
+ *    Copyright (C) 2023 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+#ifndef SIMPLEX_TREE_MULTI_H_
```
hschreiber, line 10: `Simplex_tree_multi_utilities.h` seems more precise and less misleading to me than `Simplex_tree_multi.h`. What do you think?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,229 @@
...
+		f[i+1] = default_values[i];
+	std::vector<int> simplex;
+	simplex.reserve(st.dimension()+1);
+	for (auto &simplex_handle : st.complex_simplex_range()){
+		simplex.clear();
+		for (auto vertex : st.simplex_vertex_range(simplex_handle))
+			simplex.push_back(vertex);
+
+		if (num_parameters > 0)
+		f[0] = st.filtration(simplex_handle);
```
hschreiber, line 68: This line needs an indentation for the `if`. The whole file could also be reindented to follow Gudhi's guideline?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,229 @@
...
+
+
+// Turns a multi-parameter simplextree into a 1-parameter simplextree
+template<class simplextree_std, class simplextree_multi>
+void flatten(simplextree_std &st, simplextree_multi &st_multi, const int dimension = 0){
+	for (const auto &simplex_handle : st_multi.complex_simplex_range()){
+		std::vector<int> simplex;
+		for (auto vertex : st_multi.simplex_vertex_range(simplex_handle))
+			simplex.push_back(vertex);
+		typename simplextree_multi::Options::value_type f = dimension >= 0 ? st_multi.filtration(simplex_handle)[dimension] : 0;
```
hschreiber, line 84: What is the use of `dimension` to possibly be negative?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -151,16 +151,18 @@ class Simplex_tree {
...
+    Filtration_value& filtration() { return filt_; }
    private:
     Filtration_value filt_;
   };
   struct Filtration_simplex_base_dummy {
     Filtration_simplex_base_dummy() {}
     void assign_filtration(Filtration_value GUDHI_CHECK_code(f)) { GUDHI_CHECK(f == 0, "filtration value specified for a complex that does not store them"); }
-    Filtration_value filtration() const { return 0; }
+    const Filtration_value& filtration() const  { return null_value; }
+    static constexpr Filtration_value null_value={};
```
VincentRouvreau, line 165: If we say:
```cpp
struct MyOptions : Simplex_tree_options_multidimensional_filtration {
  // is_multi_parameter but not store_filtration
};

Simplex_tree<MyOptions> stree;
```
We are facing the error:
```
gudhi-devel/src/Simplex_tree/include/gudhi/Simplex_tree.h:165:39: error: the type 'const Filtration_value' {aka 'const Gudhi::multiparameter::multi_filtrations::Finitely_critical_multi_filtration<float>'} of 'constexpr' variable 'Gudhi::Simplex_tree<MyOptions>::Filtration_simplex_base_dummy::null_value' is not literal
     static constexpr Filtration_value null_value={};
                                       ^~~~~~~~~~
```

But maybe this error is *normal* ? I was wondering to see if we should static_assert some error in this case.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,229 @@
...
+		f[0] = st.filtration(simplex_handle);
+		auto filtration_copy = f;
+		st_multi.insert_simplex(simplex,filtration_copy);
+		
+	}
+}
+
+
+
+// Turns a multi-parameter simplextree into a 1-parameter simplextree
```
hschreiber, line 77: Should precise that the filtration values are not token over (which makes sense, but still). Could also add, that for the filtration values, `linear_projection` can be used.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,157 @@
...
+
+  static std::vector<std::vector<T>> to_python(const std::vector<Finitely_critical_multi_filtration<T>>& to_convert) {
+    return std::vector<std::vector<T>>(to_convert.begin(), to_convert.end());
+  }
+
+  static std::vector<Finitely_critical_multi_filtration<T>> from_python(const std::vector<std::vector<T>>& to_convert) {
+    return std::vector<Finitely_critical_multi_filtration<T>>(to_convert.begin(), to_convert.end());
+    ;
+  }
+  void push_to(const Finitely_critical_multi_filtration<T>& x) {
```
hschreiber, line 119: Once you do the description, you could add that this method is mandatory for `Filtration_value` in the simplex tree.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: A `static_assert` could be a good idea, as the multi parameter tree only differentiate it self from the standard with its filtration values. If `store_filtration == false`, `is_multi_parameter` set to true makes no sense. But the problem could also be solved by removing `null_value` (see Issue #977), then we can just document that `is_multi_parameter` is ignored if `store_filtration == false`.

src/Simplex_tree/example/simplex_tree_multi.cpp
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#include <gudhi/Simplex_tree.h>
+#include <gudhi/Simplex_tree/Simplex_tree_multi.h>
+
+#include <iostream>
+#include <initializer_list>
+
+struct ST_MULTI {
+public:
```
hschreiber, line 18:
```suggestion
```
Just a detail: the `public` here is useless, as all members of a `struct` are by default public.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,279 @@
...
+
+
+
+namespace Gudhi::multiparameter {
+/** Model of SimplexTreeOptions.
+ *
+ * Maximum number of simplices to compute persistence is <CODE>std::numeric_limits<std::uint32_t>::max()</CODE>
+ * (about 4 billions of simplices). */
+struct Simplex_tree_options_multidimensional_filtration {
+public:
```
hschreiber, line 28:
```suggestion
```
Same as other comment from the example file.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
DavidLapous: put every filtration values of `st` to 0. May be useful to copy a simplextree when we don't care about filtration values.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,279 @@
...
+		f[i+1] = default_values[i];
+	std::vector<int> simplex;
+	simplex.reserve(st.dimension()+1);
+	for (auto &simplex_handle : st.complex_simplex_range()){
+		simplex.clear();
+		for (auto vertex : st.simplex_vertex_range(simplex_handle))
+			simplex.push_back(vertex);
+
+		if (num_parameters > 0)
+		f[0] = st.filtration(simplex_handle);
```
hschreiber, line 74:
```suggestion
		if (num_parameters > 0)
		        f[0] = st.filtration(simplex_handle);
```
This line needs an indentation for the if. The whole file could also be reindented to follow Gudhi's guideline?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,279 @@
...
+* \param coordinate_values If set to true the filtration values will be turned into coordinates in this grid
+* instead of points in this grid.
+* */ 
+void squeeze_filtration(uintptr_t splxptr, const multi_filtration_grid &grid, bool coordinate_values=true){
+	
+	Simplex_tree<options_multi> &st_multi = *(Gudhi::Simplex_tree<options_multi>*)(splxptr);
+	auto num_parameters = static_cast<unsigned int>(st_multi.get_number_of_parameters());
+	if (grid.size() != num_parameters){
+		std::cerr << "Bad grid !" << std::endl;
+		throw;
```
hschreiber, line 211:
```suggestion
		throw std::invalid_argument("Bad grid ! Its size is not coherent.");
```

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,279 @@
...
+	}
+	for (const auto &simplex_handle : st_multi.complex_simplex_range()){
+		auto& simplex_filtration = st_multi.filtration_mutable(simplex_handle);
+		find_coordinates<options_multi::value_type>(simplex_filtration, grid); // turns the simplexfiltration into coords in the grid
+		if (!coordinate_values){
+			for (auto parameter = 0u; parameter < num_parameters; parameter++)
+				simplex_filtration[parameter] = grid[parameter][simplex_filtration[parameter]];
+		}
+	}
+	return;
```
hschreiber, line 221:
```suggestion
```
The return here is useless.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,279 @@
...
+* \param degrees Only the simpleces of these dimension will be taken into account. Useful for, e.g., Rips filtrations. 
+* */ 
+std::vector<multi_filtration_grid> get_filtration_values(const uintptr_t splxptr, const std::vector<int> &degrees){
+	Simplex_tree<options_multi> &st_multi = *(Gudhi::Simplex_tree<options_multi>*)(splxptr);
+	int num_parameters = st_multi.get_number_of_parameters();
+	std::vector<multi_filtration_grid> out(degrees.size(), multi_filtration_grid(num_parameters));
+	std::vector<int> degree_index(degrees.size());
+	int count = 0;
+	for (auto degree : degrees){
+		degree_index[degree] = count; count++;
```
hschreiber, line 239:
```suggestion
		degree_index[degree] = count++;
```

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,279 @@
...
+		if (std::find(degrees.begin(), degrees.end(), degree) == degrees.end()) continue;
+		for (int parameter=0; parameter < num_parameters; parameter++){
+			out[degree_index[degree]][parameter].push_back(filtration[parameter]);
+		}
+	}
+	return out;
+
+}
+
+}	// namespace Gudhi
```
hschreiber, line 255:
```suggestion
}	// namespace Gudhi::multiparameter
```

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,166 @@
...
+    return true;
+  }
+
+  static std::vector<std::vector<T>> to_python(const std::vector<Finitely_critical_multi_filtration<T>>& to_convert) {
+    return std::vector<std::vector<T>>(to_convert.begin(), to_convert.end());
+  }
+
+  static std::vector<Finitely_critical_multi_filtration<T>> from_python(const std::vector<std::vector<T>>& to_convert) {
+    return std::vector<Finitely_critical_multi_filtration<T>>(to_convert.begin(), to_convert.end());
+    ;
```
hschreiber, line 126:
```suggestion
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1610,12 +1624,21 @@ class Simplex_tree {
       if (dim == 0) return;
       // Find the maximum filtration value in the border
       Boundary_simplex_range&& boundary = boundary_simplex_range(sh);
-      Boundary_simplex_iterator max_border = std::max_element(std::begin(boundary), std::end(boundary),
-                                                              [](Simplex_handle sh1, Simplex_handle sh2) {
-                                                                return filtration(sh1) < filtration(sh2);
-                                                              });
+      Filtration_value max_filt_border_value;
+      if constexpr (SimplexTreeOptions::is_multi_parameter) {
+        // in that case, we assume that Filtration_value has a `push_to` member to handle this.
```
VincentRouvreau, line 1629: This should also be documented in the concept file `gudhi-devel/src/Simplex_tree/concept/FiltrationValue.h`

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/finitely_critical_filtrations.h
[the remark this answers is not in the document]
DavidLapous: Oh no my bad, this can be confusing because multiparameter filtration values can be
 - finitely critical or
 - not finitely critical
This class is meant for the first kind, although it is only compatible with 1-critical filtrations for the moment (i.e. a simplex can only appear at one spot only).

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
hschreiber: Ok, I see. You should say this in the description of `dimension`, now that you added the doc.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/finitely_critical_filtrations.h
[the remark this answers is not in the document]
DavidLapous: I've made an doc addition to the FinitelyCriticalFiltrations class to mention this

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2246,6 +2269,24 @@ class Simplex_tree {
   /** \brief Upper bound on the dimension of the simplicial complex.*/
   int dimension_;
   bool dimension_to_be_lowered_ = false;
+
+  // MULTIPERS STUFF
+ public:
+  /**
+   * \brief Sets the number of parameters of the filtrations if SimplexTreeOptions::is_multi_parameter. 
+   * */
+  void set_number_of_parameters(int num) { number_of_parameters_ = num; }
```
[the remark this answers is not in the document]
DavidLapous: Returning 1 seems more natural to me. But as you wish.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -151,16 +151,18 @@ class Simplex_tree {
       Key_simplex_base;
 
   struct Filtration_simplex_base_real {
-    Filtration_simplex_base_real() : filt_(0) {}
-    void assign_filtration(Filtration_value f) { filt_ = f; }
-    Filtration_value filtration() const { return filt_; }
+    Filtration_simplex_base_real() : filt_{} {}
+    void assign_filtration(const Filtration_value& f) { filt_ = f; }
+    const Filtration_value& filtration() const { return filt_; }
+    Filtration_value& filtration() { return filt_; }
```
VincentRouvreau, line 174: To be discussed, but I am not sure we should add this method.
And if we add it we should add it to the *Filtration_simplex_base_dummy* one.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
DavidLapous: I've added that to the concept

src/Simplex_tree/include/gudhi/Simplex_tree.h
DavidLapous: I've added a few lines in the concept file.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,229 @@
...
+
+
+
+
+/// @brief turns filtration value x into coordinates in the grid
+/// @tparam out_type 
+/// @param x 
+/// @param grid 
+/// @return 
+template<typename out_type=int, typename vector_like>
```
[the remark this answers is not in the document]
mglisse: It is legal and can have good use cases, but here where out_type is not used anywhere, I don't really see the point.

src/Simplex_tree/example/simplex_tree_multi.cpp
mglisse: True, but I have seen style guides that advocate being explicit like this. So if David likes having `public:` I wouldn't force him to remove it.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,258 @@
...
+  }
+  return out;
+}
+
+}  // namespace Gudhi::multiparameter
+
+namespace std {
+
+template <>
+class numeric_limits<Gudhi::multiparameter::multi_filtration_type> {
```
VincentRouvreau, line 241: I am not sure this code is doing what it is suppose to do:
```cpp
  using Multi_Filtration_values = multi_filtrations::Finitely_critical_multi_filtration<float>;
  std::clog << std::numeric_limits<Multi_Filtration_values>::quiet_NaN() << std::endl;
  // displays: [ ]
  std::clog << std::numeric_limits<Multi_Filtration_values>::infinity() << std::endl;
  // displays: [ ]
```

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,258 @@
...
+
+    std::vector<typename simplextree_multi::Options::value_type> f = st_multi.filtration(simplex_handle);
+    if (dimension < 0) dimension = 0;
+    typename simplextree_multi::Options::value_type new_filtration = l.push_forward(f)[dimension];
+    st.insert_simplex(simplex, new_filtration);
+  }
+}
+
+/**
+ * \brief Given a point an a multiparameter discrete grid, pushes the point onto this grid.
```
VincentRouvreau, line 140: Typo
```suggestion
 * \brief Given a point on a multiparameter discrete grid, pushes the point onto this grid.
```

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
DavidLapous: I changed the position of this part of the code to the Finitely critical filtration file, which is more logical, and retested on my laptop and I get, as expected `[nan]` and `[inf]`. Does this fix works for you ?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
DavidLapous: I'm not sure what you mean, each flatten function take into account filtration values.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -604,18 +606,25 @@ class Simplex_tree {
...
     } else {
-      return std::numeric_limits<Filtration_value>::infinity();
+      return inf_;
+    }
+  }
+  static Filtration_value& filtration_mutable(Simplex_handle sh){
+    if (sh != null_simplex()) {
+      return sh->second.filtration();
+    } else {
+      return inf_;
```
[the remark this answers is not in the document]
DavidLapous: I don't have a fixed point of view on this. So its only a matter of what's the gudhi style of accessing a filtration value by reference.

src/Simplex_tree/include/gudhi/Simplex_tree.h
DavidLapous: A simple fix is to do a `static_assert(!store_filtration && is_multi_parameter)`. thoughts ?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
DavidLapous: I'm not a big fan of the `utilities` name, but the name can be changed indeed. Let's fix the other issues and we'll decide that once the content of this file is fixed.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
VincentRouvreau: Yes !

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2246,6 +2269,34 @@ class Simplex_tree {
...
+    else
+      return 1;
+  }
+
+  inline static Filtration_value inf_ = std::numeric_limits<Filtration_value>::has_infinity ? 
+      std::numeric_limits<Filtration_value>::infinity() 
+    : std::numeric_limits<Filtration_value>::max(); /**< Default infinite value. */
+
+ private:
+  int number_of_parameters_; /**< Number of parameters of the multi-filtrations when SimplexTreeOptions::is_multi_parameter.-*/
```
VincentRouvreau, line 2299: `number_of_parameters_` default value is 0. TBD, but wouldn't it be clearer from a user point of view to set it to 2 by default ?

src/Simplex_tree/concept/SimplexTreeOptions.h
```diff
@@ -31,5 +31,7 @@ struct SimplexTreeOptions {
   static const bool link_nodes_by_label;
   /// If true, Simplex_handle will not be invalidated after insertions or removals.
   static const bool stable_simplex_handles;
+  /// If true, assumes that Filtration_value is vector-like instead of float-like. This also assumes that Filtration_values is a class, which has a push_to method to push a filtration value $x$ onto $this>=0$.
```
VincentRouvreau, line 34: For a better documentation rendering:
```suggestion
  /// If true, assumes that Filtration_value is vector-like instead of float-like. This also assumes that SimplexTreeOptions::Filtration_value is a class, which has a `push_to` method to push a filtration value \f$x\f$ onto \f$this>=0\f$. 
```
image

src/Simplex_tree/include/gudhi/Simplex_tree.h
DavidLapous: Yes, why not, as the get number of parameters will always be  1 for non-multiparam simplextrees, this default value seems good.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
hschreiber: Huh? You are right, I think I completely missed a line here as I had the feeling that the filtration value was always put to 0. Which would also explain why I didn't understood the role of the dimension.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,244 @@
...
+    f[i + 1] = default_values[i];
+  std::vector<int> simplex;
+  simplex.reserve(st.dimension() + 1);
+  for (auto &simplex_handle : st.complex_simplex_range()) {
+    simplex.clear();
+    for (auto vertex : st.simplex_vertex_range(simplex_handle)) simplex.push_back(vertex);
+
+    if (num_parameters > 0) f[0] = st.filtration(simplex_handle);
+    st_multi.insert_simplex(simplex, f);
+  }
```
VincentRouvreau, line 71: We could take the opportunity to set the number of parameters here, no ?
(My suggestion may be not optimal, not tested)
```suggestion
  }
  simplex.set_number_of_parameters(std::min(static_cast<unsigned int>(default_values.size()), static_cast<unsigned int>(num_parameters - 1)));
```

DavidLapous: Good idea. Note that the filtration that we insert is always of size `num_parameters`, the default values are only meant to optionally initialize the values of this filtration f. So the number of parameter of this simplextree is indeed `num_parameters`.

DavidLapous: I'll also add a test for these functions.

DavidLapous:
I'm not sure why the ci Tangential complex test fails on git, I cannot reproduce on my laptop & it doesn't fail on the other pipelines

DavidLapous:
@VincentRouvreau I've added your tests to the end of the test file, for the NaN and I had add some checks in the multi filtration file so that the behavior of NaN and inf so that inf and nan filtration values have a proper behavior
 - NaN comparison is always false
 - The push_to method ignores NaN values
 - inf values behave as expected.

Furthermore, thanks to your test (Merci !! 😉) I noticed that there was a bug on the `make_filtration_non_decreasing` in the multiparameter case, which sometimes allows non-increasing values of simplices, that's why I had to change this line

https://github.com/GUDHI/gudhi-devel/blob/a89ff98f87eb6614cdfe8f3fcd8560319a008d71/src/Simplex_tree/test/simplex_tree_multi_unit_test.cpp#L933

I only changed this line and the last test, I hope it didn't break the spirit or your test

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+  std::vector<int> degree_index(degrees.size());
+  int count = 0;
+  for (auto degree : degrees) {
+    degree_index[degree] = count++;
+    out[degree_index[degree]].reserve(st_multi.num_simplices());
+  }
+
+  for (const auto &simplex_handle : st_multi.complex_simplex_range()) {
+    const auto filtration = st_multi.filtration(simplex_handle);
+    const auto degree = st_multi.dimension(simplex_handle);
```
mglisse, line 240: There is also `for_each_simplex` which gives you the dimension for free, but your code is fine.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+    if (std::find(degrees.begin(), degrees.end(), degree) == degrees.end()) continue;
```
mglisse, line 241: Needing this loop kind of defeats the purpose of having degree_index :disappointed:

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+ * */
+template <class simplextree_multi>
+std::vector<multi_filtration_grid> get_filtration_values(simplextree_multi& st_multi, const std::vector<int> &degrees) {
+  static_assert(simplextree_multi::Options::is_multi_parameter, "Only works for multiparameter simplextrees.");
+  int num_parameters = st_multi.get_number_of_parameters();
+  std::vector<multi_filtration_grid> out(degrees.size(), multi_filtration_grid(num_parameters));
+  std::vector<int> degree_index(degrees.size());
+  int count = 0;
+  for (auto degree : degrees) {
+    degree_index[degree] = count++;
```
mglisse, line 234: How do you know that all the elements of `degrees` are smaller than the size of `degrees`? Or was degree_index supposed to compute its size based on st_multi.dimension() (or upper_bound_dimension)?

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: OTOH, a "broken" value like 0 (or even -1) forces the user to be explicit (which can be seen as good or bad).

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2557,6 +2589,34 @@ class Simplex_tree {
   /** \brief Upper bound on the dimension of the simplicial complex.*/
   int dimension_;
   bool dimension_to_be_lowered_ = false;
+
+  // MULTIPERS STUFF
+ public:
+  /**
+   * \brief Sets the number of parameters of the filtrations if SimplexTreeOptions::is_multi_parameter. 
+   * */
+  void set_number_of_parameters(int num) {
```
mglisse, line 2598: Is it intended that we can change the number of parameters of an existing Simplex_tree, or should we ask for it in the constructor?

mglisse (commented):
A first batch of superficial comments (I don't have enough of a global view yet for deeper comments).

src/Persistent_cohomology/example/plain_homology.cpp
```diff
@@ -24,6 +24,7 @@ struct MyOptions : Gudhi::Simplex_tree_options_minimal {
   // static const bool store_filtration = false;
   // However the persistence algorithm needs this
   static const bool store_key = true;
+  static const bool is_multi_parameter = false;
```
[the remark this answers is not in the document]
DavidLapous: I had some trouble compiling, so at some point I just added this line everywhere. If you say that it's safe, we can removeit here

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -17,7 +17,19 @@
...
+  /** \brief For multiparameter filtrations, this methods pushes a filtration value 
+   * to the first moment, for operator< such that f < this. For instance, for a one critical filtration, with
+   *  - this = (1,2)
+   *  - x = (2,1)
+   *
+   * after calling this method, x should be equal to (2,2).
+   * This function is called when using, e.g. `make_filtration_non_decreasing`, as the filtration of a simplex
+   * has to be greater than the filtration of any of its faces.
+   * */ 
+  void push_to(const FiltrationValue f);
```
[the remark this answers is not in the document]
DavidLapous: `push_forward` ? I haven't though a lot about multicritical filtrations, but my first guess to implement this would be to push all of of the filtration values of `this` into each of the `f` ones, and then remove the redundant ones (for $\le$).
I think doing a concept is a good idea.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2557,6 +2589,34 @@ class Simplex_tree {
...
+    else
+      return 1;
+  }
+
+  inline static Filtration_value inf_ = std::numeric_limits<Filtration_value>::has_infinity ? 
+      std::numeric_limits<Filtration_value>::infinity() 
+    : std::numeric_limits<Filtration_value>::max(); /**< Default infinite value. */
+
+ private:
+  std::optional<int> number_of_parameters_; /**< Number of parameters of the multi-filtrations when SimplexTreeOptions::is_multi_parameter.-*/
```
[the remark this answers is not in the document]
DavidLapous: not really, I'm not even sure if the optional saves the `int` space for regular simplextrees anyway

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,322 @@
...
+namespace std {
+
+template<typename T>
+class numeric_limits<Gudhi::multiparameter::multi_filtrations::Finitely_critical_multi_filtration<T>>
+  {
+  public:
+    static constexpr bool has_infinity = true; 
+
+    static Gudhi::multiparameter::multi_filtrations::Finitely_critical_multi_filtration<T> infinity() throw(){
+      return Gudhi::multiparameter::multi_filtrations::Finitely_critical_multi_filtration<T>(1, std::numeric_limits<T>::infinity());
```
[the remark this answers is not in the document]
DavidLapous: The reason is that I need to have filtration values that can have coordinates at infinity while being less than an inf value which doesn't depend on the number of parameters. With this implementation it's also compatible with the 1-parameter case, so I'm fine with it. Also, in the multicritical case, this vector is unaware of the number of parameters.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: It doesn't save space, `optional<int>` takes the size of 2 int (an int and a bool, rounded up).

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Box.h
```diff
@@ -0,0 +1,139 @@
...
+}
+
+template <typename T>
+inline typename Box<T>::point_type& Box<T>::get_upper_corner() {
+  return upperCorner_;
+}
+
+template <typename T>
+inline bool Box<T>::contains(const point_type& point) const {
+  if (point.size() != bottomCorner_.size()) return false;
```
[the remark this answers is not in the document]
DavidLapous: I'll add a warning there

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
```diff
@@ -0,0 +1,85 @@
...
+  this->direction_.swap(v);
+}
+template <typename T>
+typename Line<T>::point_type Line<T>::push_forward(point_type x) const {  // TODO remove copy
+  x -= basepoint_;
+  T t = -std::numeric_limits<T>::infinity();
+  ;
+  for (std::size_t i = 0; i < x.size(); i++) {
+    T dir = this->direction_.size() > i ? direction_[i] : 1;
+    t = std::max(t, x[i] / dir);
```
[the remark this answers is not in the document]
DavidLapous: Yes, a check is missing there. We should be able to add orthogonal lines, and then the filtration value should be either infinity if `x` is below the basepoint of the line, or defined by the other coordinates.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * \brief Turns a 1-parameter simplextree into a multiparameter simplextree, 
+ * and keeps the 1-filtration in the 1st axis.
+ * Default values can be specified. 
+ * \ingroup multiparameter 
+ * \tparam simplextree_std A non-multi simplextree 
+ * \tparam simplextree_multi A multi simplextree 
+ * \param st Simplextree to copy 
+ * \param st_multi Multiparameter simplextree container to fill. 
+ * \param default_values If given, this vector is assume to be of size `num_parameters-1` and
```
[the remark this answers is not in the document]
DavidLapous: Yes, at the "first possible moment"

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+ * \ingroup multiparameter 
+ * \tparam simplextree_std A non-multi simplextree 
+ * \tparam simplextree_multi A multiparameter simplextree 
+ * \param st Simplextree to fill. 
+ * \param st_multi Multiparameter simplextree to convert into a 1 parameter simplex tree. 
+ * \param basepoint The basepoint of the diagonal line. 
+ * \param dimension The coordinate of the line to choose as a 1 parameter filtration (they are all equivalent).
+ * */
+template <class simplextree_std, class simplextree_multi>
+void flatten_diag(simplextree_std &st, simplextree_multi &st_multi,
```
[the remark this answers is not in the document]
DavidLapous: True, but I initially needed these function only for the python interface, so i kept it simple.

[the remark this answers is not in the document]
DavidLapous: I'll change that

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+    for (auto vertex : st_multi.simplex_vertex_range(simplex_handle)) simplex.push_back(vertex);
+
+    std::vector<typename simplextree_multi::Options::value_type> f = st_multi.filtration(simplex_handle);
+    typename simplextree_multi::Options::value_type new_filtration = l.push_forward(f)[dimension];
+    st.insert_simplex(simplex, new_filtration);
+  }
+}
+
+/**
+ * \brief Given a point on a multiparameter discrete grid, pushes the point onto this grid.
```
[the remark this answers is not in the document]
DavidLapous: I'm not aware of a terminology. If I remember correctly rivet uses "coarsening".

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+ * Turns the input point as the closest grid point, as coordinates in this grid.
+ * \ingroup multiparameter
+ * \tparam vector_like Vector like class
+ * \param x The point to push on the grid.
+ * \param grid The multiparameter grid. A vector of size `num_parameters`, whose elements are the elements of the grid
+ * for this axis.
+ * */
+template <typename vector_like>
+inline void find_coordinates(vector_like &x, const multi_filtration_grid &grid) {
+  // TODO: optimize with, e.g., dichotomy
```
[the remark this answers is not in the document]
DavidLapous: I almost never use regular grids, as the filtration values usually never fill the space regularly. I still assume that they are sorted. I'll check `lower_bound`, thanks Marc.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+ * \brief Pushes all of the filtration values of a simplextree onto a grid, c.f. \ref find_coordinates.
+ * \ingroup multiparameter
+ * \param st_multi Multiparameter simplex tree to squeeze on the grid.
+ * \param grid The multiparameter grid. A vector of size `num_parameters`,
+ * whose elements are the elements of the grid for this axis.
+ * \param coordinate_values If set to true the filtration values will be turned into coordinates in this grid
+ * instead of points in this grid.
+ * */
+template <class simplextree_multi>
+void squeeze_filtration(simplextree_multi& st_multi, const multi_filtration_grid &grid, bool coordinate_values = true) {
```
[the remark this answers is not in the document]
DavidLapous: No, maybe coarsen, like rivet, is better

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
```diff
@@ -0,0 +1,251 @@
...
+template <class simplextree_multi>
+std::vector<multi_filtration_grid> get_filtration_values(simplextree_multi& st_multi, const std::vector<int> &degrees) {
+  static_assert(simplextree_multi::Options::is_multi_parameter, "Only works for multiparameter simplextrees.");
+  int num_parameters = st_multi.get_number_of_parameters();
+  std::vector<multi_filtration_grid> out(degrees.size(), multi_filtration_grid(num_parameters));
+  std::vector<int> degree_index(degrees.size());
+  int count = 0;
+  for (auto degree : degrees) {
+    degree_index[degree] = count++;
+    out[degree_index[degree]].reserve(st_multi.num_simplices());
```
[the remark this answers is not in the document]
DavidLapous: I'm not sure that it's too bad, this function is meant to help generating a grid on which to compute "multipersistence", and the output is destroyed before doing any heavy computation. If there is a better value to put there, I can change this.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
DavidLapous: I'm not sure how to achieve the same result without this loop. Is there an iterator that is dimension increasing ?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_multi.h
DavidLapous: Thats a bug on my end yes, the degree index should be of size `st_multi.dimension()+1`.

src/Simplex_tree/include/gudhi/Simplex_tree.h
DavidLapous: It can be useful to change the filtration of an existing simplextree, e.g., adding a filtration while preserving the old ones. But I'm already asking it at construction time in the python interface.

DavidLapous:
Changes since last time:
 - Finitely_critical_filtrations renamed to One_critical_filtration
 - Added a K critical filtration alongside previous one, for the moment defined as a vector of vector, but it may be better to have it as a contiguous one : if the number of parameter is known, a np.array_split(contiguous_stuff, num_parameters) should do the trick.
 - removed some functions that are not necessary yet

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
```diff
@@ -0,0 +1,233 @@
...
+
+template <typename T> Line<T>::Line(const point_type &x) : basepoint_(x) {}
+template <typename T>
+Line<T>::Line(point_type &&x) : basepoint_(std::move(x)) {}
+template <typename T>
+Line<T>::Line(const point_type &x, const point_type &v)
+    : basepoint_(x), direction_(v) {}
+template <typename T>
+inline typename Line<T>::point_type
+Line<T>::push_forward(point_type x) const { // TODO remove copy
```
mglisse, line 68: IIUC, x is interpreted as the base point of an upper-right quadrant (a 1-critical thing), the line has a built-in parametrization, and this function returns the parameter of the (smallest) intersection point between the line and the quadrant? I'd like to see comments like that so I don't need to read the code...
(I assume this is called push_forward because we can start from x and increase ("forward") its coordinates as little as possible to bring it to the line ("project" more than "push"?))

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
```diff
@@ -0,0 +1,233 @@
...
+      x[i] = basepoint_[i] + t * direction_[i];
+  } else {
+    for (std::size_t i = 0; i < x.size(); i++)
+      x[i] = basepoint_[i] + t;
+  }
+  return x;
+}
+template <typename T>
+template <typename U>
+inline U Line<T>::push_forward2(const point_type &x) const {
```
mglisse, line 83: We need this version because the point could have integer coordinates, but you want some float type for this?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  inline point_type &basepoint() { return basepoint_; }
+  inline point_type &direction() { return direction_; }
+  inline const point_type &basepoint() const { return basepoint_; }
+  inline const point_type &direction() const { return direction_; }
+
+private:
+  point_type basepoint_; // any point on the line
+  point_type direction_; // direction of the line
```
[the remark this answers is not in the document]
DavidLapous: `direction_` is indeed assumed to be positive, although non-positive directions could be dealt with with the zigzag module.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
DavidLapous: Yes, sorry, I'll add some comments later; this is exactly it. `project` seems great to me. There is also the `push_back` method that does the same thing but with the opposite quadrant; so `project` won't be enough.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
DavidLapous: Yes! I'm not super sure that I need this, but this doesn't cost anything.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Line.h
```diff
@@ -0,0 +1,233 @@
...
+    for (std::size_t i = 0; i < x.size(); i++)
+      x[i] = basepoint_[i] + t;
+  }
+  return x;
+}
+template <typename T>
+template <typename U>
+inline U Line<T>::push_forward2(const point_type &x) const {
+  constexpr const U inf = std::numeric_limits<U>::infinity(); // This disable it for e.g.
+                                                    // ints, but that's good
```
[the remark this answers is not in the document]
DavidLapous: Indeed, I'll fix that

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
mglisse: Is "push" a standard term for posets? All I found is the pushout of 2 functors (some form of colimit), is that what you are using here? If the name matches a common definition in category theory, that can be a good reason to use it, but I don't know which one to look at.
(In my intuition, "push"ing one value does not really reflect the symmetry of the operation. Something like smallest_common_majorant (by analogy with GCD) would look more symmetric.)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2557,6 +2589,34 @@ class Simplex_tree {
...
+   * \brief Gets the number of parameters of the filtrations if SimplexTreeOptions::is_multi_parameter. 
+   * */
+  int get_number_of_parameters() const { 
+    if constexpr (SimplexTreeOptions::is_multi_parameter)
+      return *number_of_parameters_;
+    else
+      return 1;
+  }
+
+  inline static Filtration_value inf_ = std::numeric_limits<Filtration_value>::has_infinity ?
```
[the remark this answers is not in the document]
DavidLapous: This is an issue with `filtration_mutable` which has to return a reference, I don't really know what's the best way to deal with this

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
DavidLapous: I'm not following a standard term for this, so something like you proposed seems great to me.
The analogy is more with the LCM instead of GCD, but "smallest_common_majorant" seems great to me.
As calling multiple times LCM gives the LCM of a family (here `make_filtration_non_decreasing`), this is very moral.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+      return;
+    if (this->is_inf() || x.is_minus_inf()) {
+      *this = x;
+      return;
+    }
+    for (auto &stuff : *this) {
+      stuff.pull_to(x);
+    }
+  } 
+  // cannot be const, as gudhi iterators are not const
```
mglisse, line 703: Not sure what you mean. You are not iterating on a Simplex_tree here.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+  * Remove redundant values 
+  */
+  inline void clean(value_type max_error = 0) {
+    // A bit ugly, todo : erase+removeif ?
+    for (std::size_t i = 0; i < multi_filtration_.size(); i++) {
+      for (std::size_t j = 0; j < multi_filtration_.size(); j++) {
+        if (i == j)
+          continue;
+        if (dominates(multi_filtration_[j], multi_filtration_[i], max_error)) {
+          multi_filtration_[i].clear();
```
mglisse, line 771: Are future calls to `dominates` happy to encounter an empty vector? Or does that somehow not happen?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+      U projection = std::numeric_limits<U>::max();
+      for (auto &y : *this) { // cannot be const (Gudhi)
+        projection = std::min(projection, y.linear_projection(x));
+      }
+      return projection;
+    }
+  }
+
+  /*
+  * Checks if b is cleanable with respect to a
```
mglisse, line 721: (tested as `a >= b`)
In my interpretation as a union of upper-right quadrants, it is the larger point that can be removed, not the smaller one.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+      for (std::size_t j = 0; j < multi_filtration_.size(); j++) {
+        if (i == j)
+          continue;
+        if (dominates(multi_filtration_[j], multi_filtration_[i], max_error)) {
+          multi_filtration_[i].clear();
+        }
+        while (j < multi_filtration_.size() && dominates(multi_filtration_[i], multi_filtration_[j], max_error)) {
+          multi_filtration_[j].clear();
+          i--;
+        }
```
mglisse, line 776: Confusing algorithm, decrementing `i` while testing and operating on `j`.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+                             std::cout << "Keeping y=" << y << std::endl;
+                           }
+                         }
+                         return dominates(y, x);
+                       }),
+        multi_filtration_.end());
+
+    multi_filtration_.push_back(x);
+  }
+  inline void re_clean(){
```
mglisse, line 824: What is the difference with `clean()`?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+  begin() {
+    return multi_filtration_.begin();
+  }
+  inline typename std::vector<One_critical_filtration<T>>::iterator
+  end() {
+    return multi_filtration_.end();
+  }
+  inline typename std::vector<
+      One_critical_filtration<T>>::const_iterator
+  begin() const {
```
mglisse, line 875: Note that the following is equivalent (but your code is fine if you like being explicit)
```suggestion
  auto begin() const {
```

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+    return Multi_critical_filtration<U>(std::move(out));
+  }
+  inline void push_back(const OneCritical& x){
+    multi_filtration_.push_back(x);
+  }
+  inline const std::vector<One_critical_filtration<T>>& as_vector() const {
+    return multi_filtration_;
+  }
+
+  inline std::vector<std::vector<T>> as_VECTOR() const {
```
mglisse, line 927: as_vector vs as_VECTOR, I understand, but maybe better names?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+                                    std::vector<T>(this->num_parameters()));
+    for (std::size_t i = 0; i < this->num_generators(); ++i) {
+      for (std::size_t j = 0; j < this->num_parameters(); ++j) {
+        out[i][j] = multi_filtration_[i][j];
+      }
+    }
+    return out;
+  }
+
+  inline void _clean(bool keep_inf=true) {
```
mglisse, line 938: 3rd "clean" function (seems to have a different purpose).

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+  * Checks if `this`, seen as a birth curve is under the `other` birth curve,
+  */
+  inline bool operator<=(const Multi_critical_filtration<T,co> &other) const {
+    //check if this curves is below other's curve 
+    // ie for each guy in this, check if there is a guy in other that dominates him
+    for (std::size_t i = 0u; i < multi_filtration_.size(); ++i) {
+      for (std::size_t j = 0u; j < other.multi_filtration_.size(); ++j) {
+        // i <= j 
+        if (dominates(other.multi_filtration_[j], multi_filtration_[i]))
+          continue;
```
mglisse, line 988: `continue` which loop? I think you want the outer one, but C++ understands the inner one.

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+  }
+  inline friend bool
+  operator==(const Multi_critical_filtration<T,co> &self,
+             const Multi_critical_filtration<T,co> &to_compare) {
+    if (self.num_generators() != to_compare.num_generators())
+      return false;
+    auto it = to_compare.begin();
+    for (auto i = 0u; i < self.num_generators(); i++) {
+      if (self[i] != *(it++))
+        return false;
```
mglisse, line 598: 2 multi-critical lists with the same points in a different order are considered different?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+
+
+
+  // TODO : this costs a lot... optimize / cheat in some way for python ?
+  /*
+  * Checks if `this`, seen as a birth curve is under the `other` birth curve,
+  *
+  */
+  inline bool operator<(const Multi_critical_filtration<T,co> &other) const {
+    //check if this curves is below other's curve
```
mglisse, line 959: You say a < b if the curve of a is below the curve of b **and they don't touch anywhere**?
Usually, I would expect: the curve of a is below the curve of b and there is at least 1 place where they differ (they are not equal), although asking for the 2 curves not to touch at all could have some uses I guess. Can you shed some light on how this will be used?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Finitely_critical_filtrations.h
```diff
@@ -0,0 +1,1051 @@
...
+  std::vector<One_critical_filtration<T>> multi_filtration_;
+};
+
+} // namespace Gudhi::multiparameter::multi_filtrations
+
+namespace std {
+
+template <typename T>
+class numeric_limits<Gudhi::multiparameter::multi_filtrations::
+                         One_critical_filtration<T>> {
```
mglisse, line 1016: Where do you need it (trying to understand if it is also needed for multi-critical)?

src/Simplex_tree/include/gudhi/Simplex_tree/multi_filtrations/Box.h
```diff
@@ -0,0 +1,182 @@
...
+#include <ostream>
+#include <vector>
+
+#include "Finitely_critical_filtrations.h"
+
+/**
+ * @brief Simple box in \f$\mathbb R^n\f$ .
+ */
+
+namespace Gudhi::multiparameter::multi_filtrations {
```
mglisse, line 30: Can you explain how you choose what goes into `Gudhi::multiparameter::multi_filtrations` vs what goes directly in `Gudhi::multiparameter`?

src/Multi_filtration/include/gudhi/One_critical_filtration.h
```diff
@@ -0,0 +1,1378 @@
...
+  /**
+   * @brief Constructs a vector of the size of the given number of parameters with -inf as value for each entry.
+   *
+   * @warning The vector `{-inf, -inf, ...}` with \f$ n > 1 \f$ entries is not considered as "minus infinity" (the
+   * method @ref is_minus_inf() will not return true). The `-inf` are just meant as placeholders, at least one entry
+   * should be modified by the user. Otherwise, either use the static method @ref minus_inf() or set @p n to 1 instead.
+   *
+   * @param n Number of parameters.
+   */
+  One_critical_filtration(int n) : Base(n, -T_inf) {};
```
mglisse, line 96: This (and some others below?) should probably be marked `explicit`. 
Would it cause problems to merge it with the next constructor?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration.h
```diff
@@ -0,0 +1,2558 @@
...
+/**
+ * @class Multi_parameter_filtration Multi_parameter_filtration.h gudhi/Multi_parameter_filtration.h
+ * @ingroup multi_filtration
+ *
+ * @brief Class encoding the different generators, i.e., apparition times, of a \f$ k \f$-critical
+ * \f$\mathbb R^n\f$-filtration value. E.g., the filtration value of a simplex, or, of the algebraic generator of a
+ * module presentation. The encoding is compacted into a single vector, so if a lot of non trivial modifications are
+ * done (that not only consists of simply adding new generators at the end of the vector), it is probably preferable
+ * to use @ref Dynamic_multi_parameter_filtration instead. Implements the concept @ref FiltrationValue of the
+ * @ref Gudhi::Simplex_tree and the concept @ref Gudhi::multi_persistence::MultiFiltrationValue.
```
VincentRouvreau: `@ref Gudhi::multi_persistence::MultiFiltrationValue` does not exist (doxygen warning)

PR #978, opened 2023-10-02
Rework simplex tree options

### LINKED ISSUE #905: Simplex_tree_options_full_featured vs link_nodes_by_label
To be resolved BEFORE releasing GUDHI 3.9.

#877 introduced a feature of linking all the simplices that have the same maximal vertex, in order to speed up coface computations. For convenience, it disabled this feature for Simplex_tree_options_full_featured, because it is the default used almost everywhere in Gudhi.
In my opinion, the name Simplex_tree_options_full_featured implies that it should have this feature enabled. However, since this feature is costly (mostly in terms of memory, but it probably also slows things down a bit), we may not want it enabled everywhere, so we may want to create a new Simplex_tree_options_XXX to use by default or for Python's SimplexTree.
To help take a decision, it would be useful to do some benchmarking:

- check that #877 had no impact on memory or performance when link_nodes_by_label=false (it isn't supposed to change anything, this is just a sanity check)
- compare the running time and memory use of some examples that do not compute cofaces with/without link_nodes_by_label

VincentRouvreau:
* [x] move options in a separated header
 * [x] fast_persistence (no doc)
 * [x] python
 * [x] Rename Simplex_tree_interface_full_featured with Simplex_tree_python_interface
 * [x] Benchmark fast cofaces
 * [x] Replace fast_cofaces option with full_feature
 * [x] default
 * [x] minimal
 * [x] full_featured
 * [x] Benchmark stable simplex handles

Fix #905

### DIFF
--- a/src/Persistent_cohomology/example/plain_homology.cpp
+++ b/src/Persistent_cohomology/example/plain_homology.cpp
@@ -20,9 +20,9 @@
  * how to save on storage by not storing a filtration value.  */
 
 struct MyOptions : Gudhi::Simplex_tree_options_minimal {
-  // Implicitly use 0 as filtration value for all simplices
-  static const bool store_filtration = false;
-  // The persistence algorithm needs this
+  // Implicitly use 0 as filtration value for all simplices - inherited from Simplex_tree_options_minimal
+  // static const bool store_filtration = false;
+  // However the persistence algorithm needs this
   static const bool store_key = true;
   // I have few vertices
   typedef short Vertex_handle;
--- a/src/Persistent_cohomology/example/plain_homology.cpp
+++ b/src/Persistent_cohomology/example/plain_homology.cpp
@@ -16,10 +16,10 @@
 #include <cstdint>  // for std::uint8_t
 
 /* We could perfectly well use the default Simplex_tree<> (which uses
- * Simplex_tree_options_full_featured), the following simply demonstrates
+ * Simplex_tree_options_default), the following simply demonstrates
  * how to save on storage by not storing a filtration value.  */
 
-struct MyOptions : Gudhi::Simplex_tree_options_full_featured {
+struct MyOptions : Gudhi::Simplex_tree_options_minimal {
   // Implicitly use 0 as filtration value for all simplices
   static const bool store_filtration = false;
   // The persistence algorithm needs this
--- a/src/Simplex_tree/benchmark/simplex_tree_cofaces_benchmark.cpp
+++ b/src/Simplex_tree/benchmark/simplex_tree_cofaces_benchmark.cpp
@@ -105,10 +105,10 @@ int main(int argc, char *argv[]) {
     nb_vertices = atoi(argv[1]);
     
   std::clog << "** Without cofaces computation optimization" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_full_featured>>(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_default>>(nb_vertices);
 
   std::clog << "** With cofaces computation optimization" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_fast_cofaces>>(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_full_featured>>(nb_vertices);
 
   std::clog << "** With cofaces computation optimization and stable simplex handles" << std::endl;
   benchmark_stars_computation<Gudhi::Simplex_tree<Simplex_tree_options_stable_simplex_handles> >(nb_vertices);
--- a/src/Simplex_tree/benchmark/simplex_tree_cofaces_benchmark.cpp
+++ b/src/Simplex_tree/benchmark/simplex_tree_cofaces_benchmark.cpp
@@ -81,7 +81,7 @@ void benchmark_stars_computation(int nb_vertices) {
   std::clog << benchmark_cofaces << std::endl;
 }
 
-struct Simplex_tree_options_stable_simplex_handles {
+struct Stree_basic_cofaces_options {
   typedef Gudhi::linear_indexing_tag Indexing_tag;
   typedef int Vertex_handle;
   typedef double Filtration_value;
@@ -89,6 +89,15 @@ struct Simplex_tree_options_stable_simplex_handles {
   static const bool store_key = true;
   static const bool store_filtration = true;
   static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = false;
+  static const bool stable_simplex_handles = false;
+};
+
+struct Stree_fast_cofaces_options : Stree_basic_cofaces_options {
+  static const bool link_nodes_by_label = true;
+};
+
+struct Stree_fast_cofaces_stable_simplex_handles_options : Stree_basic_cofaces_options {
   static const bool link_nodes_by_label = true;
   static const bool stable_simplex_handles = true;
 };
@@ -105,13 +114,13 @@ int main(int argc, char *argv[]) {
     nb_vertices = atoi(argv[1]);
     
   std::clog << "** Without cofaces computation optimization" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_default>>(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Stree_basic_cofaces_options>>(nb_vertices);
 
   std::clog << "** With cofaces computation optimization" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_full_featured>>(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Stree_fast_cofaces_options>>(nb_vertices);
 
   std::clog << "** With cofaces computation optimization and stable simplex handles" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Simplex_tree_options_stable_simplex_handles> >(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Stree_fast_cofaces_stable_simplex_handles_options> >(nb_vertices);
 
   return EXIT_SUCCESS;
 }
--- a/src/Simplex_tree/example/mini_simplex_tree.cpp
+++ b/src/Simplex_tree/example/mini_simplex_tree.cpp
@@ -12,7 +12,7 @@
 #include <iostream>
 #include <initializer_list>
 
-struct MyOptions : Gudhi::Simplex_tree_options_full_featured {
+struct MyOptions : Gudhi::Simplex_tree_options_minimal {
   // Not doing persistence, so we don't need those
   static const bool store_key = false;
   static const bool store_filtration = false;
--- a/src/Simplex_tree/example/mini_simplex_tree.cpp
+++ b/src/Simplex_tree/example/mini_simplex_tree.cpp
@@ -12,11 +12,10 @@
 #include <iostream>
 #include <initializer_list>
 
+// With Gudhi::Simplex_tree_options_minimal filtration values are not stored and persistence cannot be computed.
+// I can still improve the memory footprint
 struct MyOptions : Gudhi::Simplex_tree_options_minimal {
-  // Not doing persistence, so we don't need those
-  static const bool store_key = false;
-  static const bool store_filtration = false;
-  // I have few vertices
+  // As I have few vertices
   typedef short Vertex_handle;
 };
 
--- a/src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
@@ -1,0 +1,79 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+ *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+ *    Author(s):       Vincent Rouvreau
+ *
+ *    Copyright (C) 2023 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#ifndef SIMPLEX_TREE_SIMPLEX_TREE_OPTIONS_H_
+#define SIMPLEX_TREE_SIMPLEX_TREE_OPTIONS_H_
+
+#include <gudhi/Simplex_tree/indexing_tag.h>
+
+#include <cstdint>
+
+namespace Gudhi {
+
+/** \addtogroup simplex_tree 
+ * Pre-defined options for the Simplex_tree.
+ *  @{
+ */
+
+/** Model of SimplexTreeOptions.
+ * 
+ * Maximum number of simplices to compute persistence is <CODE>std::numeric_limits<std::uint32_t>::max()</CODE>
+ * (about 4 billions of simplices). */
+struct Simplex_tree_options_full_featured {
+  typedef linear_indexing_tag Indexing_tag;
+  typedef int Vertex_handle;
+  typedef double Filtration_value;
+  typedef std::uint32_t Simplex_key;
+  static const bool store_key = true;
+  static const bool store_filtration = true;
+  static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = false;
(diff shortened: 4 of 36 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -0,0 +1,93 @@
...
+
+/** Model of SimplexTreeOptions.
+ * 
+ * Minimal version of the Simplex_tree. No filtration values are stored and it is impossible to compute persistence
+ * with these options. */
+struct Simplex_tree_options_minimal {
+  typedef linear_indexing_tag Indexing_tag;
+  typedef int Vertex_handle;
+  typedef double Filtration_value;
+  typedef std::uint32_t Simplex_key;
```
mglisse, line 64: Ah, yes, we are still forced to define those 2. Ok I guess.

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -0,0 +1,93 @@
...
+  static const bool store_key = false;
+  static const bool store_filtration = false;
+  static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = false;
+  static const bool stable_simplex_handles = false;
+};
+
+/** @private @brief Model of SimplexTreeOptions, faster than `Simplex_tree_options_full_featured` but note the unsafe
```
mglisse, line 72:
```suggestion
/** @private @brief Model of SimplexTreeOptions, faster than `Simplex_tree_options_default` but note the unsafe
```

src/Simplex_tree/test/simplex_tree_make_filtration_non_decreasing_unit_test.cpp
```diff
@@ -37,7 +37,7 @@ struct Simplex_tree_options_stable_simplex_handles {
 
 typedef boost::mpl::list<Simplex_tree<>,
                          Simplex_tree<Simplex_tree_options_fast_persistence>,
-                         Simplex_tree<Simplex_tree_options_fast_cofaces>,
+                         Simplex_tree<Simplex_tree_options_full_featured>,
```
mglisse, line 40: Just checking: do we have at least one test that has options link=true, stable=false?

src/python/include/Euclidean_witness_complex_interface.h
```diff
@@ -47,11 +47,11 @@ class Euclidean_witness_complex_interface {
     delete witness_complex_;
   }
 
-  void create_simplex_tree(Gudhi::Simplex_tree<>* simplex_tree, double max_alpha_square, std::size_t limit_dimension) {
+  void create_simplex_tree(Simplex_tree<Simplex_tree_options_for_python>* simplex_tree, double max_alpha_square, std::size_t limit_dimension) {
```
mglisse, line 50: Is there a reason not to use `Simplex_tree_interface` here (and related places)?

src/Persistent_cohomology/example/plain_homology.cpp
```diff
@@ -16,10 +16,10 @@
...
 
 /* We could perfectly well use the default Simplex_tree<> (which uses
- * Simplex_tree_options_full_featured), the following simply demonstrates
+ * Simplex_tree_options_default), the following simply demonstrates
  * how to save on storage by not storing a filtration value.  */
 
-struct MyOptions : Gudhi::Simplex_tree_options_full_featured {
+struct MyOptions : Gudhi::Simplex_tree_options_minimal {
   // Implicitly use 0 as filtration value for all simplices
   static const bool store_filtration = false;
```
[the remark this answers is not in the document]
VincentRouvreau: Or maybe we should not derive from any existing options and set all the options we need for this example.
Tell me your preference.

src/Simplex_tree/benchmark/simplex_tree_cofaces_benchmark.cpp
```diff
@@ -105,10 +105,10 @@ int main(int argc, char *argv[]) {
     nb_vertices = atoi(argv[1]);
     
   std::clog << "** Without cofaces computation optimization" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_full_featured>>(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_default>>(nb_vertices);
 
   std::clog << "** With cofaces computation optimization" << std::endl;
-  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_fast_cofaces>>(nb_vertices);
+  benchmark_stars_computation<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_full_featured>>(nb_vertices);
```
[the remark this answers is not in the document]
VincentRouvreau: That's clearer and won't be broken by any modification. Done on ae70563df

src/Simplex_tree/example/mini_simplex_tree.cpp
```diff
@@ -12,7 +12,7 @@
 #include <iostream>
 #include <initializer_list>
 
-struct MyOptions : Gudhi::Simplex_tree_options_full_featured {
+struct MyOptions : Gudhi::Simplex_tree_options_minimal {
   // Not doing persistence, so we don't need those
   static const bool store_key = false;
   static const bool store_filtration = false;
```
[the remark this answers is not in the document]
VincentRouvreau: More comments on 69e79561a

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -0,0 +1,93 @@
...
+  static const bool store_filtration = true;
+  static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = false;
+  static const bool stable_simplex_handles = false;
+};
+
+/** Model of SimplexTreeOptions.
+ * 
+ * Maximum number of simplices to compute persistence is <CODE>std::numeric_limits<std::uint32_t>::max()</CODE>
+ * (about 4 billions of simplices). */
```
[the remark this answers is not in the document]
VincentRouvreau: Add some details on Simplex_tree_options_default on 64cebe941

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -0,0 +1,93 @@
...
+struct Simplex_tree_options_full_featured {
+  typedef linear_indexing_tag Indexing_tag;
+  typedef int Vertex_handle;
+  typedef double Filtration_value;
+  typedef std::uint32_t Simplex_key;
+  static const bool store_key = true;
+  static const bool store_filtration = true;
+  static const bool contiguous_vertices = false;
+  static const bool link_nodes_by_label = true;
+  static const bool stable_simplex_handles = false;
```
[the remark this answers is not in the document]
VincentRouvreau: Yes, I was in wait of #980 
`true` on b534bcf22

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
VincentRouvreau: Yes. We could remove `Simplex_key`, but some modifications needs to be done in mini_simplex_tree.cpp example, in simplex_tree_remove_unit_test and simplex_tree_iostream_operator_unit_test unitary tests

src/python/include/Euclidean_witness_complex_interface.h
VincentRouvreau: Here it is ok, so I changed it on 55f9f8e51
But there is an issue with Tangential_complex _interface while using `create_complex`. I added a comment on the problematic line

src/python/include/Tangential_complex_interface.h
```diff
@@ -88,8 +88,8 @@ class Tangential_complex_interface {
       num_inconsistencies_ = tangential_complex_->number_of_inconsistent_simplices();
   }
 
-  void create_simplex_tree(Simplex_tree<>* simplex_tree) {
-    tangential_complex_->create_complex<Gudhi::Simplex_tree<Gudhi::Simplex_tree_options_full_featured>>(*simplex_tree);
+  void create_simplex_tree(Simplex_tree<Simplex_tree_options_for_python>* simplex_tree) {
+    tangential_complex_->create_complex<Simplex_tree<Simplex_tree_options_for_python>>(*simplex_tree);
```
VincentRouvreau, line 92:
```suggestion
    tangential_complex_->create_complex<Simplex_tree_interface>(*simplex_tree);
```
Here, when it calls `insert_simplex_and_subfaces`, it calls the method from Simplex_tree_interface and it fails because it is not returning the type required by tangential complex

mglisse: So the issue is that `Simplex_tree_interface::insert_simplex_and_subfaces` overrides the functions of the base class but has an incompatible API? Shouldn't we rename that function so it does not clash with the one in the base class then?

mglisse: Actually, that function does not seem to be used anywhere, can't we remove it?

src/Persistent_cohomology/example/plain_homology.cpp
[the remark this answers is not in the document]
VincentRouvreau: Rephrase simplex tree options for this example on 0a82c1ef9

src/python/include/Tangential_complex_interface.h
VincentRouvreau: Oh, ok ! That should be some dead code. I removed it and now my suggestion is possible on a6c5e0ae7

VincentRouvreau:
@hschreiber @mglisse Some benchmarks to decide what options we do activate for the Python interface
| link_nodes_by_label | stable_simplex_handles | OFF filename | Rips max_edge_length | Rips construction in sec | Rips construction memory in Mo | SimplexTree construction in sec | SimplexTree construction memory in Mo | SimplexTree number of simplices | Persistence computation in sec | Persistence computation memory in Mo |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| false | false | tore3D_300.off | 3.5 | 0.002 | 1.84 | 1.152 | 1859.43 | 57770908 | 62.260 | 14228.86 | 
| true | false | tore3D_300.off | 3.5 | 0.002 | 1.85 | 1.878 | 2741.14 | 57770908 | 67.564 | 14228.80 | 
| false | true | tore3D_300.off | 3.5 | 0.002 | 1.81 | 3.412 | 3622.38 | 57770908 | 78.363 | 14228.73 | 
| true | true | tore3D_300.off | 3.5 | 0.002 | 1.80 | 4.342 | 4503.83 | 57770908 | 81.205 | 14228.78 | 

| link_nodes_by_label | stable_simplex_handles | OFF filename |  Rips construction in sec | Rips construction memory in Mo | SimplexTree construction in sec | SimplexTree construction memory in Mo | SimplexTree number of simplices | edge collapse construction in sec | edge collapse construction memory in Mo | edge collapse number of simplices | expansion(5) construction in sec | expansion(5) construction memory in Mo | expansion(5) number of simplices | Persistence computation in sec | Persistence computation memory in Mo |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| false | false | 5-forex-2500.off | 0.151 | 150.76 | 0.076 | 103.18 | 3126250 | 14.341 | 6.08 | 9304 | 0.001 | 0.00 | 20925 | 0.005 | 0.00 | 
| true | false | 5-forex-2500.off | 0.152 | 150.75 | 0.155 | 151.65 | 3126250 | 14.141 | 5.60 | 9304 | 0.001 | 0.00 | 20925 | 0.005 | 0.00 | 
| false | true | 5-forex-2500.off | 0.153 | 150.73 | 0.245 | 190.84 | 3126250 | 14.407 | 4.74 | 9304 | 0.002 | 0.00 | 20925 | 0.006 | 0.00 | 
| true | true | 5-forex-2500.off | 0.150 | 150.72 | 0.278 | 238.53 | 3126250 | 14.293 | 6.07 | 9304 | 0.002 | 0.00 | 20925 | 0.006 | 0.00 | 

| link_nodes_by_label | stable_simplex_handles | OFF filename | Alpha precision | Alpha construction in sec | Alpha construction memory in Mo | SimplexTree construction in sec | SimplexTree construction memory in Mo | SimplexTree number of simplices | Persistence computation in sec | Persistence computation memory in Mo |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| false | false | 5-forex-2500.off | fast | 2.041 | 48.97 | 3.364 | 212.66 | 2278347 | 1.615 | 17.99 | 
| false | false | 5-forex-2500.off | safe | 2.013 | 49.80 | 7.064 | 664.78 | 2278347 | 1.644 | 0.55 | 
| false | false | 5-forex-2500.off | exact | 1.954 | 49.73 | 23.352 | 720.63 | 2278347 | 1.585 | 14.16 | 
| true | false | 5-forex-2500.off | fast | 2.113 | 49.00 | 3.802 | 249.55 | 2278347 | 1.790 | 17.99 | 
| true | false | 5-forex-2500.off | safe | 1.831 | 49.76 | 7.330 | 701.88 | 2278347 | 1.925 | 0.55 | 
| true | false | 5-forex-2500.off | exact | 1.900 | 49.71 | 23.676 | 757.68 | 2278347 | 1.776 | 0.55 | 
| false | true | 5-forex-2500.off | fast | 2.056 | 48.96 | 4.041 | 278.04 | 2278347 | 1.991 | 17.92 | 
| false | true | 5-forex-2500.off | safe | 1.903 | 49.75 | 7.582 | 730.27 | 2278347 | 2.032 | 0.54 | 
| false | true | 5-forex-2500.off | exact | 1.834 | 49.78 | 23.621 | 786.18 | 2278347 | 1.966 | 13.43 | 
| true | true | 5-forex-2500.off | fast | 2.067 | 48.95 | 4.155 | 312.95 | 2278347 | 2.121 | 17.93 | 
| true | true | 5-forex-2500.off | safe | 1.820 | 49.74 | 7.648 | 765.29 | 2278347 | 2.121 | 0.55 | 
| true | true | 5-forex-2500.off | exact | 1.817 | 49.70 | 23.759 | 821.02 | 2278347 | 2.153 | 14.87 |

mglisse:
For `stable_simplex_handles`, I think we said there was no point activating it in Python since for now it has 0 benefit.
For `link_nodes_by_label`, I think I'd be in favor of keeping it `false` for now (same behavior as previous releases) and possibly revisiting later. My reasoning is that it increases the size of the SimplexTree significantly (+50%), and while it makes finding cofaces less ridiculously slow on sparse complexes, that operation is still way slower than finding faces, so people who need really fast cofaces would be better off using a different datastructure that explicitly stores the links of the Hasse diagram. I am open to counter-arguments though.

hschreiber:
Yes, `stable_simplex_handles` should be `false` for python, as the user never accesses the simplex handles directly, so it slows down things for nothing.
For `link_nodes_by_label`, I admit, I don't know how much python user need cofaces when using the simplex tree. Do we have feedback on this?

DavidLapous:
A possibility, which is not headache-free, is to have both in the same simplextree : as the current python interface only stores the simplextree c++ pointer, we just need to add a private member in the python simplextree class,  telling  by run time how to reinterpret_cast the simplextree in c++. This may cost another python call for each method, but it should be fine. To avoid confusion of python users, it might also be useful to define `__repr__` to differentiate them.

This can also apply to PR #976 .

mglisse:
> For `link_nodes_by_label`, I admit, I don't know how much python users need cofaces when using the simplex tree. Do we have feedback on this?

A search in the issues and gudhi-users finds surprisingly little. I know some people have used cofaces on examples small enough that they did not care about performance. The few cases I remember where people cared about the speed of cofaces were for C++.

> A possibility, which is not headache-free, is to have both in the same simplextree : as the current python interface only stores the simplextree c++ pointer, we just need to add a private member in the python simplextree class, telling by run time how to reinterpret_cast the simplextree in c++.

So essentially a `std::variant` or a virtual base (though we may use other technical means to achieve it). That's a possibility. It has a cost though, both in speed (although the difference may be negligible depending on how we implement it) and maintenance (it increases and complicates the code), so it needs to be worth the trouble. But if we do decide that we want "fast" cofaces without forcing a slowdown on other operations and increasing the use of memory, that does look like a sensible approach.

For multi-persistence, while it still seems possible, my first reaction without thinking much about it is that the interfaces are likely to differ enough that a separate Python class makes more sense.

mglisse (approved):
We will need to say something in the release notes about it.

VincentRouvreau:
> We will need to say something in the release notes about it.

Ok, I will propose something

Issue #986, opened 2023-10-11
Re-generate doxygen headers


VincentRouvreau:
https://github.com/GUDHI/gudhi-devel/blob/7fb68dacfdabd39c060cd5dda9d8f5fa0aa1b250/src/common/doc/header.html#L4

http://www.w3.org/1999/xhtml is no more reachable.

VincentRouvreau:
This is not an issue, this link is reachable, my bad.

PR #987, opened 2023-10-11
New footer, header and style sheet with doxygen 1.9.5

VincentRouvreau:
cf. #986


VincentRouvreau:
> Out of curiosity, does updating these mean that doxygen 1.9.5 is the new minimum version to build the doc, or is it compatible with older versions?

This is what I want to try with the CI. I have no idea, but the changes does not seem that important.
I wanted also to see if `<html xmlns="http://www.w3.org/1999/xhtml" lang="$langISO">` was still the good way to declare xmlns, but it seems so.

Issue #882, opened 2023-05-22
[Cover complex] sklearn AgglomerativeClustering: Attribute `affinity` is deprecated


VincentRouvreau:
https://github.com/GUDHI/gudhi-devel/blob/8adafa5cb67bb116854d85a7542046ff47aa83c4/src/python/gudhi/cover_complex.py#L328-L330

Attribute `affinity` was deprecated in version 1.2 and will be removed in 1.4. Use `metric` instead.
cf. https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html

PR #990, opened 2023-10-25
fixed deprecated argument

MathieuCarriere:
Fix for #882

### DIFF
--- a/src/python/gudhi/cover_complex.py
+++ b/src/python/gudhi/cover_complex.py
@@ -325,9 +325,9 @@ class MapperComplex(CoverComplexPy):
             delta, resolutions = self.get_optimal_parameters_for_agglomerative_clustering(X=X, beta=self.beta, C=self.C, N=self.N)
             if self.clustering is None:
                 if self.input_type == "point cloud":
-                    self.clustering = AgglomerativeClustering(n_clusters=None, linkage="single", distance_threshold=delta, affinity="euclidean")
+                    self.clustering = AgglomerativeClustering(n_clusters=None, linkage="single", distance_threshold=delta, metric="euclidean")
                 else:
-                    self.clustering = AgglomerativeClustering(n_clusters=None, linkage="single", distance_threshold=delta, affinity="precomputed")
+                    self.clustering = AgglomerativeClustering(n_clusters=None, linkage="single", distance_threshold=delta, metric="precomputed")
             if self.resolutions is None:
                 self.resolutions = np.multiply(resolutions, 1./self.gains)
                 self.resolutions = np.array([int( (self.filter_bnds[ir,1]-self.filter_bnds[ir,0])/r) for ir, r in enumerate(self.resolutions)])
--- a/src/python/gudhi/cover_complex.py
+++ b/src/python/gudhi/cover_complex.py
@@ -17,6 +17,10 @@ from sklearn.cluster         import DBSCAN, AgglomerativeClustering
 from sklearn.metrics         import pairwise_distances
 from scipy.spatial.distance  import directed_hausdorff
 
+from sklearn import __version__ as sklearn_version
+from pkg_resources import packaging
+agglomerative_clustering_metric = packaging.version.parse(sklearn_version) >= packaging.version.parse('1.2.0')
+
 from . import SimplexTree, CoverComplex
 
 class CoverComplexPy(BaseEstimator):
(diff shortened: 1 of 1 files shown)


mglisse:
Just checking: do we have to mention somewhere that gudhi requires version at least x.y of this dependency? (I assume `metric` was not available in the past)

MathieuCarriere:
Yes, using argument `metric` won't work for Scikit-Learn < 1.2, which only knows `affinity`. So the code of this PR only works for Scikit-Learn version 1.2 or more.

VincentRouvreau:
> Yes, using argument `metric` won't work for Scikit-Learn < 1.2, which only knows `affinity`. So the code of this PR only works for Scikit-Learn version 1.2 or more.

I confirm : https://scikit-learn.org/stable/whats_new/v1.2.html#id12
As scikit-learn 1.2.0 is quite recent (december 2022), I won't ask for scikit-learn >= 1.2.0 as a dependency.
We could do something like:
```python
from sklearn import __version__ as sklearn_version
from pkg_resources import packaging

agglomerative_clustering_metric = packaging.version.parse(sklearn_version) >= packaging.version.parse('1.2.0')

...
            if self.clustering is None:
                if self.input_type == "point cloud":
                    kwarg =  {'metric':'euclidean'} if agglomerative_clustering_metric else {'affinity':'euclidean'}
                    self.clustering = AgglomerativeClustering(n_clusters=None, linkage="single", distance_threshold=delta, **kwarg)
                else:
                    kwarg =  {'metric':'precomputed'} if agglomerative_clustering_metric else {'affinity':'precomputed'}
                    self.clustering = AgglomerativeClustering(n_clusters=None, linkage="single", distance_threshold=delta, **kwarg)
```

MathieuCarriere:
Done!

mglisse (approved):
Another possibility, to avoid repetitions, would have been
```python
self.clustering = AgglomerativeClustering(
    n_clusters=None,
    linkage="single",
    distance_threshold=delta,
    **{
        "metric" if packaging.version.parse(sklearn_version) >= packaging.version.parse("1.2.0") else "affinity":
        "euclidean" if self.input_type == "point cloud" else "precomputed"
    })
```
but as long as it works, this branch is fine.

src/python/gudhi/cover_complex.py
```diff
@@ -17,6 +17,10 @@
 from sklearn.metrics         import pairwise_distances
 from scipy.spatial.distance  import directed_hausdorff
 
+from sklearn import __version__ as sklearn_version
+from pkg_resources import packaging
```
mglisse, line 21: Wait, does that come with Python, or is it a new dependency? It seems deprecated...

VincentRouvreau: @mglisse It seems to be the way it is done by setuptools (and it is required by our python package), cf. https://peps.python.org/pep-0386/#setuptools (I have read it is available for python 3.6+ in this [this stackoverflow](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python), but I cannot find the information)

mglisse: https://setuptools.pypa.io/en/latest/pkg_resources.html says
> Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Some useful APIs are also provided by [packaging](https://pypi.org/project/packaging) (e.g. requirements and version parsing). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

setuptools is currently a dependency for **building** gudhi, but not for **running** it, so using it there would require some configuration changes.

src/python/gudhi/cover_complex.py
```diff
@@ -17,6 +17,10 @@
...
+agglomerative_clustering_metric = packaging.version.parse(sklearn_version) >= packaging.version.parse('1.2.0')
```
VincentRouvreau, line 22: @mglisse you are right and pep 386 is superseded... Sorry @MathieuCarriere for the bad pointer... sklearn has its own `parse_version` implementation that seems to be the right one to use:
```suggestion
from sklearn.utils.fixes import parse_version
agglomerative_clustering_metric = parse_version(sklearn_version) >= parse_version('1.2.0')
```

MathieuCarriere: No worries! I've just done the change.

VincentRouvreau (approved):
@mglisse is it ok now for you ?

mglisse (approved):

PR #994, opened 2023-10-31
Fix pip OSx universal dynamic libraries

VincentRouvreau:
It seems exact file name is required


mglisse:
> ls: deps-amd64/gmp/*/lib/libgmp.*.dylib: No such file or directory

Doesn't look like it. I think the 2 warnings before that are what we should look at:
> Warning: Bottle for tag :big_sur is unavailable.
Warning: Bottle for tag :arm64_big_sur is unavailable.

VincentRouvreau:
Fixed by #995

PR #999, opened 2023-11-13
KNearestNeighbors interface homogenization + empty DTMRipsComplex

VincentRouvreau:
Motivated by [this comment](https://github.com/GUDHI/gudhi-devel/blob/e49ca57f7cd2ad3cb85baded9e47cd16f8f21d7e/src/python/gudhi/dtm_rips_complex.py#L44)

## current behaviour
```python
from gudhi.point_cloud.dtm import DistanceToMeasure
import numpy as np
DistanceToMeasure(0, implementation="ckdtree").fit_transform(numpy.random.rand(1000, 4))
# ValueError: zero-size array to reduction operation maximum which has no identity
DistanceToMeasure(0, implementation="sklearn").fit_transform(numpy.random.rand(1000, 4))
# ValueError: Expected n_neighbors > 0. Got 0
DistanceToMeasure(0, implementation="hnsw").fit_transform(numpy.random.rand(1000, 4))
# /home/gailuron/workspace/gudhi/gudhi-devel/build/src/python/gudhi/point_cloud/dtm.py:67: RuntimeWarning: invalid value encountered in true_divide
#   dtm = distances.sum(-1) / self.k
# array([nan, ..., nan],
#       dtype=float32)

# ==========================================================================================

DistanceToMeasure(1001, implementation="ckdtree").fit_transform(numpy.random.rand(1000, 4))
# array([inf, ..., inf])
DistanceToMeasure(1001, implementation="sklearn").fit_transform(numpy.random.rand(1000, 4))
# ValueError: Expected n_neighbors <= n_samples,  but n_samples = 1000, n_neighbors = 1001
DistanceToMeasure(1001, implementation="hnsw").fit_transform(numpy.random.rand(1000, 4))
# RuntimeError: Cannot return the results in a contigious 2D array. Probably ef or M is too small
```
(This behaviour is inherited from `KNearestNeighbors` class)
## Proposal
To me, `sklearn` behaves the most appropriately from a user point of view and `KNearestNeighbors` class should stick to this user interface.

### DIFF
--- a/src/python/gudhi/dtm_rips_complex.py
+++ b/src/python/gudhi/dtm_rips_complex.py
@@ -37,7 +37,7 @@ class DTMRipsComplex(WeightedRipsComplex):
         if distance_matrix is None:
             if points is None:
                 # Empty Rips construction
-                points=[]
+                points=[[]]
             distance_matrix = cdist(points,points)
         self.distance_matrix = distance_matrix
 
--- a/src/python/gudhi/dtm_rips_complex.py
+++ b/src/python/gudhi/dtm_rips_complex.py
@@ -12,40 +12,37 @@ from gudhi.weighted_rips_complex import WeightedRipsComplex
 from gudhi.point_cloud.dtm import DistanceToMeasure
 from scipy.spatial.distance import cdist
 
+
 class DTMRipsComplex(WeightedRipsComplex):
     """
-    Class to generate a DTM Rips complex from a distance matrix or a point set, 
+    Class to generate a DTM Rips complex from a distance matrix or a point set,
     in the way described in :cite:`dtmfiltrations`.
-    Remark that all the filtration values are doubled compared to the definition in the paper 
+    Remark that all the filtration values are doubled compared to the definition in the paper
     for the consistency with RipsComplex.
     :Requires: `SciPy <installation.html#scipy>`_
     """
-    def __init__(self, 
-                 points=None, 
-                 distance_matrix=None, 
-                 k=1, 
-                 q=2,
-                 max_filtration=float('inf')):
+
+    def __init__(self, points=None, distance_matrix=None, k=1, q=2, max_filtration=float("inf")):
         """
         Args:
             points (numpy.ndarray): array of points.
             distance_matrix (numpy.ndarray): full distance matrix.
-            k (int): number of neighbors for the computation of DTM. Defaults to 1, which is equivalent to the usual Rips complex.
+            k (int): number of neighbors for the computation of DTM. Defaults to 1, which is equivalent to the usual
+                Rips complex.
             q (float): order used to compute the distance to measure. Defaults to 2.
-            max_filtration (float): specifies the maximal filtration value to be considered.      
+            max_filtration (float): specifies the maximal filtration value to be considered.
         """
         if distance_matrix is None:
             if points is None:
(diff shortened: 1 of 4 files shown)


src/python/gudhi/dtm_rips_complex.py
```diff
@@ -12,40 +12,37 @@
...
             q (float): order used to compute the distance to measure. Defaults to 2.
-            max_filtration (float): specifies the maximal filtration value to be considered.      
+            max_filtration (float): specifies the maximal filtration value to be considered.
         """
         if distance_matrix is None:
             if points is None:
                 # Empty Rips construction
-                points=[]
-            distance_matrix = cdist(points,points)
+                points = [[]]
```
mglisse, line 38: I assume this is to avoid an error with `cdist([],[])`? But with this change, we get a distance matrix `[[0]]` of shape (1,1), which should correspond to the case with 1 point, whereas it would seem better to get an empty matrix (possibly of shape (0,0) if that helps). If points is None, couldn't we skip the call to cdist?

src/python/gudhi/point_cloud/knn.py
```diff
@@ -90,6 +96,10 @@ def fit(self, X, y=None):
         Args:
             X (numpy.array): coordinates for reference points.
         """
+        if self.k > len(X):
+            raise ValueError(
```
mglisse, line 100: Raising an error certainly makes sense. I am trying to think if there could be other useful behaviors. For instance, if we compute the DTM-filtration of many point sets with the same `k` and one of the point sets happens to have fewer than `k` points, what should we do? Probably lower `k` to the number of points (not sure what else we could do if we don't want an error). But then where/how should we lower `k`, in knn? in dtm? in weightedrips? in dtmfiltration? in the caller of dtmfiltration?
Just leaving this comment here so I can think about it at a better time, but probably raising an error here as you did is good.

src/python/test/test_dtm_rips_complex.py
```diff
@@ -30,3 +34,9 @@ def test_compatibility_with_rips():
     st_from_rips = rips_complex.create_simplex_tree(max_dimension=1)
     assert list(st.get_filtration()) == list(st_from_rips.get_filtration())
 
+
+def test_empty_dtm_rips_complex():
+    dtm_rips = DTMRipsComplex()
+    st = dtm_rips.create_simplex_tree(max_dimension=1)
+    assert st.num_simplices() == 1
```
mglisse, line 41: Why is there a vertex if there are no points?

src/python/gudhi/dtm_rips_complex.py
VincentRouvreau: Oh yes this is better this way, done on [98e1ce1](https://github.com/GUDHI/gudhi-devel/pull/999/commits/98e1ce14bd2ddb3475d06f88b3c61c3b846afabc)

src/python/test/test_dtm_rips_complex.py
VincentRouvreau: Was part of your other comment. Fixed on [98e1ce1](https://github.com/GUDHI/gudhi-devel/pull/999/commits/98e1ce14bd2ddb3475d06f88b3c61c3b846afabc)

mglisse (approved):

PR #1007, opened 2023-11-29
Use find_package(Python) instead of PythonInterp

### LINKED ISSUE #139: FindPythonInterp deprecated

VincentRouvreau:
Modification taken from #151 
Fix #139 
CMake ~~3.12 is now 5 years old~~ 3.15 is now 4 years old (july 2019), so it should be ok for every one.
CMake 3.15 is required for `cmake_policy(SET CMP0094 NEW)`
CMake 3.16 (nov 2019) is required if the user wants to specify a python interpreter (`cmake -DPython_EXECUTABLE=/custom/path/to/python3 ...`)


.github/workflows/pip-build-osx.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           python --version
           mkdir build
           cd build
-          cmake -DCMAKE_BUILD_TYPE=Release -DPython_ADDITIONAL_VERSIONS=3 ..
+          cmake -DCMAKE_BUILD_TYPE=Release -DPython_EXECUTABLE=$(which python) ..
```
mglisse, line 48: Why do we need that?

VincentRouvreau: I precised it on [7e33f42](https://github.com/GUDHI/gudhi-devel/pull/1007/commits/7e33f42f936e036ed2c4a5da61e3fa71c8080229) because macosx-11.7-universal2 has several [python installation available](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#python) and cmake was not finding the correct one, cf. https://github.com/GUDHI/gudhi-devel/actions/runs/7030598162/job/19130456933#step:5:60
When I precise `Python_EXECUTABLE`, the error is no more.
I wanted to do the same for Windows, but my batch command seems not correct, WIP

VincentRouvreau: I didn't notice it the first time I wrote the PR, but setting policy CMP0094 to NEW seems promising as it is using the python interpreter from my conda environment instead of the one from my system. It worth it to test it, even if it requires cmake 3.15 (july 2019). I rewrite some python cmake horrors and I will give it a try

Issue #188, opened 2020-01-04
[Python] Use numpy more in interfaces


mglisse:
reader_utils, subsampling, etc return (and take) lists of lists (that's what cython automatically does with vectors of vectors). It would be better if they returned a numpy array (preferably directly, **not** by building a list first and then converting it to numpy array).

VincentRouvreau:
WIP and started by #498

PR #1008, opened 2023-12-04
numpy inteface for read_points_from_off_file

VincentRouvreau:
Part of #188


src/python/gudhi/off_utils.pyx
```diff
@@ -32,14 +33,14 @@ def read_points_from_off_file(off_file=''):
...
 
     :returns:  The point set.
-    :rtype: List[List[float]]
+    :rtype: numpy.ndarray
     """
     if off_file:
         if os.path.isfile(off_file):
-            return read_points_from_OFF_file(off_file.encode('utf-8'))
+            points = read_points_from_OFF_file(off_file.encode('utf-8'))
+            return np.array(points, copy=False)
```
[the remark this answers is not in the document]
mglisse: I mostly object to `copy=False` because it is misleading. `copy=True` would be useless but at least it would document reality.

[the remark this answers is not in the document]
VincentRouvreau: I tried `numpy loadtxt` and it worth it, it takes `0:00:00.210199` with the same example above (against almost 6 sec).

I also added some tests.

src/python/gudhi/off_utils.pyx
```diff
@@ -32,14 +33,22 @@ def read_points_from_off_file(off_file=''):
...
+            if line.startswith("off"):
+                index = 0
+            elif line.startswith("noff"):
+                index = 1
+            else:
+                raise ValueError(f"Inconsistent OFF header for {off_file}, got '{line.rstrip()}', should be 'OFF' or 'nOFF'")
+            # Number of points is the first number ("OFF" case) or the second one ("nOFF" case) of the second line
+            nb_points = int(next(input_file).split()[index])
+        
+        return np.loadtxt(off_file, skiprows=2, max_rows=nb_points, dtype=np.floating)
```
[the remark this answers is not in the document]
VincentRouvreau: I provided 2 new methods on [be5cfce](https://github.com/GUDHI/gudhi-devel/pull/1008/commits/be5cfced2a7e26f14f10d4d34e05d68a70a815ba):
* `_get_next_line` that reaches the next line without comments
* `_read_off_file_header` that returns all necessary information from off file reader. Can be reused if we decide to get faces from OFF files with `numpy.loadtxt`

This is now conform to what is explained in wikipedia, and with OFF files that are in `data\points`

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,91 @@ __license__ = "MIT"
...
-        if os.path.isfile(off_file):
-            return read_points_from_OFF_file(off_file.encode('utf-8'))
-        else:
-            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT),
-                                    off_file)
+    skiprows, dim, nb_points = _read_off_file_header(off_file)
+    print(skiprows, dim, nb_points)
+    # usecols=list(range(dim)) stands here to avoid comments at the end of line
+    # or colors that can be added in RGB format after the points, the faces, ...
+    return np.loadtxt(off_file, dtype=np.floating, comments='#', skiprows=skiprows,
```
mglisse, line 112: loadtxt accepts a file instead of a filename, would it work to "reuse" the file we already opened?

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,95 @@ __license__ = "MIT"
...
+    if nb_vertices < 0:
+        # Number of points is the first number ("OFF" case) or the second one ("nOFF" case) of the second line
+        line = _get_next_line(input_file)
+        digits = [int(s) for s in line.split() if s.isdigit()]
+        nb_vertices = digits[0]
+        # nb_faces =  digits[1]
+        # nb_edges =  digits[2] # not used - can be ignored
+        # nb_cells =  digits[3]
+    # If comments between header and the first point
+    line = _get_next_line(input_file)
```
mglisse: Doesn't loadtxt already ignore comments?

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,95 @@ __license__ = "MIT"
...
+    """
+    nb_vertices = -1
+
+    line = _get_next_line(file_desc)
+    # First line should be "OFF" (3d case) or "nOFF" (dD case)
+    if line.lower().startswith("off"):
+        dim = 3
+    elif line.lower().startswith("4off"):
+        dim = 4
+    elif line.lower().startswith("noff"):
```
mglisse, line 68: In the C++ code, we are a bit more flexible and accept things like COFF (see http://www.geomview.org/docs/html/OFF.html). I don't know if that's important.

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,95 @@ __license__ = "MIT"
...
-        else:
-            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT),
-                                    off_file)
+    with open(off_file) as input_file:
+        dim, nb_points = _read_off_file_header(input_file)
+        print(dim, nb_points)
+        # usecols=list(range(dim)) stands here to avoid comments at the end of line
+        # or colors that can be added in RGB format after the points, the faces, ...
+        return np.loadtxt(input_file, dtype=np.floating, comments='#',
+                          usecols=list(range(dim)), max_rows=nb_points)
```
mglisse, line 117: Should we add a sanity check that the output shape is (nb_points,dim)?

mglisse, line 117: Doesn't `usecols=range(dim)` work? The documentation says it accepts a _sequence_.

VincentRouvreau: This is a good practice, specifically when numpy < 1.23.0 and comments between points for instance. Done on [0b19e5a](https://github.com/GUDHI/gudhi-devel/pull/1008/commits/0b19e5ad2a7fb0e9545c733f34f5e3ca24371675)

src/python/gudhi/off_utils.pyx
```diff
@@ -110,11 +110,12 @@ def read_points_from_off_file(off_file=''):
     """
     with open(off_file) as input_file:
         dim, nb_points = _read_off_file_header(input_file)
-        print(dim, nb_points)
         # usecols=list(range(dim)) stands here to avoid comments at the end of line
         # or colors that can be added in RGB format after the points, the faces, ...
-        return np.loadtxt(input_file, dtype=np.floating, comments='#',
-                          usecols=list(range(dim)), max_rows=nb_points)
+        points = np.loadtxt(input_file, dtype=np.floating, comments='#',
```
mglisse, line 115: It isn't very clear what dtype np.floating yields (it is documented as "Abstract base class of all floating-point scalar types."). np.float64 or np.double would be clearer.

src/python/gudhi/off_utils.pyx
mglisse: Ah, so the new code fails to handle comments when numpy is too old, but you added a hack so it still does ok as long as the comments are in the most likely places? That makes sense. Could you add a comment next to this line? (we'll want to remove this at some point)
> It is advised to use a numpy version ≥ 1.23.0 if the `off_file` contains lines without data or with comments in between the points.

"advised" may not be strong enough. Empty/comment lines between the points are only supported with numpy ≥ 1.23.0.

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,96 @@ __license__ = "MIT"
...
+    file_desc.seek(file_desc.tell() - len(line))
+    return dim, nb_vertices
+
 def read_points_from_off_file(off_file=''):
     """Read points from an `OFF file <fileformats.html#off-file-format>`_.
 
     :param off_file: An OFF file style name.
     :type off_file: string
 
     :returns:  The point set.
```
mglisse, line 103: (not for this PR) I wonder if it would make sense to have an option to return the opened file, so the caller can keep reading simplices from the right place. Maybe reading simplices really belongs in a different function though.

src/python/gudhi/off_utils.pyx
VincentRouvreau: Modify the comment and the documentation warning on bd0f3feff

src/python/gudhi/off_utils.pyx
VincentRouvreau: Clarifcation on 8d012cc5e

src/python/gudhi/off_utils.pyx
VincentRouvreau: Yes, I keep in mind some ideas to read vertices coordinates and faces from OFF files. This could be used at least by the extended persistence notebook

src/python/gudhi/off_utils.pyx
VincentRouvreau: Ok, let's consider '*OFF' that is not 'nOFF', nor '4OFF' as a 3d OFF file, just the C++ code is doing. Changed on [9732176](https://github.com/GUDHI/gudhi-devel/pull/1008/commits/973217672dad0e842d200a042d7eeaf85329fe58)

src/python/gudhi/off_utils.pyx
VincentRouvreau: Done on [e819582](https://github.com/GUDHI/gudhi-devel/pull/1008/commits/e81958260af05ceac0b2f2c42e329c479be3a09b)

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,99 @@ __license__ = "MIT"
...
+        with `comments='#'` as an argument. Empty and or comment lines between the points are only supported with numpy
+        &ge; 1.23.0.
     """
-    if off_file:
-        if os.path.isfile(off_file):
-            return read_points_from_OFF_file(off_file.encode('utf-8'))
-        else:
-            raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT),
-                                    off_file)
+    # newline='' is required for Windows, otherwise end of line with '\r\n' are only detected as '\n'
```
mglisse, line 113: I don't understand the explanation. Could you explain at what point of the parsing you get what instead of what?

VincentRouvreau: On Windows, files end of lines are `\r\n`, while `\n` on Linux/Unix.
`file_desc.readline()` is always returning `my first line of file\n` on all system files when you open it with `newline=None`, cf. [open documentation](https://docs.python.org/3/library/functions.html#open) aka. universal newline mode.
As I need the exact length of the line (to go backward in the file reading), it was failing on windows because I was missing one character (the `\r`). By setting `newline=''`, I disable universal newline mode and when I read the line, I now have:
* `my first line of file\n` on Unix files
* `my first line of file\r\n` on Windows files
And I can go backward with the correct length of string.

mglisse: > As I need the exact length of the line (to go backward in the file reading),

Ah, that's the key point, you need the length of the line for seek. Could you update the comment to clarify that? It would make it clearer that we can also remove that when we drop support for old numpy.
An alternative would be to save the position (`.tell()`) before `readline` and pass that to `seek`, but that's not really better.

VincentRouvreau: Clarified on [90a5b5c](https://github.com/GUDHI/gudhi-devel/pull/1008/commits/90a5b5cd08076a9ec135534e0196ed8dcad38109)

src/python/gudhi/off_utils.pyx
```diff
@@ -25,21 +26,100 @@ __license__ = "MIT"
...
+    :type comment: string
+
+    :returns: The next line.
+    :rtype: string
+    """
+    while True:
+        # file_desc.readline() is preferred to next(file_desc), as the second option is not compatible with
+        # seek and tell methods
+        line = file_desc.readline()
+        if not line.startswith(comment):
```
mglisse, line 44: The doc of geomview says that blank lines are acceptable anywhere. Do we do the right thing if there is a blank line? Or do we need something like `l=line.lstrip()` `if len(l) != 0 and ...`?

mglisse (commented):
ok if we already handle blank lines, otherwise ok once the PR is updated to handle them.

mglisse (approved):
I didn't reread the whole thing, but it should be fine.

PR #1009, opened 2023-12-09
Rename collapse_edges

### LINKED ISSUE #973: Python edge_collapse naming
I just realized that we have merged #903, but the discussion on how to name the function (and in what file) was not finished. Another thing to handle before the release...

mglisse:
Fix #973.
(we can still go with another name if you don't like this one)

### DIFF
--- a/src/python/doc/rips_complex_user.rst
+++ b/src/python/doc/rips_complex_user.rst
@@ -54,7 +54,7 @@ regular Rips complex.
 
 Another option which is especially useful if you want to compute persistent homology in "high" dimension (2 or more,
 sometimes even 1), is to build the Rips complex only up to dimension 1 (a graph), then use
-:func:`~gudhi.SimplexTree.flag_filtration.edge_collapse.reduce_graph` to reduce the size of this graph, and finally call
+:func:`~gudhi.SimplexTree.collapse_edges` to reduce the size of this graph, and finally call
 :func:`~gudhi.SimplexTree.expansion` to get a simplicial complex of a suitable dimension to compute its homology. This
 trick gives the same persistence diagram as one would get with a plain use of `RipsComplex`, with a complex that is
 often significantly smaller and thus faster to process.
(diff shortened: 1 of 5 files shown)


src/python/doc/rips_complex_user.rst
```diff
@@ -54,7 +54,7 @@ regular Rips complex.
 
 Another option which is especially useful if you want to compute persistent homology in "high" dimension (2 or more,
 sometimes even 1), is to build the Rips complex only up to dimension 1 (a graph), then use
-:func:`~gudhi.SimplexTree.collapse_edges` to reduce the size of this graph, and finally call
+:func:`~gudhi.SimplexTree.flag_filtration.edge_collapse.reduce_graph` to reduce the size of this graph, and finally call
```
VincentRouvreau, line 57:
```suggestion
:func:`~gudhi.flag_filtration.edge_collapse.reduce_graph` to reduce the size of this graph, and finally call
```

src/python/doc/rips_complex_user.rst
```diff
@@ -157,7 +157,7 @@ or
    cplx.expansion(2)
 
 
-This way, you can easily add a call to :func:`~gudhi.SimplexTree.collapse_edges` before the expansion,
+This way, you can easily add a call to :func:`~gudhi.SimplexTree.flag_filtration.edge_collapse.reduce_graph` before the expansion,
```
VincentRouvreau, line 160:
```suggestion
This way, you can easily add a call to :func:`~gudhi.flag_filtration.edge_collapse.reduce_graph` before the expansion,
```

src/python/doc/rips_complex_user.rst
mglisse: Oups, I messed up those 2 in my search&replace... I should probably leave them as they were, or rephrase a bit if I want to point to the new function.

VincentRouvreau (approved):

PR #1028, opened 2024-02-05
Delaunay cech utility

VincentRouvreau:


src/Cech_complex/utilities/delaunay_cech_persistence.cpp
```diff
@@ -0,0 +1,159 @@
...
+  Simplex_tree stree;
+
+  Points_off_reader off_reader(off_file_points);
+  std::vector<Point> point_cloud = off_reader.get_point_cloud();
+  Delaunay_complex delaunay_complex_from_file(point_cloud);
+  delaunay_complex_from_file.create_complex(stree, std::numeric_limits< Filtration_value >::infinity(),
+                                            // exact can be false (or true), as default_filtration_value is set to true
+                                            false, true);
+  Gudhi::cech_complex::assign_MEB_filtration(Kernel(), stree, point_cloud, exact_version);
+  stree.prune_above_filtration(max_radius);
```
mglisse, line 52: Unlike the alpha-complex, for Delaunay-Cech, we could do a bit better (don't compute filtration for simplices whose faces already have a too large filtration), but that's not important for now.

src/Cech_complex/utilities/delaunay_cech_persistence.cpp
```diff
@@ -0,0 +1,159 @@
...
+  using Delaunay_complex = Gudhi::alpha_complex::Alpha_complex<Kernel>;
+
+  Simplex_tree stree;
+
+  Points_off_reader off_reader(off_file_points);
+  std::vector<Point> point_cloud = off_reader.get_point_cloud();
+  Delaunay_complex delaunay_complex_from_file(point_cloud);
+  delaunay_complex_from_file.create_complex(stree, std::numeric_limits< Filtration_value >::infinity(),
+                                            // exact can be false (or true), as default_filtration_value is set to true
+                                            false, true);
```
mglisse, line 50: Computing a Delaunay triangulation looks a bit strange (alpha-complex with a bunch of meaningless arguments), but that's how it is for now, ok.

src/Cech_complex/utilities/delaunay_cech_persistence.cpp
```diff
@@ -0,0 +1,159 @@
...
+  po::options_description hidden("Hidden options");
+  hidden.add_options()("input-file", po::value<std::string>(&off_file_points),
+                       "Name of an OFF file containing a point set.\n");
+
+  po::options_description visible("Allowed options", 100);
+  visible.add_options()("help,h", "produce help message")(
+      "exact,e", po::bool_switch(&exact),
+      "To activate exact version of Delaunay-Cech complex (default is false, not available if fast is set)")(
+      "fast,f", po::bool_switch(&fast),
+      "To activate fast version of Delaunay-Cech complex (default is false, not available if exact is set)")(
```
mglisse, line 123: (I realize that this is probably copied from another utility)
It isn't super clear what exact/fast mean precisely. We should say somewhere:
1. fast: right combinatorics, values can be arbitrarily bad
2. default, safe: values can have a relative error at most XXX
3. exact: true values rounded to double.

src/Cech_complex/utilities/delaunay_cech_persistence.cpp
```diff
@@ -0,0 +1,159 @@
...
+using Field_Zp = Gudhi::persistent_cohomology::Field_Zp;
+using Persistent_cohomology = Gudhi::persistent_cohomology::Persistent_cohomology<Simplex_tree, Field_Zp>;
+
+void program_options(int argc, char* argv[], std::string& off_file_points, bool& exact, bool& fast,
+                     std::string& pers_diag_file, Filtration_value& max_radius, int& p,
+                     Filtration_value& min_persistence);
+
+template<class Kernel>
+Simplex_tree create_simplex_tree(const std::string &off_file_points, bool exact_version, Filtration_value max_radius) {
+  using Point = typename Kernel::Point_d;
```
mglisse, line 39: In principle, we should be able to construct the Delaunay triangulation with Epick_d always, and only vary the kernel for assign_MEB_filtration. That probably requires converting point_cloud and I don't know how much there is to win here.

mglisse (approved):

src/Cech_complex/utilities/delaunay_cech_persistence.cpp
VincentRouvreau: To be discussed on #1029

PR #1029, opened 2024-02-07
Explain better fast/exact/safe versions of utils

VincentRouvreau:


mglisse (approved):
Quite a bit of duplication... but ok, it doesn't seem easy to avoid (unless we merge those utilities at some point).
I didn't expect an exact copy-paste of my text :wink:

PR #1034, opened 2024-02-13
Fix Ruff issues on wasserstein

VincentRouvreau:


src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -203,7 +203,9 @@ def _warn_infty(matching):
     `matching=True`) about the returned matching being `None`.
     '''
     if matching:
-        warnings.warn('Cardinality of essential parts differs. Distance (cost) is +inf, and the returned matching is None.')
+        warnings.warn('''Cardinality of essential parts differs. Distance (cost) is +inf, and the returned matching is
+                      None.
+                      ''')
```
mglisse, line 208: I don't think this split does the right thing.

mglisse (commented):
Line length changes are not the most interesting.

src/python/gudhi/wasserstein/wasserstein.py
VincentRouvreau: Ah yes the warning was on 2 separate lines.
Instead I factorized the beginning of the message in a string to reduce the line length on [347a4c1](https://github.com/GUDHI/gudhi-devel/pull/1034/commits/347a4c166cbf057d9c52a66997c5507c63368149).
I agree, line length changes are not the most interesting, but this is nice to stick with a limited number of characters per line as, with some IDE, you need to scroll to see the code.

src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -349,7 +351,7 @@ def wasserstein_distance(X, Y, matching=False, order=1., internal_p=np.inf, enab
 
     # Comptuation of the ot cost using the ot.emd2 library.
     # Note: it is the Wasserstein distance to the power q.
-    # The default numItermax=100000 is not sufficient for some examples with 5000 points, what is a good value?
+    # The default numItermax=100000 is not sufficient for some examples with 5000 points, what is a good value ?
```
mglisse, line 354: Why the extra space? In French we put a space before '?' but not in English.

VincentRouvreau: Oups, sorry, I rolled back this modification

mglisse (approved):

PR #1035, opened 2024-02-13
Remove useless imports advised by ruff

VincentRouvreau:
also some line length in rips and cubical sklearn itf


mglisse (approved):
I never understood what the mermaid stuff was about so I didn't look at that part.

VincentRouvreau:
> I never understood what the mermaid stuff was about so I didn't look at that part.

If you go to https://mermaid-js.github.io/mermaid-live-editor/ and copy/paste the code without comments, it gives you a nice sequence diagram (that I find useful):
image

mermaid can also be rendered in markdown.

PR #1036, opened 2024-02-22
safer Cech for dim > ambient

### LINKED ISSUE #1030: Unsafe Cech with Epick_d?
I do not have an example at hand, but I think the computation done in Cech_complex and assign_MEB_filtration may be unsafe when the kernel is Epick_d. In the plane, with 4 points that are cocyclic (or almost), rounding errors may tell us that no point is in the MEB of the other 3. In this case, we will ask CGAL for the circumcenter of 4 points, which in the best case will give an assertion failure.
If this comment is true, we should either
1. make this more robust with Epick_d. For the specific issue described above, we could just check if we have exceeded the ambient dimension, and in that case use the max of the faces no matter what the inclusion test might say (it could even count as an optimization, although I don't know how often this function will be used on simplices of dimension higher than ambient). But we need to think what other issues could happen.
2. document how risky Epick_d is for these constructions (unlike for Alpha_complex).

PoC (doesn't handle the empty case, probably not the best way to access the dimension)
```diff
@@ -45,6 +45,7 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
   std::vector<Sphere> cache_;
   std::vector<Point_d> pts;
   CGAL::NT_converter<FT, Filtration_value> cvt;
+  int ambient_dim = std::begin(points)->dimension();
 
   auto fun = [&](Simplex_handle sh, int dim){
     if (dim == 0) complex.assign_filtration(sh, 0);
@@ -62,6 +63,13 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
       complex.assign_key(sh, cache_.size());
       complex.assign_filtration(sh, std::max(cvt(r), Filtration_value(0)));
       cache_.emplace_back(std::move(m), std::move(r));
+    } else if (dim > ambient_dim) {
+      Filtration_value maxf = 0; // max filtration of the faces
+      for (auto face : complex.boundary_simplex_range(sh)) {
+        using std::max;
+        maxf = max(maxf, complex.filtration(face));
+      }
+      complex.assign_filtration(sh, maxf);
     } else {
       Filtration_value maxf = 0; // max filtration of the faces
       bool found = false;
```

While I am here, there is a property that the circumradius cannot exceed the max of the faces by a factor larger than $1/\sqrt{1-1/d^2}$, we could use it as an upper bound to ensure that the values we return never explode:
```diff
@@ -87,7 +95,12 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
         Point_d c = k.construct_circumcenter_d_object()(pts.begin(), pts.end());
         FT r = k.squared_distance_d_object()(c, pts.front());
         if (exact) CGAL::exact(r);
+        int d2 = dim * dim;
+        Filtration_value max_sanity = maxf * d2 / (d2 - 1);
+        using std::min;
+        using std::max;
         maxf = max(maxf, cvt(r)); // maxf = cvt(r) except for rounding errors
+        maxf = min(maxf, max_sanity);
         complex.assign_key(sh, cache_.size());
         // We could check if the simplex is maximal and avoiding adding it to the cache in that case.
         cache_.emplace_back(std::move(c), std::move(r));
```
(of course it would only be done for Epick_d, not Epeck_d)
I don't know if it is interesting though.

mglisse:
Fix #1030.
After that, I think (could be wrong though) the problems caused by Epick_d are no worse for Cech than for the alpha-complex. That is, we could investigate more numerically stable ways to compute a circumradius (especially since for the Cech we can assume that the center is inside the triangle), or mitigation strategies (like the one suggested in the new comment), to get a *better* balance between speed and safety, but nothing urgent.

- [ ] Check the impact on performance. I would expect a negligible slowdown if maxdim ≤ ambient dimension, and possibly a small speed up if maxdim is larger.

### DIFF
--- a/src/Cech_complex/include/gudhi/MEB_filtration.h
+++ b/src/Cech_complex/include/gudhi/MEB_filtration.h
@@ -101,6 +101,10 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
         Point_d c = k.construct_circumcenter_d_object()(pts.begin(), pts.end());
         FT r = k.squared_distance_d_object()(c, pts.front());
         if (exact) CGAL::exact(r);
+        // For Epick_d, if the circumcenter computation is too unstable, we could compute
+        //   int d2 = dim * dim;
+        //   Filtration_value max_sanity = maxf * d2 / (d2 - 1);
+        // and use min(max_sanity, ...), which would limit how bad numerical errors can be.
         maxf = max(maxf, cvt(r)); // maxf = cvt(r) except for rounding errors
         complex.assign_key(sh, cache_.size());
         // We could check if the simplex is maximal and avoiding adding it to the cache in that case.
--- a/src/Cech_complex/include/gudhi/MEB_filtration.h
+++ b/src/Cech_complex/include/gudhi/MEB_filtration.h
@@ -46,7 +46,15 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
   std::vector<Point_d> pts;
   CGAL::NT_converter<FT, Filtration_value> cvt;
 
+  // This block is only needed to get ambient_dim
+  if(std::begin(points) == std::end(points)) {
+    // assert(complex.is_empty());
+    return;
+  }
+  int ambient_dim = k.point_dimension_d_object()(*std::begin(points));
+
   auto fun = [&](Simplex_handle sh, int dim){
+    using std::max;
     if (dim == 0) complex.assign_filtration(sh, 0);
     else if (dim == 1) {
       // For a Simplex_tree, this would be a bit faster, but that's probably negligible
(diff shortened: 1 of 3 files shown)


src/Cech_complex/include/gudhi/MEB_filtration.h
```diff
@@ -60,12 +68,18 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
...
       if (exact) CGAL::exact(r);
       complex.assign_key(sh, cache_.size());
-      complex.assign_filtration(sh, std::max(cvt(r), Filtration_value(0)));
+      complex.assign_filtration(sh, max(cvt(r), Filtration_value(0)));
       cache_.emplace_back(std::move(m), std::move(r));
+    } else if (dim > ambient_dim) {
+      // The sphere is always defined by at most d+1 points
+      Filtration_value maxf = 0; // max filtration of the faces
+      for (auto face : complex.boundary_simplex_range(sh)) {
+	maxf = max(maxf, complex.filtration(face));
```
VincentRouvreau, line 77: Bad indentation (with  tab on this line), should be with spaces

src/Cech_complex/include/gudhi/MEB_filtration.h
```diff
@@ -46,7 +46,15 @@ void assign_MEB_filtration(Kernel&&k, SimplicialComplexForMEB& complex, PointRan
...
 
+  // This block is only needed to get ambient_dim
+  if(std::begin(points) == std::end(points)) {
+    // assert(complex.is_empty());
+    return;
+  }
+  int ambient_dim = k.point_dimension_d_object()(*std::begin(points));
+
   auto fun = [&](Simplex_handle sh, int dim){
+    using std::max;
```
VincentRouvreau, line 57: I would keep std::max everywhere else max is called to be more explicit, but maybe a matter of taste. You can forget this comment if you do not agree

mglisse: Note that in this function, I want `max` to be found by ADL if possible, I do not want to use std::max if the type is `CGAL::Lazy_exact_nt` for instance (it should use `CGAL::max` in that case). I still need to make `std::max` available in case the type is built-in, like `double`, hence the `using`.
I could copy the using directive before each call to `max` instead of once for the whole lambda, but there are 4 calls spread in the various branches, so it would be noisy.

src/Cech_complex/include/gudhi/MEB_filtration.h
mglisse: Done.

src/Cech_complex/include/gudhi/MEB_filtration.h
VincentRouvreau: Ok, I got it, thanks for the precision

PR #1037, opened 2024-02-22
Fix for #1033 

### LINKED ISSUE #1033: Representations - kernel_methods : Undefined name `metric`
In `_sklearn_wrapper(metric, **kwargs)`, `metric` is undefined, cf.

https://github.com/GUDHI/gudhi-devel/blob/06bc78ae0a3d96d1273e37438cfd1f116118385a/src/python/gudhi/representations/kernel_methods.py#L89

MathieuCarriere:
Fix for a typo in `pairwise_persistence_diagram_kernels`
Fix #1033

### DIFF
--- a/src/python/gudhi/representations/kernel_methods.py
+++ b/src/python/gudhi/representations/kernel_methods.py
@@ -86,7 +86,7 @@ def pairwise_persistence_diagram_kernels(X, Y=None, kernel="sliced_wasserstein",
     elif kernel == "persistence_weighted_gaussian":
         return _pairwise(pairwise_kernels, False, XX, YY, metric=_sklearn_wrapper(_persistence_weighted_gaussian_kernel, X, Y, **kwargs), n_jobs=n_jobs)
     else:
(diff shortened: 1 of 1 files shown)


mglisse (approved):
Ideally we would also add a small test so this line actually gets executed (https://output.circle-artifacts.com/output/job/b49fad43-f4d0-4c21-9003-5d044d899b2b/artifacts/0/htmlcov/d_86dc696057d7fff0_kernel_methods_py.html#t89 seems to indicate that this is the only line in this file that isn't covered by any test), but that doesn't need to block fixing this obvious copy-pasto.

MathieuCarriere:
As far as I remember (but I might have misunderstood), I think the typo was identified because of a failing test in another PR (I can't remember which one unfortunately). @VincentRouvreau, do you know something about this?

mglisse:
No, it was identified by a tool (ruff) that does some basic static analysis on the code. The same tool also reported some unused `import`s in other files, etc.

VincentRouvreau (approved):

PR #1038, opened 2024-03-02
Incremental Delaunay complex

mglisse:
The details differ a bit from https://bitbucket.org/mkerber/function_delaunay/src/master/ , but it builds the same thing in the end.

- I tried to make the code more robust, in particular in case the first d+1 points are affinely dependent (there is some discussion about different strategies at the end of the header).
- I only did a dD version. While using CGAL's Triangulation_2/Triangulation_3 is indeed faster (because of the hierarchy and structural filtering, which both speed up point location, an important part since we cannot sort the points as in the usual Delaunay), the difference is not that huge on the data I tried, especially with CGAL-6 (I pushed a few simple optimizations there), and that part is cheap anyway, in 3d we typically spend more time inserting in the Simplex_tree than doing CGAL stuff...

Unlike most Gudhi files, I added a GPL license (+ exception for GeometryFactory customers) on this one
- since it is unusable without CGAL, we don't lose anything by doing that
- it contains code copied from CGAL and tweaked for this application
- although I did not copy code from Michael Kerber's repository, I did look at it, and his code is GPL (he did not protest against the exception)

Documentation available at https://output.circle-artifacts.com/output/job/78531865-e05c-4d25-ba74-46ccbbd3b646/artifacts/0/doxygen/index.html


src/Alpha_complex/include/gudhi/Incremental_delaunay.h
```diff
@@ -0,0 +1,137 @@
...
+ * triangulation of (p1, ..., pn), the complex also contains a simplex formed with the same points as s plus the
+ * first pj that conflicts with s (is inside its circumsphere). For points in R^d, the complex usually has
+ * dimension d+1.
+ *
+ * This function calls `complex.insert_simplex_and_subfaces` on a set of simplices large enough that, with their
+ * faces, they define the incremental Delaunay complex. However, this set is not guaranteed to be minimal, there
+ * may be redundancies.
+ *
+ * No filtration values are computed here, those can be computed separately afterwards, for instance
+ * with `Gudhi::cech_complex::assign_MEB_filtration`.
```
VincentRouvreau, line 27: doxygen does not produce a link here. But this works:
```suggestion
 * with `Gudhi::cech_complex::assign_MEB_filtration()`.
```

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Gudhi {
+
+/** \ingroup alpha_complex
+ *
+ * This function constructs the incremental Delaunay complex, as defined in \cite incremental-delaunay .
+ * Given a list of points (p1, ..., pn), the incremental Delaunay complex contains the Delaunay triangulation
+ * of (p1, ..., pk) for all k≤n. For any simplex s of the triangulation of (p1, ..., pk) that is not in the
+ * triangulation of (p1, ..., pn), the complex also contains a simplex formed with the same points as s plus the
+ * first pj that conflicts with s (is inside its circumsphere). For points in R^d, the complex usually has
```
VincentRouvreau, line 19: For a better LaTeX render:
```suggestion
* first pj that conflicts with s (is inside its circumsphere). For points in \f$\mathbb{R}^d\f$, the complex usually has
```

src/Alpha_complex/test/Incremental_delaunay_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <vector>
+
+#include <gudhi/Incremental_delaunay.h>
+#include <gudhi/Simplex_tree.h>
+#include <gudhi/random_point_generators.h>
+#include <gudhi/Alpha_complex.h>
+
+using Static_kernel_3 = CGAL::Epick_d< CGAL::Dimension_tag<3> >;
+using Dynamic_kernel = CGAL::Epick_d< CGAL::Dynamic_dimension_tag >;
+using list_of_kernel_3_variants = boost::mpl::list<Static_kernel_3, Dynamic_kernel>;
```
VincentRouvreau, line 28: Should we also test `Epeck_d` here ? Or there is no added value ?

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
mglisse: It works with doxygen 1.9.8. The CI has 1.9.1 which is starting to date a bit.

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
```diff
@@ -0,0 +1,137 @@
...
+    if (lt == Triangulation::OUTSIDE_AFFINE_HULL) {
+      // nothing disappears
+      tri.insert_outside_affine_hull(p)->data() = idx;
+    }
+    else if (lt == Triangulation::ON_VERTEX) {
+      // nothing disappears
+    }
+    else {
+      if(tri.current_dimension() == 1) {
+	if(lt == Triangulation::OUTSIDE_CONVEX_HULL) {
```
VincentRouvreau, line 76: Strange identation with tabulations from here

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
```diff
@@ -0,0 +1,137 @@
...
+      } else { // main case
+        // Adapted from Delaunay_triangulation::insert_in_conflicting_cell(Point,Full_cell_handle)
+	/* Full_cell_h_vector cs; */ cs.clear();
+	std::back_insert_iterator<Full_cell_h_vector> out(cs);
+	typename Triangulation::Facet ft = tri.compute_conflict_zone(p, s, out);
+	for(auto c : cs) {
+	  /* std::vector<int> splx; */ splx.clear();
+	  for(int i = 0; i <= tri.current_dimension(); ++i) {
+	    // We could skip the infinite vertex and output a (new) simplex, but that's not so useful
+	    if(tri.is_infinite(c->vertex(i))) goto next_conflicting_cell;
```
VincentRouvreau, line 93: I am still not a big fan of `goto` and I would prefer a solution with a `boolean` and `continue`, but we already discussed about it, the remark is minor.

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
```diff
@@ -0,0 +1,137 @@
...
+  std::vector<typename SimplicialComplex::Vertex_handle> splx;
+
+  for(int idx = 0; pt_it != pt_end; ++pt_it, ++idx) {
+    // Adapted from Delaunay_triangulation::insert(Point)
+    auto&& p = *pt_it;
+    typename Triangulation::Locate_type lt;
+    typename Triangulation::Face f(d);
+    typename Triangulation::Facet ft;
+    typename Triangulation::Full_cell_handle s = tri.locate(p, lt, f, ft);
+    // Adapted from Delaunay_triangulation::insert(Point,Locate_type,Face,Facet,Full_cell_handle)
```
VincentRouvreau, line 66: Should we mention it in the copyright ?

src/Alpha_complex/test/Incremental_delaunay_test.cpp
mglisse: There shouldn't be any added value to using Epeck_d, but that doesn't mean there cannot be in practice, and it wouldn't hurt to test at least one Epeck_d variant.

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
mglisse: Ah, I did set up my editor to avoid TABs for files in gudhi, but I probably wrote this code elsewhere and copy-pasted it later :disappointed:

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
mglisse:
```diff
@@ -88,15 +88,16 @@ void construct_incremental_delaunay(K const&k, SimplicialComplex& complex, Point
         typename Triangulation::Facet ft = tri.compute_conflict_zone(p, s, out);
         for(auto c : cs) {
           /* std::vector<int> splx; */ splx.clear();
+          bool interrupted = false;
           for(int i = 0; i <= tri.current_dimension(); ++i) {
             // We could skip the infinite vertex and output a (new) simplex, but that's not so useful
-            if(tri.is_infinite(c->vertex(i))) goto next_conflicting_cell;
+            if(tri.is_infinite(c->vertex(i))) { interrupted = true; break; }
             splx.push_back(c->vertex(i)->data());
           }
+          if (interrupted) continue;
           // We could put idx at the beginning of the vector once instead of reinserting it each time
           splx.push_back(idx);
           complex.insert_simplex_and_subfaces(splx);
-next_conflicting_cell:;
         }
         tri.insert_in_hole(p, cs.begin(), cs.end(), ft)->data() = idx;
       }
```
doesn't look too horrible so I might change this one. But note that `goto` is the official syntax to do what some would like to call `continue(2)` (`continue` but for a loop that is not the innermost one), and the usual `continue` is defined in terms of `goto` in the standard.

image

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
mglisse: Would
```
Authors: Michael Kerber, Marc Glisse, (Samuel Hornus for the original CGAL code)
```
in both places work?

VincentRouvreau: I am not sure (not a specialist), but keep it like that. I don't know if we should ask CGAL people if it fits their requirements, but you have more experience than me on this.

src/Alpha_complex/include/gudhi/Incremental_delaunay.h
VincentRouvreau: Ok, keep it like that and I will update doxygen on my side before generating the official documentation

mglisse: I already changed it, I don't think it is worth changing back...

VincentRouvreau (approved):

Issue #795, opened 2023-01-15
Randomness in Cover Complex


astamm:
I understand that behind the scene all computations for the GIC in the cover complex are performed in C. However, there is no way from the Python interface to set the C seed for reproducibility unless I missed something. Could it be added?

mglisse:
That's a general issue in Gudhi. Even in the C++ interface we seldom let users set the seed. Adding this possibility without breaking the API or making it too ugly may not always be trivial (or maybe it is...).
(search for `std::random_device` to find a few relevant places)

It isn't high on our list of priorities for now, but if someone wants to tackle it, I'll make sure we review the PR.

astamm:
I see, thanks. From a data science perspective, it is very important to be able to set the seed (and the generator algorithm) to ensure reproducibility of the results. But, maybe that *only* means adding an extra argument `seed` to any function or method that uses random numbers?
Then, we also need to make sure it is reproducible inter-platforms. Is randomness always used via `std::random`? Or do you use also the boost module for randomness? Do you know whether they use the same underlying algorithm for generating random numbers?

mglisse:
Hoping for too much reproducibility is illusory. For instance: https://pytorch.org/docs/stable/notes/randomness.html
> Completely reproducible results are not guaranteed across PyTorch releases, individual commits, or different platforms. Furthermore, results may not be reproducible between CPU and GPU executions, even when using identical seeds.

(actually, computing the "same" distance matrix between 2 point clouds repeatedly 20 times using pykeops+pytorch, I get 20 slightly different matrices)

In addition to C++ <random>, we also use random numbers from CGAL, from boost (through CGAL), from numpy (can be seeded globally). There is some randomness that comes from TBB, if you activate that.

It can be more convenient to set a seed globally instead of passing one to each function, although that may not work as well with parallelism.

While C++ defines exactly the output of generators like mt19937_64, the way it is used in uniform_int_distribution is specific to each implementation.

Some algorithms may use the order of some set of pointers, because it needs an order for efficient processing, but that order can change from one run to the next.

Etc, not trivial. Although again, if someone wants to tackle the lowest hanging fruits...

astamm:
I understand your arguments.
However, when I run a simulation study for a paper say, I need it to be reproducible, meaning that if reviewers or readers want to redo the experiment, they will end up with the same results, which makes total sense.
I understand the comment from PyTorch but in my view this is a flaw.
Reproducibility is of paramount importance in data science.
[...]

PR #1039, opened 2024-03-08
Random graph methods

VincentRouvreau:
A new helper class `Gudhi::Random` that can be constructed with a user given seed, or from a random seed if not given.
This could help for #795 resolution.

The use of `Gudhi::Random` everywhere `std::mt19937` is used in legacy code is not part of this PR, but can be done after on a different PR.

The purpose of this PR is to provide 2 new methods to construct random graph:
* `random_edges` that returns X% (cf.density argument) of permuted vertices (cf. nb_vertices argument), seen as a list of edges
* `simplex_tree_random_flag_complex` that fills a Simplex_tree with `random_edges`, but also with random filtration values

Maybe `simplex_tree_random_flag_complex` is a bit slow:
```
Random flag complex with 5000 vertices and 1881350 simplices

real    0m42,729s
user    0m42,715s
sys     0m0,012s
```

But this is not what we want to benchmark.


src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,56 @@
...
+ *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+ *    Author(s):       Vincent Rouvreau
+ *
+ *    Copyright (C) 2024 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#ifndef RANDOM__H_
```
mglisse, line 11: Shouldn't we put GUDHI in the name of the macro, to reduce the risk of conflicts?

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,56 @@
...
+        std::generate(result.begin(), result.end(), [&]() { return dis(gen_); });
+      } else if constexpr (std::is_integral_v<Type>) {
+        std::uniform_int_distribution<Type> dis(min, max);
+        std::generate(result.begin(), result.end(), [&]() { return dis(gen_); });
+      }
+      return result;
+    }
+
+   private:
+    std::mt19937 gen_;
```
mglisse, line 52: If we pick one, mt19937_64 may be better, the platforms we target are all 64 bits (and then the seed is 64 bits as well).

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,56 @@
...
+#ifndef RANDOM__H_
+#define RANDOM__H_
+
+#include <random>  // for std::random_device, std::mt19937, std::uniform_real_distribution
+#include <vector>
+#include <algorithm>  // for std::generate
+#include <cstddef>  // for std::size_t
+
+namespace Gudhi {
+  std::random_device rd;
```
mglisse, line 20: As a global object, it may cause trouble if you try to link 2 .o that both use gudhi.
You could mark it `inline` (and then also hide it as a static member of the class or in some subnamespace), or create a temporary one just where it is used.

src/common/include/gudhi/random_graph_generators.h
```diff
@@ -0,0 +1,98 @@
...
+#include <array>
+#include <random>
+#include <numeric>  // for std::iota
+#include <cstddef>  // for std::size_t
+#include <type_traits>  // for std::is_floating_point_v
+#ifdef DEBUG_TRACES
+#include <iostream>
+#endif  // DEBUG_TRACES
+
+std::random_device rd;
```
mglisse, line 28: Argh!

src/common/include/gudhi/random_graph_generators.h
```diff
@@ -0,0 +1,98 @@
...
+std::vector<std::array<Vertex_handle, 2>> random_edges(Vertex_handle nb_vertices, double density = 0.15) {
+  std::vector<std::array<Vertex_handle, 2>> permutations;
+  if (nb_vertices < 2)
+    return permutations;
+
+  std::vector<bool> to_permute(nb_vertices);
+  std::fill(to_permute.begin(), to_permute.begin() + 2, true);
+  
+  std::size_t nb_permutations = (nb_vertices * (nb_vertices - 1)) / 2;
+  auto random_values = Gudhi::Random().get_range<double>(nb_permutations);
```
mglisse, line 42: If you are going to create Random objects left and right, I don't see how this will help determinism. There is no way for the user to specify a seed...

src/common/include/gudhi/random_graph_generators.h
```diff
@@ -0,0 +1,98 @@
...
+
+template<typename Simplex_tree>
+void simplex_tree_random_flag_complex(
+    Simplex_tree& st,
+    typename Simplex_tree::Vertex_handle nb_vertices,
+    double density = 0.15) {
+  using Vertex_handle = typename Simplex_tree::Vertex_handle;
+  std::vector<Vertex_handle> vertices(nb_vertices);
+  std::iota(vertices.begin(), vertices.end(), 0); // vertices is { 0, 1, 2, ..., 99 } when nb_vertices is 100
+  st.insert_batch_vertices(vertices);
```
mglisse, line 82:
```suggestion
  st.insert_batch_vertices(boost::irange(nb_vertices));
```
https://www.boost.org/doc/libs/1_85_0/libs/range/doc/html/range/reference/ranges/irange.html (or counting_range)

src/common/include/gudhi/random_graph_generators.h
```diff
@@ -0,0 +1,98 @@
...
+#endif  // DEBUG_TRACES
+    permutations.push_back(permutation);
+    //std::cout << permutation.first << ", " << permutation.second << "\n";
+  } while (std::prev_permutation(to_permute.begin(), to_permute.end()));
+
+  return permutations;
+}
+
+template<typename Simplex_tree>
+void simplex_tree_random_flag_complex(
```
mglisse, line 75: It doesn't produce a flag complex, it produces a graph.

src/common/include/gudhi/random_graph_generators.h
```diff
@@ -0,0 +1,98 @@
...
+#ifdef DEBUG_TRACES
+#include <iostream>
+#endif  // DEBUG_TRACES
+
+std::random_device rd;
+
+namespace Gudhi {
+
+template <typename Vertex_handle>
+std::vector<std::array<Vertex_handle, 2>> random_edges(Vertex_handle nb_vertices, double density = 0.15) {
```
mglisse, line 33: The algorithm in there seems very complicated. What do you want exactly?
1. For each pair of vertices, add the edge with proba 15% (independently)
2. Get exactly 15% * n(n-1)/2 edges, chosen randomly
3. Other

PR #1046, opened 2024-04-12
bibliography clean up

hschreiber:
Clean up of the bibliography.bib file:
 - re-alignement of the citations
 - alphabetical sort by titles + all unused citations at the end of the file (also by alphabetical order)
 - removal of doubles
 - removal of unnecessary/incomplete citations when they were not used
 - addition of an url for used citations when there was none (except for one I couldn't find)

I thought also about homogenizing the authors names and titles, but I don't know if it is worth it when future citations will probably do whatever they want again...

### DIFF
--- a/biblio/bibliography.bib
+++ b/biblio/bibliography.bib
@@ -163,7 +163,7 @@
 % D
 %-----------------------------
 
-@inproceedings{dtmdtmfiltrations,
+@inproceedings{dtmfiltrations,
   author    = {Anai, Hirokazu
                and Chazal, Fr{\'e}d{\'e}ric
                and Glisse, Marc
--- a/biblio/bibliography.bib
+++ b/biblio/bibliography.bib
@@ -1,119 +1,680 @@
-@inproceedings{gudhilibrary_ICMS14,
-  author    = {Cl\'ement Maria and Jean-Daniel Boissonnat and
-              Marc Glisse and Mariette Yvinec},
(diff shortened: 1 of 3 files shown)


VincentRouvreau (commented):
* [X] I am not sure if we should keep unused biblio (47 entries and 540 lines of unused code) => *we keep it*
* [X] Should we take the opportunity to rename `tomato` and `dtm` bib entries ? (as `tomato2013` and `dtm2011` for instance) => *to be done*

mglisse (approved):
If you are both happy with it, ok.

The short labels dtm and tomato were on purpose, they are not ambiguous (they refer to the papers that defined them), and they can be used without having to open the bib file to check what random label they were given. But that's not important, ok for the new labels.

hschreiber:
> If you are both happy with it, ok.
> 
> The short labels dtm and tomato were on purpose, they are not ambiguous (they refer to the papers that defined them), and they can be used without having to open the bib file to check what random label they were given. But that's not important, ok for the new labels.

There are several citations starting with "dtm", so it is not clear why this particular one is the only one with a "simple" name. In particular when there where only one citation of `dtm` in the end, so it is not about having to write it a lot. For `tomato`, I understand a bit more, but it is true that the word appears a lot, so finding the place where the citation is used is a bit trickier (even though not impossible). But honestly, I don't mind reverting the names if you prefer.

VincentRouvreau (approved):

PR #1049, opened 2024-04-23
Use Cgal and Eigen3 cmake target

VincentRouvreau:
This PR is required by future CGAL 6.0

* Factorize `add_executable` and `target_link_libraries` with CMake targets in a new function `add_executable_with_targets`
* Use Cgal and Eigen3 cmake target - requires CGAL >=5.1 and Eigen >= 3.3
* Remove `project` from each CMake (just keep the main one) as not mandatory and it does not add any functionality

## questions :question:
~~* [ ] Do we update CGAL_VERSION_NR and EIGEN_VERSION_AT_LEAST check in code ?~~
* [X] `find_package(CGAL 5.1.0)` does not work with CGAL master (but 6.0.X). Do we open an issue on CGAL side ? https://github.com/CGAL/cgal/issues/8192


mglisse:
> Remove `project` from each CMake (just keep the main one) as not mandatory and it does not add any functionality

Does it not even help if we have tests with the same name in 2 subprojects, or something like that? Anyway, whatever you prefer.

> Do we update CGAL_VERSION_NR and EIGEN_VERSION_AT_LEAST check in code ?

I would say no. The code still works with older versions, it is only the cmake stuff that breaks.

>  find_package(CGAL 5.1.0) does not work with CGAL master (but 6.0.X). Do we open an issue on CGAL side ?
 
Yes please. I remember discussing this issue a while ago, but I don't know with whom...
If they don't want to change it before the 6.0 release, we will have to handle that differently in gudhi.

VincentRouvreau:
> Does it not even help if we have tests with the same name in 2 subprojects, or something like that? Anyway, whatever you prefer.

From what I read [here](https://cmake.org/cmake/help/latest/guide/tutorial/A%20Basic%20Starting%20Point.html#exercise-1-building-a-basic-project):
 > To start a project, we use the [project()](https://cmake.org/cmake/help/latest/command/project.html#command:project) command to set the project name. This call is required with every project and should be called soon after [cmake_minimum_required()](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required). As we will see later, this command can also be used to specify other project level information such as the language or version number.

As all of our code managed by CMake is C++, with the same gudhi version number, cmake minimal version is the same for all the library (easier to maintain),... I prefer to keep one project.

src/python/doc/bottleneck_distance_sum.inc
```diff
@@ -8,7 +8,7 @@
    |                                                   | diagonal points) such that any couple of matched points are at       |                                                                 |
    |      Bottleneck distance is the length of         | distance at most b, where the distance between points is the sup     | :License: MIT (`GPL v3 </licensing/>`_)                         |
    |      the longest edge                             | norm in :math:`\mathbb{R}^2`.                                        |                                                                 |
-   |                                                   |                                                                      | :Requires: `CGAL <installation.html#cgal>`_ :math:`\geq` 4.11.0 |
+   |                                                   |                                                                      | :Requires: `CGAL <installation.html#cgal>`_ :math:`\geq` 5.1.0  |
```
mglisse, line 11: I don't know if having the version number repeated everywhere is that useful (it isn't horrible though). A possibility would be to specify it only when it isn't the "default" one.

mglisse (approved):

src/python/gudhi/tensorflow/_layer.py
```diff
@@ -0,0 +1,16 @@
+import tensorflow
+# The sklearn parse_version function is taken from 'packaging' and this dependency is not mandatory
+from sklearn.utils.fixes import parse_version
```
mglisse, line 3: I am not sure what the comment means.
Do we now need to install sklearn to be able to use the tensorflow stuff?

src/python/gudhi/tensorflow/_layer.py
```diff
@@ -0,0 +1,16 @@
...
+
+class TensorflowKerasLayer(tensorflow.keras.layers.Layer):
+    """
+    TensorFlow Keras layer
+    """
+    def __init__(self, **kwargs):
+        """
+        Constructor for the TensorflowKerasLayer class
+        """
+        # On tensorflow < 2.15.1 set dynamic argument to True
```
mglisse, line 13: Did they really change that in a patch release, i.e. 2.15.0 wants `dynamic` but 2.15.1 rejects it? That's surprising :disappointed:

mglisse (commented):
Why is the tensorflow stuff in this unrelated PR?
IIUC, `dynamic=True` is now always enabled so they removed the option? The approach looks ok then.

VincentRouvreau:
Sorry about that bad merge... I start again a new PR from [96f7508](https://github.com/GUDHI/gudhi-devel/pull/1049/commits/96f7508e6f3b89b4cb13dcdef8e49868c578fd6f) and I close this one.
Forget the tensorflow proposal, it will come on a new PR

Issue #982, opened 2023-10-09
[Representations Module] New `__call__` strategy does not retrieve attributes


astamm:
If I understand correctly, the new strategy of the `__call__` method in the representations module is to clone the class and call `fit_transform()` on the cloned class. This effectively does not modify the attributes of the main class.

However, it also means that we do not have access to these attributes, e.g. for Betti curves, we get the y-coordinate of the curve but there is no way to retrieve on which grid these y-coordinates were computed. Unless we run the `fit()` method on the main class and then access `grid_` but this was precisely what we wanted to avoid in the first place.

astamm:
Small reprex:
```
from gudhi.representations import BettiCurve
import numpy as np

dgm = np.array([[1, 4], [2, 5]])

bc = BettiCurve()
bc(dgm)
bc.grid_
```

mglisse:
`fit` and `transform` are the main interface, with `__call__` only a shortcut for the cases where it suffices, so I don't find it too problematic if some use cases cannot be satisfied without `fit`.
I guess the opposite strategy would be to store a flag of whether `fit` has been explicitly called, and if it hasn't then `__call__` overwrites `grid_` (so we don't work on a clone anymore). That seems doable, but potentially confusing.
Do you have suggestions on a *good* behavior?
(adding @MathieuCarriere to the conversation)

astamm:
Well in my opinion, when I call `__call__()` on a single diagram from a class in the {representations} module, I want it to return all the information I need about the selected representation. Hence it means the y-coordinate (values representing the diagram in that representation) **and** the grid on which those values were computed. Currently, we are lacking this latter information. I guess it does not matter for feeding machine learning pipelines because the current behaviour produces the y-coordinate which is then used as is in feature space. But we might want to do other statistical analyses on the functional representations such as domain selection where differences lie between groups, in which case it is of primary importance to also have the domain on which the representation is defined.

More generally, I guess in my head, it depends on what a class is expected to do. In the case of the {representations} module, they are supposed to provide representations of diagrams. A *functional* representation is made of two things: x- and y-coordinates. So I think not only should `__call__()` return both but also `transform()` and `fit_transform()`. In the latter two cases, this is achieved by updating the attribute `grid_` or some other attribute (`im_range_fixed_`, etc.) which can then be accessed to actually generate the grid. But different representations update different attributes and forming the grid for each has to be done in different ways.

[...]

PR #1056, opened 2024-05-27
Fix Atol for scikit-learn >= 1.4.0

### LINKED ISSUE #1031: Atol : tests are failing with scikit-learn 1.4.0
With scikit-learn 1.4.0:
```python
from sklearn.cluster import KMeans
from gudhi.representations.vector_methods import Atol
import numpy as np
a = np.array([[1, 2, 4], [1, 4, 0], [1, 0, 4]])
b = np.array([[4, 2, 0], [4, 4, 0], [4, 0, 2]])
c = np.array([[3, 2, -1], [1, 2, -1]])
atol_vectoriser = Atol(quantiser=KMeans(n_clusters=2, random_state=202006))
atol_vectoriser.fit(X=[a, b, c]).centers
# array([[3.75, 2.  , 0.25],
#        [1.  , 2.  , 1.75]])
atol_vectoriser(a)
# array([0.00892619, 0.20804165])
atol_vectoriser(c)
# array([0.44476906, 0.05480506])
atol_vectoriser.transform(X=[a, b, c])
# array([[0.00892619, 0.20804165],
#       [1.19118877, 0.01362205],
#       [0.44476906, 0.05480506]])
```

which is not what is expected from [Atol documentation](https://gudhi.inria.fr/python/latest/representations.html#gudhi.representations.vector_methods.Atol)

VincentRouvreau:
Fix #1031 
Sorry @martinroyer I was not able to keep the history while taking your changes from #1017 but merging #1017 will still be possible in the future.

### DIFF
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -746,7 +746,10 @@ class Atol(BaseEstimator, TransformerMixin):
 
         Parameters:
             quantiser (Object): Object with `fit` (sklearn API consistent) and `cluster_centers` and `n_clusters`
-                attributes, e.g. sklearn.cluster.KMeans. It will be fitted when the Atol object function `fit` is called.
+                attributes, e.g. `sklearn.cluster.KMeans`. It will be fitted when the Atol object function `fit` is
+                called (default: `sklearn.cluster.KMeans` with 2 clusters because there are very few points, a fixed
+                `random_state` to be deterministic and `n_init=10` because it was KMeans default value for sklearn <
+                1.4.0)
             weighting_method (string): constant generic function for weighting the measure points
                 choose from {"cloud", "iidproba"}
                 (default: constant function, i.e. the measure is seen as a point cloud by default).
--- a/src/python/gudhi/representations/vector_methods.py
+++ b/src/python/gudhi/representations/vector_methods.py
@@ -15,6 +15,8 @@ from sklearn.base          import BaseEstimator, TransformerMixin
 from sklearn.exceptions    import NotFittedError
 from sklearn.preprocessing import MinMaxScaler, MaxAbsScaler
 from sklearn.metrics       import pairwise
+from sklearn.cluster import KMeans
+
 try:
     # New location since 1.0
     from sklearn.metrics     import DistanceMetric
@@ -719,13 +721,13 @@ class Atol(BaseEstimator, TransformerMixin):
     >>> a = np.array([[1, 2, 4], [1, 4, 0], [1, 0, 4]])
     >>> b = np.array([[4, 2, 0], [4, 4, 0], [4, 0, 2]])
     >>> c = np.array([[3, 2, -1], [1, 2, -1]])
-    >>> atol_vectoriser = Atol(quantiser=KMeans(n_clusters=2, random_state=202006))
+    >>> atol_vectoriser = Atol(quantiser=KMeans(n_clusters=2, random_state=202006, n_init=10))
     >>> atol_vectoriser.fit(X=[a, b, c]).centers
     array([[ 2.6       ,  2.8       , -0.4       ],
            [ 2.        ,  0.66666667,  3.33333333]])
-    >>> atol_vectoriser(a)
+    >>> atol_vectoriser._transform(a)
     array([0.42375966, 1.18168665])
-    >>> atol_vectoriser(c)
+    >>> atol_vectoriser._transform(c)
     array([1.25157463, 0.02062512])
     >>> atol_vectoriser.transform(X=[a, b, c])
     array([[0.42375966, 1.18168665],
@@ -733,7 +735,12 @@ class Atol(BaseEstimator, TransformerMixin):
            [1.25157463, 0.02062512]])
     """
     # Note the example above must be up to date with the one in tests called test_atol_doc
-    def __init__(self, quantiser, weighting_method="cloud", contrast="gaussian"):
+    def __init__(
+            self,
+            quantiser=KMeans(n_clusters=2, random_state=202312, n_init=10),
+            weighting_method="cloud",
+            contrast="gaussian"
+    ):
         """
         Constructor for the Atol measure vectorisation class.
 
@@ -751,6 +758,7 @@ class Atol(BaseEstimator, TransformerMixin):
         self.quantiser = quantiser
         self.contrast = contrast
         self.weighting_method = weighting_method
+        self._running_transform_names = ""
 
     def get_contrast(self):
         return {
@@ -790,7 +798,9 @@ class Atol(BaseEstimator, TransformerMixin):
 
         measures_concat = np.concatenate(X)
         weights_concat = np.concatenate(sample_weight)
+
         self.quantiser.fit(X=measures_concat, sample_weight=weights_concat)
+
         self.centers = self.quantiser.cluster_centers_
         # Hack, but some people are unhappy if the order depends on the version of sklearn
         self.centers = self.centers[np.lexsort(self.centers.T)]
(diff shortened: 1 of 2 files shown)


src/python/gudhi/representations/vector_methods.py
```diff
@@ -805,7 +815,7 @@ def fit(self, X, y=None, sample_weight=None):
             self.inertias = np.min(dist_centers, axis=0)/2
         return self
 
-    def __call__(self, measure, sample_weight=None):
+    def _transform(self, measure, sample_weight=None):
```
mglisse, line 818: The name `_transform` means that it won't appear in the doc, but we use it in the main example, that's not consistent. @martinroyer , could you remind me what the problem was with `__call__`?

martinroyer: Hmm, I believe it is more similar to what the people at sklearn do, which is pertinent because these are "transformers", whereas `__call__` came from nowhere when I wrote it.
So if you ask me, unless `__call__`  is consistent with other parts of gudhi I favor the sklearn style and would just use `atol_vectoriser.transform([a])` for consistency.

mglisse: We have `__call__` in most other transformers as a shortcut for the case where we want to transform a single diagram. It is clearly not the main interface, and I don't know if it was a good idea, but I don't think it hurts that much. I would be in favor of keeping it to avoid needlessly breaking code that might be using it. We could change the example to make it use `transform` instead of `__call__` though, to advertise the main API. Would that be ok with you?
(A wider discussion about the API of transformers in gudhi is in #982, still open to comments)

martinroyer: I'd like to make the case for not going that direction!

One of the things I gain from looking at code where there are `transform` and `_transform` (yes mostly in sklearn) is that I know that `transform` is going to contain most of the checks (e.g. input checks), and the real  "transform" mechanisms are found in the sub `_transform` functions. So I go through code faster because of that convention, I go directly to the level that interests me.

In this (Atol) instance it is a bit different, the `_transform` acts on a single measure and the `tranform` acts on a list of measures, which is another mechanical distinction, and probably a bad one if we do not want to hide the `_transform` on a single measure (perhaps a conversation to be had?).
But the good part is that the naming convention still shows that those two are mechanically linked. And even if I did not know the convention, I would know these functions are linked to each other because of the name, whereas `__call__` doesn't relate to `transform` at all.

So I guess my vote goes to either :
- decide that we publicly only manipulate lists of measures, therefore hide the transform on a single measure (so as is, `transform` and `_transform`),
- don't decide that but rename `_transform` to `transform_single` (or something...), not `__call__`.

(lastly `__call__` is not IDE-clickable when used someplace which is another downside!)

mglisse: There are several things discussed here, and I think considering them separately would help.
1. Fixing compatibility with recent versions of scikit-learn. This only involves passing some explicit arguments, and is the main goal of this PR. It needs to happen quickly for the next release.
2. Handling pandas output. This is essentially one new short function and an extra line of code to set a member variable, I don't mind handling it at the same time as the above, although it could make sense to add it to more classes at the same time (in a separate PR then).
3. What the interface of all those vectorization classes should be. In my opinion, whatever is decided there, it should be done in a separate PR, and handle all classes consistently at the same time, not just Atol (unless there is a good reason for the difference).

Now 3. can be subdivided into
- Should we keep providing a convenience function (not part of the standard sklearn interface) for the case where we only transform 1 diagram?
- If yes, what should its public name be? `__call__`, `transform_single` (that's going to be longer to use than `transform([x])[0]`), both...
- If as an implementation detail the core computation of `transform`/`__call__`/`transform_single` should be in a function called `_transform` (I don't care either way, I don't mind approving a PR that changes this detail (but not the API) at any time, even if it isn't very consistent between classes, if the author feels strongly about it)
- If we have a public transform_single (whatever its name), whether the examples should use it, or use the main `transform` only, to encourage using one or the other
- Assuming we decide to remove `__call__` from the documentation (either remove or rename), whether we should secretly keep it (without doc) in the code to avoid obnoxiously breaking users' code
- If IDEs that cannot handle standard magic functions like `__call__` suck (please ignore this troll)

I don't think I am really contradicting your end goal too strongly, this is mostly about the order in which to do the changes.

martinroyer: Perfectly summarized!, agree on all points except that `transform_single` is clearly way superior to `transform([x])[0]` (it makes both eyes and hands bleed...)

src/python/gudhi/representations/vector_methods.py
```diff
@@ -719,21 +721,26 @@ class Atol(BaseEstimator, TransformerMixin):
...
     >>> atol_vectoriser.transform(X=[a, b, c])
     array([[0.42375966, 1.18168665],
            [1.06330156, 0.29861028],
            [1.25157463, 0.02062512]])
     """
     # Note the example above must be up to date with the one in tests called test_atol_doc
-    def __init__(self, quantiser, weighting_method="cloud", contrast="gaussian"):
+    def __init__(
+            self,
+            quantiser=KMeans(n_clusters=2, random_state=202312, n_init=10),
```
mglisse, line 740: For the tests (and the values are repeated there), I understand that we want a fixed random_state so the output is deterministic, only 2 clusters because there are very few points, and `n_init=10` because that's what we had before. But I wonder if those are the best choices as default values (for comparison, plain `KMeans()` with all default arguments would use 8, None and 'auto' instead). Do we have a reason to believe that greedy k-means++ won't work well on diagrams and that we need more iterations (`n_init`) than sklearn would do by default? Do we have a reason to recommend a very low number of clusters (2)? Do we have a reason to force determinism?

src/python/gudhi/representations/vector_methods.py
VincentRouvreau: I renamed `_transform` as `__call__` on [e4226ea](https://github.com/GUDHI/gudhi-devel/pull/1056/commits/e4226ea14d00bdc5cc878731a4ab8d677a5445c2)

mglisse (approved):
ok, though I would appreciate a comment on https://github.com/GUDHI/gudhi-devel/pull/1056#discussion_r1634853719 (the values may be fine, but using different default values than sklearn deserves in my opinion a quick justification, for instance as a comment in the source code).

src/python/gudhi/representations/vector_methods.py
martinroyer: I put a `n_init=10` to suppress the warning sklearn had for not setting the parameter, and because `n_init=10` is very understandable contrary to `n_init='auto'`. I recommend setting either, because it can be bothersome on some upper levels if a lower code produces warning on a regular basis... Or we could not set the parameter and hide the warning...
I put `n_clusters=2` because its the lightest we can use that is still multidimensional, so because it can suffice and one should probably work upwards from 2. We could leave the default here.
And no reason to force determinism outside of tests, we can definitely remove `random_state`!

VincentRouvreau:
> ok, though I would appreciate a comment on [#1056 (comment)](https://github.com/GUDHI/gudhi-devel/pull/1056#discussion_r1634853719) (the values may be fine, but using different default values than sklearn deserves in my opinion a quick justification, for instance as a comment in the source code).

[57563ef](https://github.com/GUDHI/gudhi-devel/pull/1056/commits/57563ef4dfc0a814a8a4cdc3407e1b8a3d21771a) I put the comment in the doc. The rendering (some sphinx dependencies were updated):
image

src/python/gudhi/representations/vector_methods.py
mglisse: > I put a `n_init=10` to suppress the warning sklearn had for not setting the parameter, and because `n_init=10` is very understandable contrary to `n_init='auto'`. I recommend setting either, because it can be bothersome on some upper levels if a lower code produces warning on a regular basis... Or we could not set the parameter and hide the warning...

Since sklearn considers `auto` to be the right value (and in particular it yields `1` for the default k-means++), I think it would be nicer to use that (explicitly to avoid a warning with sklearn 1.[23]). This means requiring sklearn 1.2+, I don't know if we already do that.

> I put `n_clusters=2` because its the lightest we can use that is still multidimensional, so because it can suffice and one should probably work upwards from 2.

ok. (I may add a word to the doc to encourage people to override the default)

> And no reason to force determinism outside of tests, we can definitely remove `random_state`!

:+1:

PR #1057, opened 2024-05-27
Cgal and eigen cmake target

VincentRouvreau:
Replace #1049


VincentRouvreau:
Works fine with CGAL-6.0-I-254

PR #1062, opened 2024-06-05
Fix for #1023

### LINKED ISSUE #1023: BettiCurve: `not X` is not a safe way to test for emptiness
https://github.com/GUDHI/gudhi-devel/blob/e7b6d08f25e35891196ce104a014330da442cf12/src/python/gudhi/representations/vector_methods.py#L382
(there is another one a few lines below)
I think this was meant as a short version of `len(X)==0`, but if X is a numpy array (which is perfectly valid), `not` does not have the intended behavior.

(replacing an empty list with a list that contains a single empty diagram also looks suspicious, but I did not look at the details)

MathieuCarriere:
I think the replacement of the empty list with a list containing the empty diagram was to ensure that the `events`, `starts` and `ends` variables does not crash.

Fix #1023


mglisse:
If we `transform` a list of n diagrams, shouldn't we output a list of size n, even when n=0? I.e. why not just `return []`?
For fit_transform... how well do we support `fit([])`? With default parameters, it looks like we build a `grid_` where all the coordinates are NaN, could be worse.

MathieuCarriere:
Yes indeed, good catch. I added a few if conditions to make the outputs consistent.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -379,72 +384,80 @@ def transform(self, X):
...
-            offsets = np.zeros(1 + N, dtype=int)
-            for i in range(0, N):
-                offsets[i+1] = offsets[i] + 2*X[i].shape[0]
-            starts = offsets[0:N]
-            ends = offsets[1:N + 1] - 1
+            if sum([len(x) for x in X]) == 0:
 
-            xs = [-np.inf]
-            bettis = [[0] for i in range(0, N)]
+                return np.zeros((N, 1))
```
mglisse, line 428: Should we still set `self.grid_`? In case we want to call `transform` on this object afterwards. (one of the tests seems to fail because of that)

mglisse (commented):
test_vectorization_empty_diagrams (about Landscape) fails. I didn't check if the problem is with the code, or if the test is wrong and needs to be adapted.

MathieuCarriere:
Thank you for the comments! I've pushed a few fixes according to your suggestions (that I agree with).

src/python/gudhi/representations/vector_methods.py
```diff
@@ -356,8 +360,8 @@ def fit(self, X, y = None):
 
         if self.predefined_grid is None:
             if self.resolution is None: # Flexible/exact version
-                events = np.unique(np.concatenate([pd.ravel() for pd in X] + [[-np.inf]], axis=0))
-                self.grid_ = np.array(events)
+                print("Empty list or empty diagrams: evaluation grid only contains -infinity")
```
mglisse, line 363: Er, unless I am completely confused, this print is executed even if there is a non-empty list of non-empty diagrams :confused:

src/python/gudhi/representations/vector_methods.py
```diff
@@ -379,43 +383,16 @@ def transform(self, X):
...
-        The result is the same as fit(X) followed by transform(X), but potentially faster.
-        """
-
-        if self.predefined_grid is None and self.resolution is None:
-            if not X:
-                X = [np.zeros((0, 2))]
-
-            N = len(X)
+            if self.resolution is None:
+                return np.zeros((N, 1))
```
mglisse, line 391: I think in the exact case we can have `resolution is None` and still have a `grid_` (of length larger than 1).

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: My bad, I put it in the wrong place, thanks for catching this.

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: Yes, I added some code to define `self.grid_` in `transform` if the class has not been fitted yet.

mglisse: That's not what I meant.
```python
import numpy as np
from gudhi.representations import BettiCurve
bc=BettiCurve(resolution=None)
bc.fit([np.array([[1,2],[4,6],[9,12]])])
BettiCurve(resolution=None)
bc.grid_
```
```pre
array([-inf,   1.,   2.,   4.,   6.,   9.,  12.])
```
```python
bc.transform([np.array([[1,2],[4,6],[9,12]])])
```
```pre
array([[0, 1, 0, 1, 0, 1, 0]])
```
```python
bc.transform([])
```
```pre
Empty list or empty diagrams: output contains only zeros
array([], shape=(0, 1), dtype=float64)
```
I believe we want shape=(0, 7) here.
```python
e=np.zeros((0,2))
bc.transform([e,e])
```
```pre
Empty list or empty diagrams: output contains only zeros
array([[0.],
       [0.]])
```
```python
bc.transform([e,np.array([[1,2],[4,6],[9,12]])])
```
```pre
array([[0, 0, 0, 0, 0, 0, 0],
       [0, 1, 0, 1, 0, 1, 0]])
```
So the transform of `e` is not the same depending on the other diagrams in X, that looks wrong.

Let me post a suggestion in a separate comment.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -379,72 +383,88 @@ def transform(self, X):
...
+        if sum([len(x) for x in X]) == 0:
+            print("Empty list or empty diagrams: output contains only zeros")
+            if self.resolution is None:
+                if not self.is_fitted():
+                    self.grid_ = np.array([-np.inf])
+                return np.zeros((N, 1))
+            else:
+                if not self.is_fitted():
+                    self.grid_ = np.full(shape=[self.resolution], fill_value=-np.inf)
+                return np.zeros((N, self.resolution))
```
mglisse, line 397:
```suggestion
        if N == 0:
            return np.zeros((N, len(self.grid_)))
```

MathieuCarriere: I see, that makes sense. Just pushed a fix.

src/python/gudhi/representations/vector_methods.py
```diff
@@ -147,7 +147,12 @@ def _grid_from_sample_range(self, X):
     if not self.keep_endpoints:
         self.new_resolution_ += self.nan_in_range_.sum()
     self.sample_range_fixed_ = _automatic_sample_range(sample_range, X)
-    self.grid_ = np.linspace(self.sample_range_fixed_[0], self.sample_range_fixed_[1], self.new_resolution_)
+    if sum([len(x) for x in X]) == 0:
```
mglisse, line 150: Hmm, actually, this test doesn't seem sufficient. If there is no NaN in `sample_range`, we don't need any points in the diagrams... I'll give it another look soon.

MathieuCarriere: Something like `if sum([len(x) for x in X]) == 0 and np.any(np.isnan(self.sample_range_fixed_))` for instance?

mglisse: I think just `if np.any(np.isnan(self.sample_range_fixed_))` is a good test that `_automatic_sample_range` failed to learn the bounds and the `linspace` branch is not going to work properly (it is too bad that `numpy.linspace(-np.inf, -np.inf, N)` doesn't return an array with N copies of `-np.inf` or we could have handled everything inside `_automatic_sample_range`). And then, if both endpoints are NaN, we can use `-np.inf`, but if one endpoint is specified, maybe we should use (N copies of) that value instead (with a warning in both cases)?

(a user explicitly asking for `sample_range=[-np.inf,-np.inf]` will still get nonsense, but I'll blame that on the user)

mglisse: > (a user explicitly asking for `sample_range=[-np.inf,-np.inf]` will still get nonsense, but I'll blame that on the user)

Actually, this suggests an alternative approach:
- in `_grid_from_sample_range`, only protect linspace with `if self.sample_range_fixed_[0] != self.sample_range_fixed_[1]`, using `np.full` in the `else` branch
- in `_automatic_sample_range`: print the warning when appropriate. Return bounds that never contain NaN (possibly (-inf,-inf), or (b,b) instead)

MathieuCarriere: I see, that makes sense. I've pushed a fix implementing this (I think).

src/python/gudhi/representations/vector_methods.py
```diff
@@ -427,26 +401,65 @@ def fit_transform(self, X, y = None):
...
-                        bettis[k].append(bettis[k][-1])
+                    i += 1
+                for k in range(0, N):
+                    bettis[k].append(bettis[k][-1])
+    
+            return np.array(bettis, dtype=int)[:, 0:-1]
 
-            self.grid_ = np.array(xs)
-            return np.array(bettis, dtype=int)
+    def fit_transform(self, X):
```
mglisse, line 418: Wait what? Is this reverting #1061? Bad merge?

mglisse (commented):
Suggestion to handle the case where a user would pass `sample_range=[0,np.nan]`, if you think it makes sense.

src/python/gudhi/representations/vector_methods.py
MathieuCarriere: weird, I must have messed up with the merge, fixed it

src/python/gudhi/representations/vector_methods.py
```diff
@@ -129,8 +129,8 @@ def _automatic_sample_range(sample_range, X):
                 [Mx,My] = [pre.scalers[0][1].data_max_[0], pre.scalers[1][1].data_max_[0]]
                 return np.where(nan_in_range, np.array([mx, My]), sample_range)
             except ValueError:
-                # Empty persistence diagram case - https://github.com/GUDHI/gudhi-devel/issues/507
-                pass
+                print("Empty list or empty diagrams: sample range is [-infinity, -infinity]")
+                return np.array([-np.inf, -np.inf])
```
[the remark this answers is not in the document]
MathieuCarriere: makes sense, thanks!

src/python/gudhi/representations/vector_methods.py
mglisse: Hmm, now you seem to have a merge issue with the gudhi-deploy submodule.
(yes, git submodules are a pain)

src/python/gudhi/representations/vector_methods.py
```diff
@@ -489,7 +503,10 @@ def fit(self, X, y=None):
         """
         if self.mode == "vector":
             _grid_from_sample_range(self, X)
-            self.step_ = self.grid_[1] - self.grid_[0]
+            if sum([len(x) for x in X]) > 0:
```
mglisse, line 506: Maybe `if self.sample_range_fixed_[0] != self.sample_range_fixed_[1]` as in the other class? Or always compute `self.grid_[1] - self.grid_[0]`, and if it is NaN, replace with 0?

Otherwise I am afraid we are going to break the case where the user manually specified a valid range and then fits on an empty list since there is nothing to learn.

mglisse (commented):
Ok for me, but @VincentRouvreau will need to fix the gudhi-deploy thing before merging this branch.

VincentRouvreau:
> Ok for me, but @VincentRouvreau will need to fix the gudhi-deploy thing before merging this branch.

Reverted on [2a2e764](https://github.com/GUDHI/gudhi-devel/pull/1062/commits/2a2e764364808a5d435b3570b0f814e21a73af9a)

mglisse (approved):

PR #1058, opened 2024-05-27
Fix tensorflow 1.6.0 API change for Layer class

VincentRouvreau:
Fix #1048

PR #1063, opened 2024-06-05
Fix for #1048

### LINKED ISSUE #1048: Perslay tests are failing with tensorflow ~~2.15.1~~ 2.16
While testing a new compilation environment with up-to-date third parties libraries, I faced errors with perslay tests:

```txt
      File "/workspace/gudhi/gudhi-devel/build/src/python/gudhi/tensorflow/perslay.py", line 132, in __init__
        super().__init__(dynamic=True, **kwargs)
      File "/miniconda3/envs/custom-cgal/lib/python3.10/site-packages/keras/src/layers/layer.py", line 263, in __init__
        raise ValueError(
    ValueError: Unrecognized keyword arguments passed to GaussianPerslayPhi: {'dynamic': True}
```

The perslay python tests is ok with tensorflow 2.11.0, but fails with tensorflow 2.15.1

MathieuCarriere:
I removed the `dynamic` argument in the tensorflow models so that the code does not crash with new versions of tensorflow. I also tried running the code with an old tensorflow version (2.14) for retrocompatibility, and had no issues.

Fix #1048


mglisse:
Did you see @VincentRouvreau 's #1058 ? (there is a link to it in #1048)
IIUC, you are saying that his workaround with TensorflowKerasLayer to use `dynamic=True` only for old versions is unnecessary?

MathieuCarriere:
Ah my bad, I missed @VincentRouvreau 's #1058. What I did was indeed to check that `dynamic=True` is not mandatory even for older `tensorflow` versions. However, I did not run extensive checks (I only made sure that the tutorials + tests were working fine), so it might be safer to just check the `tensorflow` version.

mglisse:
@VincentRouvreau this one is up to you.

VincentRouvreau (approved):
And tested with tensorflow 2.12.1

PR #1065, opened 2024-06-10
Do not install debug vcpkgs

VincentRouvreau:


VincentRouvreau:
Do not work and windows-latest was updated

PR #1073, opened 2024-06-15
Rescale values for KDE to avoir underflow

### LINKED ISSUE #1060: Tomato KDE underflow
In tomato, with 'logKDE', we get values from KernelDensity.score_samples and they work fine. With 'KDE', in order to give more importance to higher densities, we naively compute the exponential of these values. However, if the values are all around -1000 (high dimension), the exponential will just send everything to 0 (underflow). I think it would make sense to renormalize first, i.e. add a constant to all scores before computing the exponential (that just multiplies the weights by a constant, which in theory (no underflow) has no impact on tomato except that it changes the numbers on the axes in the diagram plot). The question is then what that constant should be: the maximum score? something closer to the average/median? And how do we document that?

mglisse:
Fix #1060


VincentRouvreau (approved):
I you have any test case, where `exp(-1000)` was happening and that could be tested in the test suite, that would be even better !

PR #1078, opened 2024-06-20
Persistence matrix doc

VincentRouvreau:
Some updates for Persistence matrix documentation and release notes update for 3.10.0


.github/next_release.md
```diff
@@ -4,15 +4,23 @@ We are now using GitHub to develop the GUDHI library, do not hesitate to [fork t
 
 Below is a list of changes made since GUDHI 3.9.0:
 
+- [Persistence matrix](https://gudhi.inria.fr/doc/latest/group__persistence__matrix.html)
+     > Matrix API is in a bêta version and we reserve the right to modify it in the near future.
```
mglisse, line 8: "we reserve the right" sounds like wooden language. Maybe just "and will change in incompatible ways in the near future"?

mglisse: By the way, I don't think "beta" has an accent in English.

VincentRouvreau: I said `and may change`, as it is not sure to change in the future

mglisse: I had used "will" on purpose to make it more scary, but ok for "may".

PR #1080, opened 2024-06-21
code review: gimp was bluring image

VincentRouvreau:
Add simplicial complex for image to be less square


mglisse:
Using SVG may also be a way to get images that are never blurry.

PR #1083, opened 2024-06-24
[Persistence_matrix] move bar type outside

hschreiber:
To be able to use it also for Zigzag persistence (PR #917) more easily.

I also added the possibility to use `std::get` and `std::tuple_size` for the bars for the compatibility with other modules. `std::tuple_size` just doesn't work for structure inheriting from a `Persistence_interval`.

### DIFF
--- a/src/Persistence_matrix/include/gudhi/persistence_interval.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_interval.h
@@ -60,8 +60,8 @@ struct Persistence_interval {
 
 namespace std {
 
-template<>
-struct tuple_size<Gudhi::persistence_matrix::Persistence_interval<int, double> >
+template<typename dimension_type, typename event_value_type>
+struct tuple_size<Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type> >
     : std::integral_constant<std::size_t, 3> {};
 
 template <size_t I, typename dimension_type, typename event_value_type>
--- a/src/Persistence_matrix/include/gudhi/persistence_interval.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_interval.h
@@ -26,6 +26,7 @@ namespace Gudhi {
 namespace persistence_matrix {
 
 /**
+ * @class Persistence_interval persistence_interval.h gudhi/persistence_interval.h
  * @ingroup persistence_matrix
  *
  * @brief Type for an interval in a persistent diagram or barcode.
@@ -37,7 +38,11 @@ namespace persistence_matrix {
 template <typename dimension_type, typename event_value_type>
 struct Persistence_interval {
   /**
-   * @brief Stores the infinity value for birth and death events.
+   * @brief Stores the infinity value for birth and death events. Its value depends on the template parameter
+   * `event_value_type`:
+   * - if `event_value_type` has a native infinity value, it takes this value,
+   * - otherwise, if `event_value_type` is a signed type, it takes value -1,
+   * - otherwise, if `event_value_type` is a unsigned type, it takes the maximal possible value.
    *
    * Is also used as default value for birth and death attributes when not initialized.
    */
@@ -47,13 +52,13 @@ struct Persistence_interval {
 
   /**
    * @brief Default constructor. Initializes the stored dimension to -1 and the stored birth and death values
-   * to @ref Persistence_interval::inf.
+   * to @ref inf.
    */
   Persistence_interval() : dim(-1), birth(inf), death(inf) {}
 
   /**
    * @brief Constructor. Initializes the stored dimension and the stored birth to the given values and the stored
-   * death value to @ref Persistence_interval::inf.
+   * death value to @ref inf.
    * 
    * @param dim Dimension of the cycle.
    * @param birth Birth value of the cycle.
--- a/src/Persistence_matrix/include/gudhi/persistence_interval.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_interval.h
@@ -67,7 +67,15 @@ struct Persistence_interval {
 
   inline friend std::ostream &operator<<(std::ostream &stream, const Persistence_interval &interval) {
     stream << "[" << interval.dim << "] ";
-    stream << interval.birth << ", " << interval.death;
+    if constexpr (std::numeric_limits<event_value_type>::has_infinity) {
+      stream << interval.birth << " - " << interval.death;
+    } else {
+      if (interval.birth == inf) stream << "inf";
+      else stream << interval.birth;
+      stream << " - ";
+      if (interval.death == inf) stream << "inf";
+      else stream << interval.death;
+    }
     return stream;
   }
 };
--- a/src/Persistence_matrix/include/gudhi/persistence_interval.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_interval.h
@@ -29,8 +29,9 @@ namespace persistence_matrix {
  * @class Persistence_interval persistence_interval.h gudhi/persistence_interval.h
  * @ingroup persistence_matrix
  *
- * @brief Type for an interval in a persistent diagram or barcode.
- * Stores the birth, death and dimension of the interval.
+ * @brief Type for an interval in a persistent diagram or barcode. Stores the birth, death and dimension of the
+ * interval. It can be used as a tuple with `get`/`std::get` (birth, death and dimension in this order),
+ * `std::tuple_element` and `std::tuple_size`, as well as structured binding.
  * 
  * @tparam dimension_type Type of the dimension value.
  * @tparam event_value_type Type of the birth and death value.
@@ -51,29 +52,13 @@ struct Persistence_interval {
                                               : static_cast<event_value_type>(-1);
 
   /**
-   * @brief Default constructor. Initializes the stored dimension to -1 and the stored birth and death values
-   * to @ref inf.
-   */
-  Persistence_interval() : dim(-1), birth(inf), death(inf) {}
-
-  /**
-   * @brief Constructor. Initializes the stored dimension and the stored birth to the given values and the stored
-   * death value to @ref inf.
+   * @brief Constructor.
    * 
-   * @param dim Dimension of the cycle.
-   * @param birth Birth value of the cycle.
+   * @param dim Dimension of the cycle. Default value: -1.
+   * @param birth Birth value of the cycle. Default value: @ref inf.
+   * @param death Death value of the cycle. Default value: @ref inf.
    */
-  Persistence_interval(dimension_type dim, event_value_type birth) : dim(dim), birth(birth), death(inf) {}
-
-  /**
-   * @brief Constructor. Initializes the stored dimension, the stored birth value and the stored
-   * death value to the given values.
-   * 
-   * @param dim Dimension of the cycle.
-   * @param birth Birth value of the cycle.
-   * @param death Death value of the cycle.
-   */
-  Persistence_interval(dimension_type dim, event_value_type birth, event_value_type death)
+  Persistence_interval(dimension_type dim = -1, event_value_type birth = inf, event_value_type death = inf)
       : dim(dim), birth(birth), death(death) {}
 
   dimension_type dim;     /**< Dimension of the cycle.*/
@@ -85,40 +70,6 @@ struct Persistence_interval {
     stream << interval.birth << ", " << interval.death;
     return stream;
   }
-};
-
-}  // namespace persistence_matrix
-}  // namespace Gudhi
-
-namespace std {
-
-/**
- * @ingroup persistence_matrix
- *
- * @brief Overload of `std::tuple_size` for @ref Gudhi::persistence_matrix::Persistence_interval.
- * 
- * @tparam dimension_type First template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
- * @tparam event_value_type Second template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
- */
-template <typename dimension_type, typename event_value_type>
-struct tuple_size<Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type> >
-    : std::integral_constant<std::size_t, 3> {};
-
-/**
- * @ingroup persistence_matrix
- *
- * @brief Overload of `std::tuple_element` for @ref Gudhi::persistence_matrix::Persistence_interval.
- * 
- * @tparam I Index of the type to store: 0 for the birth value type, 1 for the death value type and 2 for the
- * dimension value type.
- * @tparam dimension_type First template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
- * @tparam event_value_type Second template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
- */
-template <std::size_t I, typename dimension_type, typename event_value_type>
-struct tuple_element<I, Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type> > {
-  static_assert(I < 3, "Value mismatch at argument 1 in template parameter list. Maximal possible value is 2.");
-
-  using type = typename std::conditional<I < 2, event_value_type, dimension_type>::type;
 };
 
 /**
@@ -201,6 +152,62 @@ constexpr const auto&& get(
   if constexpr (I == 0) return std::move(i.birth);
   if constexpr (I == 1) return std::move(i.death);
   if constexpr (I == 2) return std::move(i.dim);
+}
+
+}  // namespace persistence_matrix
+}  // namespace Gudhi
+
+namespace std {
+
+/**
+ * @ingroup persistence_matrix
+ *
+ * @brief Overload of `std::tuple_size` for @ref Gudhi::persistence_matrix::Persistence_interval.
+ * 
+ * @tparam dimension_type First template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
+ * @tparam event_value_type Second template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
+ */
+template <typename dimension_type, typename event_value_type>
+struct tuple_size<Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type> >
+    : std::integral_constant<std::size_t, 3> {};
+
+/**
+ * @ingroup persistence_matrix
+ *
+ * @brief Overload of `std::tuple_element` for @ref Gudhi::persistence_matrix::Persistence_interval.
+ * 
+ * @tparam I Index of the type to store: 0 for the birth value type, 1 for the death value type and 2 for the
+ * dimension value type.
+ * @tparam dimension_type First template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
+ * @tparam event_value_type Second template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
+ */
+template <std::size_t I, typename dimension_type, typename event_value_type>
+struct tuple_element<I, Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type> > {
+  static_assert(I < 3, "Value mismatch at argument 1 in template parameter list. Maximal possible value is 2.");
+
+  using type = typename std::conditional<I < 2, event_value_type, dimension_type>::type;
+};
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr auto& get(Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>& i) noexcept {
+  return Gudhi::persistence_matrix::get<I>(i);
+}
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr const auto& get(
+    const Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>& i) noexcept {
+  return Gudhi::persistence_matrix::get<I>(i);
+}
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr auto&& get(Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>&& i) noexcept {
+  return Gudhi::persistence_matrix::get<I>(i);
+}
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr const auto&& get(
+    const Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>&& i) noexcept {
+  return Gudhi::persistence_matrix::get<I>(i);
 }
 
 }  // namespace std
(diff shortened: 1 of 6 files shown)


src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * @file persistence_interval.h
+ * @author Hannah Schreiber
+ * @brief Contains @ref Gudhi::persistence_matrix::Persistence_interval class.
+ */
+
+#ifndef PM_INTERVAL_INCLUDED
+#define PM_INTERVAL_INCLUDED
+
+#include <iostream>
```
mglisse, line 20:
```suggestion
#include <ostream>
```
would be sufficient?

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+namespace persistence_matrix {
+
+/**
+ * @ingroup persistence_matrix
+ *
+ * @brief Type for an interval in a persistent diagram or barcode.
+ * Stores the birth, death and dimension of the interval.
+ */
+template <typename dimension_type, typename event_value_type>
+struct Persistence_interval {
```
mglisse, line 34: It is inside namespace persistence_matrix, so I guess it is fine.
If it was going directly in namespace Gudhi, I might object against using this general name for one specific version (the one that also stores the dimension and nothing else).

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+      : dim(dim), birth(birth), death(death) {}
+
+  dimension_type dim;     /**< Dimension of the bar.*/
+  event_value_type birth; /**< Birth index in the current filtration. */
+  event_value_type death; /**< Death index in the current filtration. */
+
+  inline friend std::ostream &operator<<(std::ostream &stream, const Persistence_interval &interval) {
+    stream << "[" << interval.dim << "] ";
+    stream << interval.birth << ", " << interval.death;
+    stream << "\n";
```
mglisse, line 53: I understand this function is only for debug, but I still find it strange to include the `'\n'` in it, I would expect users to do `cout << interval << '\n';` instead.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+  }
+};
+
+}  // namespace persistence_matrix
+}  // namespace Gudhi
+
+namespace std {
+
+template<typename dimension_type, typename event_value_type>
+struct tuple_size<Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type> >
```
mglisse, line 64: I think we should also specialize `tuple_element`.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+    : std::integral_constant<std::size_t, 3> {};
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr typename tuple_element<I, tuple<event_value_type, event_value_type, dimension_type> >::type& get(
```
mglisse, line 68:
```suggestion
constexpr auto& get(
```
? (or some other way to avoid the long, explicit and redundant type)

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+constexpr const typename tuple_element<I, tuple<event_value_type, event_value_type, dimension_type> >::type&& get(
+    const Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>&& i) noexcept {
+  if constexpr (I == 0) return i.birth;
+  if constexpr (I == 1) return i.death;
+  if constexpr (I == 2) return i.dim;
+  // does not compile if I > 2, because of tuple_element
+}
+
+template <typename dimension_type, typename event_value_type>
+constexpr dimension_type& get(
```
mglisse, line 104: I think we could skip the last 4 overloads, since we can't use the type for birth/death, they look more confusing than helpful.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+ * @ingroup persistence_matrix
+ *
+ * @brief Type for an interval in a persistent diagram or barcode.
+ * Stores the birth, death and dimension of the interval.
+ */
+template <typename dimension_type, typename event_value_type>
+struct Persistence_interval {
+  static constexpr event_value_type inf = std::numeric_limits<event_value_type>::has_infinity
+                                              ? std::numeric_limits<event_value_type>::infinity()
+                                              : static_cast<event_value_type>(-1);
```
mglisse, line 37: Maybe briefly mention this in the doc of the class?

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,129 @@
...
+  if constexpr (I == 0) return i.birth;
+  if constexpr (I == 1) return i.death;
+  if constexpr (I == 2) return i.dim;
+  // does not compile if I > 2, because of tuple_element
+}
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr typename tuple_element<I, tuple<event_value_type, event_value_type, dimension_type> >::type&& get(
+    Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>&& i) noexcept {
+  if constexpr (I == 0) return i.birth;
```
mglisse, line 88: I don't think that works, you would need `std::move(i.birth)` if the return type is an rvalue reference.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: Yes, because it is inside `Gudhi::persistence_matrix`, I thought it would be fine. Otherwise, do you have a name suggestion which is not too long and more explicit?

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: With `auto`, the compilation error with I > 2 is not automatic anymore. I could add a `static_cast` by hand, but I just kind of liked it that it adjusts itself. But if you find it more readable with `auto`, I will do it.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: I agree, it is just that if someone had the weird idea to use it for some reasons because it is possible with tuples, it would be there. But I have no problems to remove them.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: If the return type is `&&`, it is not automatically done?

mglisse: no.

src/Persistence_matrix/include/gudhi/persistence_interval.h
mglisse: I think `static_assert(i<=2, "explanation");` is more readable if you want an error. But if you really like tuple_element, ok.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * @ingroup persistence_matrix
+ *
+ * @brief Type for an interval in a persistent diagram or barcode.
+ * Stores the birth, death and dimension of the interval.
+ */
+template <typename dimension_type, typename event_value_type>
+struct Persistence_interval {
+  /**
+   * @brief Stores the infinity value for birth and death events.
```
mglisse, line 36: Ah, sorry my comment wasn't clear. `inf` doesn't need to be visible. What I wanted was a mention of the fact that "infinity" may actually mean "-1" for some types.

src/Persistence_matrix/include/gudhi/persistence_interval.h
mglisse: I'd rather they get an error if they try that.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: Ah ok. I changed it again, is it better like that? The constructors are referring to it.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,208 @@
...
+
+  using type = typename std::conditional<I < 2, event_value_type, dimension_type>::type;
+};
+
+/**
+ * @ingroup persistence_matrix
+ *
+ * @brief Overload of `std::get` for @ref Gudhi::persistence_matrix::Persistence_interval.
+ * 
+ * @tparam I Index of the value to return: 0 for the birth value, 1 for the death value and 2 for the dimension.
```
mglisse, line 129: It seems confusing that the constructor uses the order dim-birth-death while the access is in order birth-death-dim. Can't we pick one consistent order? Or are you trying to be consistent with 2 existing, incompatible types, and this is some compromise?

src/Persistence_matrix/test/pm_matrix_tests.h
```diff
@@ -1443,6 +1443,34 @@ void test_barcode() {
...
+
+  //same test but with std::get
+  bars.clear();
+  for (auto it = barcode.begin(); it != barcode.end(); ++it) {
+    bars.emplace(std::get<2>(*it), std::get<0>(*it), std::get<1>(*it));
+  }
+  it = bars.begin();
+  BOOST_CHECK_EQUAL(std::get<0>(*it), 0);
+  BOOST_CHECK_EQUAL(std::get<1>(*it), 0);
+  BOOST_CHECK_EQUAL(std::get<2>(*it), -1);
```
mglisse, line 1455: Could you also test that [structured binding](https://en.cppreference.com/w/cpp/language/structured_binding) works?
`auto [ x, y, z ] = *it` and then check the values of x, y and z.

mglisse (commented):
I guess we have different stylistic tastes. I tend to favor short:
- `Persistence_interval(dimension_type dim, event_value_type birth, event_value_type death)` doesn't need any description (it should still appear in the doc), the declaration is self-documenting.
- A comment at the beginning that Persistence_interval is usable like a tuple that stores birth, death and dim in this order with std::tuple_element, std::tuple_size, std::get and structured binding, may be clearer than explicitly and redundantly documenting each specialization and overload.

But we don't have to follow my tastes :wink:

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: I initially did all the matrix, zigzag etc. with the order dim-birth-death in mind. But when I was looking at making it consistent with the output of cohomology when adding the `std::get` overloads, I saw that the tuples are in order birth-death-coeff. So I wanted the `std::get` to stick closer to that. I thought that it won't be much of a problem as the users using the constructors them-selves would probably not use the `std::get` to access the members and therefore not be confused. But I agree, it was mostly laziness on my side, as I didn't wanted to correct all the calls to the bars in the matrix to change the order... But I could do that if you find it really to confusing.

hschreiber:
> I guess we have different stylistic tastes. I tend to favor short:
> 
>     * `Persistence_interval(dimension_type dim, event_value_type birth, event_value_type death)` doesn't need any description (it should still appear in the doc), the declaration is self-documenting.
> 
>     * A comment at the beginning that Persistence_interval is usable like a tuple that stores birth, death and dim in this order with std::tuple_element, std::tuple_size, std::get and structured binding, may be clearer than explicitly and redundantly documenting each specialization and overload.
> 
> 
> But we don't have to follow my tastes 😉

By experience I realized that the more redundant the documentation is, the better the people are understanding it... Even though I personally also find it often too much. So, I will try to make a mix of the two approaches by adding what you mentioned in the main description.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,203 @@
...
+
+  /**
+   * @brief Constructor. Initializes the stored dimension, the stored birth value and the stored
+   * death value to the given values.
+   * 
+   * @param dim Dimension of the cycle.
+   * @param birth Birth value of the cycle.
+   * @param death Death value of the cycle.
+   */
+  Persistence_interval(dimension_type dim, event_value_type birth, event_value_type death)
```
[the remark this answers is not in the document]
hschreiber: Right, even all three constructors can even be put in one.

src/Persistence_matrix/test/pm_matrix_tests.h
hschreiber: I had to move things a bit, but I was able to make it work.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,215 @@
...
+
+}  // namespace persistence_matrix
+}  // namespace Gudhi
+
+namespace std {
+
+/**
+ * @ingroup persistence_matrix
+ *
+ * @brief Overload of `std::tuple_size` for @ref Gudhi::persistence_matrix::Persistence_interval.
```
mglisse, line 165: "overload" is for functions, this is a (partial) specialization.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,215 @@
...
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr const auto& get(
+    const Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>& i) noexcept {
+  return Gudhi::persistence_matrix::get<I>(i);
+}
+
+template <size_t I, typename dimension_type, typename event_value_type>
+constexpr auto&& get(Gudhi::persistence_matrix::Persistence_interval<dimension_type, event_value_type>&& i) noexcept {
+  return Gudhi::persistence_matrix::get<I>(i);
```
mglisse, line 204:
```suggestion
  return Gudhi::persistence_matrix::get<I>(std::move(i));
```

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: Ok, I thought you could use it for both.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: Right, I did the same mistake again...

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,215 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
```
VincentRouvreau, line 1: File should be named `Persistence_interval.h`, cf. [code convention](https://github.com/GUDHI/gudhi-devel/blob/master/.github/code_conventions.md#c)

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -0,0 +1,215 @@
...
+  Persistence_interval(dimension_type dim = -1, event_value_type birth = inf, event_value_type death = inf)
+      : dim(dim), birth(birth), death(death) {}
+
+  dimension_type dim;     /**< Dimension of the cycle.*/
+  event_value_type birth; /**< Birth value of the cycle. */
+  event_value_type death; /**< Death value of the cycle. */
+
+  inline friend std::ostream &operator<<(std::ostream &stream, const Persistence_interval &interval) {
+    stream << "[" << interval.dim << "] ";
+    stream << interval.birth << ", " << interval.death;
```
VincentRouvreau, line 70: `Inf` values should be detected and print something else.
```cpp
  using Dimension_type = typename Matrix::dimension_type;
  using Pos_index = typename Matrix::pos_index;
  using Interval = Gudhi::persistence_matrix::Persistence_interval<Dimension_type, Pos_index>;

  Interval bar_default;
  std::cout << bar_default << std::endl;
  // prints: [-1] 4294967295, 4294967295
```

VincentRouvreau: Maybe the same remark than Marc, but:
```cpp
  Interval bar(0,1,2);
  std::cout << bar << std::endl;
  // print: [0] 1, 2
  std::cout << std::get<0>(bar) << ", " << std::get<1>(bar) << ", " << std::get<2>(bar) << std::endl;
  // print: 1, 2, 0
```

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: It is not the only file which doesn't follow this convention in the module. Is it fine if I correct them all in this PR or should I do another one? Also the `example_*` are not respected.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: I am not sure to understand which remark of Marc you are referring to?

src/Persistence_matrix/include/gudhi/persistence_interval.h
mglisse: Please don't mix mass renaming and actual changes in the same branch, it makes it super hard to review.

hschreiber: I agree, but as the changes of this PR are mostly in just one file, I thought there could be an exception. But no problem in doing another PR for the renamings.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: Well, I already explained why it was like that, I was waiting for an answer. So, should I change the order in the constructor to match the `std::get`s? But for the print, I like it better to have the dimension first. `[0] 1, 2` is much easier to read than `1, 2 [0]` in particular when there are a lot of bars.

src/Persistence_matrix/include/gudhi/persistence_interval.h
mglisse: > as the changes of this PR are mostly in just one file

If you rename things all over the place in other files, it will take a reviewer a long time just to make sure that this file is the only one with non-trivial changes.
(I didn't look at what the renaming was about)

hschreiber: True, but I don't think there will be other reviewesr than you two, no? But, like I said, it makes sense to make another PR out of it, it was more of a "you never know" question. The new PR is here: #1106.

hschreiber: @VincentRouvreau I am not sure anymore if I should really rename `persistence_interval.h` to `Persistence_interval`. The convention says:
```
- If a single class or function is provided in a file, its name (with the same letter case) should be used for the file name.
- If a file does not contain a single class, its name should not begin with a capital letter.
```
The file does not contain only the `Persistence_interval` class, but also different global methods (the get etc.) and global structures (tuple_element). So, the second point of the convention should apply, no?

mglisse: (I never read the convention and don't really care)
I would argue that those are part of the interface of this class. Whether they are free functions, static members, etc is a detail.

hschreiber: I agree, that is why they are in the same file. But I am not sure that is how I interpret the convention. I saw the two namings a way for the developer to search more easily: "if there is a upper case, don't need to look here for this one method related to it, it has to be somewhere else". Otherwise, I am not sure to see the utility to have two different naming convention for files...

src/Persistence_matrix/include/gudhi/persistence_interval.h
mglisse: I think I like the new consistent order better (I don't mind that printing is in a different order because the `[]` around the dimension don't make it look like a tuple).
A drawback is that we are probably less likely to be able to use the short versions of the constructor with 1 or 2 elements :shrug:

mglisse (approved):

hschreiber:
> Side remark: I am surprised your intervals _always_ come with a dimension. Can you comment on that

Until now, all use cases I had for barcodes needed the dimension somehow (to plot them separately for example), so I thought that adding the dimension is a natural thing to do. In particular when getting the dimension from the barcode using the matrix is not always possible (eg. when MatIdx != PosIdx). But it could be an option if you think that there are enough cases where the dimension is not needed and will therefore just take up useless space in the barcode.

src/Persistence_matrix/include/gudhi/persistence_interval.h
hschreiber: Yes, I used the short version in zigzag and will have to use the long version now. But it is not the end of the world. Or I could go back to have three different constructors.

mglisse:
> Until now, all use cases I had for barcodes needed the dimension somehow (to plot them separately for example), so I thought that adding the dimension is a natural thing to do. In particular when getting the dimension from the barcode using the matrix is not always possible (eg. when MatIdx != PosIdx). But it could be an option if you think that there are enough cases where the dimension is not needed and will therefore just take up useless space in the barcode.

When computing usual persistence, the matrix doesn't need to know anything about dimensions. We can have a single matrix for all dimensions, or a separate matrix for each dimension.
When storing a diagram, if we care about dimensions, instead of a list of (birth, death, dim), we could store for each dimension a list of (birth, death), that's essentially the new format we are using in Python, and it is more compact.
(I am just explaining why I was surprised, not telling you to go change everything right now)

hschreiber:
> > Until now, all use cases I had for barcodes needed the dimension somehow (to plot them separately for example), so I thought that adding the dimension is a natural thing to do. In particular when getting the dimension from the barcode using the matrix is not always possible (eg. when MatIdx != PosIdx). But it could be an option if you think that there are enough cases where the dimension is not needed and will therefore just take up useless space in the barcode.
> 
> When computing usual persistence, the matrix doesn't need to know anything about dimensions. We can have a single matrix for all dimensions, or a separate matrix for each dimension. When storing a diagram, if we care about dimensions, instead of a list of (birth, death, dim), we could store for each dimension a list of (birth, death), that's essentially the new format we are using in Python, and it is more compact. (I am just explaining why I was surprised, not telling you to go change everything right now)

True, for computing the persistence the dimensions are not needed per se,  but it can be useful for some optimization during reduction. Also with it, the matrix can be used a bit like a simplex tree by storing commonly needed information about your filtration the user don't have to manage anymore. But I agree that we could simplify the structure by removing completely the notion of dimension from it (or make it completely optional to keep the optimization of $`\partial \circ \partial = 0`$).

PR #1085, opened 2024-06-25
[Persistence_matrix] removal of the static from the column pool in compressed matrix

### LINKED ISSUE #1082: Cannot have 2 threads that both compute persistence on completely independent filtrations.
It seems like a big problem if we cannot even have 2 threads that both compute persistence on completely independent filtrations.

_Originally posted by @mglisse in https://github.com/GUDHI/gudhi-devel/pull/669#discussion_r1649558063_

hschreiber:
Fix #1082.

### DIFF
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
@@ -331,8 +331,7 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
    * @brief The disposer object function for boost intrusive container
    */
   struct delete_disposer {
-    delete_disposer() {};
-    delete_disposer(Base_matrix_with_column_compression* matrix) : matrix_(matrix) {};
+    delete_disposer(Base_matrix_with_column_compression* matrix) : matrix_(matrix) {}
 
     void operator()(Column_type* delete_this) { matrix_->columnPool_->destroy(delete_this); }
 
@@ -353,7 +352,7 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
    * @brief Column factory. Has to be a pointer as Simple_object_pool is not swappable, so their adresses have to be
    * exchanged instead.
    */
-  Simple_object_pool<Column_type>* columnPool_;
+  std::unique_ptr<Simple_object_pool<Column_type> > columnPool_;
   inline static const Column_type empty_column_;      /**< Representative for empty columns. */
 
   void _insert_column(index columnIndex);
@@ -434,7 +433,6 @@ template <class Master_matrix>
 inline Base_matrix_with_column_compression<Master_matrix>::~Base_matrix_with_column_compression() 
 {
   columnToRep_.clear_and_dispose(delete_disposer(this));
-  delete columnPool_;
 }
 
 template <class Master_matrix>
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
@@ -296,7 +296,7 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
    * @param cellConstructor Pointer to the cell factory.
    */
   void reset(Column_settings* colSettings) {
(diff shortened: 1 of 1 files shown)


src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
```diff
@@ -344,11 +350,10 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
...
   Column_settings* colSettings_;                      /**< Cell factory. */
   /**
-   * @brief Column factory.
-   * @warning As the member is static, they can eventually be problems if the matrix is duplicated in several threads.
-   * If this become necessary, the static should be removed in the future.
+   * @brief Column factory. Has to be a pointer as Simple_object_pool is not swappable, so their adresses have to be
+   * exchanged instead.
    */
-  inline static Simple_object_pool<Column_type> columnPool_;
+  Simple_object_pool<Column_type>* columnPool_;
```
mglisse, line 356:
```suggestion
  std::unique_ptr<Simple_object_pool<Column_type>> columnPool_;
```
then you can remove delete from the destructor.

src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
```diff
@@ -330,7 +331,12 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
    * @brief The disposer object function for boost intrusive container
    */
   struct delete_disposer {
-    void operator()(Column_type* delete_this) { columnPool_.destroy(delete_this); }
+    delete_disposer() {};
```
mglisse, line 334: Do you need this default constructor?

src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
```diff
@@ -330,7 +331,12 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
...
+    delete_disposer(Base_matrix_with_column_compression* matrix) : matrix_(matrix) {};
```
mglisse, line 335:
```suggestion
    delete_disposer(Base_matrix_with_column_compression* matrix) : matrix_(matrix) {}
```

src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
hschreiber: Is there any other advantage other than to remove the `delete`?

mglisse: It has a move constructor that does the right thing, which helps avoid explicitly defining one in classes that contain it.
If built with `make_unique`, it removes a potential memory leak in case of exceptions.
Mostly, it clearly documents ownership.

src/Persistence_matrix/include/gudhi/Persistence_matrix/base_matrix_with_column_compression.h
```diff
@@ -344,11 +349,10 @@ class Base_matrix_with_column_compression : protected Master_matrix::Matrix_row_
   index nextColumnIndex_;                             /**< Next unused column index. */
   Column_settings* colSettings_;                      /**< Cell factory. */
   /**
-   * @brief Column factory.
-   * @warning As the member is static, they can eventually be problems if the matrix is duplicated in several threads.
-   * If this become necessary, the static should be removed in the future.
+   * @brief Column factory. Has to be a pointer as Simple_object_pool is not swappable, so their adresses have to be
+   * exchanged instead.
```
mglisse, line 353: Does this only come up in the compressed matrix? I think you are using pools elsewhere. I didn't check what those other uses of pools are like, could you tell me why they don't have this issue with moving/swapping?

mglisse: Ah, no, it is also indirectly behind a pointer for others
```
  Column_settings* colSettings_;  //pointer because the of swap operator on matrix_ which also stores the pointer
```

mglisse (approved):

VincentRouvreau (approved):

PR #1105, opened 2024-07-08
Allow storing custom data in simplices

### LINKED ISSUE #891: Store data per simplex in Simplex_tree
It would be useful to be able to store some extra data in the simplices of the Simplex_tree (like `key` and the filtration value currently).

mglisse:
Fix #891. I've been delaying because I wanted to design something more fancy, but it isn't worth it.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -104,6 +104,8 @@ class Simplex_tree {
    *
    * Must be an integer type. */
   typedef typename Options::Simplex_key Simplex_key;
+  /** \bried Extra data stored in each simplex. */
+  typedef typename Get_simplex_data_type<Options>::type Simplex_data;
   /** \brief Type for the vertex handle.
    *
    * Must be a signed integer type. It admits a total order <. */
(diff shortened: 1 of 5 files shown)


VincentRouvreau (commented):
I should have make this comment on the issue, but too late, sorry...

Why not an external class (to avoid to modify the Simplex_tree again) `Filtration_with_data<Filtration_value, Simplex_data>` with the `operator<` on `Filtration_value` that fits the Simplex_tree concept ?

Then we are able to set/get data with:
```cpp
  Simplex_tree<Options_with_filtration_and_data> st;
  st.insert_simplex_and_subfaces({0, 1});
  st.insert_simplex_and_subfaces({2, 1});
  st.insert_simplex_and_subfaces({0, 2});
  st.filtration(st.find({0, 1})).data() = 5;
  st.expansion(3);
  st.filtration(st.find({0, 1, 2})).data() = 4;
  BOOST_CHECK(st.filtration(st.find({0, 1})).data() == 5);
```

mglisse:
`filtration` returns by value currently, so `st.filtration(st.find({0, 1})).data() = 5;` only modifies a temporary copy if it compiles. `expansion` will copy the data at the same time as the filtration value (not wanted). That's doable as a hack, but I don't think it works all that well.

The reason I waited was because I was thinking of something related (see "properties" in Boost.Graph): having only `simplex_data(sh)` (no more `filtration(sh)` and `key(sh)`), where Simplex_data would be something like boost::property (for instance `property<Filtration_tag, Filtration_value, property<Key_tag, Simplex_key, property<Color_tag, Color_type, no_property>>>`), with some way to let Simplex_tree know if some tag requires special treatment (like Filtration_tag during `expansion`), and access through `get(Color_tag(), st.simplex_data(sh))`. I still think that would be nice, if a bit overengineered, but unlikely to be ready soon.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -104,6 +104,8 @@ class Simplex_tree {
    *
    * Must be an integer type. */
   typedef typename Options::Simplex_key Simplex_key;
+  /** \bried Extra data stored in each simplex. */
```
VincentRouvreau, line 107: typo:
```suggestion
  /** \brief Extra data stored in each simplex. */
```

VincentRouvreau (approved):

PR #1106, opened 2024-07-08
[Persistence_matrix] rename files to correspond to gudhi convention

hschreiber:
We forgot to respect the gudhi file name conventions for the module.


mglisse:
Ok if @VincentRouvreau likes the new names...

VincentRouvreau (approved):

PR #1109, opened 2024-07-15
[Persistence matrix] barcode index shift fix

hschreiber:
There was a problem with the indices returned for the barcode in R/RU. The birth indices were row indices (ie IDs) while death were column indices (ie positions). In most of the cases, the two coincide, so I missed it. Now, everything should correspond to positions. I also added a test for it.

### DIFF
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_pairing.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_pairing.h
@@ -96,7 +96,8 @@ class Base_pairing : public std::conditional<
   using id_index = typename Master_matrix::id_index;
   using dictionary_type = typename Master_matrix::bar_dictionary_type;
   using base_matrix = typename Master_matrix::Boundary_matrix_type;
-  using RUM = typename std::conditional<Master_matrix::Option_list::has_removable_columns,
+  //PIDM = Position to ID Map
+  using PIDM = typename std::conditional<Master_matrix::Option_list::has_removable_columns,
                                         Face_position_to_ID_mapper<id_index, pos_index>,
                                         Dummy_pos_mapper
                                        >::type;
@@ -118,7 +119,7 @@ class Base_pairing : public std::conditional<
 };
 
 template <class Master_matrix>
-inline Base_pairing<Master_matrix>::Base_pairing() : RUM(), isReduced_(false) 
+inline Base_pairing<Master_matrix>::Base_pairing() : PIDM(), isReduced_(false) 
 {}
 
 template <class Master_matrix>
@@ -210,10 +211,10 @@ inline void Base_pairing<Master_matrix>::_remove_last(pos_index columnIndex)
     };
   }
 
-  auto it = RUM::map_.find(columnIndex);
-  if (it != RUM::map_.end()){
+  auto it = PIDM::map_.find(columnIndex);
+  if (it != PIDM::map_.end()){
     idToPosition_.erase(it->second);
-    RUM::map_.erase(it);
+    PIDM::map_.erase(it);
   }
 }
 
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_pairing.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/base_pairing.h
@@ -78,10 +78,6 @@ class Base_pairing : public std::conditional<
   const barcode_type& get_current_barcode();
 
   /**
-   * @brief Assign operator.
-   */
-  Base_pairing& operator=(Base_pairing other);
-  /**
    * @brief Swap operator.
    */
   friend void swap(Base_pairing& pairing1, Base_pairing& pairing2) {
@@ -219,17 +215,6 @@ inline void Base_pairing<Master_matrix>::_remove_last(pos_index columnIndex)
     idToPosition_.erase(it->second);
     RUM::map_.erase(it);
   }
-}
-
-template <class Master_matrix>
-inline Base_pairing<Master_matrix>& Base_pairing<Master_matrix>::operator=(Base_pairing<Master_matrix> other) 
-{
-  RUM::operator=(other);
-  barcode_.swap(other.barcode_);
-  deathToBar_.swap(other.deathToBar_);
-  idToPosition_.swap(other.idToPosition_);
-  std::swap(isReduced_, other.isReduced_);
-  return *this;
 }
 
 }  // namespace persistence_matrix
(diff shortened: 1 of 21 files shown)


src/Persistence_matrix/include/gudhi/Persistence_matrix/boundary_matrix.h
```diff
@@ -474,14 +474,16 @@ inline typename Boundary_matrix<Master_matrix>::index Boundary_matrix<Master_mat
...
-      pivot = std::prev(boundary.end())->first;
+    if (boundary.size() != 0){
+      id_index pivot;
+      if constexpr (Master_matrix::Option_list::is_z2) {
+        pivot = *std::prev(boundary.end());
+      } else {
+        pivot = std::prev(boundary.end())->first;
+      }
+      //row container
+      if (ra_opt::rows_->size() <= pivot) ra_opt::rows_->resize(pivot + 1);
```
mglisse, line 485: You had a doubling strategy elsewhere, but here it isn't needed?

src/Persistence_matrix/include/gudhi/Persistence_matrix/overlay_ididx_to_matidx.h
```diff
@@ -756,7 +756,7 @@ inline void Id_to_index_overlay<Matrix_type, Master_matrix_type>::insert_boundar
     GUDHI_CHECK(idToIndex_->find(faceIndex) == idToIndex_->end(),
                 std::invalid_argument("Id_to_index_overlay::insert_boundary - Index for simplex already chosen!"));
   } else {
-    GUDHI_CHECK((idToIndex_->size() <= faceIndex || idToIndex_[faceIndex] == static_cast<index>(-1)),
+    GUDHI_CHECK((idToIndex_->size() <= faceIndex || idToIndex_->operator[](faceIndex) == static_cast<index>(-1)),
```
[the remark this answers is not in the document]
hschreiber: I find both not really readable. So I have no preference. Do you want me to change it?

src/Persistence_matrix/include/gudhi/Persistence_matrix/boundary_matrix.h
hschreiber: I don't think I was very consistent with this strategy. There were also places where it was important that the vector had the exact right size. But I could probably use the doubling strategy here. Do you know if std::vector does already the doubling strategy by itself (that is, even if I resize by one, he reserves the double of size in case of reallocation?). If yes, do you think that doing it on our side is still useful?

src/Persistence_matrix/include/gudhi/Persistence_matrix/overlay_ididx_to_matidx.h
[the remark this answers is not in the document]
mglisse: And `_id_to_index(faceIndex)` is not good either?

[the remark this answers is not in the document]
hschreiber: No, for readability, this would probably be the best option. I just didn't thought that `idToIndex_->operator[](faceIndex)` was so unreadable that it was necessary. But no problem to change that.

src/Persistence_matrix/include/gudhi/Persistence_matrix/boundary_matrix.h
mglisse: Yes, for a `resize` that requires increasing capacity, the standard library already has an exponential strategy (the constant is not always 2, but that's a detail). I checked with libstdc++, libc++, and msvc, and I think the standard requires it (resize is specified in terms of insertions at the end, and those are specified as having amortized constant complexity). So I guess there is no point doing anything subtle in our code.

(note that the situation would be completely different with `reserve`, where the library trusts that the user knows what they are doing, and does not second guess them)

hschreiber: Ok, good to know!

src/Persistence_matrix/include/gudhi/Persistence_matrix/overlay_ididx_to_matidx.h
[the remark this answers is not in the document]
mglisse: It isn't unreadable, but it looks strange to have a lot of ugly calls to `operator[]` when the whole point of operators is their nice short syntax.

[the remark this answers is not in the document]
hschreiber: I agree, it is a bit of a pity that those operators doesn't work similarly with pointers.

src/Persistence_matrix/include/gudhi/Persistence_matrix/base_pairing.h
```diff
@@ -216,13 +213,21 @@ inline void Base_pairing<Master_matrix>::_remove_last(pos_index columnIndex)
...
+
+  auto it = RUM::map_.find(columnIndex);
+  if (it != RUM::map_.end()){
+    idToPosition_.erase(it->second);
+    RUM::map_.erase(it);
+  }
 }
 
 template <class Master_matrix>
 inline Base_pairing<Master_matrix>& Base_pairing<Master_matrix>::operator=(Base_pairing<Master_matrix> other)
```
mglisse, line 225: The comment for copy/move construction implied removing this operator as well, since having a user-declared assignment operator disables the implicit generation of a move constructor (and deprecates the copy).
If you think the default `operator=` doesn't do the right thing (calling `operator=` on each base and member), then we better bring back the constructors (possibly with `=default;` so they aren't too long).

hschreiber: Sorry, I just forgot it.

src/Persistence_matrix/include/gudhi/Persistence_matrix/base_pairing.h
```diff
@@ -83,26 +77,36 @@ class Base_pairing
...
+    pairing1.idToPosition_.swap(pairing2.idToPosition_);
     std::swap(pairing1.isReduced_, pairing2.isReduced_);
   }
 
  protected:
   using pos_index = typename Master_matrix::pos_index;
+  using id_index = typename Master_matrix::id_index;
   using dictionary_type = typename Master_matrix::bar_dictionary_type;
   using base_matrix = typename Master_matrix::Boundary_matrix_type;
+  using RUM = typename std::conditional<Master_matrix::Option_list::has_removable_columns,
```
mglisse, line 99: Does "RUM" come from RU-matrix Map? I am ok with using initials, but there should be at least one comment that helps find out what they stand for.

PR #1110, opened 2024-07-23
[Persistence_matrix] Type renaming to be more coherent with Gudhi conventions

hschreiber:
All types are now starting with a capital letter, with no "_type" or other at the end, as discussed in #917:

              So, if I understood well, your preference would be: all types start with a capitalized letter and only the name of object without any "_type" or "_t" at the end? Eg. `dimension_type` ---> `Dimension`.

_Originally posted by @hschreiber in https://github.com/GUDHI/gudhi-devel/pull/917#discussion_r1686650628_

              Yes on the description of my preference, I think that's mostly what we've been doing so far in Gudhi (with obvious exceptions like value_type for compatibility with the standard library, or Thing_type where "type" is not in the C++ sense but means "kind", "category").

_Originally posted by @mglisse in https://github.com/GUDHI/gudhi-devel/pull/917#discussion_r1686753508_

### DIFF
--- a/src/Persistence_matrix/concept/FieldOperators.h
+++ b/src/Persistence_matrix/concept/FieldOperators.h
@@ -30,8 +30,8 @@ namespace persistence_matrix {
 class FieldOperators 
 {
  public:
-  using element_type = unspecified;         /**< Type for the elements in the field. */
-  using characteristic_type = unspecified;  /**< Type for the field characteristic. */
+  using Element = unspecified;         /**< Type for the elements in the field. */
+  using Characteristic = unspecified;  /**< Type for the field characteristic. */
 
   /**
    * @brief Default constructor. If a non-zero characteristic is given, initializes the field with it.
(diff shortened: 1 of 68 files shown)


mglisse (approved):
I didn't check all of them in details, but it looks good to me.
A couple oddities:
* Boundary_container is a container to describe the content of 1 boundary, while Column_container is a container that contains many columns?
* Column::Column::iterator (we already had that before this PR)

hschreiber:
> I didn't check all of them in details, but it looks good to me. A couple oddities:
> 
>     * Boundary_container is a container to describe the content of 1 boundary, while Column_container is a container that contains many columns?
> 
>     * Column::Column::iterator (we already had that before this PR)

Now that you are saying it, it looks indeed odd. I changed `Boundary_container` to `Boundary_range` and the inner `Column` to `Column_support`. Is it better?

VincentRouvreau (approved):
I did not review them all, but this is the exact role of these code conventions (`characteristic_type` becomes `Characteristic`, the syntax says it is a type, and it is clearer)

PR #1102, opened 2024-07-05
Fix some doxygen warnings in Persistence_matrix

mglisse:
@hschreiber this reduces a bit the number of doxygen warnings. It is in no way complete, but some of the fixes would be straightforward to apply in more places. It is possible that some of the changes are wrong.
Feel free to take over...

PR #1111, opened 2024-07-29
Modify 'enum' to c++11 'enum class' for doxygen to be able to ref enumarations the way it has been done in doc

VincentRouvreau:
After #1102 there were still some warnings for matrix module.
As I prefer the `enum class` c++11 feature and as it fixes these warnings, this is my fix proposal.
I did not test this fix with rips oscillating, zig zag or multi filtration branch, but I hope it does not use some of the former `enum` weak feature (increment, addition, ...)

### DIFF
--- a/src/Persistence_matrix/include/gudhi/persistence_matrix_options.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_matrix_options.h
@@ -27,7 +27,7 @@ namespace persistence_matrix {
  *
  * @brief List of column types.
  */
-enum Column_types { 
+enum class Column_types { 
   LIST,           /**< @ref List_column "": Underlying container is a std::list<@ref Cell*>. */
   SET,            /**< @ref Set_column "": Underlying container is a std::set<@ref Cell*>. */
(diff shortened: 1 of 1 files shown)


mglisse:
Er, we had tested with @hschreiber , and there were no warnings left in Persistence_matrix. Are you using an old version of doxygen?
I tend to find `enum class` annoying, but since Hannah already used these enums as if they were enum classes, why not if Hannah agrees.

VincentRouvreau:
> Er, we had tested with @hschreiber , and there were no warnings left in Persistence_matrix. Are you using an old version of doxygen? I tend to find `enum class` annoying, but since Hannah already used these enums as if they were enum classes, why not if Hannah agrees.

Ok, sorry, I was not aware. Yes my doxygen is quite old, 1.9.3

hschreiber:
> I tend to find `enum class` annoying, but since Hannah already used these enums as if they were enum classes, why not if Hannah agrees.

I don't have a particular opinion in this particular case, as I don't think I use their implicit conversion in here and it will probably not impact any performances. So, as you both prefer.

hschreiber (approved):

PR #1112, opened 2024-07-31
CMake find_package pythoninterp is deprecated

### LINKED ISSUE #139: FindPythonInterp deprecated

VincentRouvreau:
Replaces #1007 
Modification taken from #151 
Fix #139 
CMake ~~3.12 is now 5 years old~~ 3.15 is now 4 years old (july 2019), so it should be ok for every one.
CMake 3.15 is required for `cmake_policy(SET CMP0094 NEW)`
CMake 3.16 (nov 2019) is required if the user wants to specify a python interpreter (`cmake -DPython_EXECUTABLE=/custom/path/to/python3 ...`)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -67,7 +67,7 @@ jobs:
       displayName: 'Install build dependencies'
     - script: |
         $ErrorActionPreference = 'Stop'
-        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" amd64
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
         mkdir build
         cd build
         cmake -DCMAKE_BUILD_TYPE=Release -G "Ninja" -DFORCE_EIGEN_DEFAULT_DENSE_INDEX_TYPE_TO_INT=ON $(cmakeVcpkgFlags) $(cmakeFlags) ..
--- a/src/cmake/modules/GUDHI_third_party_libraries.cmake
+++ b/src/cmake/modules/GUDHI_third_party_libraries.cmake
@@ -134,7 +134,7 @@ add_definitions( -DBOOST_SYSTEM_NO_DEPRECATED )
 
 if (WITH_GUDHI_PYTHON)
   # Find the correct Python interpreter.
-  # Can be set with -DPYTHON_EXECUTABLE=/usr/bin/python3 or -DPython_ADDITIONAL_VERSIONS=3 for instance.
+  # Can be set with -DPYTHON_EXECUTABLE=/usr/bin/python3 or for instance.
   find_package( PythonInterp )
   
   # find_python_module tries to import module in Python interpreter and to retrieve its version number
--- a/src/cmake/modules/GUDHI_third_party_libraries.cmake
+++ b/src/cmake/modules/GUDHI_third_party_libraries.cmake
@@ -134,7 +134,7 @@ add_definitions( -DBOOST_SYSTEM_NO_DEPRECATED )
 
 if (WITH_GUDHI_PYTHON)
   # Find the correct Python interpreter.
-  # Can be set with -DPYTHON_EXECUTABLE=/usr/bin/python3 or for instance.
+  # Can be set with -DPython_EXECUTABLE=/usr/bin/python3 or for instance.
   find_package( PythonInterp )
   
   # find_python_module tries to import module in Python interpreter and to retrieve its version number
--- a/src/cmake/modules/GUDHI_third_party_libraries.cmake
+++ b/src/cmake/modules/GUDHI_third_party_libraries.cmake
@@ -134,15 +134,17 @@ add_definitions( -DBOOST_SYSTEM_NO_DEPRECATED )
 
 if (WITH_GUDHI_PYTHON)
   # Find the correct Python interpreter.
-  # Can be set with -DPython_EXECUTABLE=/usr/bin/python3 or for instance.
-  find_package( PythonInterp )
-  
+  # Can be set with -DPython_EXECUTABLE=/usr/bin/python3 for instance.
+  # CMP0094 OLD finds system python interpreter instead of conda even when using '-DCMAKE_PREFIX_PATH=$CONDA_PREFIX'
+  cmake_policy(SET CMP0094 NEW)
+  find_package( Python COMPONENTS Interpreter)
+
   # find_python_module tries to import module in Python interpreter and to retrieve its version number
   # returns ${PYTHON_MODULE_NAME_UP}_VERSION and ${PYTHON_MODULE_NAME_UP}_FOUND
   function( find_python_module PYTHON_MODULE_NAME )
     string(TOUPPER ${PYTHON_MODULE_NAME} PYTHON_MODULE_NAME_UP)
     execute_process(
-            COMMAND ${PYTHON_EXECUTABLE}  -c "import ${PYTHON_MODULE_NAME}; print(${PYTHON_MODULE_NAME}.__version__)"
+            COMMAND ${Python_EXECUTABLE}  -c "import ${PYTHON_MODULE_NAME}; print(${PYTHON_MODULE_NAME}.__version__)"
             RESULT_VARIABLE PYTHON_MODULE_RESULT
             OUTPUT_VARIABLE PYTHON_MODULE_VERSION
             ERROR_VARIABLE PYTHON_MODULE_ERROR)
@@ -167,7 +169,7 @@ if (WITH_GUDHI_PYTHON)
   function( find_python_module_no_version PYTHON_MODULE_NAME )
     string(TOUPPER ${PYTHON_MODULE_NAME} PYTHON_MODULE_NAME_UP)
     execute_process(
-            COMMAND ${PYTHON_EXECUTABLE}  -c "import ${PYTHON_MODULE_NAME}"
+            COMMAND ${Python_EXECUTABLE}  -c "import ${PYTHON_MODULE_NAME}"
             RESULT_VARIABLE PYTHON_MODULE_RESULT
             ERROR_VARIABLE PYTHON_MODULE_ERROR)
     if(PYTHON_MODULE_RESULT EQUAL 0)
@@ -182,7 +184,7 @@ if (WITH_GUDHI_PYTHON)
     endif()
   endfunction( find_python_module_no_version )
   
-  if( PYTHONINTERP_FOUND )
+  if( TARGET Python::Interpreter )
     find_python_module("cython")
     find_python_module("pytest")
     find_python_module("matplotlib")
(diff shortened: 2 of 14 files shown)


src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -134,15 +134,17 @@ add_definitions( -DBOOST_SYSTEM_NO_DEPRECATED )
 
 if (WITH_GUDHI_PYTHON)
   # Find the correct Python interpreter.
-  # Can be set with -DPYTHON_EXECUTABLE=/usr/bin/python3 or -DPython_ADDITIONAL_VERSIONS=3 for instance.
-  find_package( PythonInterp )
-  
+  # Can be set with -DPython_EXECUTABLE=/usr/bin/python3 for instance.
+  # CMP0094 OLD finds system python interpreter instead of conda even when using '-DCMAKE_PREFIX_PATH=$CONDA_PREFIX'
+  cmake_policy(SET CMP0094 NEW)
+  find_package( Python COMPONENTS Interpreter)
```
mglisse, line 140: Why not the following, to make it less likely to pick a *bad* version of Python?
```suggestion
  find_package( Python COMPONENTS Interpreter Development.Module NumPy)
```

VincentRouvreau: I do not see a lot of added value with Development.Module or NumPy. In the [documentation](https://cmake.org/cmake/help/latest/module/FindPython.html), you can get `Python_NumPy_INCLUDE_DIRS`, that I get in `setup.py.in` with the python code `from numpy import get_include as numpy_get_include`, and `Python_NumPy_VERSION`, that I get with the CMake function `find_python_module` (that performs `${Python_EXECUTABLE}  -c "import ${PYTHON_MODULE_NAME}; print(${PYTHON_MODULE_NAME}.__version__`), and by the way I also need this function for the other python dependencies to decide if a test must be passed or not. Maybe we could also skip tests when `import` throw `ImportError`.

I am not sure this would help to understand this ugly CMakeLists.txt... But tell me if you disagree, I can try

mglisse: > I do not see a lot of added value with Development.Module or NumPy.

I didn't suggest them for the variables they define, but because if they are present, find_package is supposed to find a version of Python that has the dev stuff and numpy installed. If there are several versions of Python installed, this makes it less likely to pick one that is unusable for our purpose. (we do need the dev stuff to compile a plugin)

VincentRouvreau: Ok, that makes sense. I added your proposal on [23a393b](https://github.com/GUDHI/gudhi-devel/pull/1112/commits/23a393bf23af0fc7345fff227a615112ec6018ff)

azure-pipelines.yml
```diff
@@ -56,41 +57,44 @@ jobs:
...
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        mkdir build
+        cd build
+        cmake -DCMAKE_BUILD_TYPE=Release -G "Ninja" -DFORCE_EIGEN_DEFAULT_DENSE_INDEX_TYPE_TO_INT=ON $(cmakeVcpkgFlags) $(cmakeFlags) ..
+        IF %ERRORLEVEL% NEQ 0 EXIT 1
+        ninja
+        IF %ERRORLEVEL% NEQ 0 EXIT 1
+        ctest --output-on-failure -C Release -E diff_files
+        IF %ERRORLEVEL% NEQ 0 EXIT 1
+        for /f %%i in ('where python') do set PYTHON_EXE=%%i
```
mglisse, line 83: Er, this sets PYTHON_EXE to the *last* entry returned by `where python`? That's a bit strange.

src/python/CMakeLists.txt
```diff
@@ -41,663 +62,632 @@ function( add_gudhi_debug_info DEBUG_INFO )
...
+endif()
+if(SCIPY_FOUND)
+  add_gudhi_debug_info("Scipy version ${SCIPY_VERSION}")
+else()
+  disable_python_documentation("scipy")
+endif()
+if(SKLEARN_FOUND)
+  add_gudhi_debug_info("Scikit-learn version ${SKLEARN_VERSION}")
+else()
+  disable_python_documentation("sklear")
```
mglisse, line 124:
```suggestion
  disable_python_documentation("sklearn")
```

azure-pipelines.yml
VincentRouvreau: Because it seems to be the _good way_ to do it in batch :sweat_smile: cf. https://devblogs.microsoft.com/oldnewthing/20120731-00/?p=7003

If you prefer, I can do:
```batch
          for /f %%i in ('where python') do (set PYTHON_EXE=%%i & goto :next)
          :next
```

mglisse: I guess my question is not so much the syntax (this is CI, I don't much care), it is more trying to understand why we need to specify python now when we didn't before. What versions of Python exist under what names on that machine, which one do we want (apparently the only one called plain `python`?), and which one does CMake choose (and ideally why). Again, I don't really care about the CI case in particular, I care more that this is a fallout of the cmake change which could also affect users.

VincentRouvreau: You are right, and I did not talk too much about my experiments around this new feature, my fault too.
If you have a look to https://github.com/VincentRouvreau/play_with_azure_pipelines/blob/159a010d99352edd169949c700af23940b5eabe0/azure-pipelines.yml#L44-L58

In a CMakeLists.txt, I just call `find_package( Python COMPONENTS Interpreter)`.

1. I first call `cmake` without any arguments, I get on [Azure CI](https://dev.azure.com/vincentrouvreau/play_with_azure_pipelines/_build/results?buildId=1171&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=3d94a178-cdfb-5c90-a971-5deeccbbe043) (you may not have access to the logs):
`-- Found Python: C:/hostedtoolcache/windows/Python/3.12.4/x64/python3.exe (found version "3.12.4") found components: Interpreter
`
Note this is not correct, because I said in the yaml file:
```yml
    steps:
      - task: UsePythonVersion@0
        displayName: Use Python $(pythonVersion)
        inputs:
          versionSpec: 3.9
          addToPath: true
          architecture: "x64"
```

2. Then I call:
```batch
for /f %%i in ('where python') do echo %%i
# Prints: C:\hostedtoolcache\windows\Python\3.9.13\x64\python.exe
for /f %%i in ('where python') do (set PYTHON_EXE=%%i & goto :next)
:next
cmake -DPython_EXECUTABLE=%PYTHON_EXE% ..
# -- Found Python: C:\hostedtoolcache\windows\Python\3.9.13\x64\python.exe (found version "3.9.13") found components: Interpreter
```

I don't know how it will affect the users, but for sure it won't affect main users that are only using conda or pip packages.
I also don't know how the python installation is performed on the Azure Windows virtual machines, but the behaviour is quite strange... I can have a look if there is something that could replace `UsePythonVersion` or if its installation is not well performed.

src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -134,23 +134,25 @@ add_definitions( -DBOOST_SYSTEM_NO_DEPRECATED )
 
 if (WITH_GUDHI_PYTHON)
   # Find the correct Python interpreter.
-  # Can be set with -DPYTHON_EXECUTABLE=/usr/bin/python3 or -DPython_ADDITIONAL_VERSIONS=3 for instance.
-  find_package( PythonInterp )
-  
+  # Can be set with -DPython_EXECUTABLE=/usr/bin/python3 for instance.
+  # CMP0094 OLD finds system python interpreter instead of conda even when using '-DCMAKE_PREFIX_PATH=$CONDA_PREFIX'
+  cmake_policy(SET CMP0094 NEW)
+  find_package( Python COMPONENTS Interpreter Development.Module NumPy)
```
mglisse, line 140: It looks like `Development.Module` (as opposed to plain `Development` which also includes `Development.Embed`) requires CMake 3.18, so we should either require that, or say `Development` (and put a comment next to it to remind us to update it in a couple years), or remove it.

src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -134,23 +134,25 @@ add_definitions( -DBOOST_SYSTEM_NO_DEPRECATED )
 
 if (WITH_GUDHI_PYTHON)
   # Find the correct Python interpreter.
-  # Can be set with -DPYTHON_EXECUTABLE=/usr/bin/python3 or -DPython_ADDITIONAL_VERSIONS=3 for instance.
-  find_package( PythonInterp )
-  
+  # Can be set with -DPython_EXECUTABLE=/usr/bin/python3 for instance.
+  # CMP0094 OLD finds system python interpreter instead of conda even when using '-DCMAKE_PREFIX_PATH=$CONDA_PREFIX'
+  cmake_policy(SET CMP0094 NEW)
```
mglisse, line 139: Maybe the comment could say:
Default Python_FIND_STRATEGY to LOCATION: Stops lookup as soon as a version satisfying version constraints is founded (as opposed to VERSION: Try to find the most recent version in all specified locations.)

(Supposedly Python_FIND_VIRTUALENV defaults to FIRST since 3.15, so if you run cmake in an active conda environment the conda version of python should be selected anyway, but setting the policy to NEW looks good in any case.)

azure-pipelines.yml
mglisse: So the system has other versions of Python installed in addition to the one we request, that's fine. The one we request is the only one in the default search path, which is good. However, Python_FIND_REGISTRY is FIRST by default, meaning "Try to use registry before environment variables.", and the registry probably knows both versions of Python and not our preferences. It looks like we could pass `-DPython_FIND_REGISTRY=LAST` (or NEVER) to cmake in this CI. Does that work?

VincentRouvreau: I used `-DPython_FIND_REGISTRY=LAST` and `-DPython_FIND_FRAMEWORK=LAST` on [d190e7d](https://github.com/GUDHI/gudhi-devel/pull/1112/commits/d190e7d2a3bff37608cc6da429d570d61b69004c)

src/cmake/modules/GUDHI_third_party_libraries.cmake
VincentRouvreau: `find_package( Python COMPONENTS Interpreter Development NumPy)` and a comment on [630094b](https://github.com/GUDHI/gudhi-devel/pull/1112/commits/630094bfea2811de28292a81cfa9459a02885d39)

src/cmake/modules/GUDHI_third_party_libraries.cmake
VincentRouvreau: I took your comment proposal on [1b4b4fa](https://github.com/GUDHI/gudhi-devel/pull/1112/commits/1b4b4fac2179b0eeaa134c0cf4a84b74823b963a)

azure-pipelines.yml
```diff
@@ -8,6 +8,10 @@ jobs:
     variables:
       pythonVersion: "3.9"
       cmakeBuildType: Release
+      gudhiCmakeOptions: -DWITH_GUDHI_EXAMPLE=ON -DWITH_GUDHI_TEST=ON -DWITH_GUDHI_UTILITIES=ON -DWITH_GUDHI_PYTHON=ON -DWITH_GUDHI_REMOTE_TEST=ON
+      # On this VM, 2 versions of python are installed. Default Python_FIND_FRAMEWORK is FIRST, which means asks frameworks first
+      # LAST means consult frameworks in last resort, use the standard libraries or headers first
```
mglisse, line 13: I would expect it uses PATH, not "standard libraries or headers", since it is looking for a python exe?

mglisse (approved):

PR #1119, opened 2024-08-13
[Simplex tree] Const Simplex_tree

hschreiber:
Possibility to have a `const Simplex_tree`, but still be able to iterate over it or access information.

It is a draft, as it probably needs more testing (ignore the tests added to `simplex_tree_unit_test.cpp`). But I would like to have an opinion on it and see if I didn't forgot something obvious.

The modifications should not impact old code, if I didn't missed something important. The only big difference is that `Simplex_handle` is now a `const_iterator` instead of an `iterator`. If the users used the interface as expected, than there should be no problems. It will only be a problem if a user used the `Simplex_handle` as an iterator instead of an handle, ie, directly did `sh->second.assign_filtration(1)` instead of using `st.assign_filtration(sh, 1)`. Which is a bit risky and therefore unlikely?


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -174,14 +196,16 @@ class Simplex_tree {
...
    * of a simplex may invalidate any other Simplex_handle in the complex,
    * unless Options::stable_simplex_handles == true. */
-  typedef typename Dictionary::iterator Simplex_handle;
+  typedef typename Dictionary::const_iterator Simplex_handle;
 
  private:
   typedef typename Dictionary::iterator Dictionary_it;
+  typedef typename Dictionary::const_iterator Const_dictionary_it;
   typedef typename Dictionary_it::value_type Dit_value_t;
+  typedef typename Const_dictionary_it::value_type Const_dit_value_t;
```
mglisse, line 205: That's the same type as Dit_value_t.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -292,9 +322,12 @@ class Simplex_tree {
 
   /** \brief Returns a range over the vertices of the simplicial complex. 
    * The order is increasing according to < on Vertex_handles.*/
+  Const_complex_vertex_range complex_vertex_range() const {
+    return Const_complex_vertex_range(boost::make_transform_iterator(root_.members_.begin(), return_first()),
+                                      boost::make_transform_iterator(root_.members_.end(), return_first()));
+  }
   Complex_vertex_range complex_vertex_range() {
```
mglisse, line 329: Why does this one have a non-const version (while most others don't)?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -373,6 +406,11 @@ class Simplex_tree {
     return Boundary_simplex_range(Boundary_simplex_iterator(this, sh),
                                   Boundary_simplex_iterator(this));
   }
+  template<class SimplexHandle>
+  Boundary_simplex_range boundary_simplex_range(SimplexHandle sh) const {
```
mglisse, line 410: Why keep both a const and non-const version if they are equivalent?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -613,29 +660,70 @@ class Simplex_tree {
...
+  /**
+   * @brief Returns true if and only if the filtration value associated to the given simplex is infinity.
+   * If the given handle is @ref null_simplex(), returns true. If @ref SimplexTreeOptions::store_filtration is set
+   * to false, returns always false.
+   * 
+   * @param sh Simplex handle.
+   * @return true If the filtration value is equal to infinity or @p sh is @ref null_simplex().
+   * @return false Otherwise.
+   */
+  static bool has_filtration_value_infinity(Simplex_handle sh){
```
mglisse, line 707: When do we want this? Would exposing get_infinity help?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1489,7 +1590,7 @@ class Simplex_tree {
     {
       Node& node_u = static_cast<Node&>(node_as_hook); //corresponding node, has label u
       Simplex_handle sh_u = simplex_handle_from_node(node_u);
-      Siblings * sib_u = self_siblings(sh_u);
+      Siblings* sib_u = const_cast<Siblings*>(self_siblings(sh_u));
```
mglisse, line 1593: Needing const_cast here looks ugly. Maybe self_siblings should still return a non-const pointer, at least if the Simplex_tree is not const?

mglisse (commented):
Ah, I thought you wanted to make it more similar to standard containers that have both iterator and const_iterator, but here you are basically moving from nothing is const to everything is const (Dictionary_it is not public). It doesn't look bad, but I'll need to think a bit.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -147,22 +147,44 @@ class Simplex_tree {
...
-    Extended_filtration_data(Filtration_value vmin, Filtration_value vmax): minval(vmin), maxval(vmax) {}
+    Extended_filtration_data(const Filtration_value& vmin, const Filtration_value& vmax): minval(vmin), maxval(vmax) {}
   };
   typedef typename std::conditional<Options::store_key, Key_simplex_base_real, Key_simplex_base_dummy>::type
       Key_simplex_base;
 
+  /**
+   * @brief Equals @ref Filtration_value if @ref SimplexTreeOptions::store_filtration is false or the type
+   * corresponds to a native arithmetic type (int, double, etc.). Equals to a reference of @ref Filtration_value
+   * otherwise.
```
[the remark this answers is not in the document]
hschreiber: The idea was just to avoid referencing values which are faster to be copied than redirected. But I have no idea if that really brings that much performance wise. Removing this alias would make things easier, so if you don't think it is necessary, I will do so (in PR #1122 first, I will not really work on here before finishing the others).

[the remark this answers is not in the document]
mglisse: Most small functions are inlined, and after inlining it doesn't matter if the int was passed by reference or value. For most big functions, the time taken by an indirection for an int parameter is negligible. So trying to pass by value conditionally is not worth it. If the type is explicitly `int`, sure, pass by value (less typing). But if it is a template that may sometimes be heavy and require a reference, let's *always* use the reference.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Oh, right! Did it without thinking...

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: There isn't really a reason, I removed the non-const version.

VincentRouvreau (approved):

PR #1121, opened 2024-08-13
[Simplex tree] Additional copy constructor for Simplex_tree

hschreiber:
Another copy constructor for the simplex tree copying from a simplex tree with eventually different options as template parameter. Takes also a method converting the filtration values. Will be used to replace the methods `multify` and `flatten` in PR #976.

I also changed the default copy constructor such that it copies all the keys. For now, it copied the keys of the vertices (with the default copy constructors of the node), but not for the other faces (the nodes were reconstructed from scratch). Copying the keys is not adding to the complexity, so I think that it is worth correcting the inconsistency. Also, I can imagine scenarios where having the keys can be useful when flatten a multi simplex tree (to keep the correspondences of the boundaries in the boundary matrix used for most of the computations).

Edit: we should also think about the attribute 'data' in the nodes to be copied or not.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -10,7 +10,7 @@
  *      - 2023/05 Clément Maria: Edge insertion method for flag complexes
  *      - 2023/05 Hannah Schreiber: Factorization of expansion methods
  *      - 2023/08 Hannah Schreiber (& Clément Maria): Add possibility of stable simplex handles.
- *      - 2024/08 Hannah Schreiber: Addition of customable copy constructor.
+ *      - 2024/08 Hannah Schreiber: Addition of customizable copy constructor.
  *      - YYYY/MM Author: Description of the modification
  */
 
@@ -133,6 +133,7 @@ class Simplex_tree {
 
   struct Key_simplex_base_real {
     Key_simplex_base_real() : key_(-1) {}
+    Key_simplex_base_real(Simplex_key k) : key_(k) {}
     void assign_key(Simplex_key k) { key_ = k; }
     Simplex_key key() const { return key_; }
    private:
@@ -140,6 +141,7 @@ class Simplex_tree {
   };
   struct Key_simplex_base_dummy {
     Key_simplex_base_dummy() {}
+    Key_simplex_base_dummy([[maybe_unused]] Simplex_key k) {}
     // Undefined so it will not link
     void assign_key(Simplex_key);
     Simplex_key key() const;
@@ -155,6 +157,7 @@ class Simplex_tree {
 
   struct Filtration_simplex_base_real {
     Filtration_simplex_base_real() : filt_(0) {}
+    Filtration_simplex_base_real(Filtration_value f) : filt_(f) {}
     void assign_filtration(Filtration_value f) { filt_ = f; }
     Filtration_value filtration() const { return filt_; }
    private:
@@ -162,6 +165,7 @@ class Simplex_tree {
   };
   struct Filtration_simplex_base_dummy {
     Filtration_simplex_base_dummy() {}
+    Filtration_simplex_base_dummy([[maybe_unused]] Filtration_value f) {}
     void assign_filtration(Filtration_value GUDHI_CHECK_code(f)) { GUDHI_CHECK(f == 0, "filtration value specified for a complex that does not store them"); }
     Filtration_value filtration() const { return 0; }
   };
@@ -514,11 +518,17 @@ class Simplex_tree {
     auto root_source = complex_source.root_;
 
     // root members copy
+    root_.members().reserve(root_source.size());
     for (auto& p : root_source.members()){
-      auto it = root_.members().try_emplace(root_.members().end(), p.first);
-      it->second.assign_children(&root_);
-      it->second.assign_filtration(translate_filtration_value(p.second.filtration()));
-      if constexpr (Options::store_key && OtherSimplexTreeOptions::store_key) it->second.assign_key(p.second.key());
+      if constexpr (Options::store_key && OtherSimplexTreeOptions::store_key) {
+        auto it = root_.members().try_emplace(
+            root_.members().end(),
+            p.first,
+            Node(&root_, translate_filtration_value(p.second.filtration()), p.second.key()));
+      } else {
+        auto it = root_.members().try_emplace(
+            root_.members().end(), p.first, Node(&root_, translate_filtration_value(p.second.filtration())));
+      }
     }
 
     rec_copy<OtherSimplexTreeOptions::store_key>(&root_, &root_source, translate_filtration_value);
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -245,7 +245,7 @@ class Simplex_tree {
   typedef boost::iterator_range<Simplex_vertex_iterator> Simplex_vertex_range;
   /** \brief Range over the cofaces of a simplex. */
   typedef typename std::conditional<Options::link_nodes_by_label,
-                                    Optimized_cofaces_simplex_filtered_range,  // faster implem
+                                    Optimized_cofaces_simplex_filtered_range,  // faster implementation
                                     std::vector<Simplex_handle>>::type Cofaces_simplex_range;
 
   /** \private
@@ -345,7 +345,7 @@ class Simplex_tree {
   /** \brief Returns a range over the vertices of a simplex.
    *
    * The order in which the vertices are visited is the decreasing order for < on Vertex_handles,
-   * which is consequenlty
+   * which is consequently
    * equal to \f$(-1)^{\text{dim} \sigma}\f$ the canonical orientation on the simplex.
    */
   Simplex_vertex_range simplex_vertex_range(Simplex_handle sh) const {
@@ -401,6 +401,29 @@ class Simplex_tree {
       root_(nullptr, null_vertex_),
       filtration_vect_(),
       dimension_(-1) { }
+
+  /**
+   * @brief Construct the simplex tree as the copy of a given simplex tree with eventually different template
+   * parameters.
+   * Therefore, should provide a method converting the filtration values of one tree to the another. All other values
+   * are already implicitly convertible if the concept of @ref SimplexTreeOptions is respected (note that there is
+   * an eventual loss of precision or an undefined behaviour if a value is converted into a new type too small to
+   * contain it). Any extra data (@ref Simplex_data) stored in the simplices are ignored in the copy for now.
+   * 
+   * @tparam OtherSimplexTreeOptions Options of the given simplex tree.
+   * @tparam F Method taking an OtherSimplexTreeOptions::Filtration_value as input and returning an
+   * Options::Filtration_value.
+   * @param complex_source Simplex tree to copy.
+   * @param translate_filtration_value Method taking an OtherSimplexTreeOptions::Filtration_value from the source tree
+   * as input and returning the corresponding Options::Filtration_value in the new tree.
+   */
+  template<typename OtherSimplexTreeOptions, typename F>
+  Simplex_tree(const Simplex_tree<OtherSimplexTreeOptions>& complex_source, F&& translate_filtration_value) {
+#ifdef DEBUG_TRACES
+    std::clog << "Simplex_tree custom copy constructor" << std::endl;
+#endif  // DEBUG_TRACES
+    copy_from(complex_source, translate_filtration_value);
+  }
 
   /** \brief User-defined copy constructor reproduces the whole tree structure. */
   Simplex_tree(const Simplex_tree& complex_source) {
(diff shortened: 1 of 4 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -471,27 +495,58 @@ class Simplex_tree {
...
+  template<typename OtherSimplexTreeOptions, typename F>
+  void copy_from(const Simplex_tree<OtherSimplexTreeOptions>& complex_source, F&& translate_filtration_value) {
+    null_vertex_ = complex_source.null_vertex_;
+    filtration_vect_.clear();
+    dimension_ = complex_source.dimension_;
+    auto root_source = complex_source.root_;
+
+    // root members copy
+    for (auto& p : root_source.members()){
+      auto it = root_.members().try_emplace(root_.members().end(), p.first);
```
mglisse, line 518: If you can't use the constructor with `boost::container::ordered_unique_range` (it should be possible, but maybe ugly with the transform_iterator), could you at least `reserve()` before the loop?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
```diff
@@ -33,20 +32,23 @@ namespace Gudhi {
...
+      boost::empty_value<typename SimplexTree::Simplex_data> {
   typedef typename SimplexTree::Siblings Siblings;
   typedef typename SimplexTree::Filtration_value Filtration_value;
   typedef typename SimplexTree::Simplex_key Simplex_key;
   typedef typename SimplexTree::Simplex_data Simplex_data;
 
   Simplex_tree_node_explicit_storage(Siblings * sib = nullptr,
-                                     Filtration_value filtration = 0)
+                                     Filtration_value filtration = 0,
+                                     [[maybe_unused]] Simplex_key key = -1)
```
mglisse, line 48:
```suggestion
                                     [[maybe_unused]] Simplex_key key = SimplexTree::null_key())
```
?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -471,27 +495,58 @@ class Simplex_tree {
...
+      it->second.assign_children(&root_);
+      it->second.assign_filtration(translate_filtration_value(p.second.filtration()));
+      if constexpr (Options::store_key && OtherSimplexTreeOptions::store_key) it->second.assign_key(p.second.key());
+    }
+
+    rec_copy<OtherSimplexTreeOptions::store_key>(&root_, &root_source, translate_filtration_value);
```
mglisse, line 524: Don't we want something like
```suggestion
    rec_copy<OtherSimplexTreeOptions::store_key && Options::store_key>(&root_, &root_source, translate_filtration_value);
```
? Or is that already tested elsewhere?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -471,27 +495,58 @@ class Simplex_tree {
...
+    null_vertex_ = complex_source.null_vertex_;
+    filtration_vect_.clear();
+    dimension_ = complex_source.dimension_;
+    auto root_source = complex_source.root_;
+
+    // root members copy
+    for (auto& p : root_source.members()){
+      auto it = root_.members().try_emplace(root_.members().end(), p.first);
+      it->second.assign_children(&root_);
+      it->second.assign_filtration(translate_filtration_value(p.second.filtration()));
```
mglisse, line 520: For multi, do we have that `emplace` creates a vector of size 1 containing +inf, then `translate` creates a vector of size 2 (for instance), then `assign_filtration` copies that vector, for a total of 3 allocations?

mglisse (commented):
some quick comments

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I only care if the source complex stores the key or not. If the target stores keys but not the source, the `Node` constructor will automatically store -1 as key.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Emplace will create an empty vector I think, but yes, that is an unnecessary number of allocations. It is just that it was how the loop was done initially. Why not do `auto it = root_.members().try_emplace(root_.members().end(), p.first, Node(&root));` instead of `auto it = root_.members().try_emplace(root_.members().end(), p.first); it->second.assign_children(&root_);` ? I was wondering if I missed something and just continued the pattern...

hschreiber: Deleted old answer, as I just realized I remembered it all wrong... And I don't remember why I made this choice. I will try the constructor version, that should avoid at least one allocation.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Done.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -132,13 +133,15 @@ class Simplex_tree {
...
     Key_simplex_base_real() : key_(-1) {}
+    Key_simplex_base_real(Simplex_key k) : key_(k) {}
     void assign_key(Simplex_key k) { key_ = k; }
     Simplex_key key() const { return key_; }
    private:
     Simplex_key key_;
   };
   struct Key_simplex_base_dummy {
     Key_simplex_base_dummy() {}
+    Key_simplex_base_dummy([[maybe_unused]] Simplex_key k) {}
```
mglisse, line 144: Your version is ok, but if you want, in general, I use the less verbose option of not naming the argument (except when naming it is needed for assertions, doc, etc)
```suggestion
    Key_simplex_base_dummy(Simplex_key) {}
```

mglisse, line 144: Actually, isn't it a bit dangerous to allow this and silently ignore the argument? `assign_key` exists but using it causes an error. For dummy filtrations we have an assertion that we don't assign a non-default value.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -154,13 +157,15 @@ class Simplex_tree {
...
     Filtration_simplex_base_real() : filt_(0) {}
+    Filtration_simplex_base_real(Filtration_value f) : filt_(f) {}
     void assign_filtration(Filtration_value f) { filt_ = f; }
     Filtration_value filtration() const { return filt_; }
    private:
     Filtration_value filt_;
   };
   struct Filtration_simplex_base_dummy {
     Filtration_simplex_base_dummy() {}
+    Filtration_simplex_base_dummy([[maybe_unused]] Filtration_value f) {}
```
mglisse, line 168: Same comment as above, shouldn't we have the same assertion as in `assign_filtration`?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I already found it weird that `Filtration_simplex_base_dummy` and `Key_simplex_base_dummy` work so differently. The filtration dummy allows to use `assign_filtration` and `filtration` in release mode and `assign_filtration` throws in debug mode if you try to assign something else than 0. Meanwhile, the key dummy throws systematically if you try to call `assign_key` or `key`. So I am a bit confused on what to do.

But I can indeed already add `GUDHI_CHECK` in both constructors.

By the way, `GUDHI_CHECK_code` does not seem to do much for now, no? Is it necessary?

mglisse: `assert(n>=0)` doesn't mean that `n<0` is allowed in release mode.
`GUDHI_CHECK_code` is for code that is only used by a later GUDHI_CHECK.
The current state is the result of a compromise. On the one hand, when the class has no filtration/key, we want the user to realize quickly that something is wrong is they try to use it. On the other hand, we don't want to write `if constexpr` every other line (and the code predates C++17, so it was even worse), so it can be convenient to have dummy functions. To reconcile the 2 points, the dummy functions produce an error if used, preferably at compile time, otherwise at runtime (in debug mode to avoid slowing down release mode).
Also, filtration and key have different semantics: with no key, we have no key, that's it, while with no filtration, we chose to interpret it as everything having the same filtration (arbitrarily: 0). Of course we could have made a different choice...

hschreiber: > `assert(n>=0)` doesn't mean that `n<0` is allowed in release mode.

I know, it depends on the compiling flags if I remember well. But when compiling Gudhi in release mode, it didn't triggered the throw when I called `assign_filtration(n)` at some point during my tests, it only did it when I recompiled in Debug mode.

> On the other hand, we don't want to write `if constexpr` every other line (and the code predates C++17, so it was even worse),

But precisely because there is this difference, I had to use the `if constexpr (Options::store_key)` at different places. It would make more sense to do the same with the dummy key than with the dummy filtration values.

mglisse: Compiler flags and release/debug mode are the same thing. This is a question of perspective, but let me explain how I see things.
When we have `GUDHI_CHECK(f==0)` in a function, it means that the caller should **never** pass anything but `0` for `f`, whatever the compiler options. If you pass 1, in release mode, it may let you (it won't waste time checking), but whatever follows is undefined behavior; in debug mode, it may stop you, the way it stops you happens to be an exception, but I was against documenting it because this isn't like `vector::at()` where you can reliably catch the exception, here it is an early error that helps you debug your code by telling you about your mistake.

> It would make more sense to do the same with the dummy key than with the dummy filtration values.

Maybe. A compile-time error is preferable to a runtime one whenever possible, but if the runtime one really simplifies things (I didn't check)...

hschreiber: > Compiler flags and release/debug mode are the same thing. 

Hm, I am not sure. If I remember well, you can use de release mode (which is a flag I know) and still have the assert working by using an additional flag.

> This is a question of perspective, but let me explain how I see things. When we have `GUDHI_CHECK(f==0)` in a function, it means that the caller should **never** pass anything but `0` for `f`, whatever the compiler options. If you pass 1, in release mode, it may let you (it won't waste time checking), but whatever follows is undefined behavior; in debug mode, it may stop you, the way it stops you happens to be an exception, but I was against documenting it because this isn't like `vector::at()` where you can reliably catch the exception, here it is an early error that helps you debug your code by telling you about your mistake.

I agree, but this changes nothing to my comment, no? For dummy filtration values, you **should not** use `assign_filtration`, whereas for dummy keys you **cannot** use `assign_key`. In the end, there is the possibility to handle them differently which is not completely coherent in my point of view (even if I get that they don't have the same role). And now, the dummy key is even less coherent as you can "assign" a value by using the constructor (even though not wanted as they is a GUDHI_CHECK), but assigning a value via `assign_key` always throws.

> 
> > It would make more sense to do the same with the dummy key than with the dummy filtration values.
> 
> Maybe. A compile-time error is preferable to a runtime one whenever possible, but if the runtime one really simplifies things (I didn't check)...

I also prefer a compile time error over a run time error. And I personally like the `if constexpr` even if they not always make things very readable. I would have preferred this solution even for the filtration values. It is just that you said that you don't like to use them when not necessary, and I can understand why. I was just highlighting that we use two opposite solutions to more or less the same problem...

mglisse: * "Release mode" is a more or less arbitrary set of flags (typically "-O3 -DNDEBUG" with gcc, it mixes "optimize", "no assertions", "no debug symbols"). Different flags (say just "-O3") can enable optimizations without disabling assertions, if that's what you mean.
* "assigning a value via assign_key always throws." please don't use "throws" (which seems to imply an exception thrown at runtime from this function) for "fails to compile/link" (what currently happens with the dummy assign_key). You don't really need `Key_simplex_base_dummy(Simplex_key)`, you can split out the constructor of Simplex_tree_node_explicit_storage that takes a key, this avoids making it "even less coherent".
* You favor more consistency, we went with something slightly inconsistent to get an earlier error for keys (where is was possible) / less verbose code for filtrations. Both approaches make sense.

hschreiber: > * "Release mode" is a more or less arbitrary set of flags (typically "-O3 -DNDEBUG" with gcc, it mixes "optimize", "no assertions", "no debug symbols"). Different flags (say just "-O3") can enable optimizations without disabling assertions, if that's what you mean.

Yes, that is what I mean.

>     * "assigning a value via assign_key always throws." please don't use "throws" (which seems to imply an exception thrown at runtime from this function) for "fails to compile/link" (what currently happens with the dummy assign_key). You don't really need `Key_simplex_base_dummy(Simplex_key)`, you can split out the constructor of Simplex_tree_node_explicit_storage that takes a key, this avoids making it "even less coherent".

Right, "throw" is not the right word. It was just meant to make a difference between "should not" and "cannot".
And yes, I know that the constructor is not needed, I didn't used it first. But it felt like pushing the problem somewhere else and in case we wanted to change the current approach, it would have made more sense with the constructor. But I know that it is subjective, I have no problem to change it back.

>     * You favor more consistency, we went with something slightly inconsistent to get an earlier error for keys (where is was possible) / less verbose code for filtrations. Both approaches make sense.

There is perhaps a misunderstanding here. I made my initial remark only because I was not sure how I was suppose to handle the changes in those two classes as they work so differently while being so similar. Just to make sure: there was no criticism here. I wanted to make sure that the current behaviors were wanted and not just artifacts which went under the radar. So, if I understand well: I should continue to push key errors as much as possible to compile time while keeping filtration value errors at run time? Even if I find it weird, I am fine going with it.

mglisse: > I wanted to make sure that the current behaviors were wanted and not just artifacts which went under the radar.

Maybe not strongly wanted, but they were conscious choices, yes.

> So, if I understand well: I should continue to push key errors as much as possible to compile time while keeping filtration value errors at run time?

For the time being, I would say yes, unless you can show a significant advantage to changing it (maybe such an advantage will appear naturally after changing the meaning of insert, or some other future change, I don't know).

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
```diff
@@ -33,21 +32,22 @@ namespace Gudhi {
...
   typedef typename SimplexTree::Simplex_key Simplex_key;
   typedef typename SimplexTree::Simplex_data Simplex_data;
 
   Simplex_tree_node_explicit_storage(Siblings * sib = nullptr,
-                                     Filtration_value filtration = 0)
-      : children_(sib) {
-    this->assign_filtration(filtration);
-  }
+                                     Filtration_value filtration = 0,
+                                     [[maybe_unused]] Simplex_key key = SimplexTree::null_key())
```
mglisse, line 48:
```suggestion
                                     Simplex_key key = SimplexTree::null_key())
```
it seems to be used.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -132,17 +133,22 @@ class Simplex_tree {
...
     Key_simplex_base_real() : key_(-1) {}
+    Key_simplex_base_real(Simplex_key k) : key_(k) {}
     void assign_key(Simplex_key k) { key_ = k; }
     Simplex_key key() const { return key_; }
    private:
     Simplex_key key_;
   };
   struct Key_simplex_base_dummy {
     Key_simplex_base_dummy() {}
+    Key_simplex_base_dummy(Simplex_key k) {
```
mglisse, line 144: I think the conclusion was to remove this constructor?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -471,27 +506,64 @@ class Simplex_tree {
...
+    auto root_source = complex_source.root_;
+
+    // root members copy
+    if constexpr (!Options::stable_simplex_handles) root_.members().reserve(root_source.size());
+    for (auto& p : root_source.members()){
+      if constexpr (Options::store_key && OtherSimplexTreeOptions::store_key) {
+        auto it = root_.members().try_emplace(
+            root_.members().end(),
+            p.first,
+            Node(&root_, translate_filtration_value(p.second.filtration()), p.second.key()));
```
mglisse, line 534: Not necessary for this PR (the next one affects of lot of `emplace` calls), but note that an advantage of `try_emplace` is that it builds the mapped type automatically from the arguments you pass, you don't need to construct it yourself.
```suggestion
        auto it = root_.members().try_emplace(
            root_.members().end(),
            p.first,
            &root_, translate_filtration_value(p.second.filtration()), p.second.key());
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Yes, it should be done now.

mglisse (approved):

PR #1122, opened 2024-08-19
[Simplex tree] Generalization of filtration values in Simplex_tree

hschreiber:
Enables `Filtration_value` to be something else than `double` / `float` in the simplex tree. In particular, is thought as a preparation for the multi simplex tree of PR #976, where the filtration value will be entirely new classes.


src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+  /**
+   * @brief Has to construct the default value of FiltrationValue. E.g., 0 for a numerical value or {} for a vector.
+   */
+  FiltrationValue(0);
+
+  // only for default ordering of filtration_vect_ in initialize_filtration
+  /**
+   * @brief Strictly smaller operator. If the filtration values are totally ordered, should be a StrictWeakOrdering.
+   */
+  bool operator<(FiltrationValue f1, FiltrationValue f2);
```
mglisse, line 35: (not new in this PR) the syntax seems bogus, this is a member function, so it takes 3 arguments...

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+ * f(\sigma)\f$ whenever \f$\tau \subseteq \sigma\f$. Ordering the simplices
+ * by increasing filtration values (breaking ties so as a simplex appears after
+ * its subsimplices of same filtration value) provides an indexing scheme
+ * (see IndexingTag).
+ */
+struct FiltrationValue {
+  /**
+   * @brief Has to construct the default value of FiltrationValue. E.g., 0 for a numerical value or {} for a vector.
+   */
+  FiltrationValue(0);
```
mglisse, line 29:
```suggestion
  FiltrationValue();
```
?

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+  bool operator==(FiltrationValue f1, FiltrationValue f2);
+  /**
+   * @brief Not equal operator
+   */
+  bool operator!=(FiltrationValue f1, FiltrationValue f2);
+
+  /**
+   * @brief Given two filtration values at which a simplex exists, returns the minimal union of births generating
+   * a lifetime including those two values. The overload for native arithmetic types like `double` or `int` is
+   * already implemented.
```
mglisse, line 53: ... as the minimum?

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+   * If the filtration is 1-critical and totally ordered (as in one parameter persistence), the union is simply
+   * the minimum of the two values. If the filtration is 1-critical, but not totally ordered (possible for
+   * multi-persistence), than the union is also the minium if the two given values are comparable and the
+   * method should throw an error if they are not, as a same simplex should not exist at those two values.
+   * Finally, if the filtration is k-critical, FiltrationValue should be able to store an union of values and this
+   * method adds the values of @p f2 in @p f1 and removes the values from @p f1 which are comparable and greater than
+   * other values.
```
mglisse, line 60: I think I would prefer describing the general case (k-critical), and then say that if the type cannot represent this result (limited to 1-critical), it should throw.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+   */
+  friend void unify_births(FiltrationValue& f1, const FiltrationValue& f2);
+
+  /**
+   * @brief Given two filtration values, stores in the first value the greatest common upper bound of the two values.
+   * The overload for native arithmetic types like `double` or `int` is already implemented.
+   */
+  friend void push_to_greatest_common_upper_bound(FiltrationValue& f1, const FiltrationValue& f2);
```
mglisse, line 68: The greatest common upper bound it +inf. The smallest common upper bound, on the other hand...
(`make_no_smaller_than`?)

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+   * already implemented.
+   * If the filtration is 1-critical and totally ordered (as in one parameter persistence), the union is simply
+   * the minimum of the two values. If the filtration is 1-critical, but not totally ordered (possible for
+   * multi-persistence), than the union is also the minium if the two given values are comparable and the
+   * method should throw an error if they are not, as a same simplex should not exist at those two values.
+   * Finally, if the filtration is k-critical, FiltrationValue should be able to store an union of values and this
+   * method adds the values of @p f2 in @p f1 and removes the values from @p f1 which are comparable and greater than
+   * other values.
+   */
+  friend void unify_births(FiltrationValue& f1, const FiltrationValue& f2);
```
mglisse, line 62: The other function name starts with `push_to_`, probably to indicate that it modifies its first argument, while this one doesn't. The 2 functions seem similar enough that more consistent names might help? (this is more a "to be discussed", not "I want this name")

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -873,19 +898,26 @@ class Simplex_tree {
     GUDHI_CHECK(*vi != null_vertex(), "cannot use the dummy null_vertex() as a real vertex");
     res_insert = curr_sib->members_.emplace(*vi, Node(curr_sib, filtration));
     if (!res_insert.second) {
-      // if already in the complex
-      if (res_insert.first->second.filtration() > filtration) {
+      if constexpr (Options::store_filtration){
+        // if already in the complex
+        auto oldFil = res_insert.first->second.filtration();
+        unify_births(res_insert.first->second.filtration_raw(), filtration);
+        if (oldFil == res_insert.first->second.filtration()) {
```
mglisse, line 905: For multi, this copies a vector. Should we make `unify_births` return a bool saying if things changed? (does that come almost for free?)

mglisse (commented):
Some initial comments.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: You are right, this should be a friend. Completely missed it.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Well, I wanted to tell here that if the constructor is called with `0` as argument, we should get the desired default value for the filtration value. This is because `double()` or `int()` are not guaranteed to give us `0`, so I had to replace the default arguments `= 0` with `= Filtration_value(0)` instead of `= Filtration_value()`. Do you know a better way to do this?

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Not sure to understand the question? 
Perhaps, my description is not clear? I had trouble to explain this, as it is not really an union of points but an union of infinite bars represented by their birth.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Ah! Sorry, should indeed be `smallest`.....

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: They are very similar in 1D, but in the k-critical case, I find their role quite different.
But it is true that this one does not indicate as much that the first element is modified.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Should be possible, as `add_point` in the multi-critical filtration could also return a bool to indicate if a point was actually added or not.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Is the formulation better now?

src/Simplex_tree/concept/FiltrationValue.h
mglisse: > This is because `double()` or `int()` are not guaranteed to give us `0`

Where did you read that?

src/Simplex_tree/concept/FiltrationValue.h
mglisse: In the k-critical case (i.e. the live region is a union of upper-right quadrants), one function performs a union and one an intersection, that feels rather similar to me.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: I am pretty sure I saw that at several places. But perhaps it is not true anymore? I remember `int i;` giving me quite some random values for `i` at some point. But this is a few years ago and perhaps it was C, not C++ ?

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Ah, I never saw that as an intersection. But now that you say it, it is! So, it should be something like `intersect_births`? That sounds strange. We need perhaps something more precise. Like `unify/intersect_quadrants_by_birth` but with something better than "quadrants". Cones?

src/Simplex_tree/concept/FiltrationValue.h
mglisse: `int i;` does not initialize `i`, so if you try to read it, you can get anything. But `int()` has always zero-initialized, as far as I can tell.

hschreiber: Oh, ok, good to know that there is a difference! Then, I can remove it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1948,8 +2024,8 @@ class Simplex_tree {
    * than it was before. However, `upper_bound_dimension()` will return the old value, which remains a valid upper
    * bound. If you care, you can call `dimension()` to recompute the exact dimension.
    */
-  bool prune_above_filtration(Filtration_value filtration) {
-    if (std::numeric_limits<Filtration_value>::has_infinity && filtration == std::numeric_limits<Filtration_value>::infinity())
+  bool prune_above_filtration(const Filtration_value_rep filtration) {
```
mglisse, line 2027: In the multiparameter case (partial order), I wonder what makes the most sense: removing everything that is larger than something, or removing everything that is not smaller than something? (In 1d the difference only matters for NaN)

hschreiber: The less risky would be to remove everything larger than something as those have to be pruned for sure.

A compromise would be to remove everything larger and to set to infinity all parameters of a filtration value whose value is larger.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_siblings.h
```diff
@@ -78,10 +76,11 @@ class Simplex_tree_siblings {
...
    * present in the node.
    */
-  void insert(Vertex_handle v, Filtration_value filtration_value) {
+  void insert(Vertex_handle v, const Filtration_value& filtration_value) {
     auto ins = members_.emplace(v, Node(this, filtration_value));
-    if (!ins.second && filtration(ins.first) > filtration_value)
-      ins.first->second.assign_filtration(filtration_value);
+    if (!ins.second){
+      unify_births(ins.first->second.filtration_raw(), filtration_value);
+    }
```
mglisse, line 83: Would something like this work and reduce copying a bit?
```c++
template<class Filt>
void insert(Vertex_handle v, Filt&& filtration_value) {
  auto ins = members_.try_emplace(v, this, std::forward<Filt>(filtration_value));
  if (!ins.second)
    unify_births(ins.first->second.filtration_raw(), filtration_value);
}
```
(this reuses `filtration_value` after `forward` only in the case where `try_emplace` did not use it)

hschreiber: This method is literally never used somewhere in Gudhi. And as the "Siblings" can theoretically be accessed from outside, but are not really supposed to be, I am not even sure that some outside code uses it. So I was tempted to ask if you don't want to remove it? Otherwise, I will add a test calling it in the unit test.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,67 @@
...
-  * f(\sigma)\f$ whenever \f$\tau \subseteq \sigma\f$. Ordering the simplices 
-  * by increasing filtration values (breaking ties so as a simplex appears after 
-  * its subsimplices of same filtration value) provides an indexing scheme 
-  * (see IndexingTag).
-  */
-  struct FiltrationValue {
-    /** \brief Operator < is a StrictWeakOrdering. */
-    bool operator<(FiltrationValue f1, FiltrationValue f2);
-  };
+ *
```
VincentRouvreau, line 13: We need to modify SimplexTreeOptions concept, cf.
https://github.com/GUDHI/gudhi-devel/blob/2a89fd56885ccabde60044816c4e8e0d1613b2cc/src/Simplex_tree/concept/SimplexTreeOptions.h#L22-L23

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_siblings.h
hschreiber: I looked at the places were Nodes are inserted in the members. There are not so many of them where the filtration values are also updated. More or less just in `insert_simplex_raw` and `rec_insert_simplex_and_subfaces_sorted`. Which is probably the reason why this method got forgotten. I can use it there if you want or just remove it. Or add a second version where filtration values are not updated (but perhaps other things instead)?

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: I updated it. As the new interface is a bit more complicated, I just referenced the concept. Is it fine?

mglisse: I expect that's what was originally intended, otherwise I don't know if the concept was used anywhere.

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_siblings.h
mglisse: For just 2 uses, I guess it is ok to duplicate the logic.
Although if we can pack emplace+unify_births+update_simplex_tree_after_node_insertion in one function to replace those 2 cases, that could be nice.
As you like.

hschreiber: Ok, I can try unify all this. It would make the methods more readable.

hschreiber: Hm, the only thing is that the siblings are not really aware of the simplex tree they are contained in. So, calling `update_simplex_tree_after_node_insertion` from the sibling would be difficult. But I could move the insert method for siblings out into the simplex tree instead as a private method.

hschreiber: Ok, there are some `if constexpr` in it, but the functions calling the insert are quite simpler now. What do you think?

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Or just `unify` and `intersect` and let the explanation to the doc ? I don't find a good name.

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -99,6 +101,56 @@ template <class, class=void> struct Get_simplex_data_type { typedef No_simplex_d
...
+ * Because the filtration values are totally ordered then, the upper bound is always the maximum of the two values.
+ */
+template <typename Floating_filtration_value,
+          typename = std::enable_if_t<std::is_arithmetic_v<Floating_filtration_value> > >
+bool push_to_smallest_common_upper_bound(Floating_filtration_value& f1, Floating_filtration_value f2)
+{
+  if constexpr (std::is_floating_point_v<Floating_filtration_value>) {
+    if (std::isnan(f1)) {
+      f1 = f2;
+      return !std::isnan(f1);
```
mglisse, line 136: typo?
```suggestion
      return !std::isnan(f2);
```

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -99,6 +101,56 @@ template <class, class=void> struct Get_simplex_data_type { typedef No_simplex_d
...
+    f1 = f2;
+    return true;
+  }
+  return false;
+}
+
+/**
+ * @private
+ * @brief Given two filtration values, stores in the first value the greatest common upper bound of the two values.
+ * If a filtration value has value `NaN`, it should be considered as the lowest value possible.
```
mglisse, line 125: Why is that? Ah, that's what make_filtration_non_decreasing expects (IIRC we only added a little bit of support for NaN in Simplex_tree to help Alpha_complex). Well actually, NaN means "not set" there (so it should be overridden by anything), which happens to be (almost?) equivalent to -inf in this context, but if we wanted to handle it in unify_births it might correspond to +inf in that context.
I wonder if we should try to formalize a bit what NaN means in the Simplex_tree before extending it to more than a hack in make_filtration_non_decreasing? On the other hand, as long as we don't document the behavior publicly, we can still change it later.
I think in the current uses only f1 can be a NaN (in case we decide to give an asymmetric definition...).

mglisse (commented):
Some comments (on everything but the crucial issue of naming the 2 functions :wink: )

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -854,42 +915,25 @@ class Simplex_tree {
...
       GUDHI_CHECK(*vi != null_vertex(), "cannot use the dummy null_vertex() as a real vertex");
-      res_insert = curr_sib->members_.emplace(*vi, Node(curr_sib, filtration));
-      if (res_insert.second) {
-        // Only required when insertion is successful
-        update_simplex_tree_after_node_insertion(res_insert.first);
-      }
-      if (!(has_children(res_insert.first))) {
-        res_insert.first->second.assign_children(new Siblings(curr_sib, *vi));
-      }
+      res_insert = insert_node_<false, true, false>(curr_sib, *vi, filtration);
```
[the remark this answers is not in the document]
hschreiber: I agree... But I found it better than having 4 methods doing more or less the same with just small variations. The other solution would have been to just keep the `try_emplace` and the `update_simplex_tree_after_node_insertion` in the method and letting the rest in the calling methods, but this would have lessen the point of `insert_node_`. Methods like `insert_simplex_raw` are so much more readable now.
If someone has a nicer solution, they are welcome!

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,67 @@
...
+   */
+  friend bool operator<(const FiltrationValue& f1, const FiltrationValue& f2);
+  /**
+   * @brief Equality operator
+   */
+  friend bool operator==(const FiltrationValue& f1, const FiltrationValue& f2);
+  /**
+   * @brief Not equal operator
+   */
+  friend bool operator!=(const FiltrationValue& f1, const FiltrationValue& f2);
```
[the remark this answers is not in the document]
hschreiber: We don't have to, it is just that somewhere in the Simplex tree I saw that `!=` was used, so I added it. We can replace it with `!(==)` if you want.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1958,21 +2028,24 @@ class Simplex_tree {
...
     bool emptied = false;
     Simplex_handle last;
 
     auto to_remove = [this, filt](Dit_value_t& simplex) {
-      if (simplex.second.filtration() <= filt) return false;
-      if (has_children(&simplex)) rec_delete(simplex.second.children());
-      // dimension may need to be lowered
-      dimension_to_be_lowered_ = true;
-      return true;
+      //if filt and simplex.second.filtration() are non comparable, should return false.
```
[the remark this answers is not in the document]
hschreiber: It was to match the multi-parameter case and I forgot somehow about the NaN case... If we disable this for multi-parameter instead, I would roll back the changes.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -146,23 +145,42 @@ class Simplex_tree {
...
   typedef typename std::conditional<Options::store_key, Key_simplex_base_real, Key_simplex_base_dummy>::type
       Key_simplex_base;
 
   struct Filtration_simplex_base_real {
     Filtration_simplex_base_real() : filt_(0) {}
-    void assign_filtration(Filtration_value f) { filt_ = f; }
-    Filtration_value filtration() const { return filt_; }
+    void assign_filtration(const Filtration_value& f) { filt_ = f; }
+    const Filtration_value& filtration() const { return filt_; }
+    Filtration_value& filtration_raw() { return filt_; }
```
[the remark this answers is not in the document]
hschreiber: I needed it before, when I had the std::conditional the return type. But now that we removed it, we don't need those two methods indeed.

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -99,6 +101,56 @@ template <class, class=void> struct Get_simplex_data_type { typedef No_simplex_d
...
+ * This is the overload for when `Filtration_value` is a native arithmetic type, like double, float, int etc.
+ * Because the filtration values are totally ordered then, the upper bound is always the maximum of the two values.
+ */
+template <typename Floating_filtration_value,
+          typename = std::enable_if_t<std::is_arithmetic_v<Floating_filtration_value> > >
+bool push_to_smallest_common_upper_bound(Floating_filtration_value& f1, Floating_filtration_value f2)
+{
+  if constexpr (std::is_floating_point_v<Floating_filtration_value>) {
```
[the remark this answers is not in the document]
hschreiber: `std::isnan` does not work with `int` on Windows........

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
hschreiber: Yes, it was just to comply to what `make_filtration_non_decreasing` expected.
NaN meaning "not set" makes sense to me. Otherwise, I am not sure what other use NaN could have when not "there is a problem here". 

But you are right, we should decide, if we want to handle NaN in general or not and how (except if we keep everything in the backend as for now and in general, we just pretend that NaN is not possible). For example, `unify_birth` does not handle NaN, because we didn't before, but if we introduce NaN to `push_to_smallest_common_upper_bound` because needed in `make_filtration_non_decreasing`, it looks a bit inconsistent for an "outside use".

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: I added a condition to includes the NaNs also. If you prefer that we disable this methods for any filtration value which is not an arithmetic native type, I will just put back the old conditions. Otherwise, I would add in the doc of `prune_above_filtration` that `operator<` is used to judge if a simplex should be removed or not.

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
mglisse: > `!(NaN == NaN)` is different than `NaN != NaN`

Are you sure? Both evaluate to `true`.

There is also the option of going in the other direction: require that all the comparison operators work.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: "arithmetic native type" is not the right condition. It is perfectly fine to use this function when the filtration value type is `CGAL::Lazy_exact_nt` or `mpq_class` or any other non-native scalar type.
The problematic case is multiparameter (which in the other branch can be tested with something like `Options::is_multi_parameter`), where the code will need heavier modifications (according to David's description of what the function should do, for instance), and thus it is safer to disable the function for now.
The main drawback of the old code is that it uses `<=` which isn't currently part of the concept (discussed in another comment).

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
hschreiber: > > `!(NaN == NaN)` is different than `NaN != NaN`
> 
> Are you sure? Both evaluate to `true`.

Huh? I was pretty sure that NaN is supposed to return false with whatever operator, so also for !=. But now, you make me doubt, I have to test it.

> There is also the option of going in the other direction: require that all the comparison operators work.

Sure, but that can be more annoying for someone who wants to create its own class. But I am fine with both approaches. Creating all the counterparts are usually not the most difficult task.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: True, "arithmetic native type" is not the right condition, but I don't see how to discriminate the "right" classes from the "wrong" classes otherwise (I can imagine that someone can come up with other weird cases for filtration other than multi-peristence). In particular, if the reason we want to forbid multi-persistence is that the condition "<" is not always the right choice. "<" is not wrong per-se for multi-persistence, we are just not sure if it is the usual use case for it, if there is one.
So, I was just suggesting that we either explicitly forbid it for everythinh non trivial or that we allow it for everyone, but we highlight in the documentation, that "above" means "operator<" and if it is not what the user wants, he shouldn't use it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1906,26 +1982,20 @@ class Simplex_tree {
     bool modified = false;
```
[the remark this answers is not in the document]
VincentRouvreau: I performed [this benchmark](https://gist.github.com/VincentRouvreau/a2df412c4369098d7cfc9f6274f1c843#file-gistfile1-txt) and it seems this implementation is even faster than `master`'s one:
# master
```txt
(gudhi-devel) gailuron@precision:~/workspace/gudhi/gudhi-devel/build/src/Simplex_tree/benchmark$ make make_filtration_non_decreasing && ./make_filtration_non_decreasing
Building CXX object src/Simplex_tree/benchmark/CMakeFiles/make_filtration_non_decreasing.dir/make_filtration_non_decreasing.cpp.o
Linking CXX executable make_filtration_non_decreasing
Built target make_filtration_non_decreasing
num_simplices= 3799085
... make_filtration_non_decreasing: 0.712s
... make_filtration_non_decreasing: 0.711s
... make_filtration_non_decreasing: 0.71s
... make_filtration_non_decreasing: 0.712s
... make_filtration_non_decreasing: 0.712s
(gudhi-devel) gailuron@precision:~/workspace/gudhi/gudhi-devel/build/src/Simplex_tree/benchmark$ make make_filtration_non_decreasing && ./make_filtration_non_decreasing
Building CXX object src/Simplex_tree/benchmark/CMakeFiles/make_filtration_non_decreasing.dir/make_filtration_non_decreasing.cpp.o
Linking CXX executable make_filtration_non_decreasing
Built target make_filtration_non_decreasing
num_simplices= 38001212
... make_filtration_non_decreasing: 10.347s
... make_filtration_non_decreasing: 10.389s
... make_filtration_non_decreasing: 10.588s
... make_filtration_non_decreasing: 10.648s
... make_filtration_non_decreasing: 10.643s
```

# generalization of filtration branch
```txt
(gudhi-devel) gailuron@precision:~/workspace/gudhi/gudhi-devel/build/src/Simplex_tree/benchmark$ make make_filtration_non_decreasing && ./make_filtration_non_decreasing
Building CXX object src/Simplex_tree/benchmark/CMakeFiles/make_filtration_non_decreasing.dir/make_filtration_non_decreasing.cpp.o
Linking CXX executable make_filtration_non_decreasing
Built target make_filtration_non_decreasing
num_simplices= 3799085
... make_filtration_non_decreasing: 0.642s
... make_filtration_non_decreasing: 0.644s
... make_filtration_non_decreasing: 0.645s
... make_filtration_non_decreasing: 0.644s
... make_filtration_non_decreasing: 0.628s
(gudhi-devel) gailuron@precision:~/workspace/gudhi/gudhi-devel/build/src/Simplex_tree/benchmark$ make make_filtration_non_decreasing && ./make_filtration_non_decreasing
Building CXX object src/Simplex_tree/benchmark/CMakeFiles/make_filtration_non_decreasing.dir/make_filtration_non_decreasing.cpp.o
Linking CXX executable make_filtration_non_decreasing
Built target make_filtration_non_decreasing
num_simplices= 38001212
... make_filtration_non_decreasing: 9.763s
... make_filtration_non_decreasing: 9.833s
... make_filtration_non_decreasing: 9.747s
... make_filtration_non_decreasing: 9.76s
... make_filtration_non_decreasing: 9.734s
```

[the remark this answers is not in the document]
mglisse: Nice unexpected gain!
I think most of the difference comes from initializing the max computation from the old value (instead of computing the max of the faces before comparing it to the old value), which makes comparisons more predictable.
I played a bit with it, and using `!(a<=b)` or `isnan(b)||b<a` didn't make a difference, so we can ignore performance for this discussion.

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
hschreiber: As `!=` was only used in two places, I just removed it from the requirements for now. We can always put it back if it becomes to annoying to not use it.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+ * returns `false`.
+ *
+ * A <EM>filtration</EM> of a cell complex (see FilteredComplex) is
+ * a function \f$f:\mathbf{K} \rightarrow \mathbb{R}\f$ satisfying \f$f(\tau)\leq
+ * f(\sigma)\f$ whenever \f$\tau \subseteq \sigma\f$. Ordering the simplices
+ * by increasing filtration values (breaking ties so as a simplex appears after
+ * its subsimplices of same filtration value) provides an indexing scheme
+ * (see IndexingTag).
+ */
+struct FiltrationValue {
```
mglisse, line 28: I sometimes wonder if we should rename it to FiltrationValueForSimplexTree. But the name of concepts is not that important, it can always be changed later since it does not appear in users' code.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2215,13 +2291,18 @@ class Simplex_tree {
...
   /** \brief Extend filtration for computing extended persistence. 
    * This function only uses the filtration values at the 0-dimensional simplices, 
    * and computes the extended persistence diagram induced by the lower-star filtration 
    * computed with these values. 
    * \post Note that after calling this function, the filtration 
    * values are actually modified. The function `decode_extended_filtration()` 
    * retrieves the original values and outputs the extended simplex type.
+   *
+   * @warning Currently only works for @ref SimplexTreeOptions::Filtration_value which are native
+   * float types like `float` or `double`.
```
mglisse, line 2304: Where does this limitation come from?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2595,17 +2676,15 @@ class Simplex_tree {
     if (members_size > 0) {
       if constexpr (!Options::stable_simplex_handles) sib->members_.reserve(members_size);
       Vertex_handle vertex;
-      Filtration_value filtration;
+      Filtration_value filtration(0);
```
mglisse, line 2679: Why?

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: I also like solution 2 the best. Even though I was referring to `rec_prune_above_filtration` in the comment above which does not make use of intersect/unify. But the spirit is the same.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: If the concepts of FiltrationValue are not the same for other classes, it would make sense to rename it for the long run. But I have the feeling that for now, everything is more or less complying with the needs of the simplex tree, as it is mostly used as `double`.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1081,11 +1164,51 @@ class Simplex_tree {
...
 #ifdef GUDHI_USE_TBB
-    tbb::parallel_sort(filtration_vect_.begin(), filtration_vect_.end(), is_before_in_filtration(this));
+    tbb::parallel_sort(filtration_vect_.begin(), filtration_vect_.end(), is_before_in_totally_ordered_filtration(this));
+#else
+    std::stable_sort(filtration_vect_.begin(), filtration_vect_.end(), is_before_in_totally_ordered_filtration(this));
+#endif
+  }
+
+  // TODO: is there a possibility to give `is_before_in_totally_ordered_filtration(this)()` as default parameter for
+  // `is_before_in_filtration` to avoid having two methods?
```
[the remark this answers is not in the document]
hschreiber: It is.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+   * this method is expected to throw if the two given elements in the filtration values are not comparable.
+   * If they are comparable, the union is simply the minimum of both.
+   *
+   * @return True if and only if the values in @p f1 were actually modified.
+   */
+  friend bool unify(FiltrationValue& f1, const FiltrationValue& f2);
+
+  /**
+   * @brief Given two filtration values, stores in the first value the greatest common upper bound of the two values.
+   * The overload for native arithmetic types like `double` or `int` is already implemented as the maximum of the two
```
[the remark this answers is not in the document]
hschreiber: It is not a bad idea. I added a reference to the overloads in the documentation in case someone wants to use them. They are a global methods which are not directly affecting the simplex tree or others, so I don't think you can do much wrong with them.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,65 @@
...
+   * given values.
+   *
+   * @return True if and only if the values in @p f1 were actually modified.
+   */
+  friend bool intersect(FiltrationValue& f1, const FiltrationValue& f2);
```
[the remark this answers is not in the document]
hschreiber: I agree, I just couldn't decide. Are you fine with `intersect_lifetimes`/`unify_lifetimes`?

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
```diff
@@ -99,6 +101,58 @@ template <class, class=void> struct Get_simplex_data_type { typedef No_simplex_d
...
+ * If a filtration value has value `NaN`, it should be considered as the lowest value possible.
+ * This is the overload for when `Filtration_value` is a native arithmetic type, like double, float, int etc.
+ * Because the filtration values are totally ordered then, the upper bound is always the maximum of the two values.
+ */
+template <typename Arithmetic_filtration_value,
+          typename = std::enable_if_t<std::is_arithmetic_v<Arithmetic_filtration_value> > >
+bool intersect(Arithmetic_filtration_value& f1, Arithmetic_filtration_value f2)
+{
+  if constexpr (std::is_floating_point_v<Arithmetic_filtration_value>) {
+    if (std::isnan(f1)) {
```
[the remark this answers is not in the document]
hschreiber: So, should I change the `if constexpr` condition to `has_quite_NaN` instead to allow NaNs also for other types than native floating types? I used `std::is_floating_point_v` to avoid the problem with Windows refusing to compile `std::isnan` with integers, so we probably need to keep some kind of condition to avoid this problem again.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1149,11 +1228,51 @@ class Simplex_tree {
...
+   * which can be cleared with @ref clear_filtration().
+   * 
+   * @tparam Comparator Method type taking two Simplex_handle as input and returns a bool.
+   * @param is_before_in_filtration Method used to compare two simplices with respect to their position in the
+   * wanted filtration. Takes two Simplex_handle as input and returns true if and only if the first simplex appears
+   * strictly before the second simplex in the resulting 1-parameter filtration.
+   * @param ignore_infinite_values If true, simplices with filtration values at infinity are not cached.
+   */
+  template<typename Comparator>
+  void initialize_filtration(Comparator&& is_before_in_filtration, bool ignore_infinite_values = false) {
```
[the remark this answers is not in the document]
hschreiber: Sure, we could also use some arbitrary predicate. Some method telling us if a simplex should be ignored or not. Would it replace the `ignore_infinite_values` version or add another overload with it?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: The "native" is probably wrong. But it has to be some kind of number (i.e. multi-parameter values make no sense for the algorithm). And I excluded integers for some reasons I cannot remember. Perhaps I wanted just to exclude unsigned integers, as the sign seems to play a role. It is also not impossible that `std::numeric_limits<Filtration_value>::quiet_NaN()` played a role, as it looks quite specific (i.e. why not just put `0`, if the value does not matter and only `p.second` does in that case?), so I just assumed that an existing NaN was important.

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
mglisse: Yes (unless David hates it).

src/Simplex_tree/include/gudhi/Simplex_tree/simplex_tree_options.h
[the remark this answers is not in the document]
mglisse: ok for has_quite_NaN.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: I see. Yes, an empty vector is good, since the only thing we will do with it is overwrite it. That constructor is confusing though. When I see `Filtration_value(n)`, I don't really expect {-inf,-inf,...,-inf} n times. For plain `double`, `Filtration_value filtration;` (without `(0)`) is nicer because it gives sanitizers a chance to warn if there is a case where we forget to assign a value.
But ok for now, we can rediscuss that in the multiparameter branch.
Actually, this seems to be in the deserialization function, does that work in the multiparameter branch?

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Integer does not work indeed, since we scale things to the interval [-2,-1]. NaN is not important; this value is not supposed to be used, and NaN can help detect a mistaken use, but if the type does not support NaN, it would be fine to store 0 or whatever else `quiet_NaN()` returns instead.
Bah, as long as this is just the documentation and not a static_assert, saying that it only works for float/double (and not any field number type) doesn't hurt much, ok.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: I don't know. `initialize_filtration(bool ignore_infinite_values=false)` is already in a release, so we shouldn't break it. But the new overload that takes a comparator is for advanced users and already takes a functor, so it shouldn't hurt much to pass it a lambda as second argument. So I would tend to say "replace", not "add". But I am open to other choices.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,67 @@
...
+   *
+   * @return True if and only if the values in @p f1 were actually modified.
+   */
+  friend bool unify(FiltrationValue& f1, const FiltrationValue& f2);
+
+  /**
+   * @brief Given two filtration values, stores in the first value the lowest common upper bound of the two values.
+   * The overload for arithmetic types like `double` or `int` is already implemented as the maximum of the two
+   * given values and can also be used for non native arithmetic types like `CGAL::Gmpq` as long as it has an
+   * `operator<`. The overload is available with @ref Gudhi::intersect in @ref simplex_tree_options.h "".
```
mglisse, line 66: I don't think we should mention simplex_tree_options.h here
* the name of the file does not match the content (the 2 functions look like they could have their own header, although it doesn't matter if we don't advertise their location too much)
* users should `#include <Gudhi/Simplex_tree.h>` if they want this function, they should not directly include simplex_tree_options.h which is an implementation detail. Doxygen "forces" us to show more details than desirable, but let's limit the damage where possible.

Making the doc of the functions public means that the NaN behavior is documented now. I hope this won't come back to bite us in the [rear end]...

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: Ok. I replaced the second parameter with a functor in the new overload of `initialize_filtration` and also added some unit tests to make sure it works as intended.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1121,39 +1148,120 @@ class Simplex_tree {
...
-   * which can be cleared with clear_filtration().  */
+   * which can be cleared with clear_filtration().
+   */
   void initialize_filtration(bool ignore_infinite_values = false) {
+    if (ignore_infinite_values){
+      initialize_filtration(is_before_in_totally_ordered_filtration(this), [&](Simplex_handle sh) -> bool {
+        return filtration(sh) == Filtration_simplex_base_real::get_infinity();
+      });
+    } else {
+      initialize_filtration(is_before_in_totally_ordered_filtration(this), [](Simplex_handle sh) -> bool {
```
mglisse, line 1215:
```suggestion
      initialize_filtration(is_before_in_totally_ordered_filtration(this), [](Simplex_handle) -> bool {
```
if you don't want a warning about the unused `sh`.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1121,39 +1148,120 @@ class Simplex_tree {
...
+   * Any insertion, deletion or change of filtration value invalidates this cache,
+   * which can be cleared with @ref clear_filtration().
+   * 
+   * @tparam Comparator Method type taking two Simplex_handle as input and returns a bool.
+   * @tparam Ignorer Method type taking one Simplex_handle as input and returns a bool.
+   * @param is_before_in_filtration Method used to compare two simplices with respect to their position in the
+   * wanted filtration. Takes two Simplex_handle as input and returns true if and only if the first simplex appears
+   * strictly before the second simplex in the resulting 1-parameter filtration.
+   * @param ignore_simplex Method taking a simplex handle as input and returns true if and only if the corresponding
+   * simplex should be ignored in the filtration and therefore not be cached.
```
mglisse, line 1247: Maybe we could remind users here that it is their responsibility to ensure that the result is a filtration, i.e. when a face of a simplex is ignored, the simplex should be ignored as well.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2215,13 +2280,18 @@ class Simplex_tree {
...
   /** \brief Extend filtration for computing extended persistence. 
    * This function only uses the filtration values at the 0-dimensional simplices, 
    * and computes the extended persistence diagram induced by the lower-star filtration 
    * computed with these values. 
    * \post Note that after calling this function, the filtration 
    * values are actually modified. The function `decode_extended_filtration()` 
    * retrieves the original values and outputs the extended simplex type.
+   *
+   * @warning Currently only works for @ref SimplexTreeOptions::Filtration_value which are
+   * float types like `float` or `double`, or **signed** integers.
```
mglisse, line 2293: I thought we just said it does not work for integers, because it expects to scale [min,max] to [-2,-1]?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I remember now another reason why I did this. Sometimes `filtration` does not get initialized by default with 0 exactly, but with just a very small value close to 0. And then, the constructor of `Filtration_simplex_base_dummy` complains and throws in Debug mode.

> Actually, this seems to be in the deserialization function, does that work in the multiparameter branch?

I naively though that it does as the methods looked quite general, but it does not... So I added some methods to the FiltrationValue concept (saying that here are only necessary in the case of de/serialization) and also some tests in `simplex_tree_serialization_unit_test.cpp` to make it work.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: I removed the mention of the file. For the NaN behaviour, I could take it out of the description and put it aside in a simple `//` comment which will be ignored by doxygen. Otherwise, I did not found a way to mix public and private with doxygen for a same description.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I added a comment and also added a constructor to the concept. Does it look fine?

I also added `Gudhi::simplex_tree::serialize_trivial` etc. in the simplex_tree group, because otherwise, they didn't appear in the documentation for some reason.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
VincentRouvreau: I did this experiment https://gist.github.com/VincentRouvreau/2b406c867e3494eb3e4d4f50e566d50e on [bcdad85](https://github.com/GUDHI/gudhi-devel/pull/1122/commits/bcdad855d4d42319b4e65851716f225d6d74d23f) and it seems good to me.
Only serialize/deserialize would require some extra code for CGAL::Gmpq (serialize_trivial, deserialize_trivial, get_serialization_size_of and operator<<) to make it work properly.

VincentRouvreau (approved):

VincentRouvreau:
Marc asked me for a small benchmark for serialization/deserialization of a simplex tree.
```python
from gudhi import AlphaComplex
import pickle

gen_points = points.sphere(n_samples = 200000, ambient_dim = 3, radius = 10., sample = "random")

stree = AlphaComplex(points = gen_points).create_simplex_tree()
print(f"num_simplices = {stree.num_simplices()}")
start = timer()
with open('stree.pkl','wb') as f:
    pickle.dump(stree, f)

with open('stree.pkl','rb') as f:
    st_copy = pickle.load(f)

print(timedelta(seconds=timer()-start))
assert stree == st_copy
```

On master:
```bash
$ PYTHONPATH=. python test_pickle.py
num_simplices = 3214259
0:00:00.232157
$ PYTHONPATH=. python test_pickle.py
num_simplices = 3214039
0:00:00.247274
$ PYTHONPATH=. python test_pickle.py
num_simplices = 3211691
0:00:00.247153
$ ll -h stree.pkl
-rw-rw-r-- 1 gailuron gailuron 50M nov.  19 16:57 stree.pkl
```

On generalization branch:
```bash
$ PYTHONPATH=. python test_pickle.py
num_simplices = 3213039
0:00:00.233847
$ PYTHONPATH=. python test_pickle.py
num_simplices = 3212403
0:00:00.253642
$ PYTHONPATH=. python test_pickle.py
num_simplices = 3210995
0:00:00.251737
$ ll -h stree.pkl
-rw-rw-r-- 1 gailuron gailuron 49M nov.  19 17:04 stree.pkl
```

The development does not seem to have any impact on the serialization.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,102 @@
...
+struct FiltrationValue {
+  /**
+   * @brief Has to construct the default value of FiltrationValue.
+   */
+  FiltrationValue();
+  /**
+   * @brief Has to be able to take `0` as input at construction and in this case, construct an empty object.
+   * E.g., 0 for a numerical value or {} for a vector.
+   */
+  FiltrationValue(Any_arithmetic_type v);
```
mglisse, line 37: I already said that elsewhere, but I am not fond of this. The notion that FiltrationValue() will be 0 for 1-parameter and -inf for multiparameter while FiltrationValue(0) will be 0 for 1-parameter and *empty* for multiparameter is confusing.
Would `FiltrationValue(empty_value_t);` (loosely inspired by `std::nullopt_t`) look ok to you? `empty_value_t` could have a template conversion operator so by default it converts to T(0), but at least the uses would look more natural I think.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,102 @@
...
+  /**
+   * @brief Only necessary when serializing the simplex tree. Serialize the given value and insert it at start position.
+   * Overloads for native arithmetic types or other simple types are already implemented with
+   * @ref Gudhi::simplex_tree::serialize_trivial "".
+   * 
+   * @param value The value to serialize.
+   * @param start Start position where the value is serialized.
+   * @return The new position in the array of char for the next serialization.
+   */
+  friend char* serialize_trivial(const FiltrationValue& value, char* start);
```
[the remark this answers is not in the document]
hschreiber: `serialize_object` or  `serialize_to_buffer`? Or is still too generic?

src/Simplex_tree/include/gudhi/Simplex_tree/Simplex_tree_node_explicit_storage.h
```diff
@@ -43,7 +43,13 @@ struct GUDHI_EMPTY_BASE_CLASS_OPTIMIZATION Simplex_tree_node_explicit_storage
   typedef typename SimplexTree::Simplex_key Simplex_key;
   typedef typename SimplexTree::Simplex_data Simplex_data;
 
-  Simplex_tree_node_explicit_storage(Siblings* sib = nullptr, Filtration_value filtration = 0)
+  // Simplex_tree_node_explicit_storage() : children_(nullptr) {}
```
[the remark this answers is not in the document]
hschreiber: Because it was in conflict with the other constructor. But I just have to remove the default value there and than it should work. Those constructors changed a lot back and forth...

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: I understand that it is a bit confusing, but I am not sure how we can be more consistent with the default values?
But even without multi-persistence, right now, `rec_deserialize` will fail in debug mode, if we do not explicitly put the value to 0. So we need either to document its behavior, or we should use an additional `if` to avoid constructing a Node with a filtration value when `store_filtration` is false. Which is probably the better solution here?
I am by the way a bit surprise that the problem does not arise at other places too. I guess, it's the precision randomness of `float`s...

mglisse: IIRC, what I had in mind was closer to
```c++
struct empty_filtration_value_t {
  template <class T> operator T() const { return T(0); }
} empty_filtration_value;
```
And users could add a constructor `MyType(empty_filtration_value_t)` to the definition of their custom filtration value type.
But 1. I may misremember and 2. I don't know if it is the best idea.

I promise this PR is still very high on my TODO-list, just below teaching...

hschreiber: Could this method bring some ambiguity problems if the custom filtration value has several constructors which are not explicit? Or will the constructor always privilege `T = Filtration_value` if we do `Filtration_value f = empty_filtration_value;`?

mglisse: In a code like
```c++
struct empty { template<class T>operator T()const; };
struct A { A(empty); };
void f(){
  A a{empty{}};
}
```
the compiler will always prefer `A(empty)` (not template).
It may still be better to mark the conversion operator `explicit` though.

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
mglisse: They could work. For safety, I would favor something explicit about who uses that function like `serialize_filtration_value_for_gudhi_simplex_tree`, though maybe not that extreme...

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,102 @@
...
+   * @ref Gudhi::simplex_tree::deserialize_trivial "".
+   * 
+   * @param value The value where to deserialize based on its type.
+   * @param start Start position where the value is serialized.
+   * @return The new position in the array of char for the next deserialization.
+   */
+  friend const char* deserialize_trivial(FiltrationValue& value, const char* start);
+
+  /**
+   * @brief Only necessary when (de)serializing the simplex tree. Returns the serialization size of the given object.
```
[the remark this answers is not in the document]
mglisse: For deserialization, you would call it after calling `deserialize` (you need an object, so you cannot call it before)? But the return value of `deserialize` already gives you that information for free.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: The assertion is `GUDHI_CHECK(f == Filtration_value()`, not `GUDHI_CHECK(f == Filtration_value(0)`. I don't know if we should have here `Filtration_value filtration{};`, or change the assertion to compare to `Filtration_value(0)`, or remove the assertion, I just wanted to point out this inconsistency for now.

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
hschreiber: It is only used by the simplex tree for now, but now that the serialization methods exists, I could imagine that we want to serialize other structures containing filtration values too in the future? If we implement filtration value classes (like the one for multi-parameter filtrations), with a friend method `serialize_filtration_value_for_gudhi_simplex_tree` it makes it sound as if the serialization method is very specific to the simplex tree and another one would be needed for another structure.
If we want to be more safe, we could perhaps specify more how the serialization is done? Something like `serialize_value_byte_by_byte_to_char_buffer`?

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
hschreiber: Yes, it is just that I had the unitary tests as examples in mind where it is used for both, but indeed it is not very useful for deserialization.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: I added the `empty_filtration_value_t` class and the constructor to the FiltrationValue concept.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: The new `empty_filtration_value` is now used in the check. This should be more consistent.

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
```diff
@@ -0,0 +1,117 @@
...
+ * @brief Serialize the given value and insert it at start position using
+ * @ref Gudhi::simplex_tree::serialize_trivial "".
+ * 
+ * @param[in] value The value to serialize.
+ * @param[in] start Start position where the value is serialized.
+ * @return The new position in the array of char for the next serialization.
+ * 
+ * @warning It is the user's responsibility to provide a pointer to a buffer with enough memory space.
+ */
+template <typename Trivial_filtration_value>
```
mglisse, line 94: There is the question of whether we should constrain this overload (enable_if, maybe with is_arithmetic, maybe with something else). I guess that can wait.

mglisse (commented):
(Note to self: just check the replies, don't reread the diff)

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,106 @@
...
+   * @ref Gudhi::deserialize_value_to_char_buffer "".
+   * 
+   * @param value The value where to deserialize based on its type.
+   * @param start Start position where the value is serialized.
+   * @return The new position in the array of char for the next deserialization.
+   */
+  friend const char* deserialize_value_to_char_buffer(FiltrationValue& value, const char* start);
+
+  /**
+   * @brief Only necessary when serializing the simplex tree. Returns the serialization size of the given object.
```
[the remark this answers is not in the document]
hschreiber: It should be fine as long as:
- the upper bound is consistent from one call to another (i.e., same for everyone or depends on the "visible" part of the object only, and not from some backend stuff, like pre-allocation size or so) 
- the serialize and the deserialize method is aware of it: they have to both return the next position with respect to the upper bound and the deserialize has to know how to handle a buffer which is potentially not full.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2632,44 +2728,76 @@ class Simplex_tree {
...
+
+  /** @private @brief Deserialize the array of char (flatten version of the tree) to initialize a Simplex tree.
+   * It is the user's responsibility to provide an 'empty' Simplex_tree, there is no guarantee otherwise.
+   * 
+   * @tparam F Method taking an @ref Filtration_value and a `const char*` as input and returning a
+   * `const char*`.
+   * @param[in] buffer A pointer on a buffer that contains a serialized Simplex_tree.
+   * @param[in] buffer_size The size of the buffer.
+   * @param[in] deserialize_filtration_value To provide if the type of @ref Filtration_value is not trivially
+   * convertible from the filtration value type of the serialized simplex tree. Takes the filtration value to fill and
```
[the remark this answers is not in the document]
hschreiber: With "trivially" I meant that either both filtration value types are the same, or, `deserialize_value_from_char_buffer` is able to handle the new filtration value type (of the simplex tree we are deserializing into) even though is does not corresponds to the old/serizalized one (because they are close enough, have same underlying structure or because the method was thought this way). 

`deserialize_filtration_value` is supposed to be a method which knows the serialized filtration value type and the current simplex tree filtration value type and knows how to read the old one from the buffer into the new type.

I guess, the description is not clear, so I will try the reformulate.

src/Simplex_tree/include/gudhi/Simplex_tree/serialization_utils.h
```diff
@@ -12,13 +12,14 @@
...
-#include <iostream>
 
 namespace Gudhi {
 
 namespace simplex_tree {
 
-/** \brief Serialize the given value and insert it at start position.
+/**
+ * @ingroup simplex_tree
+ * @brief Serialize the given value and insert it at start position.
```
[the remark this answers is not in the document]
hschreiber: I marked also `serialize_value_to_char_buffer` and `deserialize_value_from_char_buffer` as private, as all three methods are now only implementation details to correspond to the FiltrationValue interface.

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
```diff
@@ -0,0 +1,117 @@
...
+char* serialize_value_to_char_buffer(Trivial_filtration_value value, char* start) {
```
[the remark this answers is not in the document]
hschreiber: Before renaming the methods, all three methods were in serialize_utils. I just didn't thought of moving `get_serialization_size_of` as it was the only one not changing names. That's it. I moved everything together now.

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
hschreiber: Now that they are marked private, I guess that it is less important now. In the mean time, I added a description of the template parameter at least.

src/Simplex_tree/concept/FiltrationValue.h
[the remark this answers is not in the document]
mglisse: Ah, I think we are looking at 2 different options.
1. `get_serialization_size_of` defines the size the serialization will use. If `serialize_value_to_char_buffer` needs less, it should add some padding. `deserialize_value_from_char_buffer` needs to be able to advance the pointer by this same amount. At the end of the process, the whole allocated buffer is used, but there are small holes (characters that don't carry information) all over.
2. `get_serialization_size_of` only determines a sufficient amount of memory to allocate. `serialize_value_to_char_buffer` is allowed to use less, as long as `deserialize_value_from_char_buffer` can interpret that and advance by the right amount. At the end of the process, there may remain some unused space at the end of the buffer, but the part that carries information is compact (no hole).

IIUC you were thinking of the first and I was thinking of the second.
(we are not going to change that now anyway)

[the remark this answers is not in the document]
hschreiber: Yes, I talked about the first option, because it is the only one where it would work without changing the current assertion in the serialize method of the simplex tree (as it was part of you question). At the end of the method, it tests if the size given as argument corresponds to the difference between the start and end pointer, the end pointer pointing not to the end of the buffer, but to the "next position to write to if it wasn't the end".
But I guess, changing this assert wouldn't be too difficult.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2632,44 +2728,76 @@ class Simplex_tree {
...
+                              [[maybe_unused]] F&& deserialize_filtration_value)
+  {
     // In case buffer is just a 0 char
     if (members_size > 0) {
       if constexpr (!Options::stable_simplex_handles) sib->members_.reserve(members_size);
       Vertex_handle vertex;
-      Filtration_value filtration;
+      // Set explicitly to zero to avoid false positive error raising in debug mode when store_filtration == false
+      // and to force array like Filtration_value value to be empty.
+      Filtration_value filtration = Gudhi::simplex_tree::empty_filtration_value;
```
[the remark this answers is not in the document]
hschreiber: Windows doesn't like it... For some reason, he doesn't know if he should copy or move the object. I will probably have to add an explicit constructor for it, let's see...

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2667,44 +2763,77 @@ class Simplex_tree {
    * Serialize/Deserialize ignores any extra data (@ref Simplex_data) stored in the simplices for now.
    */
   void deserialize(const char* buffer, const std::size_t buffer_size) {
+    deserialize(buffer, buffer_size, [](Filtration_value& filtration, const char* ptr) {
+      return deserialize_value_from_char_buffer(filtration, ptr);
+    });
+  }
+
+  /** @private @brief Deserialize the array of char (flatten version of the tree) to initialize a Simplex tree.
```
mglisse, line 2771: I'd spell it "flat" or "flattened".

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: My point was that if the argument `deserialize_filtration_value` is supposed to read a filtration value in type FV1 (the filtration value that was used to serialize) and convert it to type FV2 (the current filtration value), it seems that it would be easier for the user to provide the information: "FV1 is this type" rather than having to provide a function that reads the char buffer.
But I see that the function is "private" now, so the conversation can wait.

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
```diff
@@ -0,0 +1,136 @@
...
+    if (f1 < f2){
+      f1 = f2;
+      return true;
+    }
+    return false;
+  }
+}
+
+/**
+ * @private
```
mglisse, line 84: It is private but referenced from the public concept?
Things look strange now, we have a concept that talks about serialization, but AFAICS all the serialization functions are private.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
mglisse: I think making the conversion operator (inside `empty_filtration_value_t`) `explicit` also lets windows compile. That seems better than having to add a constructor to every class we want to use as filtration value.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: I think, I see now what you meant: the idea is to only provide a way to convert FV1 into FV2, and internally call `deserialize_value_from_char_buffer` with FV1, convert in only then to FV2 with `deserialize_filtration_value`, instead of reading directly from the buffer with `deserialize_filtration_value`.
That could indeed be easier for the user. It would force the construction of FV1, but I guess, that in most cases, it won't be too big of an object.

By the way, why are those method private now? Was it necessary for the last release?

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
hschreiber: I don't know why all serizalization methods are private now. I only marked the ones in `filtration_value_utils.h` private.
And I just realized that I corrected the doc of the public concept to remove private references in the wrong branch...

Should I put `serialize_value_to_char_buffer`,  `deserialize_value_from_char_buffer` and `get_serialization_size_of` back to public public? Otherwise, I will redo the correction of the public concept, in the right branch this time.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: With the `explicit` keyword on the conversion operator, the conversion don't work at all anymore.

For example, the line `const Filtration_value null_ = Gudhi::simplex_tree::empty_filtration_value;` won't compile because "explicit conversion function was not considered".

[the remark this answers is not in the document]
mglisse: Yes, that's the point, that's why I suggested writing `Filtration_value filtration(empty_filtration_value);` instead of `Filtration_value filtration = Gudhi::simplex_tree::empty_filtration_value;` at the beginning of this thread.
Is there a particular reason to want implicit conversions?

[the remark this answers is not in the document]
hschreiber: It is for the instantiation of the static attribut `null_` in `Filtration_simplex_base_dummy`.  But I guess, I can just use `static_cast<Filtration_value>(Gudhi::simplex_tree::empty_filtration_value)` instead.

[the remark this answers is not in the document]
mglisse: `inline static const Filtration_value null_ { Gudhi::simplex_tree::empty_filtration_value} ;`
?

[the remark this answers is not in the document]
hschreiber: Oh, I knew that using `()` wouldn't work, but I didn't thought of `{}`!

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -5,19 +5,106 @@
...
+   */
+  FiltrationValue();
+  /**
+   * @brief Only necessary if `FiltrationValue(0)` does not correspond to an "empty" filtration value, or,
+   * if there is an ambiguity at compile time between several constructors if
+   * @ref Gudhi::simplex_tree::empty_filtration_value_t is passed.
+   * Should construct a representative of what is an "empty" object. E.g., 0 for a numerical value or {} for a vector.
+   * Two empty objects should evaluate to true if compared for equality.
+   */
+  FiltrationValue(Gudhi::simplex_tree::empty_filtration_value_t e);
```
mglisse, line 40: Should we add `explicit` (non-explicit also works, but it is only needed in explicit contexts)?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2567,6 +2646,22 @@ class Simplex_tree {
     }
   }
 
+  std::size_t num_simplices_and_filtration_size(Siblings const* sib, std::size_t& fv_byte_size) const {
```
mglisse, line 2649: Could the name reflect the fact that this is about serialization? "size" is not very specific.

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
mglisse: This seems to be the main thing missing (the other thread depends on what we decide here). So what are the options?
1. Everything related to serialization is "private" (for internal use by the Python plugin, not documented for our C++ users), we hide it in the concept, class, etc. Does the Simplex_tree still work if we provide a custom Filtration_value without serialization customization (and we don't call the serialization functions obviously)?
2. Only the serialization customization in the concept is public. (In 'Overloads for native arithmetic types or other simple types are already implemented with @ref Gudhi::deserialize_value_from_char_buffer "".', you could stop at 'already implemented.')
3. As above, plus the basic functions in the Simplex_tree ((de)serialize and get_serialization_size).
4. As above, plus the converting deserialize. The current version implies also documenting deserialize_value_from_char_buffer?
5. Everything?

If 1. works and we didn't document serialization in a previous release (I didn't check), that could be the easiest and quickest for now (adding a few `@private`). What is your opinion?

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: > A functor is more general indeed, but it might be sufficient to just construct a FV2 from the FV1, relying on existing constructors (or conversion operators).

This wouldn't work if you want to go from a multi to 1d and vice versa, e.g., retrieve only parameter p from a multi simplex tree which was serialized.
 
> It could be passed as `st.deserialize_from<FV1>(...);` (possibly with `template` after the dot) or `st.deserialize_convert(...,(FV1*)nullptr)`.

Yes, that is what I meant, we somehow need to specify which was the original type, probably via a template.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Ok for `explicit`, it makes sense. In the end, the class author can do whatever they prefer anyway.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: What do you think of `num_simplices_and_filtration_serialization_size`?

mglisse: That's fine (it is private, it doesn't need to be pretty).

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
mglisse: Our code interfacing with Python is even more private (it isn't even installed by `make install`), I don't think it can justify making something else public. Also, wouldn't it fail to compile?

In my opinion, it is a bit all-or-nothing: if we document even one small piece, we are setting in stone the protocol we use for serialization, and then there isn't much point keeping pieces hidden. I was suggesting (1) as quickest, but if you don't like it, I think I'd rather go to (3) so users can at least test the custom function they provide to the concept (or (4) or more if we resolve the other thread).

hschreiber: Honestly, I don't have a strong opinion about it, as it won't be a common issue. It is just weird to hide methods in a concept. We could perhaps keep everything private for now (with existing doc, just with `@private`), but if the serialization becomes a topic again, make the documentation public then.

src/Simplex_tree/include/gudhi/Simplex_tree.h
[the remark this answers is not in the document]
hschreiber: In the meantime, I implemented the version you proposed (i.e., a template for the original type + functor converting filtration value instead of functor constructing from buffer). The only overhead there is, is the extra construction of the original filtration type.
So, in the end, do you want me to rollback to the current version?

mglisse (approved):

PR #1125, opened 2024-08-26
[Persistence Interval] fix of structural binding bug in windows compiler

hschreiber:
Just moved the `get` from the `Gudhi::persistence_matrix` namespace to a member method, as Windows confuses them with the ones in the `std` namespace otherwise. The `git diff` got very confused by the changes, the best is to probably just look directly at the file.

### DIFF
--- a/src/Persistence_matrix/include/gudhi/persistence_interval.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_interval.h
@@ -229,7 +229,7 @@ constexpr const auto& get(const Gudhi::persistence_matrix::Persistence_interval<
  */
 template <size_t I, typename Dimension, typename Event_value>
 constexpr auto&& get(Gudhi::persistence_matrix::Persistence_interval<Dimension, Event_value>&& i) noexcept {
-  return std::move(i.template get<I>());
+  return std::move(i).template get<I>();
 }
 
 /**
--- a/src/Persistence_matrix/include/gudhi/persistence_interval.h
+++ b/src/Persistence_matrix/include/gudhi/persistence_interval.h
@@ -88,87 +88,67 @@ struct Persistence_interval {
     }
     return stream;
   }
+
+  /**
+   * @brief Specialization of `get` for @ref Gudhi::persistence_matrix::Persistence_interval.
+   *
+   * @tparam I Index of the value to return: 0 for the birth value, 1 for the death value and 2 for the dimension.
+   * @return Either the birth value if @p I == 0, the death value if @p I == 1 or the dimension if @p I == 2.
+   */
+  template <std::size_t I>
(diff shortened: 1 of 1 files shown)


mglisse:
> The `git diff` got very confused by the changes

[Cette version](https://github.com/GUDHI/gudhi-devel/pull/1125/files?w=1) du diff est assez lisible.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -88,140 +88,164 @@ struct Persistence_interval {
...
  * @tparam Event_value Second template parameter of @ref Gudhi::persistence_matrix::Persistence_interval.
+ * @param i Interval from which the value should be returned.
+ * @return Either the birth value if @p I == 0, the death value if @p I == 1 or the dimension if @p I == 2.
  */
-template <typename Dimension, typename Event_value>
-struct tuple_size<Gudhi::persistence_matrix::Persistence_interval<Dimension, Event_value> >
-    : integral_constant<size_t, 3> {};
+template <size_t I, typename Dimension, typename Event_value>
+constexpr auto&& get(Gudhi::persistence_matrix::Persistence_interval<Dimension, Event_value>&& i) noexcept {
+  return std::move(i.template get<I>());
```
mglisse, line 232: A direct translation would have been
```suggestion
  return std::move(i).template get<I>();
```
but the result should be equivalent, so ok.

mglisse (approved):
An alternative would be: do *not* specialize tuple_size, get rid of the unnecessary `get` members, reorder the members of the class, and rely on "case 3" of structured binding. But this version is good.

src/Persistence_matrix/include/gudhi/persistence_interval.h
```diff
@@ -88,140 +88,164 @@ struct Persistence_interval {
...
-
-template <size_t I, typename Dimension, typename Event_value>
-constexpr auto&& get(Gudhi::persistence_matrix::Persistence_interval<Dimension, Event_value>&& i) noexcept {
-  return Gudhi::persistence_matrix::get<I>(move(i));
-}
-
 template <size_t I, typename Dimension, typename Event_value>
 constexpr const auto&& get(const Gudhi::persistence_matrix::Persistence_interval<Dimension, Event_value>&& i) noexcept {
-  return Gudhi::persistence_matrix::get<I>(move(i));
+  return std::move(i.template get<I>());
```
mglisse, line 248: I was fine with not changing anything, but now you changed the other one (`&&`), it is inconsistent with here (`const&&`).

mglisse (approved):

VincentRouvreau (approved):

PR #1117, opened 2024-08-12
PersistenceLengths and its unitary tests

VincentRouvreau:

Issue #771, opened 2022-12-16
Filtration values of the alpha complex (squared radius)


CharlesArnal:
I assume there are good reasons why the alpha complex uses the squared radius (rather than the radius) of the smallest circumspheres as filtration values, but this is very counter-intuitive, and does not agree with the definition most often found in the literature.
I know of several researchers who were unpleasantly surprised upon discovering that. Do the benefits of this convention really outweigh its disadvantages ?

In any case, the convention is explained in the reference manual, but not in the user manual https://gudhi.inria.fr/python/latest/alpha_complex_user.html# - it might be worth it to add a line or two about it.
Moreover, it says that the alpha complex is equivalent to the Cech complex - this will only be true if the Cech complex also uses the squared radius convention. Is it the case ?

Best,

Charles

mglisse:
See also #80. If we use squared radii, people are surprised. If we do something different for unweighted and for weight 0, people are surprised. If we apply a square root to negative numbers for the weighted case, people are surprised. Something has to give... Squared radius is also what CGAL produces. But yes, improving the doc is always a good idea.
We can compute an alpha-complex using rational numbers, which would not support sqrt. However, when we convert to double, there is less reason not to do the sqrt.
The Cech uses sqrt, that's inconsistent indeed.
The equivalence could be in a loose sense, if you could deduce one from the other with a factor of 2, a sqrt, or some other similar transformation, I would still call them equivalent.

CharlesArnal:
But I think that if someone reads that two filtrations are equivalent, they will expect e.g. the persistence diagrams to be same.

I understand your point about other use cases and parameters configurations, but I would have expected the standard case to be the one that is used most often - I don't have any solid ground for thinking that, though.

VincentRouvreau:
Fixed by #1172

PR #1126, opened 2024-08-28
Cech persistence sklearn

VincentRouvreau:


mglisse:
> CI will fail as it is uses and requires #1117

It actually passes. I guess because it isn't in CMakeLists.txt yet?

DavidLapous:
Could we also add typing to this pipeline (cf #920) ?
Also, I'm not a big fan of the "max_alpha_square"; I agree that we should be coherent with alpha complexes, 
but I'm already not a fan of the alpha square there (cf #771 @CharlesArnal). 
Furthermore, Rips complexes are also expected to be coherent with Cech complexes.
I'm more in favor to "threshold", but this can be discussed.

mglisse:
> Could we also add typing to this pipeline (cf #920) ?

As I said in the other PR, the interaction with the doc is suboptimal currently, Vincent is looking into options to make it nicer so we can start adding typing all over the place.

> Also, I'm not a big fan of the "max_alpha_square";

An option would be to remove it, it is not very useful for alpha-complex / delaunay-cech (what this class currently uses), but it seems likely that we will add a pure Cech option at some point (useful in high dimension with a very low threshold) where the threshold will be essential.

> I agree that we should be coherent with alpha complexes, but I'm already not a fan of the alpha square there (cf #771 @CharlesArnal).

Ah, right, this was supposed to be the occasion to tackle this sqrt business, so let's do that.

> Furthermore, Rips complexes are also expected to be coherent with Cech complexes.

Note that even with the square root, we will likely stick to a convention where the edges have a filtration twice as large in the Rips than in the Cech (diameter vs radius).

---

So, what would be a good way to handle the square/sqrt issue?
The squared output is necessary for some weighted cases, so it cannot disappear. But users often want non-squared filtration values.
We could have an option (`output_squared_filtrations`, `do_sqrt`, other suggestions?), and depending on this option, we would apply `np.sqrt` to the diagram before outputting it? We could also add the option to AlphaComplex (computing the sqrt in the SimplexTree is probably more costly than on a numpy array, so from CechPersistence we would likely build the AlphaComplex with squared values always and compute the square root on the diagram).
Another question is what we want by default. It looks like there are a lot of vote in favor of sqrt-by-default. If so, we will need to be super explicit (and redundant) in the doc about what the default is, and the fact that it is not the same as in AlphaComplex.

DavidLapous:
> Note that even with the square root, we will likely stick to a convention where the edges have a filtration twice as large in the Rips than in the Cech (diameter vs radius).

IIRC, in the literature, this is often presented as 
```math
\forall t \ge 0, \quad \mathrm{\v Cech}(X)_t \simeq \mathrm{Alpha}(X)_t\quad \textnormal{ and }\quad
\mathrm{\v Cech}(X)_t  \subseteq \mathrm{VR}(X)_{2t}\subseteq \mathrm{\v Cech}(X)_{2t},
```
so this factor 2 is to be expected (but could be indeed be discussed).

>So, what would be a good way to handle the square/sqrt issue?

I think a good solution to this squared issue would be to have two separate classes in the interface,
e.g., `CechComplex` and `WeightedAlphaComplex`. This would also make the `threshold` clear.

mglisse:
> I think a good solution to this squared issue would be to have two separate classes in the interface, e.g., `CechComplex` and `WeightedAlphaComplex`. This would also make the `threshold` clear.

Do I understand correctly that you would split squared / not squared as follows, without options?

* classes that build an unweighted complex: CechComplex, AlphaComplex, DelaunayCechComplex -> sqrt
* separate weighted classes: WeightedCechComplex, WeightedAlphaComplex, WeightedDelaunayCechComplex -> squared
* classes that give the persistence diagram directly: CechPersistence -> sqrt, WeightedCechPersistence -> squared

This means that 
* CechComplex gives a different output than WeightedCechComplex with all 0 weights, it is still confusing
* we cannot get AlphaComplex with squared values (the current behavior), or WeightedThing with sqrt values
* it forces to have twice as many classes (that's much more code, but it may make the documentation a bit simpler, I don't know)

It is an acceptable possibility, but I wouldn't go so far as to call it "good".

VincentRouvreau:
> As I said in the other PR, the interaction with the doc is suboptimal currently, Vincent is looking into options to make it nicer so we can start adding typing all over the place.

cf. #1137

DavidLapous:
>  - CechComplex gives a different output than WeightedCechComplex with all 0 weights, it is still confusing

A possibility is to put directly the `square` in the name of the concerned classes. I've seen stuff like `sqnorm` in machine learning libraries for instance, which seems reasonable. I don't know if there is a perfect solution to this issue indeed. 

>  - we cannot get AlphaComplex with squared values (the current behavior), or WeightedThing with sqrt values

Yes, but I feel that users that want to compute weighted alpha complexes, or squared alpha complexes specifically are not neophyte to TDA (I've never seen someone use weights, or want the square), so we can expect them to read the documentation for this specific use case. --- I may be mistaken on these stats.

> or WeightedThing with sqrt values

Q. Is it possible to define WeightedAlphaComplexes with squared values, and preserving the weights on the nodes? 
If that's a natural definition, we can just remove the square in all cases, and put a flag to enable this. 

>  - it forces to have twice as many classes (that's much more code, but it may make the documentation a bit simpler, I don't know)

If the number of classes is an issue, we can just alias classes to other with proper default values. Doing multiple classes also reduces the cyclomatic complexity, so depending on what we are doing, it's not clear what is the best.

VincentRouvreau:
Replaced by #1202

PR #1131, opened 2024-09-17
Symbolic links for cython and pybind11 compilation

### LINKED ISSUE #206: [cython] cython generates .cpp files in the source tree instead of the build tree

VincentRouvreau:
Fix #206 
To be done:
* [X] verify wheel sizes
* [X] test wheels (linux and windows ok)
* [X] do not symlink if source and target is the same file (can happen if build in sources)


mglisse (approved):
Ah, you chose to do 1 symlink per file and not one to the directory. Ok.
Did you need to enable developer mode on the ci.inria windows vm?

VincentRouvreau:
> Ah, you chose to do 1 symlink per file and not one to the directory. Ok. Did you need to enable developer mode on the ci.inria windows vm?

If I symlink the directory, cpp files will still be generated in sources. This is the main reason of this choice.
On the ci.inria windows vm you need to be admin to create symlinks.

mglisse:
> On the ci.inria windows vm you need to be admin to create symlinks.

Not quite, you can enable developer mode and then anyone can create it.
Should we use cmake's option to copy the file when doing a symlink failed?

(this PR seems to cause problem for the circleCI test)

VincentRouvreau:
> > On the ci.inria windows vm you need to be admin to create symlinks.
> 
> Not quite, you can enable developer mode and then anyone can create it. Should we use cmake's option to copy the file when doing a symlink failed?

Why not. Let's see if my proposal is sufficient enough.

> (this PR seems to cause problem for the circleCI test)

Hère, on circle ci, the issue is that we build in sources `cmake .` and the actual code is trying to symlink a file to itself, which leads to an error. I will fix it.

mglisse:
> Why not. Let's see if my proposal is sufficient enough.

Sufficient for what? I think we already have an example of 1 machine/account where building gudhi used to work but not creating symlinks.

PR #1132, opened 2024-09-20
copy_on_error if symbolic link failed to be created

VincentRouvreau:
Fix last issue of #1131 on windows where you need to have admin rights, or in developper mode to be able to make symbolic links.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -63,7 +63,9 @@ function( add_gudhi_symbolic_links GLOBBING_EXPRESSION )
   endif()
   file(GLOB GUDHI_GLOB_FILES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${GLOBBING_EXPRESSION})
   foreach(GUDHI_GLOB_FILENAME ${GUDHI_GLOB_FILES})
-    file(CREATE_LINK "${CMAKE_CURRENT_SOURCE_DIR}/${GUDHI_GLOB_FILENAME}" "${CMAKE_CURRENT_BINARY_DIR}/${GUDHI_GLOB_FILENAME}" SYMBOLIC)
+    # COPY_ON_ERROR is activated because on windows, user needs to be root or in dev mode to be able to make symbolic links
+    # This case can be problematic, because if you modify the sources, build is still done on the copy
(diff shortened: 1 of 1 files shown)


src/python/CMakeLists.txt
```diff
@@ -63,7 +63,9 @@ function( add_gudhi_symbolic_links GLOBBING_EXPRESSION )
   endif()
   file(GLOB GUDHI_GLOB_FILES RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} ${GLOBBING_EXPRESSION})
   foreach(GUDHI_GLOB_FILENAME ${GUDHI_GLOB_FILES})
-    file(CREATE_LINK "${CMAKE_CURRENT_SOURCE_DIR}/${GUDHI_GLOB_FILENAME}" "${CMAKE_CURRENT_BINARY_DIR}/${GUDHI_GLOB_FILENAME}" SYMBOLIC)
+    # COPY_ON_ERROR is activated because on windows, user needs to be root or in dev mode to be able to make symbolic links
```
mglisse, line 66:
```suggestion
    # COPY_ON_ERROR is activated because on windows 10, user needs to be root or in dev mode to be able to make symbolic links
```
ISTR that windows 11 allows anyone to create a symlink, but I could misremember (I know something changed in the permissions).

src/python/CMakeLists.txt
```diff
@@ -63,7 +63,9 @@ function( add_gudhi_symbolic_links GLOBBING_EXPRESSION )
...
+    # This case can be problematic, because if you modify the sources, build is still done on the copy
```
mglisse, line 67: Ah, I was going to ask what happens in that case. I guess the link/copy is done during `cmake` generation, not during the actual build, so rerunning just `make` has no chance to redo the copy :disappointed: 
Bah, few developers use windows, even fewer will have non-working symlinks, even fewer will want partial rebuilds...

mglisse (approved):

src/python/CMakeLists.txt
VincentRouvreau: Yes that's it, you need to cmake each time you do a modification. I was asking myself if we should explain it somewhere .

PR #1135, opened 2024-09-24
Cmake improvement with importlib metadata

### LINKED ISSUE #580: [Python] Enhance third parties version detection
Python third parties version detection could be enhanced in [`find_python_module`](https://github.com/GUDHI/gudhi-devel/blob/7f1b8eb706c72921141b53e607d6e2aa28e2bf19/src/cmake/modules/GUDHI_third_party_libraries.cmake#L103) using:
```python
from importlib.metadata import version; print(version('hnswlib'))
# 0.6.0
```

* This would remove `find_python_module_no_version`
* Requires to do `find_python_module("scikit-learn")`
* Requires to do `find_python_module("POT")`

But [importlib metadata](https://docs.python.org/3/library/importlib.metadata.html) requires python 3.8, and for the moment gudhi requires 3.5.

VincentRouvreau:
Fix #580

```bash
git checkout master
time cmake .
# real	0m10.676s
# user	0m13.474s
# sys	0m13.880s

git checkout cmake_improvement_with_importlib_metadata
time cmake .
# real	0m2.100s
# user	0m1.833s
# sys	0m2.096s
```


mglisse:
As expected, we need a special case for pykeops, which can be installed on windows even if it cannot be imported. I don't know if we should always disable it on windows, or if we should try the `import` just for this package.

VincentRouvreau:
> As expected, we need a special case for pykeops, which can be installed on windows even if it cannot be imported. I don't know if we should always disable it on windows, or if we should try the `import` just for this package.

I did so on [848d853](https://github.com/GUDHI/gudhi-devel/pull/1135/commits/848d853159083041511b070b85389ebb8cb217a6)

src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -167,46 +168,37 @@ if (WITH_GUDHI_PYTHON)
...
   endif()
 
+  # Specific case for PyKeops that can be imported on Windows, but fails because it uses fcntl (not available on Windows)
+  # Also fcntl has no metadata, so find_python_module does not work
+  # "import fcntl" is about 1 sec. faster than "import pykeops"
+  execute_process(
+          COMMAND ${Python_EXECUTABLE}  -c "import fcntl"
+          RESULT_VARIABLE FCNTL_IMPORT_MODULE_RESULT
+          OUTPUT_VARIABLE FCNTL_IMPORT_MODULE_OUPUT)
+  if(FCNTL_IMPORT_MODULE_RESULT EQUAL 0)
```
mglisse, line 198: Is that better than `if (NOT WIN32)`?

VincentRouvreau: No, you are right. I wanted to do something smarter, like "the day fcntl will be available on windows, pykeops will work on windows", but that was not that smart...
`if (NOT WIN32)` on [b4694d9](https://github.com/GUDHI/gudhi-devel/pull/1135/commits/b4694d930d810c26c5aa5234dc059ce2f771595f)

src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -167,46 +168,31 @@ if (WITH_GUDHI_PYTHON)
...
     find_python_module("eagerpy")
-    find_python_module_no_version("hnswlib")
+    find_python_module("hnswlib")
     find_python_module("tensorflow")
     find_python_module("sphinx_paramlinks")
     find_python_module("pydata_sphinx_theme")
-    find_python_module_no_version("sphinxcontrib.bibtex")
+    find_python_module("sphinxcontrib.bibtex")
     find_python_module("networkx")
+    # Specific case for PyKeops that can be imported on Windows, but fails because it uses fcntl (not available on Windows)
```
mglisse, line 189: I think it can be **installed** (`pip install pykeops`) but not **imported** (`import pykeops`). That's a detail though.

.github/next_release.md
```diff
@@ -20,6 +20,7 @@ Below is a list of changes:
 
 - Installation
      - CMake &ge; 3.15 is now required (was &ge; 3.8).
+     - Python &ge; 3.8 is now required (was &ge; 3.5), because of `importlib.metadata`.
```
mglisse, line 23: Maybe the use of type hints also requires 3.8? I don't remember.

mglisse (approved):

.github/next_release.md
VincentRouvreau: In Python 3.5 for [type hints](https://docs.python.org/3/library/typing.html)

mglisse: Ok. Note that they added new type hinting possibilities with almost every release, so it is hard to be sure we aren't using modern things that were absent in older releases we are not testing anymore. Anyway, let's ignore this.

PR #1137, opened 2024-09-26
Use parameters type hints automatically with sphinx 

VincentRouvreau:
Only for RipsPersistence for the moment.

image


src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -52,25 +52,25 @@ def __init__(
...
-                Default is 'point cloud'.
-            num_collapses (int|str): Specify the number of iterations of :func:`~gudhi.flag_filtration.edge_collapse.reduce_graph`
-                (edge collapse) to perform on the graph. Default value is 'auto'.
-            homology_coeff_field (int): The homology coefficient field. Must be a prime number. Default value is 11.
-            n_jobs (Optional[int]): Number of jobs to run in parallel. `None` (default value) means `n_jobs = 1` unless in a
-                joblib.parallel_backend context. `-1` means using all processors. cf.
-                https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html for more details.
+        """Constructor for the RipsPersistence class.
+
+        :param homology_dimensions: The returned persistence diagrams dimension(s).
```
mglisse, line 57: I see you switched from Google-style to plain RST. Is that because it wasn't possible while sticking to Google-style?

VincentRouvreau: I was just following what is explained in [sphinx-autodoc-typehints documentation](https://github.com/tox-dev/sphinx-autodoc-typehints). Google style is also working, so I reverted to it on [52aa602](https://github.com/GUDHI/gudhi-devel/pull/1137/commits/52aa6022583a68c781ce925a904d7870c5206c62)

src/python/doc/conf.py
```diff
@@ -8,18 +8,21 @@
...
     "sphinx.ext.mathjax",
     "sphinx.ext.ifconfig",
     "sphinx.ext.viewcode",
-    "sphinx.ext.napoleon",
     "sphinxcontrib.bibtex",
     "sphinx_paramlinks",
 ]
 
 autodoc_class_signature = "separated"
+autodoc_typehints = "none"
```
mglisse, line 25: Looking at https://github.com/tox-dev/sphinx-autodoc-typehints#options, it suggests we should add `napoleon_use_rtype = True`. It has 0 effect currently, but it might become relevant when we add more annotations?

`always_use_bars_union = True` looks a bit prettier than the default (False), but that's subjective.

VincentRouvreau:
What I found nice (and thanks @mglisse ) for pointing the question, I have:
Screenshot from 2024-10-02 10-20-58
But if I precise a type, like for homology_dimensions in this example:
```
        Parameters:
            homology_dimensions (int or list of int): The returned persistence diagrams dimension(s).
```
The type is substituted, with the one that is explicitly given in code, we will have:
image

mglisse (approved):

PR #1140, opened 2024-10-04
Fix '#define far' is performed in windows.h

VincentRouvreau:
Oscillating rips was failing because of this, but other unitary tests was compiling and passing

### DIFF
--- a/src/Subsampling/include/gudhi/choose_n_farthest_points.h
+++ b/src/Subsampling/include/gudhi/choose_n_farthest_points.h
@@ -168,7 +168,7 @@ using radius_priority_ds =
                           boost::heap::mutable_<true>, boost::heap::constant_time_size<false>>;
 template<class FT>
 struct Landmark_info {
-  std::size_t far_idx; FT radius;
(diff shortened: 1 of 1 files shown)


mglisse (commented):
Indeed, we have to rename it. Since we don't have any other name talking about "index", for consistency, I think I'd rather name it "farthest". Is that ok for you?

VincentRouvreau:
> Indeed, we have to rename it. Since we don't have any other name talking about "index", for consistency, I think I'd rather name it "farthest". Is that ok for you?

That's fine to me and I did so on [7fb35b8](https://github.com/GUDHI/gudhi-devel/pull/1140/commits/7fb35b8a511e27407188399a075262d70ffb0fe8). I thought it was a kind of index on the farthest. But `farthest` is fine to me.

mglisse (approved):

mglisse:
> I thought it was a kind of index on the farthest

It is, but everything is indexes in this function, so it looks strange to have far_index but not parent_index, neighbor_index, etc.

PR #1154, opened 2024-11-08
[Persistence Matrix] Addition of small vectors as a column type

hschreiber:
Replacing `std::vector` with `boost::container::small_vector` in `Naive_vector_column` showed some run time improvements on some particular use cases in [multipers](https://github.com/DavidLapous/multipers). So I added the possibility in the column choices.

### DIFF
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
@@ -224,9 +224,9 @@ class Naive_vector_column : public Master_matrix::Row_access_option,
 };
 
 template <class Master_matrix>
-using STD_naive_vector_column = Naive_vector_column<Master_matrix, std::vector<typename Master_matrix::Matrix_entry*> >;
+using Naive_std_vector_column = Naive_vector_column<Master_matrix, std::vector<typename Master_matrix::Matrix_entry*> >;
 template <class Master_matrix>
-using Small_naive_vector_column =
+using Naive_small_vector_column =
     Naive_vector_column<Master_matrix, boost::container::small_vector<typename Master_matrix::Matrix_entry*, 8> >;
 
 template <class Master_matrix, class Support>
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
@@ -42,7 +42,6 @@ namespace persistence_matrix {
  * are stored uniquely in the underlying container.
  *
  * @tparam Master_matrix An instantiation of @ref Matrix from which all types and options are deduced.
- * @tparam Entry_constructor Factory of @ref Entry classes.
  */
 template <class Master_matrix>
 class Naive_vector_column : public Master_matrix::Row_access_option,
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
@@ -25,6 +25,7 @@
 #include <utility>    //std::swap, std::move & std::exchange
 
 #include <boost/iterator/indirect_iterator.hpp>
+#include <boost/container/small_vector.hpp>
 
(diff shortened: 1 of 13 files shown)


src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
```diff
@@ -224,7 +224,13 @@ class Naive_vector_column : public Master_matrix::Row_access_option,
 };
 
 template <class Master_matrix>
-inline Naive_vector_column<Master_matrix>::Naive_vector_column(Column_settings* colSettings)
+using STD_naive_vector_column = Naive_vector_column<Master_matrix, std::vector<typename Master_matrix::Matrix_entry*> >;
+template <class Master_matrix>
+using Small_naive_vector_column =
+    Naive_vector_column<Master_matrix, boost::container::small_vector<typename Master_matrix::Matrix_entry*, 8> >;
```
mglisse, line 230: It seems strange to hardcode 8.

src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
```diff
@@ -224,7 +224,13 @@ class Naive_vector_column : public Master_matrix::Row_access_option,
 };
 
 template <class Master_matrix>
-inline Naive_vector_column<Master_matrix>::Naive_vector_column(Column_settings* colSettings)
+using STD_naive_vector_column = Naive_vector_column<Master_matrix, std::vector<typename Master_matrix::Matrix_entry*> >;
```
mglisse, line 227: What do you think of Naive_std_vector_column and Naive_small_vector_column, to keep the qualifier of vector next to the word vector?

mglisse (commented):
Would the same templatization make sense for Vector_column? Maybe it does but you consider that class useless?

hschreiber:
> Would the same templatization make sense for Vector_column? Maybe it does but you consider that class useless?

It makes indeed also sense for Vector_column. It's just that Vector_column only works better when a lot of entries are manually cleared inside the column, which for now only happens during vineyard with RU. And in that case, the small vector do not seem to make things much better. At least in multipersistence. And I am wondering a bit, if the class is not indeed useless as you say, as using a set kind of works better in that particular case anyway...

src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
hschreiber: I agree. It would be better to have something general here, but I didn't wanted to add an option for that... And 8 was what worked the best in my benchmarks to compute a barcode. Do you see an easy way to give the possibility to customize this value for the user without having to add an option just for this particular case?

src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
hschreiber: Sounds much better! Changed it.

src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
mglisse: Not "easy": in my opinion, the original issue is that using an `enum` makes the design not extensible (a user cannot provide their own column type). If you changed the design enough for it to be extensible, then there would be no limitation on parameterizing the column types you provide.
(IIRC, Boost.Graph has a similar limitation with a hard-coded list of possible containers for some data structures)
Without going all the way to an extensible design, using tag classes instead of an enum (i.e. empty classes that are only used with `is_same<T,vector_column_tag>` or similar) might not require huge changes and could accommodate `small_vector_tag<8>`.
Anyway, I'll approve this for now, but please make a note somewhere about this (an issue, or a comment in the code, I don't know if you already have a list of potential improvements somewhere).

mglisse (approved):

src/Persistence_matrix/include/gudhi/Persistence_matrix/columns/naive_vector_column.h
VincentRouvreau: We will keep track of this on #1167

PR #1155, opened 2024-11-15
Move OSx CI on appveyor

VincentRouvreau:
Due to to some issue on azure with OSx


VincentRouvreau:
AppVeyor plans offers only 60 minutes quota per build job.

PR #1163, opened 2024-12-19
Use circleCI anchors to DRY. Activate assertions on all tests

VincentRouvreau:


mglisse:
> x86_64-linux-gnu-g++ **-UNDEBUG** -fPIC -DNPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION -I/opt/venv/lib/python3.12/site-packages/numpy/_core/include -I/root/project/build/version/python/gudhi/ -I/usr/include/eigen3 -I/usr/include -I/usr/local/include -I/root/project/build/version/ext/hera/include -I/usr/include -I/usr/include/x86_64-linux-gnu -I/usr/include -I/root/project/build/version/include -I/root/project/build/version/python/include -I/usr/include -I/opt/venv/include -I/usr/include/python3.12 -c gudhi/euclidean_strong_witness_complex.cpp -o build/temp.linux-x86_64-cpython-312/gudhi/euclidean_strong_witness_complex.o -DBOOST_RESULT_OF_USE_DECLTYPE -DBOOST_ALL_NO_LIB -DBOOST_SYSTEM_NO_DEPRECATED -std=c++17 -frounding-math -DCGAL_EIGEN3_ENABLED -DCGAL_HEADER_ONLY -DCGAL_USE_GMP -DCGAL_USE_GMPXX -DCGAL_USE_MPFR -DGUDHI_USE_TBB

Notice that in the python/setuptools part, there is no `-DNDEBUG`, and no `-O` either despite this being a Release build (I think). Depending on the release of setuptools, sometimes it adds C(XX)FLAGS to its default, sometimes it completely drops the default flags if C(XX)FLAGS is set. Having `-UNDEBUG` does not hurt in the second case, and is useful in the first case, so that seems good. The lack of -O may be more problematic, if we don't have -O, we might as well do a Debug build, not a Release one, and then it may be unnecessary to play with C(XX)FLAGS. If we do want to optimize (so the tests run faster?), maybe we should add `-O2` or `-O3` to C(XX)FLAGS?

For the C++ tests, make is not run in verbose mode, so I don't see the exact command that is run, that makes it harder to check if this is doing what we want.

PR #1164, opened 2025-01-08
Deprecation of BottleneckDistance(epsilon=...), replaced with BottleneckDistance(e=...)

### LINKED ISSUE #343: bottleneck_distance name of epsilon argument
bottleneck_distance takes a parameter called e, while BottleneckDistance names it epsilon.
```python
    pairwise_persistence_diagram_distances(l, e=0.00001)
    BottleneckDistance(epsilon=0.00001).fit_transform(l)
```
so we end up with confusing name changes depending on the exact syntax.

VincentRouvreau:
In order to be consistent with `bottleneck_distance` and `pairwise_persistence_diagram_distances`.
Fix #343


hschreiber (approved):
Except for the file conflicts (probably just a simple sync needed), looks fine to me.

PR #1172, opened 2025-02-10
output squared filtration values true/false mechanism for Delaunay complexes

### LINKED ISSUE #80: [Alpha_complex] new do_sqrt argument
The filtration values we use in the alpha complex are squared radii. Most applications are more interested in a (non-squared) radius. We use squares because if we add weights, taking the square root becomes less meaningful, in particular if the value becomes negative. But I think we should make it as easy as possible for people to get the square root they expect, and a parameter seems to be the easiest, passing `do_sqrt=True` if we want the square root. The name `do_sqrt` is just a placeholder for the report, better names are welcome.

VincentRouvreau:
Replace #1138 (remove function instead of classes)
While keeping backward compatibilities of former Delaunay complexes (cpp, utils and python)

Fix #80 and #771

### DIFF
--- a/src/Alpha_complex/include/gudhi/Alpha_complex.h
+++ b/src/Alpha_complex/include/gudhi/Alpha_complex.h
@@ -355,8 +355,8 @@ class Alpha_complex {
    * It also computes the filtration values accordingly to the \ref createcomplexalgorithm if default_filtration_value
    * is not set.
    *
-   * \tparam output_squared_values If `true` (default value), it assigns to each simplex a filtration value equal to
-   * \f$ \alpha^2 \f$, or to \f$ \alpha \f$ when `output_squared_values` is `false`. cf. \ref createcomplexalgorithm
+   * \tparam Output_squared_values If `true` (default value), it assigns to each simplex a filtration value equal to
+   * \f$ \alpha^2 \f$, or to \f$ \alpha \f$ when `Output_squared_values` is `false`. cf. \ref createcomplexalgorithm
    * \tparam SimplicialComplexForAlpha must meet `SimplicialComplexForAlpha` concept.
    *
    * @param[in] complex SimplicialComplexForAlpha to be created.
@@ -376,10 +376,10 @@ class Alpha_complex {
    *
    * Initialization can be launched once.
    *
-   * \note Weighted Alpha complex can have negative filtration values. If `output_squared_values` is set to `false`,
-   * filtration values will be `NaN` in this case.
+   * \exception std::invalid_argument In case of a Weighted Alpha complex with `Output_squared_values` set to `false`,
+   * as Weighted Alpha complex can have negative filtration values.
    */
-  template <bool output_squared_values = true,
+  template <bool Output_squared_values = true,
             typename SimplicialComplexForAlpha,
             typename Filtration_value = typename SimplicialComplexForAlpha::Filtration_value>
   bool create_complex(SimplicialComplexForAlpha& complex,
@@ -388,6 +388,9 @@ class Alpha_complex {
                       bool default_filtration_value = false) {
     // Filtration_value must be capable to represent the special value "Not-A-Number"
     static_assert(std::numeric_limits<Filtration_value>::has_quiet_NaN);
+    // Forbids weighted and output square root filtration values - cannot static_assert for python purpose
+    if constexpr(Weighted && !Output_squared_values)
+      throw std::invalid_argument("Weighted Alpha complex cannot output square root filtration values");
     // To support more general types for Filtration_value
     using std::isnan;
 
@@ -467,7 +470,7 @@ class Alpha_complex {
                 if(exact) CGAL::exact(sqrad);
 #endif
                 alpha_complex_filtration = cgal_converter(sqrad);
-                if constexpr (!output_squared_values) {
+                if constexpr (!Output_squared_values) {
                   alpha_complex_filtration = sqrt(alpha_complex_filtration);
                 }
               }
@@ -492,7 +495,7 @@ class Alpha_complex {
         // Only in not exact version, cf. https://github.com/GUDHI/gudhi-devel/issues/57
         complex.make_filtration_non_decreasing();
       // Remove all simplices that have a filtration value greater than max_alpha_square
-      if constexpr (!output_squared_values) {
+      if constexpr (!Output_squared_values) {
         complex.prune_above_filtration(sqrt(max_alpha_square));
       } else {
         complex.prune_above_filtration(max_alpha_square);
--- a/src/Alpha_complex/include/gudhi/Alpha_complex.h
+++ b/src/Alpha_complex/include/gudhi/Alpha_complex.h
@@ -6,6 +6,7 @@
  *
  *    Modification(s):
  *      - 2019/08 Vincent Rouvreau: Fix issue #10 for CGAL and Eigen3
+ *      - 2024/10 Vincent Rouvreau: Add square root filtration values interface
  *      - YYYY/MM Author: Description of the modification
  */
 
@@ -65,14 +66,14 @@ template<typename D> struct Is_Epeck_D<CGAL::Epeck_d<D>> { static const bool val
 /**
  * \class Alpha_complex Alpha_complex.h gudhi/Alpha_complex.h
  * \brief Alpha complex data structure.
- * 
+ *
  * \ingroup alpha_complex
- * 
+ *
  * \details
- * The data structure is constructing a CGAL Delaunay triangulation (for more information on CGAL Delaunay 
+ * The data structure is constructing a CGAL Delaunay triangulation (for more information on CGAL Delaunay
  * triangulation, please refer to the corresponding chapter in page http://doc.cgal.org/latest/Triangulation/) from a
  * range of points or from an OFF file (cf. Points_off_reader).
- * 
+ *
  * Please refer to \ref alpha_complex for examples.
  *
  * The complex is a template class requiring an <a target="_blank"
@@ -84,7 +85,7 @@ template<typename D> struct Is_Epeck_D<CGAL::Epeck_d<D>> { static const bool val
  * href="https://doc.cgal.org/latest/Kernel_d/structCGAL_1_1Epeck__d.html">CGAL::Epeck_d</a>
  * < <a target="_blank" href="http://doc.cgal.org/latest/Kernel_23/classCGAL_1_1Dynamic__dimension__tag.html">
  * CGAL::Dynamic_dimension_tag </a> >
- * 
+ *
  * \remark
  * - When Alpha_complex is constructed with an infinite value of alpha, the complex is a Delaunay complex.
  * - Using the default `CGAL::Epeck_d` makes the construction safe. If you pass exact=true to create_complex, the
@@ -161,10 +162,10 @@ class Alpha_complex {
 
  public:
   /** \brief Alpha_complex constructor from an OFF file name.
-   * 
-   * Uses the Points_off_reader to construct the Delaunay triangulation required to initialize 
+   *
+   * Uses the Points_off_reader to construct the Delaunay triangulation required to initialize
    * the Alpha_complex.
-   * 
+   *
    * Duplicate points are inserted once in the Alpha_complex. This is the reason why the vertices may be not contiguous.
    *
    * @param[in] off_file_name OFF file [path and] name.
@@ -183,9 +184,9 @@ class Alpha_complex {
    *
    * The vertices may be not contiguous as some points may be discarded in the triangulation (duplicate points,
    * weighted hidden point, ...).
-   * 
+   *
    * @param[in] points Range of points to triangulate. Points must be in Kernel::Point_d or Kernel::Weighted_point_d.
-   * 
+   *
    * The type InputPointRange must be a range for which std::begin and std::end return input iterators on a
    * Kernel::Point_d or Kernel::Weighted_point_d.
    */
@@ -198,11 +199,11 @@ class Alpha_complex {
    *
    * The vertices may be not contiguous as some points may be discarded in the triangulation (duplicate points,
    * weighted hidden point, ...).
-   * 
+   *
    * @param[in] points Range of points to triangulate. Points must be in Kernel::Point_d.
-   * 
+   *
    * @param[in] weights Range of points weights. Weights must be in Kernel::FT.
-   * 
+   *
    * The type InputPointRange must be a range for which std::begin and std::end return input iterators on a
    * Kernel::Point_d.
    */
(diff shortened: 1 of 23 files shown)


src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -366,14 +369,18 @@ class Alpha_complex {
...
+   * @return `true` if creation succeeds, `false` otherwise.
+   *
    * @pre Delaunay triangulation must be already constructed with dimension strictly greater than 0.
    * @pre The simplicial complex must be empty (no vertices)
-   * 
+   *
    * Initialization can be launched once.
+   *
+   * \note Weighted Alpha complex can have negative filtration values. If `output_squared_values` is set to `false`,
+   * filtration values will be `NaN` in this case.
```
hschreiber, line 380: If I understood correctly, contrary to the regular Alpha filtration, the filtration values in a weighted Alpha filtration are not distances and the option of having $\alpha$ squared or not has not much meaning for it. It would make more sense in that case to ignore `output_squared_values`. If the user did not want the filtration values (what the NaN represent, I guess?), he would just use a simplex tree with `store_filtration` at `false`, no?

Or are the NaN needed for some cases? Or did I simply misunderstood something?

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -389,6 +396,9 @@ class Alpha_complex {
     using Simplex_handle = typename SimplicialComplexForAlpha::Simplex_handle;
     using Vector_vertex = std::vector<Vertex_handle>;
 
+    // For users to be able to define their own sqrt function on their desired Filtration_value type
+    using std::sqrt;
```
hschreiber, line 400: We could perhaps add in the concept of `SimplicialComplexForAlpha` that for "custom" filtration value types, having its own `sqrt` is possible and if not existing `std::sqrt` is used instead?

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
```diff
@@ -29,8 +30,9 @@ using Simplex_tree = Gudhi::Simplex_tree<>;
 using Filtration_value = Simplex_tree::Filtration_value;
 
 void program_options(int argc, char *argv[], std::string &off_file_points, bool &exact, bool &fast,
-                     std::string &weight_file, std::string &output_file_diag, Filtration_value &alpha_square_max_value,
-                     int &coeff_field_characteristic, Filtration_value &min_persistence);
+                     bool& output_squared_root_values, std::string &weight_file, std::string &output_file_diag,
```
hschreiber, line 33: Shouldn't it be `output_squared_values` instead of `output_squared_root_values`? To be more consistent with the `create_complex` interface.

And is the correct term not `square_root` (without de `d`)? Or is it used as some sort of verb?

**Edit:** it seems that for the utilities in general (also for cech etc.), the option is inverted to the notion for the option. Was there a particular reason?

src/Alpha_complex/test/Weighted_alpha_complex_unit_test.cpp
```diff
@@ -124,4 +124,48 @@ BOOST_AUTO_TEST_CASE(Weighted_alpha_complex_3d_comparison) {
...
+      std::clog << "   ( ";
+      for (auto vertex : stree.simplex_vertex_range(f_simplex)) {
+        std::clog << vertex << " ";
+      }
+      std::clog << ") -> " << "[" << stree.filtration(f_simplex) << "]\n";
+
+      BOOST_CHECK(!std::isnan(stree.filtration(f_simplex)));
+    }
+  }
+  std::clog << "Weighted alpha complex with output_squared_values\n";
```
hschreiber, line 157: Isn't it the other way around? "**without** output_squared_values", as `create_complex<false>` is used?

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
```diff
@@ -61,10 +63,12 @@ std::vector<double> read_weight_file(const std::string &weight_file) {
 
 template<class Kernel>
 Simplex_tree create_simplex_tree(const std::string &off_file_points, const std::string &weight_file,
-                                 bool exact_version, Filtration_value alpha_square_max_value) {
+                                 bool exact_version, Filtration_value alpha_square_max_value,
+                                 bool output_squared_root_values) {
```
hschreiber, line 67: Again for `output_squared_root_values`.

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
```diff
@@ -75,18 +79,26 @@ Simplex_tree create_simplex_tree(const std::string &off_file_points, const std::
...
-    if (!alpha_complex_from_file.create_complex(stree, alpha_square_max_value, exact_version)) {
-      std::cerr << "Alpha complex simplicial complex creation failed." << std::endl;
-      exit(-1);
-    }
+    if (output_squared_root_values)
+      complex_creation = alpha_complex_from_file.template create_complex<false>(stree, alpha_square_max_value,
+                                                                               exact_version);
+    else
+      complex_creation = alpha_complex_from_file.template create_complex<true>(stree, alpha_square_max_value,
+                                                                                exact_version);
```
hschreiber, line 87: It is in the right order for `output_squared_root_values`, but if we change it to `output_squared_values`, it needs to be the other way around.

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
```diff
@@ -75,18 +79,26 @@ Simplex_tree create_simplex_tree(const std::string &off_file_points, const std::
...
```
hschreiber, line 97: Same than comment above.

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
```diff
@@ -97,12 +109,13 @@ int main(int argc, char **argv) {
   std::string output_file_diag;
   bool exact_version = false;
   bool fast_version = false;
+  bool output_squared_root_values = false;
```
hschreiber, line 112: `output_squared_root_values`
There are also several below, I will stop pointing them out.

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
hschreiber: Ah, I just realized that this code will only be used precompiled with command lines, it is not a method to include. The only way to keep the old behaviour without inverting the argument would be to use `-squared-filtrations=ON` by default and explicity put `-squared-filtrations=OFF` instead of `--square-root-filtrations` in the command line arguments.

If this is too annoying to do, just keep it like that then. But I would still correct `output_squared_root_values` to `output_square_root_values`. Or be more explicit about the meaning of the option with `do_not_square_filtration_values`.

On this note, we could also change `output_squared_values` to `use_squared_filtration_values` / `output_squared_filtration_values`. As just `value` sounds a bit vague. But that is perhaps a bit subjective.

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
VincentRouvreau: Modified by [44a6747](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/44a6747c1c2ffd0d80db14bfd65d64cd9f0e6e88), [4227322](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/4227322a044ffd87c2a0d3743d8bd9a6f1d66bf1) and [404cb39](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/404cb3951e1c9f6283a2ccc3c39ce340b65f9ddd)

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
VincentRouvreau: Modified by [44a6747](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/44a6747c1c2ffd0d80db14bfd65d64cd9f0e6e88), [4227322](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/4227322a044ffd87c2a0d3743d8bd9a6f1d66bf1) and [404cb39](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/404cb3951e1c9f6283a2ccc3c39ce340b65f9ddd)

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
VincentRouvreau: Modified by [44a6747](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/44a6747c1c2ffd0d80db14bfd65d64cd9f0e6e88), [4227322](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/4227322a044ffd87c2a0d3743d8bd9a6f1d66bf1) and [404cb39](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/404cb3951e1c9f6283a2ccc3c39ce340b65f9ddd)

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
VincentRouvreau: Modified by [44a6747](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/44a6747c1c2ffd0d80db14bfd65d64cd9f0e6e88), [4227322](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/4227322a044ffd87c2a0d3743d8bd9a6f1d66bf1) and [404cb39](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/404cb3951e1c9f6283a2ccc3c39ce340b65f9ddd)

src/Alpha_complex/include/gudhi/Alpha_complex.h
VincentRouvreau: I did so on [44a6747](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/44a6747c1c2ffd0d80db14bfd65d64cd9f0e6e88), but I had to print a warning to the user in case he sets weights but don't want square filtration values.

src/Alpha_complex/include/gudhi/Alpha_complex.h
VincentRouvreau: I specified it on 021561a6d

src/Alpha_complex/utilities/alpha_complex_persistence.cpp
VincentRouvreau: I did so on Modified by [44a6747](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/44a6747c1c2ffd0d80db14bfd65d64cd9f0e6e88), [4227322](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/4227322a044ffd87c2a0d3743d8bd9a6f1d66bf1) and [404cb39](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/404cb3951e1c9f6283a2ccc3c39ce340b65f9ddd)

src/Alpha_complex/test/Weighted_alpha_complex_unit_test.cpp
VincentRouvreau: I fixed the log on [a8d8634](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/a8d86348b4e8fd161506b18e462872cd245c292b)

VincentRouvreau:
@mglisse In the Weighted case, do you think we should always set `output_squared_values` to `true` ? As the distances can be negative, we would never have `NaN` values like this. This would be another way of coding the feature that may be (or not) less confusing for the end-user, no ?

VincentRouvreau:
@hschreiber I took your proposal on [4ee4878](https://github.com/GUDHI/gudhi-devel/pull/1172/commits/4ee4878eb1a8f88d90a5f3a45980c40c059d18e1) to always set `output_squared_values` to `true` in the weighted case, I just throw an exception otherwise. Does it fit what you were proposing ? I tried to `static_assert`, but it doesn't fit our need with the python interface. I could also ignore the user's choice in this case, but exceptions are quite convenient in Python.

src/Alpha_complex/include/gudhi/Alpha_complex.h
```diff
@@ -376,10 +376,10 @@ class Alpha_complex {
    *
    * Initialization can be launched once.
    *
-   * \note Weighted Alpha complex can have negative filtration values. If `output_squared_values` is set to `false`,
-   * filtration values will be `NaN` in this case.
+   * \exception std::invalid_argument In case of a Weighted Alpha complex with `Output_squared_values` set to `false`,
+   * as Weighted Alpha complex can have negative filtration values.
```
hschreiber, line 380: Radical, but I am fine with the new proposition (even though, the value should be the other way around, as the filtration value is not squared (couldn't be negative otherwise), but technically it would not fit with the default value of the non weighted case). But here I would write "..., as the filtration values of a Weighted Alpha complex results from a power product and not from an euclidean distance. Therefore this template argument should be kept to the default value." instead of "..., as Weighted Alpha complex can have negative filtration values."

hschreiber (approved):

src/python/gudhi/delaunay_complex.pyx
```diff
@@ -99,63 +102,65 @@ cdef class DelaunayComplex:
...
-            point. The vertices may not be numbered contiguously as some points may be discarded in the triangulation
-            (duplicate points, weighted hidden point, ...).
-        :rtype: SimplexTree
+        Args:
+            max_alpha_square: The maximum alpha square threshold the simplices shall not exceed. Default is set to
+                infinity, and there is very little point using anything else since it does not save time.
+            filtration: Set this value to `None` (default value) if filtration values are not needed to be computed
+                (will be set to `NaN`). Set it to `alpha` to compute the filtration values with the Alpha complex, or
+                to `cech` to compute the Delaunay Cech complex.
+            output_squared_values: Square filtration values when `True`. Default is `True`.
```
mglisse, line 119: Does the Python doc mention somewhere that passing `False` can be dangerous with negative weights?

hschreiber:
> > @mglisse In the Weighted case, do you think we should always set `output_squared_values` to `true` ? As the distances can be negative, we would never have `NaN` values like this. This would be another way of coding the feature that may be (or not) less confusing for the end-user, no ?
> 
> If a user knows they always have positive weights and wants the square root, it seems a bit disappointing to say NO when it would actually work just fine.

But what is the meaning of a "square rooted filtration value" of a weighted alpha complex? Would we return something like `||x-y||² - w` when `output_squared_values` is true and `||x-y|| - w` when false? Or `sqrt(||x-y||² - w)`? I don't even know if CGAL makes it possible. In both cases, I don't feel like it is Gudhi's responsability to handle how users want to post-process the power product when there is no clear convention for it. Or is there?

PR #1176, opened 2025-02-18
[Simplex tree] adds possibility to different insertion strategies

hschreiber:
Adds different strategies for `insert_simplex_and_subfaces` to handle the filtration values of the simplices inserted.

Includes PR #1122 so it should be easier to review once the latter is merged.

Necessary for the multi-simplex tree from #976.


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1056,81 +1134,165 @@ class Simplex_tree {
     return insert_simplex_raw(copy, filtration);
   }
 
-    /** \brief Insert a N-simplex and all his subfaces, from a N-simplex represented by a range of
+  /**
+   * @brief List of insertion strategies of a simplex \f$ \sigma \f$.
+   */
+  enum class Insertion_strategies {
```
mglisse, line 1140: Is it a good idea to make it plural?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1056,81 +1134,165 @@ class Simplex_tree {
...
+     * Then, the filtration values of all faces of \f$ \sigma \f$ are pushed to the union of their old value and of
+     * the new value of \f$ \sigma \f$.
+     */
+    LOWEST,
+    /**
+     * @brief Let \f$ f \f$ be the filtration value given as argument. Inserts the simplex \f$ \sigma \f$ as follows:
+     * - if \f$ \sigma \f$ was not inserted yet, then \f$ \sigma \f$ and all its non existing faces are inserted at the
+     * first possible filtration value with minimum filtration value \f$ f \f$.
+     * - if \f$ \sigma \f$ existed already, then it and all of its potentially already inserted cofaces are pushed to the
+     * intersection of their old value and \f$ f \f$.
```
mglisse, line 1156: Is that useful?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1056,81 +1134,165 @@ class Simplex_tree {
...
+     * be @ref FiltrationValue::unify_lifetimes (\f$ f \f$, \f$ f' \f$).
+     *
+     * Then, the filtration values of all faces of \f$ \sigma \f$ are pushed to the union of their old value and of
+     * the new value of \f$ \sigma \f$.
+     */
+    LOWEST,
+    /**
+     * @brief Let \f$ f \f$ be the filtration value given as argument. Inserts the simplex \f$ \sigma \f$ as follows:
+     * - if \f$ \sigma \f$ was not inserted yet, then \f$ \sigma \f$ and all its non existing faces are inserted at the
+     * first possible filtration value with minimum filtration value \f$ f \f$.
```
mglisse, line 1154: It feels a bit strange that it still inserts missing faces: this means that a missing face and a face with filtration value +infinity have very different consequences.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I don't know, is it better in singular? I just made it plural because it is an enumeration with more than one element. I can make it singular if that is the convention.

mglisse: The standard uses for instance `https://en.cppreference.com/w/cpp/types/numeric_limits/float_round_style` singular. I don't know if we already have enums in Gudhi and what conventions they follow.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: It can make sense, but in my mind that's not an insertion anymore. If anything, it is closer to a removal (and the code and complexity are probably more similar to a removal as well).

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: There is one with `Extended_simplex_type`, so I will make it singular.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: But then, even with other strategies, inserting an existing simplex is not really an insertion. The old version also updated the filtration values of everyone, just the other way around. 

But yes, the complexity is going towards a removal, as it needs to iterate through the cofaces (just not always to the very end, as it can stop whenever it is not necessary anymore). But that is to expect if you want this type of strategy, I don't think the user would be surprised.

But if you really don't like it, I can remove it. I don't have a use for it for now, it just didn't hurt to add it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Yes, a function called `insert_*_subfaces` should insert subfaces, but the different behavior between missing faces and faces with a high filtration value feels inconsistent (I would understand better if the filtration value was increased to +inf when a face is missing, with the idea that a missing face is the same as a face with infinitely large filtration value).
How does David use the function exactly?

hschreiber: If one face is at +inf, then all non existing cofaces of that face (which are faces of the given simplex) will be inserted at +inf, the given filtration value is just a minimum in that case (that is what I meant with "first possible filtration value with minimum filtration value $f$". It makes no sense to have sigma at a finite filtration value when some of its faces are at +inf. I am still struggling to understand if you misunderstood something (which means I have to better explain it in the description), or if I still don't understand your question...

hschreiber: On a simplex example: say you have a triangle {0,1,2}. The vertices {0}, {1} and {2} are already inserted at filtration value 0. Same for edge {0,1} that is inserted at value 5. Now, we use `insert_simplex_and_subfaces({0,1,2}, 4, HIGHEST)`. This will insert edge {0,2} and {1,2} at filtration value 4, but the triangle {0,1,2} at filtration value 5.

Before, `insert_simplex_and_subfaces({0,1,2}, 4)` (equivalent now to `insert_simplex_and_subfaces({0,1,2}, 4, LOWEST)`) would have done the same for {0,2} and {1,2}, but would have pushed down {0,1} to 4 et inserted {0,1,2} also at 4.

And finally, `insert_simplex_and_subfaces({0,1,2}, 4, FIRST_POSSIBLE)` is inserting {0,2} and {1,2} at value 0 and {0,1,2} at value 4.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Maybe to explain my perspective: if you think of a filtered simplicial complex as a set of simplicial complexes (indexed by the filtration value, and with an inclusion when filtration values are comparable), lowering the filtration value of a simplex means inserting it in some of the complexes where it was not yet present, while increasing the filtration value means removing it from some of the complexes.
I am perfectly happy with such functionality being added, but I'd much rather give it a different name than `insert`.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: > On a simple example: say you have a triangle {0,1,2}. The vertices {0}, {1} and {2} are already inserted at filtration value 0. Same for edge {0,1} that is inserted at value 5. Now, we use `insert_simplex_and_subfaces({0,1,2}, 4, HIGHEST)`. This will insert edge {0,2} and {1,2} at filtration value 4, but the triangle {0,1,2} at filtration value 5.

Would a 4th strategy that inserts 02, 12 and 012 all at filtration value 5 make sense?

> Before, `insert_simplex_and_subfaces({0,1,2}, 4)` (equivalent now to `insert_simplex_and_subfaces({0,1,2}, 4, LOWEST)`) would have done the same for {0,2} and {1,2}, but would have pushed down {0,1} to 4 et inserted {0,1,2} also at 4.
> 
> And finally, `insert_simplex_and_subfaces({0,1,2}, 4, FIRST_POSSIBLE)` is inserting {0,2} and {1,2} at value 0 and {0,1,2} at value 5.

Is `insert(x,*,first_possible)` equivalent to `insert(x,-inf,lowest)`?

I don't know if the names highest/lowest properly convey the main difference: lowest may modify the filtration value of existing simplices (so it can really insert it at the time requested), while the others all keep them fixed. Ah, but that property only exists in my head, since you have cases were insert increases the filtration value of existing cofaces...

> If you have now "missing simplices" in the sense that they are included but at value +inf, then `HIGHEST` will insert all their future cofaces also at infinity, while `LOWEST` will force their appearances by changing their filtration value to the given one while inserting the cofaces.

I find it a bit confusing that, without 02, I get 012 at time 5, while with 02 at time +inf (supposedly equivalent to not having 02?), I get 012 at time +inf. But maybe that's still useful. Or maybe it could yield a 5th strategy.

I'd still be interested in a reminder of what cases David really cared about when he came up with a different strategy.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: From an algebraic point of view, I agree that it is a removal and not an insertion. But from a data structural point of view, both strategies do the same, just on different nodes. As a compromise, I would propose to use `Update_strategy` instead of `Insertion_strategy`. What do you think? Because I am not convinced that it is a good idea to add a new method for this.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: > > On a simple example: say you have a triangle {0,1,2}. The vertices {0}, {1} and {2} are already inserted at filtration value 0. Same for edge {0,1} that is inserted at value 5. Now, we use `insert_simplex_and_subfaces({0,1,2}, 4, HIGHEST)`. This will insert edge {0,2} and {1,2} at filtration value 4, but the triangle {0,1,2} at filtration value 5.
> 
> Would a 4th strategy that inserts 02, 12 and 012 all at filtration value 5 make sense?

It would make sense as it is not wrong. It would just even more disregard the input value than before. But as an additional strategy, it is valid. Complexity wise, I don't know if it would be better, I would have to verify again how the traversal in the recursion is done.

> > Before, `insert_simplex_and_subfaces({0,1,2}, 4)` (equivalent now to `insert_simplex_and_subfaces({0,1,2}, 4, LOWEST)`) would have done the same for {0,2} and {1,2}, but would have pushed down {0,1} to 4 et inserted {0,1,2} also at 4.
> > And finally, `insert_simplex_and_subfaces({0,1,2}, 4, FIRST_POSSIBLE)` is inserting {0,2} and {1,2} at value 0 and {0,1,2} at value 5.
> 
> Is `insert(x,*,first_possible)` equivalent to `insert(x,-inf,lowest)`?

No, it is equivalent to `insert(x,-inf,highest)`. But now that I see it, I could factorise both.

> I don't know if the names highest/lowest properly convey the main difference: lowest may modify the filtration value of existing simplices (so it can really insert it at the time requested), while the others all keep them fixed. Ah, but that property only exists in my head, since you have cases were insert increases the filtration value of existing cofaces...

The idea of highest/lowest was that one strategy inserts at highest choice (between given and potentially already existing value) and the other at lowest choice and then updates the filtration such that it remains valid. It also highlight the duality between both strategies, as one is a bit the "co" version of the other. But I have nothing against a name change.

> > If you have now "missing simplices" in the sense that they are included but at value +inf, then `HIGHEST` will insert all their future cofaces also at infinity, while `LOWEST` will force their appearances by changing their filtration value to the given one while inserting the cofaces.
> 
> I find it a bit confusing that, without 02, I get 012 at time 5, while with 02 at time +inf (supposedly equivalent to not having 02?), I get 012 at time +inf. But maybe that's still useful. Or maybe it could yield a 5th strategy.

I don't see "not inserted yet faces" and "missing faces" (at +inf) as something equivalent. A missing face is purposely missing, while a not existing face is a face where you still have to indicate where it should go (at +inf for example).

Perhaps, it would be a good idea to describe me the strategies you would like to have and then I see, if I can replace one of my proposition with yours or if I can add this strategy to the existing ones?

> I'd still be interested in a reminder of what cases David really cared about when he came up with a different strategy.

What mattered to David, is that if he constructed a filtration in the simplex tree and later wanted to add a simplex, but without really caring about where, he didn't wanted `insert_simplex_and_subfaces(s)` to flat partially the filtration down to 0, but to insert the simplex at the first position the already existing filtration permits it. Then, in python, he would potentially modify some filtration values by hand. So, basically the `FIRST_POSSIBLE` strategy. The idea behind `HIGHEST` was just to give the possibility to give a lower bound on where to insert the simplex. The subcase of `HIGHEST` where the simplex already exists was just a natural extension, as this case has to be covered (even though not useful for multipers).

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Maybe we should ask for more opinions. It seems strange to me to repurpose a function called **insert**\_simplex\_and\_**sub**faces to do something that
1. corresponds to a removal
2. increases the filtration value of a simplex and its **co**faces

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1056,81 +1135,167 @@ class Simplex_tree {
...
+     *
+     * Then, the filtration values of all faces of \f$ \sigma \f$ are pushed to the union of their old value and of
+     * the new value of \f$ \sigma \f$.
+     */
+    LOWEST,
+    /**
+     * @brief Let \f$ f \f$ be the filtration value given as argument. Inserts the simplex \f$ \sigma \f$ as follows:
+     * - if \f$ \sigma \f$ was not inserted yet, then \f$ \sigma \f$ and all its faces, which are not already included
+     * in the complex, are inserted at either \f$ f \f$ or at the first possible filtration value when \f$ f \f$ is too
+     * low (to insure the validity of the filtration).
```
mglisse, line 1157:
```suggestion
     * low (to ensure the validity of the filtration).
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: > > Would a 4th strategy that inserts 02, 12 and 012 all at filtration value 5 make sense?
>
> It would make sense as it is not wrong. It would just even more disregard the input value than before.

I think it just does things in a slightly different order. It first asks: what is the earliest time I can insert 012? And then does `insert_simplex_and_subfaces(simplex_012, that_time)`. Whereas your version recursively inserts the subfaces with the same strategy.
I mostly suggested it because I have trouble figuring out what strategies are likely to be useful.

> Complexity wise, I don't know if it would be better,

Mine should be a bit slower, the simplest way I can see to implement it is to follow yours, inserting all the new `Simplex_handle` in a vector, and updating their filtration value at the end once we know the final value for the main simplex.

> I don't see "not yet inserted faces" and "missing faces" (at +inf) as something equivalent. A missing face is purposely missing, while a not existing face is a face where you still have to indicate where it should go (at +inf for example).

A philosophical difference of opinions...

> Perhaps, it would be a good idea to describe me the strategies you would like to have and then I see, if I can replace one of my proposition with yours or if I can add this strategy to the existing ones?

I don't have particular strategies I want. We could easily add a dozen strategies. I am just afraid that at most a couple of them would be useful and we would confuse users into picking the wrong one, or at least waste time trying to figure out what to use. But maybe that's still what we should do.

> What mattered to David, is that if he constructed a filtration in the simplex tree and later wanted to add a simplex, but without really caring about where, he didn't wanted `insert_simplex_and_subfaces(s)` to flat partially the filtration down to 0,

It is true that, when we don't explicitly specify a filtration value, this behavior can be annoying. I guess it wouldn't have been a problem if we had used +inf as the default filtration value (although David favors -inf...).

> but to insert the simplex at the first position the already existing filtration permits it.

I guess that makes sense, although it isn't completely obvious in this case if we want each new subface to appear also as early as possible (equivalent to calling the function on each missing subface first) or if it is better to have everything appear at the same time (it is one insertion, turning it into multiple insertions at multiple times could be unexpected).

> Then, in python, he would potentially modify some filtration values by hand.

I am unsure how that part works. One possibility, if you are going to overwrite the filtration values anyway, would be to insert at +inf. Then at least you can identify the new subfaces as those with +inf filtration value. Or you could be satisfied with the values that were computed for the subfaces and not modify them. But I suspect that maybe he was only using this function when all the subfaces were already present?

> The idea behind `HIGHEST` was just to give the possibility to give a lower bound on where to insert the simplex.

Makes sense.

> The subcase of `HIGHEST` where the simplex already exists was just a natural extension, as this case has to be covered (even though not useful for multipers).

I see... The way HIGHEST is defined in the normal case, if we try to see what should happen if the simplex is already there with a lower filtration value, the options seem to be
1. do nothing (this violates the idea that the value should be a lower bound, but we could easily say that this lower bound only applies to **new** simplices, we don't increase the filtration value of faces that are already there after all)
2. increase the filtration value of the simplex (and that of the cofaces if necessary)
3. refuse to do it (throw an exception or something)

hschreiber: > I don't have particular strategies I want. We could easily add a dozen strategies. I am just afraid that at most a couple of them would be useful and we would confuse users into picking the wrong one, or at least waste time trying to figure out what to use. But maybe that's still what we should do.

I agree, we shouldn't try to write all possible strategies here. We should choose among the strategies we think are most commonly useful. That is what I tried to do with those three options, as I now that they are useful for multipers (except for the modifications of the cofaces, as this method is never used on already existing simplices there). So let me reformulate: what do you think would be useful strategies to replace or complement the ones I suggested?
 
> I guess that makes sense, although it isn't completely obvious in this case if we want each new subface to appear also as early as possible (equivalent to calling the function on each missing subface first) or if it is better to have everything appear at the same time (it is one insertion, turning it into multiple insertions at multiple times could be unexpected).

I know that David needed the first version, but I agree, the second one could also be interesting. I cannot say what would be better: implement both or chose one.

> I am unsure how that part works. One possibility, if you are going to overwrite the filtration values anyway, would be to insert at +inf. Then at least you can identify the new subfaces as those with +inf filtration value. Or you could be satisfied with the values that were computed for the subfaces and not modify them. But I suspect that maybe he was only using this function when all the subfaces were already present?

I admit, I don't really remember the details, in particular because he changed his strategy several times. I just remember that for a while, he modified the method such that it just ignored the filtration values and then he overwrote everything in Python. Later, he asked me to modify the insertion method so that he can avoid having to do that in most cases.

> I see... The way HIGHEST is defined in the normal case, if we try to see what should happen if the simplex is already there with a lower filtration value, the options seem to be
> 
>     1. do nothing (this violates the idea that the value should be a lower bound, but we could easily say that this lower bound only applies to **new** simplices, we don't increase the filtration value of faces that are already there after all)
> 
>     2. increase the filtration value of the simplex (and that of the cofaces if necessary)
> 
>     3. refuse to do it (throw an exception or something)

Yes. I went for 2, but it does not has to be that way. If you think that this solution violates the idea of insertion, we could just go for 1, as the simplex would already be considered inserted in the filtered complex at filtration value "lower bound". Throwing would somehow go against the current strategy, where it does not throw when a simplex is already inserted in a lower complex than given by the argument filtration value (which would also be a removal otherwise). It just does nothing.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I think that is still a matter of point of view: I see the simplex tree as a structure which stores the simplices of a complex and when `store_filtration` is enabled, it additionally stores the appearance time of the simplices when the complex is considered filtered. From this point of view, `insert` is already wrong when a simplex is already inserted, as all the method is doing in that case, is updating the stored appearance time, not inserting it in the simplex tree. But if you allow the method to do so, updating the appearance time of cofaces seems fine too, as the idea the same. And I am pretty sure that a user with the same point of view will not be confused by it.
Things are different when you look at it as an algebraic object of course. But it is difficult to say which point of view is the more intuitive here.
But well, from the discussion below, this will probably not be a problem anymore anyway.

mglisse: @DavidLapous also seems to find it strange if `insert` increases the filtration value.

hschreiber: I already agreed with option (1) in the discussion below (which does not increase anything). What I need now is a confirmation from you (or/and @DavidLapous if he wants to review) on the strategies to implement.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -746,6 +747,44 @@ class Simplex_tree {
     }
   }
 
+  /**
+   * @brief Returns a non-const reference to the filtration value of a simplex.
+   * 
+   * @warning Cannot be called on a null_simplex.
+   * @warning Only available if SimplexTreeOptions::store_filtration is true.
```
VincentRouvreau: Just a suggestion but I found the doc rendering less *aggressive* (`@warning` means a big red square) with:
```suggestion
   * @note It will not compile if @ref SimplexTreeOptions::store_filtration is false.
   * @exception std::invalid_argument if sh is a @ref null_simplex.
```

image

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -746,6 +747,44 @@ class Simplex_tree {
...
+  }
+
+  /**
+   * @brief Returns a const reference to the filtration value of a simplex. It is a more strict variant of
+   * @ref filtration, as it will not compile if @ref SimplexTreeOptions::store_filtration is false and it will throw if
+   * the given simplex is a @ref null_simplex.
+   * 
+   * @warning Cannot be called on a null_simplex. If this option is necessary, use @ref filtration instead.
+   * @warning Only available if @ref SimplexTreeOptions::store_filtration is true. If false is necessary,
+   * use @ref filtration instead.
```
VincentRouvreau: Just a suggestion but I found the doc rendering less *aggressive* (`@warning` means a big red square) with:
```suggestion
   * @brief Returns a const reference to the filtration value of a simplex.
   *
   * @note It is a more strict variant of @ref filtration, as it will not compile if
   * @ref SimplexTreeOptions::store_filtration is false and it will throw if
   * the sh is a @ref null_simplex.
   * @exception std::invalid_argument if sh is a @ref null_simplex. Note that @ref filtration method returns infinity in this
   * case.
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -746,6 +747,44 @@ class Simplex_tree {
     }
   }
 
+  /**
+   * @brief Returns a non-const reference to the filtration value of a simplex.
+   * 
+   * @warning Cannot be called on a null_simplex.
+   * @warning Only available if SimplexTreeOptions::store_filtration is true.
+   */
+  Filtration_value& get_filtration_value(Simplex_handle sh) {
```
VincentRouvreau: Just a curiosity question: Why this new method ? It is not tested nor used somewhere...
I know we have to live with the existing code, but:
```cpp
auto  e = st.find({0, 2});
st.get_filtration_value(e) = 2;
```
seems quite strange to me. I would prefer:
```cpp
auto  e = st.find({0, 2});
st.filtration(e) = 2;
```
Even if it implies that the behaviour is a bit different because it can throw an exception (if called with null_simplex) but it is justified by the feature.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1096,69 +1135,138 @@ class Simplex_tree {
...
     dimension_ = (std::max)(dimension_, static_cast<int>(std::distance(copy.begin(), copy.end())) - 1);
 
-    return rec_insert_simplex_and_subfaces_sorted(root(), copy.begin(), copy.end(), filtration);
+    if constexpr (Options::store_filtration){
+      if (insertion_strategy == Insertion_strategy::LOWEST)
+        return _rec_insert_simplex_and_subfaces_sorted(root(), copy.begin(), copy.end(), filtration);
+      else if (insertion_strategy == Insertion_strategy::HIGHEST)
+        return _insert_simplex_and_subfaces_at_highest(root(), copy.begin(), copy.end(), filtration);
+      else
+        return _insert_simplex_and_subfaces_at_highest(root(), copy.begin(), copy.end(), 0);
```
VincentRouvreau: In [this discussion with Marc](https://github.com/GUDHI/gudhi-devel/pull/1176#discussion_r1963409345), you say IIUC:
 > `insert(x,*,first_possible)` is equivalent to `insert(x,-inf,highest)`

But here I see that you `insert(x, 0,highest)`. I am confused.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1096,69 +1135,138 @@ class Simplex_tree {
...
-        GUDHI_CHECK(v != null_vertex(), "cannot use the dummy null_vertex() as a real vertex");
-    )
+    GUDHI_CHECK_code(for (Vertex_handle v : copy)
+                         GUDHI_CHECK(v != null_vertex(), "cannot use the dummy null_vertex() as a real vertex"););
     // Update dimension if needed. We could wait to see if the insertion succeeds, but I doubt there is much to gain.
     dimension_ = (std::max)(dimension_, static_cast<int>(std::distance(copy.begin(), copy.end())) - 1);
 
-    return rec_insert_simplex_and_subfaces_sorted(root(), copy.begin(), copy.end(), filtration);
+    if constexpr (Options::store_filtration){
+      if (insertion_strategy == Insertion_strategy::LOWEST)
```
VincentRouvreau: Just a matter of taste, but here we have a nice enum class, I would have used a `switch/case` instead of nested condition. It would allow to decide what to do in the `default` case.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1096,69 +1135,138 @@ class Simplex_tree {
...
+                                                                          ForwardVertexIterator last,
+                                                                          const Filtration_value& filt) {
     // An alternative strategy would be:
     // - try to find the complete simplex, if found (and low filtration) exit
     // - insert all the vertices at once in sib
     // - loop over those (new or not) simplices, with a recursive call(++first, last)
     Vertex_handle vertex_one = *first;
 
-    if (++first == last) return insert_node_<true, false, true>(sib, vertex_one, filt);
+    if (++first == last) return insert_node_<update_fil, false, update_fil>(sib, vertex_one, filt);
```
VincentRouvreau, line 1320: A small comment would be welcome here, I was lost with the templates arguments :sweat_smile: 
(just like the one below where you just say `we are building the whole subtree from scratch`)

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: There is also `\pre` in doxygen to document preconditions (although I guess that's more runtime stuff (e.g. pointer not null) than compile-time).

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: It was used in `_make_subfiltration_non_decreasing`, but I replaced this line with `_to_node_it(sh)->second.filtration()` and removed the methods for now. I will need them in another PR, but we can discuss it when the time comes.

The reason why I called them differently than `filtration()` was because of a discussion we had (I thing in PR #976?), where the result was that we should call it differently, because the mechanism with `null_simplex` etc. is not handled the same.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I assumed somehow that the simplex tree didn't allow negative values and so the smallest non negative is 0. But it was perhaps just the cohomology algorithm who didn't allow them? Because I don't see why this should be a problem for the tree after thinking about it. So I replace the value with -inf.

But this has the consequence that we have to add the inverse operator to the `FiltrationValue` concept.
Which becomes problematic for `unsigned` values... I think `-max` for `unsigned int` is 1 and not 0, no? Do you see a workaround for it which would work for any possible FiltrationValue?

A possibility would be to ask for the inverse operator for filtration values with `std::numeric_limits<FiltrationValue>::has_infinity` at `true` and `std::numeric_limits<FiltrationValue>::lowest` for the others.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I don't mind, so I changed to a switch. For the default case, I added a throw as it should never be reached.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I added a comment explaining the template arguments above. Is it okay like this?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Ahhh, I remember another problem related to multi-parameter filtration and infinity values. As it is related to the above and it impacts the FiltrationValue concept, it is perhaps the right moment to discuss about it here:

When comparing two filtration values with multiple parameters, two infinity values should be different if the number of parameters is different. At least intuitively, can some of the experts confirm (@DavidLapous for example?) ? If not, we can forget about the discussion and we just use the solution in the comment above. But if it the case, the infinity value is related to the number of parameters and therefore `std::numeric_limits<FiltrationValue>::infinity()` cannot give it to us anymore (at least not canonically). We would need a `FiltrationValue get_infinity_value([[maybe_unused]] const FiltrationValue& f)` method, which would return the infinity value with the same number of parameters as `f`, that we can use to do the comparisons with `f` (used at 2 places in the simplex tree).
Then we would simply ask additionally for a `FiltrationValue get_minus_infinity_value([[maybe_unused]] const FiltrationValue& f)`.

mglisse: > When comparing two filtration values with multiple parameters, two infinity values should be different if the number of parameters is different.

When does it make sense to compare filtration values that don't live in the same space (don't have the same dimension)? (except possibly tricks like when David stored a single +inf instead of d to denote all infinities)

hschreiber: I agree. I though this more as some sort of security. That, if for some reasons the user compares two values with different number of parameters, it could be that we want the program to react negatively somehow (by return false for example, even if both are at infinity).

In the simplex tree, we allocate the `inf_` variable independently from the number of parameters. In that case, the above "secure behavior" would obstruct some simplex tree method which compare to `inf_`.

mglisse: I think we can react "negatively" by asserting / throwing an exception...

hschreiber: Yes, but this is exactly the problem. Everything would throw as soon as we want to do `f == inf_` in `initialize_filtration` or `prune_above_filtration` because `inf_` has 1 parameter (defined at the construction of the simplex tree, when we don't know the number of parameters) whereas `f` can have whatever was decided on later...

mglisse: We could define, like David did originally, that a vector with a single element that is ±inf actually represents a vector of the appropriate dimension with d copies of that element. It may make comparisons a little bit more complicated, but remains manageable. (Other magic values might be possible, like an empty vector)

Otherwise, yes, we need the dimension to create an infinity vector of the right dimension, whether we get that information from the simplex tree or from another filtration value.

hschreiber: Exactly! That was the point of my discussion: do we choose the first or second option? I.e., do we prefer that infinity values are equal independently in which space they are living or do we have a need to discriminate them? So that I can change things (this and other PRs) accordingly.

mglisse: They wouldn't be guaranteed to be equal. When the filtration type is `array<double,2>` vs `array<double,3>`, it would either not be comparable or be different. But, for a dynamic dimension, when meaninglessly comparing values of different dimensions, if there are cases that we cannot diagnose and where we return some arbitrary result, I don't see that as a problem.
I would document that only comparison between filtrations of the same dimension is supported. What happens if users break that precondition... too bad for them, no guarantee.

mglisse: For future-proofing, if we wanted get_infinity to return something that depends on the dimension, would we really need to pass it some argument? Doesn't `*this` allow to get to the dimension somehow?

hschreiber: > They wouldn't be guaranteed to be equal. When the filtration type is `array<double,2>` vs `array<double,3>`, it would either not be comparable or be different. But, for a dynamic dimension, when meaninglessly comparing values of different dimensions, if there are cases that we cannot diagnose and where we return some arbitrary result, I don't see that as a problem. I would document that only comparison between filtrations of the same dimension is supported. What happens if users break that precondition... too bad for them, no guarantee.

That is how it is implemented for now in the multi-parameter classes: there is an assert and if the assert is skipped, the result is undefined. But I would have changed that if we want to allow comparison between everyone such that there is no problem with the simplex tree.

> For future-proofing, if we wanted get_infinity to return something that depends on the dimension, would we really need to pass it some argument? Doesn't `*this` allow to get to the dimension somehow?

There will be a `number_of_parameter` attribute in the simplex tree later. But even with the information of the dimension, we need to be able to construct this infinity value with it without making weird very special cases (which goes against generality). Having a method taking a filtration value as parameter is a good way to have specializations the compiler can easily find by-itself. It would work the same way all other methods we added in the concept works. And we would get rid of the condition of specializing `std::numeric_limits` (except for extended_filtration, but this works only with doubles anyway).

hschreiber: In the simplex tree, the comparisons would look like this: `filtration(sh) == get_infinity_value(filtration(sh))`.

And we would have something like this in `filtration_value_utils.h`:
```C++
template <typename Arithmetic_filtration_value>
constexpr Arithmetic_filtration_value get_infinity_value([[maybe_unused]] const Arithmetic_filtration_value& f)
{
  return std::numeric_limits<Arithmetic_filtration_value>::has_infinity
             ? std::numeric_limits<Arithmetic_filtration_value>::infinity()
             : std::numeric_limits<Arithmetic_filtration_value>::max();
}
```

In the `Multi_parameter_filtration` class, there is already this method:

```C++
friend Multi_parameter_filtration get_infinity_value(const Multi_parameter_filtration &f)
{
  return Multi_parameter_filtration::inf(f.num_parameters());
}
```

mglisse: > In the simplex tree, the comparisons would look like this: `filtration(sh) == get_infinity_value(filtration(sh))`.

If that's supposed to become a common operation it seems expensive, creating a new vector just to check if the existing one contains infinity.

hschreiber: To avoid creating the vector every time, we could just update the `inf_` value once with the `get_infinity_value` method, when the user sets the number of parameters and from there on, assume that he will only try to insert simplices with that number of parameters. The only problem with this is that the user has to update the number of parameters when he want more than 1 parameter. He cannot just ignore it, if he didn't needed to store it.

mglisse: A function is_positive_inf that directly tests for infinity instead of relying on get_infinity and operator== could also help, if that's something we intend to check often.

hschreiber: Oh, right, that could be a good compromise! So obvious that I didn't though about it. As I think we have this problem only when we want to compare to infinity in the simplex tree.

hschreiber: So, I added a `is_positive_infinity` method to this PR to see how it looks like. What do you think?

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -49,6 +50,17 @@ struct FiltrationValue {
...
   friend bool operator==(const FiltrationValue& f1, const FiltrationValue& f2);
 
+  /**
+   * @brief Returns true if and only if the given filtration value is at infinity.
+   */
+  friend bool is_positive_infinity(const FiltrationValue& f);
+
+  /**
+   * @brief Only necessary if `std::numeric_limits<FiltrationValue>::has_infinity` returns `true`.
+   * Inverses the value. Only used on the infinity filtration value to obtain minus infinity.
```
mglisse:
```suggestion
   * Negates the value. Only used on the infinity filtration value to obtain minus infinity.
```
For me, "inverse" is 1/x.

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
```diff
@@ -30,6 +31,25 @@ inline struct empty_filtration_value_t {
 
 }  // namespace simplex_tree
 
+/**
+ * @ingroup simplex_tree
+ * @brief Given two filtration values at which a simplex exists, stores in the first value the minimal union of births
+ * generating a lifetime including those two values.
+ * This is the overload for when @ref FiltrationValue is an arithmetic type, like double, int etc.
+ * Because the filtration values are totally ordered then, the union is simply the minimum of the two values.
```
mglisse: Doc for the wrong function?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1096,69 +1105,177 @@ class Simplex_tree {
     return insert_simplex_raw(copy, filtration);
   }
 
-    /** \brief Insert a N-simplex and all his subfaces, from a N-simplex represented by a range of
+  /**
+   * @brief List of insertion strategies for @ref insert_simplex_and_subfaces, which takes a simplex \f$ \sigma \f$
+   * and a filtration value \f$ f \f$ as argument.
+   */
+  enum class Insertion_strategy : std::uint8_t {
```
mglisse: Reading the current version
* FIRST_POSSIBLE looks good.
* HIGHEST is something like FIRST_POSSIBLE but no new simplex may be lower than f.
    * Should the doc mention intersect_lifetimes?
    * Should the doc say that FIRST_POSSIBLE is equivalent to HIGHEST with argument -inf?
    * Why is that called "HIGHEST"?
* LOWEST is the current behavior, minimally extends the lifetime of simplices so we have a filtration and sigma's lifetime contains f.
    * Why is that called "LOWEST"?
* FORCE: all new simplices have value f, plus sigma is given the value f.

If I try to describe them differently: in all cases, we insert whatever is missing from sigma and its faces, the only question is what happens to filtration values.
* LOWEST: sigma and its faces have their (possibly empty if they were not present) filtration extended by unify_lifetimes(f,*)
* HIGHEST: the filtration value of a new simplex tau is the intersection of f and all the existing faces of tau. No existing simplex changes filtration. The new filtration values are as small as possible (but no smaller than f) with the constraint that existing values cannot change (and we must have a filtration).
* FIRST_POSSIBLE is the special case of HIGHEST where f is "always" or -inf (can be ignored in the intersection).
* FORCE: all new simplices get f, all old simplices remain the same, except that sigma is set to f. That's a bit awkward. Maybe that's useful, but I just want to make sure that it wouldn't be better to do just the new insertions (with value f), and let the caller do `set_filtration` afterwards if they really want to change the value of sigma.

Would it make sense to merge HIGHEST and FIRST_POSSIBLE, where the second would happen when we call insert without its second argument? Or would that complicate things too much?

Do you have other ideas for names that might be more intuitive? Maybe LOWEST→INCREASE_LIFETIME, HIGHEST→FIRST_POSSIBLE?

src/Simplex_tree/include/gudhi/Simplex_tree/filtration_value_utils.h
hschreiber: Ah, yes, wrong copy/paste. Is corrected now.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: > - Should the doc mention intersect_lifetimes?

That is a good idea, as `intersect_lifetimes` can be not trivial for complex types, so the user should know.

> - Should the doc say that FIRST_POSSIBLE is equivalent to HIGHEST with argument -inf?

Why not. It can help clarify.

> - Why is that called "HIGHEST"?
> - Why is that called "LOWEST"?

I saw HIGHEST and LOWEST as two dual behaviour, where one was pushing up values to fulfill the validity of a filtration and the other pushes down those values for the same purpose. But that is not really the case anymore. I don't mind at all to change the names.

> - FORCE: all new simplices get f, all old simplices remain the same, except that sigma is set to f. That's a bit awkward. Maybe that's useful, but I just want to make sure that it wouldn't be better to do just the new insertions (with value f), and let the caller do `set_filtration` afterwards if they really want to change the value of sigma.

I just had the case in mind, where the user uses some loop to insert and could potentially insert several times the same simplex (because some coface was inserted first for example, and then the rest of the iterations corrects the values of the faces). This would make sure that the last value would be stored. But storing the first only also makes sense, so I don't mind here.

> Would it make sense to merge HIGHEST and FIRST_POSSIBLE, where the second would happen when we call insert without its second argument? Or would that complicate things too much?

Would that not destroy retro-compatibility? In the current main branch, calling `insert_simplex_and_subfaces(sig)` is possible and has the behaviour of `LOWEST` with a filtration value at 0/{}. Or would you not use the strategy as third argument, but as template argument? Then it would brake the syntax, no? (ie., you would need to do `insert_simplex_and_subfaces<>(sig)` instead of `insert_simplex_and_subfaces(sig)`)

What we could do, is just remove `FIRST_POSSIBLE` and let the user use `HIGHEST` with a filtration value at -inf. `FIRST_POSSIBLE` is more of a commodity than a necessity if we keep `HIGHEST`.

> Do you have other ideas for names that might be more intuitive? Maybe LOWEST→INCREASE_LIFETIME, HIGHEST→FIRST_POSSIBLE?

I am very bad at finding names. If we merge `HIGHEST` and  `FIRST_POSSIBLE` then `FIRST_POSSIBLE` is fine. `INCREASE_LIFETIME` is not wrong for `LOWEST`, but does not feel very intuitive either...

mglisse: > > * Should the doc mention intersect_lifetimes?
> 
> That is a good idea, as `intersect_lifetimes` can be not trivial for complex types, so the user should know.

I was mostly saying this for symmetry, since the other policy mentioned `unify_lifetimes`.

> I saw HIGHEST and LOWEST as two dual behavior, where one was pushing up values to fulfill the validity of a filtration and the other pushes down those values for the same purpose. But that is not really the case anymore. I don't mind at all to change the names.

Trying to come up with other names:
Filtration_maintenance::{LOWER_EXISTING, INCREASE_NEW, IGNORE}
In all cases, we insert the same simplices, we try to give the value `f` to the new simplices, and the question is how to maintain the property that we have a filtration. In one case we do that by lowering some existing filtration values, in another by increasing the value we give to some new simplices, and in the last case we don't try to maintain the filtration property (I was thinking of putting `BREAK` in the name, but that seems too active like we are going to replace existing filtration values with nonsense, instead of just doing nothing).

> > * FORCE: all new simplices get f, all old simplices remain the same, except that sigma is set to f. That's a bit awkward. Maybe that's useful, but I just want to make sure that it wouldn't be better to do just the new insertions (with value f), and let the caller do `set_filtration` afterwards if they really want to change the value of sigma.
> 
> I just had the case in mind, where the user uses some loop to insert and could potentially insert several times the same simplex (because some coface was inserted first for example, and then the rest of the iterations corrects the values of the faces). This would make sure that the last value would be stored. But storing the first only also makes sense, so I don't mind here.

I have a weak preference for what I described (don't update any existing filtration, use `f` for new simplices), but if a clearer use case comes up, I don't mind some other version.

> > Would it make sense to merge HIGHEST and FIRST_POSSIBLE, where the second would happen when we call insert without its second argument? Or would that complicate things too much?
> 
> Would that not destroy retro-compatibility? In the current main branch, calling `insert_simplex_and_subfaces(sig)` is possible and has the behaviour of `LOWEST` with a filtration value at 0/{}. Or would you not use the strategy as third argument, but as template argument? Then it would brake the syntax, no? (ie., you would need to do `insert_simplex_and_subfaces<>(sig)` instead of `insert_simplex_and_subfaces(sig)`)

If you give the template parameter a default value, it should be possible to call `f(x)`, not just `f<>(x)`.

I was thinking, with s=simplex, p=policy, f=filtration, i=insert:
i(s) and i(s,f) keep the current behavior
i(p,s,f) (or `i<p>(s,f)`, or with the arguments shuffled, that's not what I am discussing here) does what it says
i(p,s) is equivalent to i(p,s,f) with a value of f that depends on p, maybe: +inf for the policy that lowers existing filtrations, -inf for the one that increases new filtrations, ?? for the last policy (`Filtration_value()`? forbidden?).

> What we could do, is just remove `FIRST_POSSIBLE` and let the user use `HIGHEST` with a filtration value at -inf. `FIRST_POSSIBLE` is more of a commodity than a necessity if we keep `HIGHEST`.

If implemented separately, it may save quite a few calls to `intersect_lifetimes(*,-inf)`.

hschreiber: > Trying to come up with other names: Filtration_maintenance::{LOWER_EXISTING, INCREASE_NEW, IGNORE}

I am fine with those new names.

> I have a weak preference for what I described (don't update any existing filtration, use `f` for new simplices), but if a clearer use case comes up, I don't mind some other version.

I don't mind too much for now either, so I just changed it.

> I was thinking, with s=simplex, p=policy, f=filtration, i=insert: i(s) and i(s,f) keep the current behavior i(p,s,f) (or `i<p>(s,f)`, or with the arguments shuffled, that's not what I am discussing here) does what it says i(p,s) is equivalent to i(p,s,f) with a value of f that depends on p, maybe: +inf for the policy that lowers existing filtrations, -inf for the one that increases new filtrations, ?? for the last policy (`Filtration_value()`? forbidden?).

I tried to obtain the behavior you described here. Does it look fine to you?

> > What we could do, is just remove `FIRST_POSSIBLE` and let the user use `HIGHEST` with a filtration value at -inf. `FIRST_POSSIBLE` is more of a commodity than a necessity if we keep `HIGHEST`.
> 
> If implemented separately, it may save quite a few calls to `intersect_lifetimes(*,-inf)`.

I changed a few things to `_make_subfiltration_non_decreasing` such that it avoids unnecessary recursion and  `intersect_lifetimes` calls. But I don't see how the special case of -inf can avoid us more calls to `intersect_lifetimes` than the other cases, except for the first call in each loop (+additional calls if the first intersection is still -inf). Do you think, it is really worth doing a special implementation for it? Or did I missed a trick?

hschreiber: Arf, like many other words, `IGNORE` is a already reserved word (probably macro) for Windows... So, we cannot use it without weird compilation error messages. I changed it to `IGNORE_VALIDITY`. Is it still fine?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1114,69 +1123,251 @@ class Simplex_tree {
...
+     * @brief Let \f$ f \f$ be the filtration value given as argument. Inserts the simplex \f$ \sigma \f$ as follows:
+     * - if \f$ \sigma \f$ was not inserted yet, then \f$ \sigma \f$ and all its faces, which are not already included
+     * in the complex, are inserted at either \f$ f \f$ or at the first possible filtration value when \f$ f \f$ is too
+     * low (to insure the validity of the filtration). This is computed with @ref FiltrationValue::intersect_lifetimes.
+     * - if \f$ \sigma \f$ existed already, then nothing is done.
+     *
+     * So, this strategy can be interpreted as "insert the new simplices as soon as the current filtration allows,
+     * with a lower bound at \f$ f \f$". If no lower bound is desired, \f$ f \f$ should be set to -inf or not specified
+     * at all.
+     */
```
VincentRouvreau, line 1151: There is an issue in the documentation rendering, cf. this rendering (the same `\_form#272` at home):

Image

It seems to come from the double quotes `So, this strategy can be interpreted as " ... \f$ f \f$ ..."` that are not on the same line. I think your sentence still makes sense without the double quotes, or you can modify with, for example: `So, this strategy can be interpreted as: Insert the ... \f$ f \f$ ...`

mglisse:
I recently wished we had some kind of try_insert / insert_if_subfaces_are_present (actually a batch version of that in Python), a function that inserts only the simplex provided, and only if all of the faces are already present. In some sense, that would be similar to INCREASE_NEW, except that missing simplices are interpreted as having filtration +inf (not quite since we can have simplices present with infinite filtration, but that's still relevant for the intuition), so a missing face forces the filtration value of the new simplex to +inf, i.e. it cancels the insertion. It isn't directly related to this PR since it wouldn't be implemented in `insert_simplex_and_subfaces` (it most certainly does not insert subfaces), but close enough that I thought it would be worth mentioning here.

In case someone is curious, the use case is building a (weak, non-relaxed) witness complex, where the definition says that a simplex is in the complex if it has a witness and all its faces also have a witness.
1. each witness witnesses a vertex, we insert these vertices
2. each witness witnesses an edge, we insert these as well, but only if the 2 extremities are already inserted (i.e. the faces have witnesses)
3. each witness witnesses a triangle, we insert these as well, but only if the 3 edges are already inserted (and thus also the vertices, i.e. all the faces have witnesses)
4. etc.

(removing triangles from the list when an edge is missing from the previous list can probably be done in Python so we only need an unchecked insertion afterwards, but it is not as convenient, and possibly slower)

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1114,69 +1123,251 @@ class Simplex_tree {
...
+        case Filtration_maintenance::INCREASE_NEW:
+          return Filtration_simplex_base_real::get_minus_infinity();
+        case Filtration_maintenance::IGNORE_VALIDITY:
+          return Filtration_value();
+        default:
+          throw std::invalid_argument("Given insertion strategy is not available.");
+      }
+    };
+
+    return insert_simplex_and_subfaces(insertion_strategy, n_simplex, get_default_value(insertion_strategy));
```
VincentRouvreau, line 1239: A lambda is not required here.
```suggestion
    switch (insertion_strategy) {
      case Filtration_maintenance::LOWER_EXISTING:
        return insert_simplex_and_subfaces(insertion_strategy, n_simplex, Filtration_simplex_base_real::get_infinity());
      case Filtration_maintenance::INCREASE_NEW:
        return insert_simplex_and_subfaces(insertion_strategy, n_simplex, Filtration_simplex_base_real::get_minus_infinity());
      case Filtration_maintenance::IGNORE_VALIDITY:
        return insert_simplex_and_subfaces(insertion_strategy, n_simplex, Filtration_value());
      default:
        throw std::invalid_argument("Given insertion strategy is not available.");
    }
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1114,69 +1123,251 @@ class Simplex_tree {
...
+          return _rec_insert_simplex_and_subfaces_sorted(root(), copy.begin(), copy.end(), filtration);
+        case Filtration_maintenance::INCREASE_NEW:
+          return _insert_simplex_and_subfaces_at_highest(root(), copy.begin(), copy.end(), filtration);
+        case Filtration_maintenance::IGNORE_VALIDITY:
+          return _insert_simplex_and_subfaces_forcing_filtration_value(root(), copy.begin(), copy.end(), filtration);
+        default:
+          throw std::invalid_argument("Given insertion strategy is not available.");
+      }
+    } else {
+      return _rec_insert_simplex_and_subfaces_sorted(root(), copy.begin(), copy.end(), filtration);
```
VincentRouvreau, line 1298: `insertion_strategy` is ignored here ? Maybe just a comment

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1114,69 +1123,251 @@ class Simplex_tree {
...
+    INCREASE_NEW,
+    /**
+     * @brief If the simplex to insert:
+     * - already exists in the simplex tree, nothing is done,
+     * - does not exists yet, it and all its non-existing faces are inserted at the given filtration value and
+     * none of the already inserted faces are touched.
+     *
+     * So, in this case, the property of filtration is not actively maintained.
+     *
+     * This option is mainly usefull when the methods @ref FiltrationValue::intersect_lifetimes or
```
VincentRouvreau:
```suggestion
     * This option is mainly useful when the methods @ref FiltrationValue::intersect_lifetimes or
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -1114,69 +1123,251 @@ class Simplex_tree {
...
     // An alternative strategy would be:
     // - try to find the complete simplex, if found (and low filtration) exit
     // - insert all the vertices at once in sib
     // - loop over those (new or not) simplices, with a recursive call(++first, last)
     Vertex_handle vertex_one = *first;
 
-    if (++first == last) return insert_node_<true, false, true>(sib, vertex_one, filt);
+    // insert_node_<bool update_fil, bool update_children, bool set_to_null>(sib, ...)
+    // update_fil: if true, calls `unify_lifetimes` on the new and old filtration value of the node
+    // update_children: if true, assign a child to the node if he didn't had one
```
VincentRouvreau:
```suggestion
    // update_children: if true, assign a child to the node if it did not have one
```

VincentRouvreau (commented):
I also did a quick benchmark and no performance was lost with this new feature on legacy code.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: That is the case, when the filtration is not stored. I added a comment.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Yes, but I found it prettier with it, as `insert_simplex_and_subfaces(insertion_strategy, n_simplex, Filtration_simplex_base_real::get_minus_infinity());` is long and not as readable.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -171,6 +173,13 @@ class Simplex_tree {
     Filtration_value& filtration() { return filt_; }
 
     static const Filtration_value& get_infinity() { return inf_; }
+    static Filtration_value get_minus_infinity() {
+      if constexpr (std::numeric_limits<Filtration_value>::has_infinity) {
+        return -inf_;
+      } else {
+        return std::numeric_limits<Filtration_value>::lowest();
+      }
+    }
```
VincentRouvreau: Why not mirroring positive infinity code, something like:
```cpp
    static const Filtration_value& get_minus_infinity() { return minus_inf_; }
   private:
    inline static const Filtration_value minus_inf_ = std::numeric_limits<Filtration_value>::has_infinity
                                                    ? -std::numeric_limits<Filtration_value>::infinity()
                                                    : std::numeric_limits<Filtration_value>::lowest();
```
Maybe a matter of taste here.

src/Simplex_tree/concept/FiltrationValue.h
```diff
@@ -48,6 +49,18 @@ struct FiltrationValue {
...
+  * @brief Only necessary if `std::numeric_limits<FiltrationValue>::has_infinity` returns `true`.
+  * Negates the value. Only used on the infinity filtration value to obtain minus infinity.
+  */
+  friend FiltrationValue operator-(const FiltrationValue& f);
+
+  /**
+   * @brief Returns true if and only if the given filtration value is at infinity.
+   * Overloads for native arithmetic types are already implemented using `std::numeric_limits`.
+   */
+  friend bool is_positive_infinity(const FiltrationValue& f);
```
VincentRouvreau, line 63: I do not understand this concept.
This could be inside the Simplex_tree (maybe as static if required from outside), something like:
```cpp
 public:
  static const bool is_positive_infinity(const FiltrationValue& f) { return f == inf_; }
```
If it is for multi-filtration, this method could be expanded later, no ?

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Ok, maybe a matter of taste

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: It was just to avoid having to store a variable more, as we don't need a reference to it anywhere (for now?). But I am fine with this other version too. And if we end up needing a reference, it would already be ready.

src/Simplex_tree/concept/FiltrationValue.h
hschreiber: Strictly speaking, this should have been included in PR #1200 for generalized filtration values, but we missed it...

Having it inside the simplex tree is not possible for multi-filtration, because two filtration values could be infinity and still not be equal, because they don't have the same amount of parameters.

VincentRouvreau: Ok I got it thanks !

src/Simplex_tree/include/gudhi/Simplex_tree.h
VincentRouvreau: Done on [c77f1f6](https://github.com/GUDHI/gudhi-devel/pull/1176/commits/c77f1f636e0f4797a40b76ec5d535849e946d071)

PR #1182, opened 2025-03-07
Detect when user sets CXXFLAGS and propagate it correctly to setup.py

VincentRouvreau:
Could fix https://github.com/conda-forge/gudhi-feedstock/issues/71 (to investigate)


mglisse:
Did setuptools (or other) do all these manipulations before? (removing -I, -isystem, \n, deduplicate, etc) They seem rather risky.

PR #1188, opened 2025-03-14
Propagate CXXFLAGS correctly to setup.py

VincentRouvreau:
When CXXFLAGS environment variable is set, with setuptools >= 74.0.0, python compilation flags are removed.
Take the ones coming from CFLAGS (as setuptools is doing before overwriting them with the CXXFLAGS environment variable value).
It also adds `CMAKE_CXX_FLAGS_RELEASE` | `CMAKE_CXX_FLAGS_DEBUG` to setup.py `extra_compile_args` in order to add '-O3' in release mode and speed up computation.

Could fix https://github.com/conda-forge/gudhi-feedstock/issues/71 (to investigate)


src/python/CMakeLists.txt
```diff
@@ -363,6 +363,17 @@ if(DEBUG_TRACES)
   set(GUDHI_PYTHON_EXTRA_COMPILE_ARGS "${GUDHI_PYTHON_EXTRA_COMPILE_ARGS}'-DDEBUG_TRACES', ")
 endif(DEBUG_TRACES)
 
+if(CMAKE_BUILD_TYPE MATCHES Debug)
+  set(GUDHI_COMPILATION_FLAGS_STRING ${CMAKE_CXX_FLAGS_DEBUG})
+else()
+  set(GUDHI_COMPILATION_FLAGS_STRING ${CMAKE_CXX_FLAGS_RELEASE})
+endif()
+# Compilation flags are in a string. Transform it to a list and append them to GUDHI_PYTHON_EXTRA_COMPILE_ARGS
+string(REPLACE " " ";" GUDHI_COMPILATION_FLAGS_LIST ${GUDHI_COMPILATION_FLAGS_STRING})
```
mglisse, line 372: I assume it breaks if the flags contain `-I/some/dir with/a space/`? It is probably already broken elsewhere in our build stuff though.

mglisse (approved):

PR #1190, opened 2025-03-18
Simple_mdspan

hschreiber:
As we would like to flatten some `vector<vector>` or even `vector<vector<vector>>>` in the future multi persistence module, for a better communication with Python, I recreated the `std::mdspan` class which comes for C++23 such that it works with C++17. I tried to keep the class as close as possible to the original. The idea is to replace this class with the standard `std::mdspan` with a minimum of effort once Gudhi changes the minimum compiler version standard to 23.

### DIFF
--- a/src/common/include/gudhi/simple_mdspan.h
+++ b/src/common/include/gudhi/simple_mdspan.h
@@ -1,0 +1,277 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+ *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+ *    Author(s):       Hannah Schreiber, David Loiseaux
+ *
+ *    Copyright (C) 2025 Inria
+ *
+ *    Modification(s):
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+/**
+ * @private
+ * @file simple_mdspan.h
+ * @author Hannah Schreiber, David Loiseaux
+ */
+
+#ifndef GUDHI_SIMPLE_MDSPAN_H_
+#define GUDHI_SIMPLE_MDSPAN_H_
+
+#include <cstddef>      // std::size_t
+#include <type_traits>  // std::remove_cv_t
+#include <limits>
+#include <initializer_list>
+#include <vector>
+
+#include <gudhi/Debug_utils.h>
+
+namespace Gudhi {
+
+/**
+ * @private
+ * @brief Reproduces the behaviour of C++23 `std::layout_right` class.
+ */
+class layout_right
+{
+ public:
+  class mapping
+  {
+   public:
+    using index_type = std::size_t;
+    using extents_type = std::vector<index_type>;
+    using size_type = typename extents_type::size_type;
+    using rank_type = std::size_t;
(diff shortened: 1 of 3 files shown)


mglisse:
> In particular, because some compiler optimizations of the original are just not possible with C++17.

Could you clarify what you are talking about here?

hschreiber:
> Mainly I meant, that `std::mdspan` is able to do a lot at compile time, even things concerning the dynamic extents. But now, that I rethink about it, I have a slight doubt about the non feasibility. In particular, as I just realized that `sizeof...` exists in C++11 already and is compile time. I will retry to do a few things.

So, I added an `extents` and made all the compile time things work. I missed the possibility at first, as I was convinced that you couldn't initialize the variadic template parameters with help of the variadic run time arguments.

But in the end, what took me most of the time was to implement the deduction guidelines for `Simple_mdspan`...

Only thing left is to see if everything compiles well also on macos and Windows.

mglisse:
IIUC, if we want to replace it with std::mdspan in the future, we will have 2 things to do (in addition to the obvious change in the class name and includes)
1. replace `s(i,j)` with `s[i,j]`
2. remove the calls to `update_extent` and `update_data`

Is that all?

The 2. part looks problematic. Does it mean that we will lose performance by switching to std::mdspan?

hschreiber:
> IIUC, if we want to replace it with std::mdspan in the future, we will have 2 things to do (in addition to the obvious change in the class name and includes)
> 
>     1. replace `s(i,j)` with `s[i,j]`
> 
>     2. remove the calls to `update_extent` and `update_data`
> 
> 
> Is that all?

Yes, I think so.

> The 2. part looks problematic. Does it mean that we will lose performance by switching to std::mdspan?

Not that much, as we can give our own mapping class to `std::mdspan`. We can just replace `view.update_extent()` with `view.mapping.update_extent()`. 
For now, I only needs those two methods for multi filtrations values, as only rank 0 changes potentially and updating it is constant time which is much better than reconstructing even a small array at each request to a value in the filtration value (as those can be quite frequent).

mglisse:
> Not that much, as we can give our own mapping class to `std::mdspan`. We can just replace `view.update_extent()` with `view.mapping.update_extent()`.

Ah, good :smiley: 
Could you sketch out somewhere (maybe as a comment in the file, or in an issue, some place we have a chance of finding it) what that would look like? It doesn't have to be complete and compile, but enough for someone not familiar with mdspan to be able to do it quickly in a few years (let's take advantage of the fact that you are currently very familiar with the design).

src/common/include/gudhi/simple_mdspan.h
```diff
@@ -0,0 +1,484 @@
...
+
+inline constexpr std::size_t dynamic_extent = std::numeric_limits<std::size_t>::max();
+
+template <class IndexType, std::size_t... Extents>
+class extents;
+
+namespace detail {
+
+  template <std::size_t v>
+  struct is_dynamic : std::integral_constant<std::size_t, 0> {};
```
mglisse, line 40: Why not `std::false_type`?

mglisse (approved):
Ok, I'm not going to review the details, since the interface follows std::mdspan it should be fine.

PR #1191, opened 2025-03-18
Multi parameter filtrations

hschreiber:
Extracts the `Multi_filtration` module from PR #976 as quite some changes were done since. So, PR #976 will only contain the addition of the `Multi_persistence` module.

Only a draft for now, as the main page for the documentation is still missing, as well as examples. But feedback could already be useful (in particular, @DavidLapous you wanted another name for the degree Rips filtration class?). Suggestions for examples are also welcome! 

**Edit 19/06/26:**
PR dependencies:
- #1190
- #1342


src/Multi_filtration/include/gudhi/Degree_rips_bifiltration.h
```diff
@@ -0,0 +1,2311 @@
...
+        if (a.num_generators() == 0) return true;
+        if (b.num_generators() == 0) return false;
+        if (a.generators_[0] < b.generators_[0]) return true;
+        if (b.generators_[0] < a.generators_[0]) return false;
+        return a.num_generators() < b.num_generators();
+      }
+    }
+
+    for (std::size_t i = 0U; i < std::min(a.num_generators(), b.num_generators()); ++i) {
+      if constexpr (inverse) i = std::min(a.num_generators(), b.num_generators()) - 1 - i;
```
VincentRouvreau: Maybe a function that does the inversion would help with the readability here, something like:
```cpp
    auto min_num_generators = std::min(a_num_generators(), b.num_generators());
    for (std::size_t i = 0U; i < min_num_generators; ++i) {
      if constexpr (inverse) i = invert_i(min_num_generators, i);
      ...
      // Like that we see clearly `i` is inverted twice
      if constexpr (inverse) i = invert_i(min_num_generators, i);
    }
```

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -3078,13 +3104,38 @@ class Simplex_tree {
 // Print a Simplex_tree in os.
 template<typename...T>
 std::ostream& operator<<(std::ostream & os, const Simplex_tree<T...> & st) {
+  using handle = typename Simplex_tree<T...>::Simplex_handle;
+  // lexicographical order to ensure total order even with custom filtration values
+  st.initialize_filtration(
+      [&](handle sh1, handle sh2) {
```
mglisse: We already have reverse_lexicographic_order, do we need a second one?

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -3078,13 +3104,38 @@ class Simplex_tree {
 // Print a Simplex_tree in os.
 template<typename...T>
 std::ostream& operator<<(std::ostream & os, const Simplex_tree<T...> & st) {
+  using handle = typename Simplex_tree<T...>::Simplex_handle;
+  // lexicographical order to ensure total order even with custom filtration values
+  st.initialize_filtration(
```
mglisse: * It might be simpler to iterate on `complex_simplex_range` instead of `filtration_simplex_range` so we don't need to sort. Or use `for_each_simplex` for a different order.
* What happens if a user calls `filtration_simplex_range` directly? Are there some functions where we should assert that num_parameters is 1? Is that actually the right condition, since we could have a single parameter of a non-totally ordered type?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: `reverse_lexicographic_order` is private. I either have to make it public or to make `operator<<` a friend to the `Simplex_tree` class. Making `operator<<` a friend seems the cleaner version for me. What do you think?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: > * It might be simpler to iterate on `complex_simplex_range` instead of `filtration_simplex_range` so we don't need to sort. Or use `for_each_simplex` for a different order.

Using `complex_simplex_range` was also my first thought, but this is not compatible with the dual method `operator>>` which needs a face to come before its cofaces.

> * What happens if a user calls `filtration_simplex_range` directly? Are there some functions where we should assert that num_parameters is 1? Is that actually the right condition, since we could have a single parameter of a non-totally ordered type?

If `filtration_simplex_range` is called directly when the filtration value has no `operator<` which is total, the behavior is undefined. If `filtration_vect_` was already filled way before correctly, then it will be fine, but if it needs to be constructed, `operator<` will not fill the requirements for the internal sort method (`tbb::parallel_sort`/`std::stable_sort`).

Using `num_parameter == 1` as a test should work most of the time, but it is not impossible that someone has a scenario where `num_parameter != 1`, but `operator<` is still implemented such that it induces a total ordering. The opposite situation, where `num_parameter == 1`, but the order of `<` is not total, seems much more unlikely to me, but who knows? 
The good news is that this point seems to only matter for the filtration initialization.
The cleanest way to make the call to `filtration_simplex_range` safe, would be to find a sort algorithm/condition which works whatever `operator<` induces a total order or not. Something which uses `intersect_lifetimes` or `unify_lifetimes` instead ?

In the meantime, I will update the documentation of `filtration_simplex_range`.

hschreiber: I replaced `filtration_simplex_range` with `for_each_simplex` for `operator<<` and the method is much simpler that way!
Now, it being `friend` is not that useful anymore, but as it still simplifies the syntax, I would still keep it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Making << a hidden friend is generally a good thing, whether it is needed for other reasons or not.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: > If `filtration_simplex_range` is called directly when the filtration value has no `operator<` which is total, the behavior is undefined.

Some form of assertion would be nice then.

> Using `num_parameter == 1` as a test should work most of the time, but it is not impossible that someone has a scenario where `num_parameter != 1`, but `operator<` is still implemented such that it induces a total ordering. The opposite situation, where `num_parameter == 1`, but the order of `<` is not total, seems much more unlikely to me, but who knows?

SimplexTreeOptions/FiltrationValue could say whether the order is total (or provide a separate total order), if that information is needed.

> The good news is that this point seems to only matter for the filtration initialization. The cleanest way to make the call to `filtration_simplex_range` safe, would be to find a sort algorithm/condition which works whatever `operator<` induces a total order or not.

I think those are costly, can you do better than testing all pairs to build a graph then topologically sort it?
As long as we do not have a clear use case in mind for such functionality, I'd rather disable stuff (initialize_filtration with the default order?) for the non-totally-ordered case.

> Something which uses `intersect_lifetimes` or `unify_lifetimes` instead ?

Those invent new values, that's not really sorting.

src/Multi_filtration/example/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_executable_with_targets(Multi_filtration_example_multi_simplex_tree example_multi_simplex_tree.cpp TBB::tbb)
```
mglisse, line 1: If we don't sort anymore, we might not need TBB.

src/Multi_filtration/example/example_multi_simplex_tree.cpp
```diff
@@ -0,0 +1,49 @@
...
+  /*    1              */
+  /*    o              */
+  /*   /X\             */
+  /*  o---o---o        */
+  /*  2   0   3        */
+
+  auto triangle012 = {0, 1, 2};
+  auto edge03 = {0, 3};
+  // Inserts the triangle with multi filtration value {1,2,3}
+  st.insert_simplex_and_subfaces(triangle012, Multi_filtration_value({1, 2, 3}));
```
mglisse:
```suggestion
  st.insert_simplex_and_subfaces(triangle012, Multi_filtration_value({1., 2., 3.}));
```
Making it visually obvious what is an index and what is a filtration might make the example slightly easier to read.

src/Multi_filtration/doc/COPYRIGHT
mglisse, line 1: Uh, we randomly have one of those in half of the subdirectories? Strange...

src/Multi_filtration/example/example_multi_simplex_tree.cpp
```diff
@@ -0,0 +1,49 @@
...
+  // The filtration value can also be given as an initializer list
+  st.insert_simplex_and_subfaces(edge03, std::initializer_list<double>{4, 5, 6});
```
mglisse: Does this work?
```suggestion
  st.insert_simplex_and_subfaces(edge03, {4., 5., 6.});
```
`initializer_list` is not something users are meant to type.

src/Multi_filtration/example/example_multi_simplex_tree.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+  auto edge02 = {0, 2};
+  ST::Simplex_handle e = st.find(edge02);
+  // Multi_filtration_value has an operator<<
+  std::cout << st.filtration(e) << std::endl;
+  GUDHI_CHECK(st.filtration(st.find(edge03)) == ST::Filtration_value({4, 5, 6}),
```
mglisse: Do we use GUDHI_CHECK in existing examples? I think it is more for internal stuff, while examples are what the user reads, which could use assert. Now if something is both an example and a test...

src/Multi_filtration/example/CMakeLists.txt
hschreiber: Even if the example does not use any method with TBB, the file won't compile without the link (or perhaps I missed the place where it is used).

src/Multi_filtration/example/example_multi_simplex_tree.cpp
hschreiber: To do so, I have to add a constructor in the filtration classes with `std::initializer_list<T>` as explicit argument (instead of just template `<class ValueRange = std::initializer_list<T>>`). Which is fine, but do you know if there would have been another solution ?

For info, the compiler error otherwise is a `no matching function for call to ...` error because `cannot convert ‘{4.0e+0, 5.0e+0, 6.0e+0}’ (type ‘<brace-enclosed initializer list>’) to type ‘const Gudhi::Simplex_tree<...>::Filtration_value&’` even though it has no problem compiling `Filtration_value({4., 5., 6.})`. So it is probably a double deduction problem.

src/Multi_filtration/example/example_multi_simplex_tree.cpp
hschreiber: We use `assert` in other examples, I guess I replaced it by reflex (this example is more or less just a copy of the `src/Simplex_tree/example/mini_simplex_tree.cpp` example). `assert` is probably less confusing than `GUDHI_CHECK` for an external user, so I will change it.

The examples are used in the test suite. At least they are called by ctest.

src/Multi_filtration/doc/COPYRIGHT
hschreiber: I just copied the one in the Simplex tree, I have no clue if they are supposed to be there or not... 

@VincentRouvreau This file is indeed not in all modules. Is it supposed to be in all or none, or just in particular ones? If the latter, what is the critieria (e.g., does it have to be in this one)?

src/Multi_filtration/example/CMakeLists.txt
mglisse: Ah, right, our cmake scripts add -DGUDHI_USE_TBB unconditionally for every file if TBB is found during the configuration. Ok to keep TBB::tbb then, cleaning this stuff is not for this PR.

src/Multi_filtration/example/example_multi_simplex_tree.cpp
mglisse: Since we can specify the simplex as just `{1,2,3}`, I guess we could do the same for filtration values, i.e. have insert_simplex_and_subfaces take an argument of template type FiltrationValue (defaulting to initializer_list) instead of the fixed Filtration_value. Although it might get messy now that there are plenty of overloads of this function, and I didn't check if that does the right thing for single-parameter persistence. What you propose makes sense to me, if the person who provides a multi-filtration type wants to be able to construct them conveniently, it is up to them to provide the appropriate constructor, and we can do it for the types we provide.
(I didn't think about it long, there may be better solutions)

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,422 @@
...
+  using const_iterator = unspecified;
+  /**
+   * @brief Returns the type of @ref StoragePolicy if @ref value_type is equal to the template parameter `U`.
+   */
+  template <typename U>
+  using As_type = unspecified;
+
+  /**
+   * @brief Value considered +infinity for an element in the filtration value.
+   * Carefull: `-T_inf` does **not** have to be equal to `T_m_inf`.
```
VincentRouvreau: ~~Carefull~~ -> Careful
You could also use `@note`,  `@remark` ( `@warning` is displayed in red, so maybe to strong)

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,422 @@
...
+   *
+   * @tparam U Type of the element. Default value: @ref value_type.
+   */
+  template <typename U = value_type>
+  constexpr static const U T_inf;
+
+  /**
+   * @brief Value considered -infinity for an element in the filtration value.
+   * Carefull: `-T_m_inf` does **not** have to be equal to `T_inf`.
```
VincentRouvreau: ~~Carefull~~ -> Careful
You could also use `@note`,  `@remark` ( `@warning` is displayed in red, so maybe to strong)

src/Multi_filtration/include/gudhi/Multi_filtration/multi_filtration_conversions.h
```diff
@@ -0,0 +1,308 @@
...
+namespace Gudhi {
+
+namespace multi_filtration {
+
+/**
+ * @ingroup multi_filtration
+ *
+ * @brief Converts the given @ref Flat_array_filtration into the @ref StoragePolicy given as template argument.
+ * It is assumed that the given value is simplified (i.e. minimal and ordered lexicographically). It is also assumed
+ * that the given value has the right format for the choosen output type. E.g., if `OutStoragePolicy` is
```
VincentRouvreau: ~~choosen~~ -> chosen

src/Multi_filtration/include/gudhi/Multi_filtration/multi_filtration_conversions.h
```diff
@@ -0,0 +1,308 @@
...
+    throw std::invalid_argument("Given out multi filtration value is not available.");
+  }
+}
+
+/**
+ * @ingroup multi_filtration
+ *
+ * @brief Converts the given @ref Nested_array_filtration into the @ref StoragePolicy given as template argument.
+ * It is assumed that the given value is simplified (i.e. minimal and ordered lexicographically). It is also assumed
+ * that the given value has the right format for the choosen output type. E.g., if `OutStoragePolicy` is
```
VincentRouvreau: ~~choosen~~ -> chosen

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+   */
+  [[nodiscard]] bool is_nan() const { return generators_.is_nan(); }
+
+  /**
+   * @brief Returns `true` if and only if the filtration value is not considered as plus infinity,
+   * minus infinity or NaN.
+   */
+  [[nodiscard]] bool is_finite() const { return generators_.template is_finite<Co>(); }
+
+  // COMPARAISON OPERATORS
```
VincentRouvreau:
```suggestion
  // COMPARISON OPERATORS
```

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+
+  /**
+   * @brief Adds the given generator to the filtration value such that the set remains sorted.
+   * If @ref StoragePolicy::has_minimal_set_representation is true, it also ensures that it remains minimal. In that
+   * case, it is therefore possible that the generator is ignored if it does not generated any new lifetime or that
+   * old generators disappear if they are overshadowed by the new one.
+   *
+   * @tparam Iterator Iterator class satisfying the requirements of the standard `LegacyForwardIterator`.
+   * The dereferenced type has to be convertible to `T`.
+   * @param genStart Iterator pointing to the begining of the range.
```
VincentRouvreau, line 1425:
```suggestion
   * @param genStart Iterator pointing to the beginning of the range.
```

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+   *
+   * Fails to compile if `Ensure1Criticality` is true.
+   *
+   * @warning If @ref StoragePolicy::has_minimal_set_representation is true and the resulting set of generators is
+   * not minimal or sorted after this modification, some methods will have an undefined behaviour. Be sure to
+   * call @ref simplify() before using them. Same if the internal order is not preserved.
+   *
+   * @tparam Iterator Iterator class satisfying the requirements of the standard `LegacyForwardIterator`.
+   * The dereferenced type has to be convertible to `T`.
+   * @param genStart Iterator pointing to the begining of the range.
```
VincentRouvreau, line 1495:
```suggestion
   * @param genStart Iterator pointing to the beginning of the range.
```

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+      generators_.emplace_back(genStart, genEnd);
+    } else {
+      generators_.template emplace<Co>(genStart, genEnd, _get_default_null_value());
+    }
+  }
+
+  /**
+   * @brief If @ref StoragePolicy::has_minimal_set_representation is true, simplifies the current set of generators
+   * such that it becomes minimal. Also orders it in increasing lexicographical order. Only necessary if generators
+   * were added "by hand" without verification either trough the constructor or with @ref add_guaranteed_generator "",
```
VincentRouvreau:
```suggestion
   * were added "by hand" without verification either through the constructor or with @ref add_guaranteed_generator "",
```

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+    if (!coordinate && num_generators() > 1) {
+      if constexpr (StoragePolicy::has_minimal_set_representation)
+        simplify();
+      else
+        generators_.sort();
+    }
+#endif
+  }
+
+  // FONCTIONNALITIES
```
VincentRouvreau:
```suggestion
  // FUNCTIONALITIES
```

VincentRouvreau (commented):
And thanks for the refactoring !

src/Multi_filtration/doc/Intro_multi_filtration_values.h
```diff
@@ -0,0 +1,36 @@
...
+ * \author    Hannah Schreiber, David Loiseaux
+ *
+ * \section multifiltrationintro Multi-parameter Filtration Values
+ *
+ * TODO: introduction 
+ * 
+ * \subsection multifiltrationexamples Examples
+ * 
+ * Here is a list of examples using the module:
+ * \li TODO?.
```
VincentRouvreau, line 29: This TODO needs to be fixed before merge.
Do you need the help of @DavidLapous ?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+    if constexpr (Ensure1Criticality) {
+      if constexpr (Co) {
+        return f1.pull_to_greatest_common_lower_bound(f2);
+      } else {
+        return f1.push_to_least_common_upper_bound(f2);
+      }
+    } else if constexpr (StoragePolicy::has_an_implicit_axis) {
+      bool modified = false;
+      for (size_type p = 0; p < f1.num_parameters(); ++p) {
+        // TODO: assumes that f1(*, implicit_axis) == f2(*, implicit_axis), it should throw if it is not the case
```
VincentRouvreau: Maybe we could add a `GUDHI_CHECK(...)` here to be sure tests are not falling in this case in debug mode ?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2596 @@
...
+    if (a.is_nan() || b.is_nan()) return false;
+    if (&a == &b) return !strict;
+
+    if constexpr (StoragePolicy::has_an_implicit_axis) {
+      // works because values at implicit axis are unique and the generators are ordered
+      // by increasing value at that axis
+      GUDHI_CHECK(StoragePolicy::implicit_axis() < b.num_parameters(),
+                  std::runtime_error("Implicit axis out of bound??"));
+      for (size_type p = 0; p < b.num_parameters(); ++p) {
+        // TODO: if a(g, implicit_axis) != b(g, implicit_axis), the `if` probably has to be removed
```
VincentRouvreau: It seems to me this comment was not updated regarding the refactorization, no ?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
VincentRouvreau: Maybe we could add a `GUDHI_CHECK(...)` here to be sure tests are not falling in this case in debug mode ?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: This comment was added because of the factorization. `a(g, implicit_axis) != b(g, implicit_axis)` can never happen with the classes we have right now, but now that it was generalized, it is possible that it can happen with other user classes (or if I add some shifting arguments to `Degree_bifiltration` for example).

I wanted for now the factorization to work. But before merging, I should take care of that case, yes.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: Same as just above.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: This and the point below should be done now. I used a bit naive way of solving the more general problem, there is possibly a more efficient algorithm, but I don't want to lose too much time on it.

(I also added some shift/step argument to `Degree_bifiltration` to test the possibility)

src/common/include/gudhi/serialization_utils.h
```diff
@@ -0,0 +1,175 @@
...
+inline std::size_t get_serialization_size_of([[maybe_unused]] T value) {
+  return sizeof(T);
+}
+
+template <typename T>
+inline char *serialize_value_to_char_buffer(const std::vector<T> &range, char *start) {
+  const std::size_t sizeSize = sizeof(std::size_t);
+  const std::size_t length = range.size();
+  char *curr = start;
+  memcpy(curr, &length, sizeSize);
```
mglisse, line 119: Why not `char *curr = serialize_value_to_char_buffer(start, length);`?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+namespace multi_filtration {
+
+/**
+ * @ingroup multi_filtration
+ *
+ * @brief Concept for the first template parameter of @ref Multi_parameter_filtration_value.
+ * The internal order of generators and parameters has to be fixed, whatever order it is,
+ * i.e. if two instances of @ref StoragePolicy are equal, the generators and parameters are ordered the same for both.
+ */
+class StoragePolicy {
```
mglisse, line 29: I think so far we have mostly not used namespaces for concepts, and tried to have explicit names, like SimplicialComplexForWitness instead of Witness::SimplicialComplex.
The name of concepts only matters for documentation purposes. In this case, what happens if we have a separate Simplex_tree::StoragePolicy, will the documentation work and look ok?
If you think everything will be fine, I don't mind if the name is a bit generic. Besides, concept names can be changed quite easily without breaking user code.

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+ public:
+  /**
+   * @brief Type of an element of the filtration value at a parameter (in a generator).
+   */
+  using value_type = unspecified;
+  /**
+   * @brief Type of the underlying element container.
+   */
+  using Underlying_container = unspecified;
```
mglisse, line 38: Can't we derive a lot of the other types from this one?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+  /**
+   * @brief Iterator type for a generator of the filtration value. Has to be at least LegacyForwardIterator.
+   */
+  using iterator = unspecified;
+  /**
+   * @brief Const iterator type for a generator of the filtration value. Has to be at least LegacyForwardIterator.
+   */
+  using const_iterator = unspecified;
+  /**
+   * @brief Returns the type of @ref StoragePolicy if @ref value_type is equal to the template parameter `U`.
```
mglisse, line 61: Unclear sentence.

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   * the same value for this parameter.
+   */
+  constexpr static const bool has_an_implicit_axis;
+  /**
+   * @brief True if and only if @ref sort sorts lexicographically.
+   */
+  constexpr static const bool has_lexicographical_storage;
+  /**
+   * @brief True if and only if the storage strategy allows to simplify the set of generators such that it becomes
+   * minimal.
```
mglisse, line 99: What does "allow" mean in this context?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   */
+  constexpr static const bool has_quiet_NaN;
+
+  /**
+   * @brief Constructs an empty container.
+   */
+  StoragePolicy();
+
+  /**
+   * @brief Constructs a single generator with given number of parameters and all values at given value.
```
mglisse, line 123: I presume defining "generator" is part of the TODO in the introduction?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   * @param numberOfParameters Number of parameters.
+   */
+  template <class Iterator, class = std::enable_if_t<!std::is_arithmetic_v<Iterator> > >
+  StoragePolicy(Iterator itBegin, Iterator itEnd, size_type numberOfParameters);
+
+  /**
+   * @brief Builds filtration value with given number of parameters and values copied from the given
+   * @ref Underlying_container container.
+   */
+  StoragePolicy(const Underlying_container &generators, size_type numberOfParameters);
```
mglisse, line 155: Does it have an advantage compared to taking as parameter an arbitrary range? Or maybe ranges are just too inconvenient before C++20?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+  /**
+   * @brief Builds filtration value with given number of parameters and values moved from the given
+   * @ref Underlying_container container.
+   */
+  StoragePolicy(Underlying_container &&generators, size_type numberOfParameters);
+
+  /**
+   * @brief Copy constructor.
+   */
+  StoragePolicy(const StoragePolicy &other);
```
mglisse, line 166: Listing all the standard methods is too verbose. Can't we say that the type is a regular type, or copyable or something, and be done with it?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   * the \f$ p \f$ next elements to the second generator and so on... So the length of the range has to be a multiple
+   * of \f$ p \f$ and the number of generators will be \f$ length / p \f$. The range is represented by two iterators.
+   *
+   * @tparam Iterator LegacyForwardIterator.
+   * @param itBegin Begin iterator of the range.
+   * @param itEnd End iterator of the range.
+   * @param numberOfParameters Number of parameters.
+   */
+  template <class Iterator, class = std::enable_if_t<!std::is_arithmetic_v<Iterator> > >
+  StoragePolicy(Iterator itBegin, Iterator itEnd, size_type numberOfParameters);
```
mglisse, line 149: (comment in a random place)
This doesn't feel like a policy, it feels like a container, the name sounds misleading to me.

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+  /**
+   * @brief Returns the serialization size of the given filtration value.
+   */
+  friend std::size_t get_serialization_size_of(const StoragePolicy &value);
+};
+
+/**
+ * @brief Only necessary if @ref Multi_parameter_filtration_value::as_type has to be used. Returns a copy of the given
+ * @ref StoragePolicy but in the format of another class (or same with different template parameters) respecting the
+ * concept. Has to exist in the `Gudhi::multi_filtration` namespace.
```
mglisse, line 431: ADL is not enough to find it elsewhere?

src/Multi_filtration/include/gudhi/Multi_filtration/Flat_array_filtration.h
```diff
@@ -0,0 +1,373 @@
...
+ * and throws otherwise.
+ * 
+ * @tparam T Arithmetic type of an entry for one parameter of a filtration value. Has to be **signed** and
+ * to implement `std::isnan(T)`, `std::numeric_limits<T>::has_quiet_NaN`, `std::numeric_limits<T>::quiet_NaN()`,
+ * `std::numeric_limits<T>::has_infinity`, `std::numeric_limits<T>::infinity()` and `std::numeric_limits<T>::max()`.
+ * If `std::numeric_limits<T>::has_infinity` returns `false`, a call to `std::numeric_limits<T>::infinity()`
+ * can simply throw. Examples are the native types `double`, `float` and `int`.
+ */
+template <typename T>
+class Flat_array_filtration {
```
mglisse, line 65: I wonder if using array for vector may be a bit confusing.

mglisse (commented):
First batch.

src/Multi_filtration/include/gudhi/Multi_filtration/Degree_bifiltration.h
```diff
@@ -0,0 +1,477 @@
...
+ */
+
+/**
+ * @file Degree_bifiltration.h
+ * @author Hannah Schreiber, David Loiseaux
+ * @brief Contains the @ref Gudhi::multi_filtration::Degree_bifiltration class.
+ */
+
+#ifndef MF_DEGREE_BIFILTRATION_H_
+#define MF_DEGREE_BIFILTRATION_H_
```
mglisse, line 18: This might not be an official policy in Gudhi, but I really think all header guards (and other macros) should start with `GUDHI_`, in particular for new files.

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+ * i.e. if two instances of @ref StoragePolicy are equal, the generators and parameters are ordered the same for both.
+ */
+class StoragePolicy {
+ public:
+  /**
+   * @brief Type of an element of the filtration value at a parameter (in a generator).
+   */
+  using value_type = unspecified;
+  /**
+   * @brief Type of the underlying element container.
```
mglisse, line 36: Are there any requirements on what a container means? Is it `[container.reqmts]` from the standard? Are there operations that Multi_parameter_filtration_value (or other code further down the chain) does directly on the container without going through the wrappers in StoragePolicy?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   * @brief Type of the container size.
+   */
+  using size_type = unspecified;
+  /**
+   * @brief Reference type for an element in the filtration value.
+   */
+  using reference = unspecified;
+  /**
+   * @brief Const reference type for an element in the filtration value.
+   * Does not really have to be a r-value (e.g., can just be the same than `value_type`, that is a copy).
```
mglisse, line 49: The use of "r-value" does not make sense here.

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   */
+  using const_reference = unspecified;
+  /**
+   * @brief Iterator type for a generator of the filtration value. Has to be at least LegacyForwardIterator.
```
mglisse, line 53: "for a generator of the filtration value" is not super clear, maybe explain that it is meant to iterate on the values (coordinates?) of a single generator.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2757 @@
...
+
+  static constexpr Filtration_value max() noexcept(false) {
+    throw std::logic_error(
+        "The max value cannot be represented with no finite numbers of parameters."
+        "Use `max(numberOfParameters)` instead");
+  };
+
+  static constexpr Filtration_value max(std::size_t p) {
+    if constexpr (has_infinity) {
+      return Filtration_value(p, std::numeric_limits<T>::max());
```
mglisse, line 2738: `max` is supposed to be the largest finite value. Seeing it below a test for `has_infinity` is suspicious.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2757 @@
...
+    return true;
+  }
+};
+
+}  // namespace Gudhi::multi_filtration
+
+namespace std {
+
+template <class StoragePolicy, bool Co, bool Ensure1Criticality>
+class numeric_limits<Gudhi::multi_filtration::Multi_parameter_filtration_value<StoragePolicy, Co, Ensure1Criticality>> {
```
mglisse, line 2716: Just leaving a note for myself for now: adding non-standard things to `numeric_limits<our_type>` sounds questionable, I'll have to look at how it is used.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
mglisse, line 1: If I want to represent a 1-critical bifiltration, which should be rather common, do I have any efficient option, something similar to `array<double,2>`?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2757 @@
...
+    throw std::logic_error(
+        "The max value cannot be represented with no finite numbers of parameters."
+        "Use `max(numberOfParameters)` instead");
+  };
+
+  static constexpr Filtration_value max(std::size_t p) {
+    if constexpr (has_infinity) {
+      return Filtration_value(p, std::numeric_limits<T>::max());
+    } else {
+      throw std::logic_error("No biggest value possible for Co-filtrations yet.");
```
mglisse, line 2740: What are co-filtrations? Are they the only case that doesn't have has_infinity?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+   */
+  using const_iterator = unspecified;
+  /**
+   * @brief Returns the type of @ref StoragePolicy if @ref value_type is equal to the template parameter `U`.
+   */
+  template <typename U>
+  using As_type = unspecified;
+
+  /**
+   * @brief Value considered +infinity for an element in the filtration value.
```
mglisse, line 67: Do we expect different storage to have a different notion of infinity?

src/Multi_filtration/concept/StoragePolicy.h
```diff
@@ -0,0 +1,443 @@
...
+
+  /**
+   * @brief True if and only if exactly one of the parameter values is just implicitly "stored" and is (therefore)
+   * fixed/not updatable. See for example @ref Gudhi::multi_filtration::Degree_bifiltration.
+   * For the comparaison operators `>`, `>=`, `<=` and `<` of @ref Multi_parameter_filtration_value to properly work
+   * if needed, additional constrains are: the implicit values are integers (even if @ref value_type is not), the
+   * generators are internally always ordered by increasing implicit values and no two distinct generators can have
+   * the same value for this parameter.
+   */
+  constexpr static const bool has_an_implicit_axis;
```
mglisse, line 92: The values are implicit in terms of storage, but appear explicitly in the various accessors (iterators for example), right?

mglisse (commented):
Second batch.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
```diff
@@ -0,0 +1,2757 @@
...
+
+/**
+ * @class Multi_parameter_filtration_value Multi_parameter_filtration_value.h gudhi/Multi_parameter_filtration_value.h
+ * @ingroup multi_filtration
+ *
+ * @brief Class encoding the different generators, i.e., apparition times, of a \f$ k \f$-critical
+ * \f$\mathbb R^n\f$-filtration value. E.g., the filtration value of a simplex, or, of the algebraic generator of a
+ * module presentation.
+ * Implements the concept @ref FiltrationValue of the @ref Gudhi::Simplex_tree and the concept
+ * @ref Gudhi::multi_persistence::MultiFiltrationValue.
```
VincentRouvreau, line 67: `@ref Gudhi::multi_persistence::MultiFiltrationValue` does not exist (doxygen warning)

src/Multi_filtration/include/gudhi/Multi_filtration/Degree_bifiltration.h
```diff
@@ -0,0 +1,477 @@
...
+ * - `::infinity(int)` returns `Degree_bifiltration::inf(size_type)`,
+ * - `::minus_infinity(int)` returns `Degree_bifiltration::minus_inf(size_type)`,
+ * - `::max(int)` throws if `Co` is true and otherwise returns a @ref Degree_bifiltration with one generator with
+ * first parameter 0 and second parameter `std::numeric_limits<T>::max()`,
+ * - `::quiet_NaN(int)` returns `Degree_bifiltration::nan(size_type)`.
+ *
+ * At construction, the format will be \f$ [(v_0,0), (v_1,1), (v_2,2), ..., (v_{k-1},k-1)] \f$, but it is possible
+ * to set a shit and step value to get
+ * \f$ [(v_0,shift+0*step), (v_1,shift+1*step), (v_2,shift+2*step), ..., (v_{k-1},shift+(k-1)*step)] \f$ using
+ * @ref Degree_bifiltration::set_mapping:
```
VincentRouvreau, line 71:
```suggestion
 * @ref Degree_bifiltration::set_mapping
```

The last 2 dots makes doxygen fail to find the reference

src/common/include/gudhi/serialization_utils.h
hschreiber: I just forgot it. Done.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: I am not sure about this? For example, the concept for filtration values for the simplex tree is just called `FiltrationValue` even though it requirements are very specific to the simplex tree. Same with the concept `CoefficientField` and `FilteredComplex` for `Persistent_cohomology`. Those names are also very generic.
I found that the namespace was it good compromise to avoid very long names in the templated classes whose documentation has to link to the concept (and therefore needs templates with the same name). But honestly I don't really care, so I can change that if we have a rule about it.

I already use the namespaces for the Matrix module, so you can see how it looks like in the current documentation.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: The `size_type` I guess? And `reference` probably also. But that's it. In `Degree_bifiltration` for exemple, `const_reference`, `iterator` and `const_iterator` have not much to do with `Underlying_container`.
I wanted to keep `reference` open in case someone needs to add a transformation to `Underlying_container::reference` for some reasons. And `size_type` in case someone wants to enforce something smaller than `std::size_t`?
In the end, it was just to make it look more symmetric, so I can simplify it for those two if you prefer.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: I reformulated it, tell me if it is clearer.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: I reformulated it, tell me if it is clearer.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: Yes. It is used a lot, so we have to define it there.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: This was just to enforce that this constructor has to be used with a container which is ordered and structured like the internal container should be (i.e., it is mostly for internal use or for user knowing what they do). I don't want people to confuse it with `Multi_parameter_filtration_value(Iterator itBegin, Iterator itEnd, int numberOfParameters)` (which had a `Multi_parameter_filtration_value(Range range, int numberOfParameters)` version before, but I ended up never using it).

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: Ah, yes, we can do that. You mean for `StoragePolicy()`, `StoragePolicy(const StoragePolicy &other)`, `StoragePolicy(StoragePolicy &&other)`, `StoragePolicy &operator=(const StoragePolicy &other)` and `StoragePolicy &operator=(StoragePolicy &&other)` or are there also other points?

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: I am also not very satisfied with it, but that was the closest I could find. Just "container" felt wrong, as the container can be the same but used differently (like for `Flat_array_filtration` and `Degree_bifiltration`). Do you have a suggestion?

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: Sadly not. At least I could not make it work. I think (not so sure anymore) it was because it has the same name than the class method using it even if the template arguments are different. I lost so much time finding that out, that I didn't tried changing them. I will probably do that at some point.

src/Multi_filtration/include/gudhi/Multi_filtration/Flat_array_filtration.h
hschreiber: I guess if you think about the C++ array (which makes sense...), a bit yes. Should I rename them `Flat_vector_filtration` and `Nested_vector_filtration`?

src/Multi_filtration/include/gudhi/Multi_filtration/Degree_bifiltration.h
hschreiber: I changed all the guard names. If I forget to do that also for the other PRs, remind me of it.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: `Multi_parameter_filtration_value` never uses the container directly, it always access it through the iterators or `operator()(g, p)`. Only the python interface uses `get_underlying_container()` for now, specifically knowing what StoragePolicy is used, to wrap it in a numpy container. The type really just needs to be defined for the signature of the `Multi_parameter_filtration_value(Underlying_container &&generators, int numberOfParameters)` constructor.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: I reformulated it, tell me if it is better.

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: I reformulated it, tell me if it is clearer.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: To be honest, this specialization is a mess. I wrote it like that because we were supposed to use it (with not much convictions), but things changed so much that it is not even true anymore I realized. So I went through multipers and verified where we still use this: only for the simplex tree. And even for that, that interface does not work, as `Simplex_tree` assumes that `-std::numeric_limits<Filtration_value>::infinity()` is equal to minus infinity, which is not true here when `value_type` is an integral... So I removed everything just now and I redid it for the `Simplex_tree` class specifically. It it cleaner now. And as it is not documented, it should not be a problem that it makes less (?) sense for a general use.

I will mark as resolved the other comments about this class, don't hesitate to re-look at the new version.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: I had one version like that before, but as it ended up not being used (we needed something with flexible amount of parameters), I removed it and did not adapted the factorized version to it indeed.
But I think, we just need to add a way to forbid trying to modify the size of the container with such type of storage classes. The template parameter `Ensure1Criticality` already makes it possible to skip useless steps.

Should I do so?

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: The `Co` will also be explained in the missing introduction.

Co-filtration values are multi filtration values for which we want to look at the negative cones instead of the positive ones. They represent deaths instead of births in a sense. This changes mostly how comparisons are done.

For the infinities, it is more linked to the storage method than to the fact it is co or not. The message was a wrong copy/paste...

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: Not really, it is just that they need to be coherent in the composition. The storage class needs them and also the filtration value class. So one has to pass its way of representing it to the other to agree on it. Is there a more standard way to do so? Or should I just enforce a value for both (which I describe then in the documentation of the concept class instead of the attribute description)?

src/Multi_filtration/concept/StoragePolicy.h
hschreiber: Yes. The values are not stored, but `operator()(g, p)` and `begin(g)`/`end(g)` will explicitly return them.

src/Multi_filtration/include/gudhi/Multi_parameter_filtration_value.h
hschreiber: Yes, it is because it exists in the next PR... I am just afraid of forgetting to correct this if I don't link it now. Or do you prefer I remove it for now?

src/Multi_filtration/include/gudhi/Multi_filtration/Degree_bifiltration.h
hschreiber: Corrected!

PR #1196, opened 2025-04-11
New nanobind binding for the python interface

hschreiber:
It is replacing pybind11 and cython completely.

We are still benchmarking run times, but for now, the times looks equivalent to before. The compile time is much faster.

We had to enable the `NOMINSIZE` option to force nanobind to compile with -03, as the nanobind overrides it with -Os otherwise and it did worsen the performances a lot. But the nanobind documentation recommends to not do so and to pre-compile the performance heavy part instead: https://nanobind.readthedocs.io/en/latest/api_cmake.html#command:nanobind_add_module After the PR, we should open a discussion on this topic to decide if we keep the `NOMINSIZE`s or if we implement a pre-compiled C++ interface (user could also use for other purposes?) we will then link to the nanobind bindings which will remain -Os in that case.

PR #1198, opened 2025-04-15
Pre-pull request to nanobind bindings

hschreiber:
Just to make PR #1196 more readable. Includes all formatting changes, credits update etc. which do not change the code itself.

There are some `???` and `20??` here and there in the credits, when no credits were in the file originally. If someone knows the authors, please let me know. Otherwise I will probably replace them with something generic like "Gudhi Contributor" or similar.


src/python/gudhi/nerve_gic.pyx
```diff
@@ -1,15 +1,20 @@
-# This file is part of the Gudhi Library - https://gudhi.inria.fr/ -
-# which is released under MIT.
-# See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full
-# license details.
+from __future__ import print_function
```
mglisse, line 1: Why does this specific import end up in a different place than the others? Is it the result of some automatic tool?
(We should get rid of it (so its position does not matter) since we do not support python2 anymore, but not this PR)

CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ add_gudhi_module(Toplex_map)
 add_gudhi_module(Witness_complex)
 add_gudhi_module(Nerve_GIC)
 add_gudhi_module(Persistence_matrix)
-add_gudhi_module(Zigzag_persistence)
+#add_gudhi_module(Zigzag_persistence)
```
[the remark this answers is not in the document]
VincentRouvreau: I rolled back this modification (and in src/CMakeLists.txt too) that was here just to try to understand why the CI is failing on Azure OSx.

src/python/gudhi/nerve_gic.pyx
VincentRouvreau: I removed all the `from __future__ import print_function`

hschreiber: This file will disappear anyway, but just as information:

> Python mandates that future-imports must appear in the module before any other code except docstrings.

That is why this line got separated from the rest, even though it could have probably have been put below the comment (and before the __licence__ etc.).
PEP recommandations are: first docstring, then future-imports, then any metadata dunders, then any imports in order of global to local.

mglisse: Thanks, that's exactly what I was missing :smiley:

VincentRouvreau (approved):

PR #1200, opened 2025-04-24
[Persistence cohomology] Enabling more general FilteredComplex types

hschreiber:
For now, the persistence cohomology module only accepts complexes storing all dimensions from 0 to the maximal desired one and only simplicial/cubical ones (i.e. with a predictable boundary size). But the core algorithm it-self does not have this restrictions, it can more or less work with any valid boundary matrix. Which will be necessary for multi-persistence for example. But even for standard persistence, it can be useful.

The problems come from the optimizations for 0/1 homology, which need those restrictions. So for now, I just added the possibility to deactivate them. I am not sure if it is the best way to do this, but it works and should maintain retro-compatibility.

This PR is mostly to open the discussion about how we want to handle this.

### DIFF
--- a/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
+++ b/src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
@@ -165,6 +165,7 @@ class Persistent_cohomology {
    *
    * Assumes that the filtration provided by the simplicial complex is
    * valid. Undefined behavior otherwise. */
+  template<bool optimizations_possible = true>
   void compute_persistent_cohomology(Filtration_value min_interval_length = 0) {
     if (dim_max_ <= 0)
       return;  // --------->>
@@ -177,16 +178,20 @@ class Persistent_cohomology {
       cpx_->assign_key(sh, ++idx_fil);
       dsets_.make_set(cpx_->key(sh));
       int dim_simplex = cpx_->dimension(sh);
-      switch (dim_simplex) {
-        case 0:
-          vertices.push_back(idx_fil);
-          break;
-        case 1:
-          update_cohomology_groups_edge(sh);
-          break;
-        default:
-          update_cohomology_groups(sh, dim_simplex);
-          break;
+      if constexpr (optimizations_possible) {
+        switch (dim_simplex) {
+          case 0:
+            vertices.push_back(idx_fil);
+            break;
+          case 1:
+            update_cohomology_groups_edge(sh);
+            break;
+          default:
+            update_cohomology_groups(sh, dim_simplex);
+            break;
+        }
+      } else {
+        update_cohomology_groups(sh, dim_simplex);
       }
     }
     // Compute infinite intervals of dimension 0
--- a/src/Persistent_cohomology/test/persistent_cohomology_unit_test.cpp
+++ b/src/Persistent_cohomology/test/persistent_cohomology_unit_test.cpp
@@ -1,7 +1,5 @@
 #include <iostream>
 #include <string>
-#include <algorithm>
-#include <utility> // std::pair, std::make_pair
 #include <cmath> // float comparison
 #include <limits>
 #include <cstdint>  // for std::uint8_t
@@ -20,6 +18,19 @@ using namespace Gudhi::persistent_cohomology;
 using namespace boost::unit_test;
 
 typedef Simplex_tree<> typeST;
+
+template<bool with_optimizations>
+std::string get_persistence_string(typeST& st, int coefficient, int min_persistence){
+  Persistent_cohomology<typeST, Field_Zp> pcoh(st);
+  pcoh.init_coefficients( coefficient );  // initializes the coefficient field for homology
+  // Compute the persistent homology of the complex, with given minimal lifetime of homology feature to be recorded.
+  pcoh.compute_persistent_cohomology<with_optimizations>( min_persistence );
+  std::ostringstream ossPers;
+
+  pcoh.output_diagram(ossPers);
+  std::string strPers = ossPers.str();
+  return strPers;
+}
 
 std::string test_persistence(int coefficient, int min_persistence) {
   // file is copied in CMakeLists.txt
@@ -41,15 +52,8 @@ std::string test_persistence(int coefficient, int min_persistence) {
   st.initialize_filtration();
 
   // Compute the persistence diagram of the complex
-  Persistent_cohomology<Simplex_tree<>, Field_Zp> pcoh(st);
-
-  pcoh.init_coefficients( coefficient );  // initializes the coefficient field for homology
-  // Compute the persistent homology of the complex
-  pcoh.compute_persistent_cohomology( min_persistence );  // Minimal lifetime of homology feature to be recorded.
-  std::ostringstream ossPers;
-
-  pcoh.output_diagram(ossPers);
-  std::string strPers = ossPers.str();
+  auto strPers = get_persistence_string<true>(st, coefficient, min_persistence);
+  BOOST_CHECK_EQUAL(strPers, get_persistence_string<false>(st, coefficient, min_persistence));
   return strPers;
 }
 


mglisse:
> only simplicial/cubical ones (i.e. with a predictable boundary size)

I think more precisely: cells of dimension 0 must have an empty boundary, and cells of dimension 1 must have 2 vertices on their boundary (or 0, passing the same vertex twice). With the possible workaround of giving things a fake dimension (true_dim+2, or 42 for everyone).

> The problems come from the optimizations for 0/1 homology, which need those restrictions. So for now, I just added the possibility to deactivate them. I am not sure if it is the best way to do this, but it works and should maintain retro-compatibility.

Is your change only meant for Z/2Z? If we want to pass an arbitrary boundary matrix, we need to be able to pass the coefficients, which the current interface does not allow (it hardcodes that the coefficients alternate between +1 and -1). It could make sense to tackle both issues together since they affect the API.
(I had started something a long time ago, see for instance https://github.com/GUDHI/branches/tree/orient-boundary-glisse or possibly other branches there)

While we are looking into it: IIRC, for the cubical complex, recomputing the dimension of a cell from a handle has a non-trivial cost. It could be nice if there was an easy way to avoid it (although now that I think about it, the issue was probably more in `sort` than when iterating in pcoh).

mglisse:
> The problem of giving a fake dimension is that this fake dimension will also be used in the barcode, so it needs to be translated back.

Sure, it isn't ideal, but at least converting back (if that's really necessary for the application) is not too hard.

> Also it is used to determine the start sign of the alternated sum for the boundary, so it should be bot completely random.

(in Z/2Z the sign doesn't matter)
Does that actually ever matter? When does negating the signs of a whole column change the diagram?

> Indeed, the algorithm decides the initial coefficients even when it is not Z2 (with (-1) % p). So, there is a lack of generality here we will need to think of. My solution only works when those initial coefficients are good enough, which is the case in multipers. It was just thought as a fast patch the time we decide on a real solution.

If it is only for temporary internal use, then I would prefer adding a "private" (in the doxygen sense) function that does not appear in the doc at all so we can remove it as soon as we have an alternative. My concern with a "fast patch" that affects the API is the risk that it makes a cleaner solution harder.

Otherwise, I'd like some discussion of what the long-term solution would look like (even if we don't do it now), to see how the quick patch interferes with it.

> > While we are looking into it: IIRC, for the cubical complex, recomputing the dimension of a cell from a handle has a non-trivial cost. It could be nice if there was an easy way to avoid it (although now that I think about it, the issue was probably more in `sort` than when iterating in pcoh).
> 
> Hm, I am not sure how easy that would be. It is used once for every cell in the main algorithm to distinguish 0, 1 and 2+ dimensions, but also to alternate the starting sign for each dimension (to have coherent boundary signs). All other calls are for diagram/betti number outputs.

(assuming some flexibility in changing the API)

* It should be up to the complex/filtration to provide the coefficients
* If useful, we could also ask the complex to provide the dimension at the same time as it provides the simplex and its boundary
* Internally, we could store the diagram by dimension so we don't need to recompute the dimension again for the output.

src/Persistent_cohomology/test/persistent_cohomology_unit_test.cpp
```diff
@@ -41,15 +52,8 @@ std::string test_persistence(int coefficient, int min_persistence) {
...
-
-  pcoh.init_coefficients( coefficient );  // initializes the coefficient field for homology
-  // Compute the persistent homology of the complex
-  pcoh.compute_persistent_cohomology( min_persistence );  // Minimal lifetime of homology feature to be recorded.
-  std::ostringstream ossPers;
-
-  pcoh.output_diagram(ossPers);
-  std::string strPers = ossPers.str();
+  auto strPers = get_persistence_string<true>(st, coefficient, min_persistence);
+  BOOST_CHECK_EQUAL(strPers, get_persistence_string<false>(st, coefficient, min_persistence));
```
VincentRouvreau, line 61: Add a comment to say that we test here that `compute_persistent_cohomology` returns the same values when `optimizations_possible` is true or false

hschreiber:
> > Also it is used to determine the start sign of the alternated sum for the boundary, so it should be bot completely random.
> 
> (in Z/2Z the sign doesn't matter) Does that actually ever matter? When does negating the signs of a whole column change the diagram?

I should not matter indeed, but as the code goes out of his way to compute the first sign with a formula, I though it would bring something. Perhaps heuristically?

> > Indeed, the algorithm decides the initial coefficients even when it is not Z2 (with (-1) % p). So, there is a lack of generality here we will need to think of. My solution only works when those initial coefficients are good enough, which is the case in multipers. It was just thought as a fast patch the time we decide on a real solution.
> 
> If it is only for temporary internal use, then I would prefer adding a "private" (in the doxygen sense) function that does not appear in the doc at all so we can remove it as soon as we have an alternative. My concern with a "fast patch" that affects the API is the risk that it makes a cleaner solution harder.

I am totally fine with doing something "privately" for now. I could add a non documented version of `compute_persistent_cohomology` which does not use any optimization. And we see later how we want to solve the problem really.

> Otherwise, I'd like some discussion of what the long-term solution would look like (even if we don't do it now), to see how the quick patch interferes with it.

I am also in favor for a discussion about it. Computing barcodes is a central point of the library and we should not hurry into something even uglier than before. But I think it is important that we do something, precisely because it is central.

> (assuming some flexibility in changing the API)
>     * It should be up to the complex/filtration to provide the coefficients
>     * If useful, we could also ask the complex to provide the dimension at the same time as it provides the simplex and its boundary
>     * Internally, we could store the diagram by dimension so we don't need to recompute the dimension again for the output.

If it is okay to add changes in the API (with retro-compatibility I guess), I am in favor of your proposition.
Honestly, I am also very tempted to rewrite the class into two : one completely independent from the simplex tree and which role is just to computes barcode with what is given to him, and another which can take a simplex tree/cubical and calls the other class on it. It would make it easier to build other API's on it while maintaining the old one.

mglisse:
> I should not matter indeed, but as the code goes out of his way to compute the first sign with a formula, I though it would bring something. Perhaps heuristically?

I think the code was trying to follow the usual mathematical formula, and nobody wondered if the micro-optimization of changing the sign would have any impact (or maybe it does have an impact I haven't thought about).

> Honestly, I am also very tempted to rewrite the class into two : one completely independent from the simplex tree and which role is just to computes barcode with what is given to him, and another which can take a simplex tree/cubical and calls the other class on it. It would make it easier to build other API's on it while maintaining the old one.

pcoh was already supposed to be generic, but I guess there is still space for a lower level, which would indeed make things easier for your case.

VincentRouvreau:
Timings with `alpha_complex_persistence` utility are almost the same between the 2 versions:

| dataset | # vertices | # simplices | compute_persistent_cohomology | _without_optimizations |
| --- | --- | --- | --- | --- |
| tore3D_1307.off | 1.307 | 46.079 | 0.015 sec. | 1.568 sec. |
| spiral_3d_10k.off | 10.000 | 2.716.431 | 0.016 sec. | 1.581 sec. |

No difference in the output

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -208,6 +208,44 @@ class Persistent_cohomology {
     }
   }
 
+  /**
+   * @private
+   * Temporary patch to make the computation of more general FilteredComplex possible
+   * Will be removed when a better solution was agreed on.
+   */
+  void compute_persistent_cohomology_without_optimizations(Filtration_value min_interval_length = 0) {
```
VincentRouvreau, line 216: The modification is small, well localized and doesn't seem to make a difference in the output, nor in the timings.
I think it would be safer to not repeat ourself here:
```cpp
  template <bool Optimization = true>
  void compute_persistent_cohomology(Filtration_value min_interval_length = 0) {
  ...
        if (Optimization) {
          switch (dim_simplex) {
          ...
          }
        } else {
          update_cohomology_groups(sh, dim_simplex);
        }
  ...
  }
```
Or

```cpp
  void compute_persistent_cohomology(Filtration_value min_interval_length = 0, bool optimization = true) {
```

Maybe with a warning in the documentation to say that `optimization = false` is experimental or something like that.

mglisse: The discussion with Hannah concluded that for now we would not touch the documented API, on purpose, until we have a clearer picture of what we want in the long term.
If you want to avoid repetitions, we could have a private helper function that both delegate to (the helper and the undocumented function could even be a single function if that's convenient).

VincentRouvreau: Ok, I have no strong opinion on this, but almost copy/paste `compute_persistent_cohomology` (except the switch case) was looking strange to me.

hschreiber: I made a simple copy paste because the method will very probably disappear anyway. Additionally, it is a "private" (public but not documented) patch, so I didn't felt the need to factorize it, just to undo everything later. As we discussed it with Marc, we should do a meeting about this to decide on something. But as this can take a bit more time, in the meantime, this hidden method would exists for multi-persistence / matrix module purposes.

hschreiber:
> Timings with `alpha_complex_persistence` utility are almost the same between the 2 versions:

| dataset | # vertices | # simplices | compute_persistent_cohomology | _without_optimizations |
| --- | --- | --- | --- | --- |
| tore3D_1307.off | 1.307 | 46.079 | 0.015 sec. | 1.568 sec. |
| spiral_3d_10k.off | 10.000 | 2.716.431 | 0.016 sec. | 1.581 sec. |

> No difference in the output

Is the table correct? Because then the timings are not the same at all: 0.015 vs. 1.568, that is x100, no?

VincentRouvreau:
> Is the table correct? Because then the timings are not the same at all: 0.015 vs. 1.568, that is x100, no?

Good catch ! I completely missed my markdown table... The correct one is:

| dataset | # vertices | # simplices | compute_persistent_cohomology | _without_optimizations |
| --- | --- | --- | --- | --- |
| tore3D_1307.off | 1.307 | 46.079 | 0.015 sec. | 0.016 sec. |
| spiral_3d_10k.off | 10.000 | 2.716.431 | 1.568 sec. | 1.581 sec. |

mglisse:
> Timings with `alpha_complex_persistence` utility are almost the same between the 2 versions:

Do you measure only the time taken by the function compute_persistent_homology, or the whole program? IIRC, for the alpha-complex, persistence computation is fairly negligible (we do expensive work to get a small filtration, for which persistence has little work to do). Persistence should be less negligible for rips or cubical.

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -208,6 +208,44 @@ class Persistent_cohomology {
...
+    Simplex_key idx_fil = -1;
+    std::vector<Simplex_key> vertices; // so we can check the connected components at the end
+    // Compute all finite intervals
+    for (auto sh : cpx_->filtration_simplex_range()) {
+      cpx_->assign_key(sh, ++idx_fil);
+      dsets_.make_set(cpx_->key(sh));
+      int dim_simplex = cpx_->dimension(sh);
+      update_cohomology_groups(sh, dim_simplex);
+    }
+    // Compute infinite intervals of dimension 0
```
mglisse: How come you still have code specific for dimension 0, I thought the point was to handle all dimensions the same?

hschreiber: Right, I forgot to remove this loop.

mglisse: What about the loop on `zero_cocycles_`?

hschreiber: I went too fast, sorry. Should be fine this time...

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
```diff
@@ -208,6 +208,31 @@ class Persistent_cohomology {
     }
   }
 
+  /**
+   * @private
+   * Temporary patch to make the computation of more general FilteredComplex possible
+   * Will be removed when a better solution was agreed on.
+   */
+  void compute_persistent_cohomology_without_optimizations(Filtration_value min_interval_length = 0) {
+    if (dim_max_ <= 0)
```
mglisse, line 217: I guess we don't generalize to the point of allowing cells with negative dimension? I don't know if that could ever appear or if it would cause trouble somewhere in the code.

mglisse (approved):

src/Persistent_cohomology/include/gudhi/Persistent_cohomology.h
hschreiber: What would a cell with negative dimension look like? I remember that a (-1)-dimensional "fictional" cell could be used to describe the border of 0-dimensional cells, but I never encountered other ones. `dim_max_` is mostly used to decide if a cell should be token into account or not, so having it negative won't disturb the algorithm. Except if the theoretical algorithm does not work with them?

mglisse: > What would a cell with negative dimension look like?

I don't know, but then I don't know what an edge with 3 vertices looks like either.

> `dim_max_` is mostly used to decide if a cell should be token into account or not, so having it negative won't disturb the algorithm. Except if the theoretical algorithm does not work with them?

The generalization seems to be mostly about handling all dimensions the same, ignoring any special property that some dimensions (0-1 in particular) might have. In that sense, it seems natural to remove this test for `dim_max_ <= 0`. I don't think negative dimensions would cause any problem (although they might in the future if we start storing persistence intervals per dimension, in a vector, indexed by the dimension, since the index cannot be negative).

Anyway this function is private, so if this doesn't look obvious and you know you won't need negative dimensions in your application, this can all wait.

VincentRouvreau (approved):

Issue #1199, opened 2025-04-18
#include <boost/version.hpp> before checking BOOST_VERSION


mglisse:
https://github.com/GUDHI/gudhi-devel/blob/3cff425e026d4991dd5b0f73fe507c8280bdfcba/src/Zigzag_persistence/include/gudhi/zigzag_persistence.h#L28-L29

There is no guarantee that including some arbitrary boost header will define BOOST_VERSION, we should be explicit. This breaks the compilation of some tests with recent versions of boost.

PR #1201, opened 2025-04-28
Fix ZigZag for boost unordered flat map include

VincentRouvreau:


mglisse:
You forgot to write "fix #1199" in the description. I added it, but it only acts as a mention now (probably because the branch is already merged, so I'll go close it by hand.

PR #1208, opened 2025-06-02
[Persistence_matrix] Zp U construction fix

hschreiber:
I just realized that I completely confused my right and left while doing this code and constructed U in the RU decomposition the wrong way around. This should fix this.
It went unseen until now, because RU with Zp coefficients is not really used (vineyard is only implemented for Z2 and the representative cycles interface is not ready)

If someone wants to verify the maths to ensure I didn't do some stupid mistake again, they are welcome...

### DIFF
--- a/src/GudhUI/gui/gudhui.cpp
+++ b/src/GudhUI/gui/gudhui.cpp
@@ -10,7 +10,6 @@
 
 #include "MainWindow.h"
 #include <QApplication>
-#include <CGAL/Qt/resources.h>
 
 int main(int argc, char** argv) {
   QApplication application(argc, argv);
--- a/src/GudhUI/model/Model.h
+++ b/src/GudhUI/model/Model.h
@@ -197,8 +197,6 @@ class Model {
     std::clog << "Num edges : " << complex_.num_edges() << std::endl;
     std::clog << "Num connected components : " << complex_.num_connected_components() << std::endl;
     std::clog << "Min/avg/max degree : " << min_degree() << "/" << avg_degree() << "/" << max_degree() << std::endl;
-    std::clog << "Num connected components : " << complex_.num_connected_components() << std::endl;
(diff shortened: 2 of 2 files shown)


mglisse:
This only touches the unreviewed part of the code and passes the testsuite, so feel free to merge whenever you want.
(sorry, I am unlikely to verify the logic in the short term)

src/Persistence_matrix/include/gudhi/Persistence_matrix/ru_rep_cycles.h
```diff
@@ -135,40 +135,22 @@ inline RU_representative_cycles<Master_matrix>::RU_representative_cycles(
...
-    representativeCycles_.resize(c);
-    for (Index i = 0; i < _matrix()->mirrorMatrixU_.get_number_of_columns(); i++) {
-      for (const auto& entry : _matrix()->mirrorMatrixU_.get_column(i)) {
-        auto idx = birthToCycle_[entry.get_row_index()];
-        if (idx != Master_matrix::template get_null_value<Index>()) {
-          representativeCycles_[idx].push_back(i);
-        }
-      }
+  birthToCycle_.clear();
+  birthToCycle_.resize(_matrix()->reducedMatrixR_.get_number_of_columns(), -1);
```
VincentRouvreau, line 139: `-1` is a magic number ? Maybe it should be defined as `constexpr DEFAULT_MATRIX_VALUE = -1`, no ? Could also be in matrix options as we could define `Index` type as an unsigned for example.

hschreiber: The whole method will change in a follow up PR, so I didn't gave much attention to it. But yes, it will be changed to `Master_matrix::template get_null_value<Index>()`.

VincentRouvreau (approved):

PR #1213, opened 2025-06-13
Do not promote __root_source_dir__ for tangential complex

VincentRouvreau:
Fix partially #179

### DIFF
--- a/src/python/doc/tangential_complex_user.rst
+++ b/src/python/doc/tangential_complex_user.rst
@@ -119,11 +119,10 @@ The result can be obtained after the computation of the Tangential complex
 itself and/or after the perturbation process.
 
 
-Simple example with perturbation
---------------------------------
+Simple example
+--------------
 
-This example builds the Tangential complex of a point set, then tries to solve inconsistencies by perturbing the
-positions of points involved in inconsistent simplices.
+This example builds the Tangential complex of point set.
 
 .. testcode::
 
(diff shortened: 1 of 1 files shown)


src/python/doc/tangential_complex_user.rst
```diff
@@ -119,26 +119,35 @@ The result can be obtained after the computation of the Tangential complex
...
-        ' simplices - ' + repr(tc.num_vertices()) + ' vertices.'
-    print(result_str)
+    print(f'Tangential contains {tc.num_simplices()} simplices - {tc.num_vertices()} vertices.')
+
+    if tc.num_inconsistent_simplices() > 0:
+        print('Tangential contains inconsistencies.')
+
+    tc.fix_inconsistencies_using_perturbation(10, 60)
+    if tc.num_inconsistent_simplices() == 0:
+        print('Inconsistencies has been fixed.')
```
mglisse:
```suggestion
        print('Inconsistencies have been fixed.')
```
It doesn't seem to work on this dataset though. It might be nicer to use a dataset where the inconsistencies do get fixed by the perturbation? (replacing the 3rd point with [4,3] seems to work some of the time, but not often enough)

VincentRouvreau: It was working for me, I was just wanting to test on CI (I should make this PR as draft, sorry).
I removed `__root_source_dir__` and rewrote the python example while keeping the same example on [d831ee5](https://github.com/GUDHI/gudhi-devel/pull/1213/commits/d831ee5d15a7f54da6673914701fe013a3d39c03)

mglisse: It works sometimes (which seems surprising to me, I would expect more consistency, but I don't have time to investigate).
The new version was a good idea. Currently we have one example with inconsistencies that we do not try to fix, and one example without inconsistency (but where the message confusingly claims we fixed them). A single example that originally has inconsistencies but where we manage to fix them sounds good, if we can get a suitable dataset...

VincentRouvreau: Ok, I rolled back the example and fix the conjugation issue. The idea of this PR was just to use something else than `__root_source_dir__`, but yes, the example could be improved

hschreiber (approved):

PR #1214, opened 2025-06-16
Do not promote __root_source_dir__ for example and persistence graphical tools

VincentRouvreau:
This PR contains:

1. Seed management for random point generator (based on CGAL)

- I used a singleton class to set them all, but this is discutable, I could also add a `seed` argument for each function, but this way I found it more convenient.
- `gudhi.datasets.generators.points.seed()` could also set `numpy.random.seed` to simplify the documentation

*Remark:* `c3dtorus` could also be a pure numpy function, which would also simplify this PR.

2. Use of `gudhi.datasets.generators.points.c3dtorus` with a seed instead of reading some files from `__root_source_dir__`

Fix partially #179

3. Remove Witness complex example generators file because part of the code was duplicated and/or dead code


src/common/include/gudhi/CGAL_Random_Singleton.h
```diff
@@ -0,0 +1,51 @@
...
+    }
+    static void set_seed(unsigned int seed) {
+      if(singleton_ != nullptr){
+        delete singleton_;
+      }
+      singleton_ = new CGAL::Random(seed);
+    }
+   private:
+    CGAL_Random_Singleton() {};
+    static CGAL::Random* singleton_;
```
mglisse, line 44: If we are going to use CGAL for random number generation, why create a new singleton instead of using the one in CGAL? (possibly with wrapper functions in gudhi if we want to be able to change the backend at some point, but still using the object returned by `CGAL::get_default_random()`)
Or if we roll our own, it would make more sense to base in on the standard (since C++11) `<random>` instead of CGAL.

src/python/gudhi/datasets/generators/_points.cc
```diff
@@ -122,4 +153,34 @@ PYBIND11_MODULE(_points, m)
...
+          :param R: The big radius.
+          :type R: float
+          :param r: The small radius.
+          :type r: float
+          :returns: the generated points on a torus.
+
+          The typical donut shape is obtained with an aspect ratio (R/r) of the torus equal to 3/2.
+          )pbdoc");
+
+  m.def("seed",
```
mglisse, line 176: It looks inconvenient to hide it in a submodule. Why not at the top level, like `gudhi.set_random_seed`, or maybe `gudhi.random.seed`? Otherwise it looks like we may have to set one seed for the generators, one seed for the internal use of subsampling, one seed for the internal use of mapper, etc, and then we might as well pass a generator argument.

mglisse:
> I used a singleton class to set them all, but this is discutable, I could also add a `seed` argument for each function, but this way I found it more convenient.

If we pass some argument to every function, I think it should be a random number generator (see numpy.random.Generator or https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator.html), not a seed. Normally, the seed is something you use once to initialize the RNG, and then you reuse the same RNG as argument to many functions.

mglisse:
One complication is that we use random numbers both in python (via numpy, maybe also via the standard library) and in C++ (sometimes via CGAL, sometimes directly). If we add a RNG argument to functions, some functions will expect a numpy RNG, some (a wrapper to) a C++ RNG, some (a wrapper to) a CGAL RNG... That makes it confusing for a user.
I think it is possible to wrap a C++ or CGAL RNG in something that numpy can use, and it is possible to wrap numpy's generator as a bit generator that C++ can use (for CGAL, it looks uglier, I think the best that can be done is use the next number from the external RNG as a seed for a new instance of a CGAL RNG), so it might be possible to have a single RNG that is accepted by all python functions in gudhi.

The global RNG (like CGAL does) is not as good as passing RNG arguments (and indeed the global RNG is considered legacy in numpy (https://numpy.org/doc/stable/reference/random/generated/numpy.random.seed.html), but it sure saves a lot of work...

Issue #18, opened 2019-03-07
[Persistence_representations] Missing `inline` for many functions


VincentRouvreau:
Trying to link together 2 files that both include gudhi headers, I get many "multiple definition" errors from the Persistence_representations module. Some examples:
`Gudhi::Persistence_representations::read_persistence_intervals_in_one_dimension_from_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, int, double)`, `Gudhi::Persistence_representations::PSSK::construct(std::vector<std::pair<double, double>, std::allocator<std::pair<double, double> > > const&, std::vector<std::vector<double, std::allocator<double> >, std::allocator<std::vector<double, std::allocator<double> > > >, unsigned long, double, double)`, `Gudhi::Persistence_representations::Persistence_intervals::Persistence_intervals(char const*, unsigned int)`, etc.

There is also `Gudhi::Persistence_representations::epsi`. Making this variable const would work. If it is supposed to remain writable, since we can't use C++17 inline variables yet, it is a bit more complicated.

Issue #34, opened 2019-03-07
[Persistence_representations - Persistence_heat_maps] create_Gaussian_filter ignores its second argument


VincentRouvreau:
```c++
#include <iostream>
#include <gudhi/Persistence_heat_maps.h>

int main(){
  int n=2;
  auto a=Gudhi::Persistence_representations::create_Gaussian_filter(n,.1);
  auto b=Gudhi::Persistence_representations::create_Gaussian_filter(n,10);
  for(int i=0;i<a.size();++i){
    for(int j=0;j<a[0].size();++j){
      std::cout << a[i][j] << ' ';
    }
    std::cout << '\n';
  }
  std::cout << '\n';
  for(int i=0;i<b.size();++i){
    for(int j=0;j<b[0].size();++j){
      std::cout << b[i][j] << ' ';
    }
    std::cout << '\n';
  }
}
```
prints
<pre>
0.000106789 0.00214491 0.00583047 0.00214491 0.000106789
0.00214491 0.0430817 0.117108 0.0430817 0.00214491
0.00583047 0.117108 0.318333 0.117108 0.00583047
0.00214491 0.0430817 0.117108 0.0430817 0.00214491
0.000106789 0.00214491 0.00583047 0.00214491 0.000106789

0.000106789 0.00214491 0.00583047 0.00214491 0.000106789 
0.00214491 0.0430817 0.117108 0.0430817 0.00214491 
0.00583047 0.117108 0.318333 0.117108 0.00583047 
0.00214491 0.0430817 0.117108 0.0430817 0.00214491 
0.000106789 0.00214491 0.00583047 0.00214491 0.000106789 
</pre>
i.e. the same thing whatever `sigma` we use. The code looks like it is using sigma, but it actually simplifies out (up to rounding errors of course).

Hind-M:
Is this bug still relevant ? Is it related to the mathematics involved (bad formula) or something else ?

mglisse:
Yes, I don't believe the bug was fixed.

There are strange things about the formula. Computing sqrt then immediately squaring it back, the normalization factor `1/(pi*sigma²)` (useless since there is a renormalization right after) doesn't match the usual `1/(sigma*sqrt(2pi))`, it looks closer to its square but not quite either.

There are contradictory comments and I am not sure what is intended. From this comment
>   // we are computing the kernel mask to 2 standard deviations away from the center. We discretize it in a grid of a
  // size 2*pixel_radius times 2*pixel_radius.

[...]

Issue #19, opened 2019-03-07
[Persistence_representations - Persistence_heat_maps] Simplify distance_from_diagonal_scaling


VincentRouvreau:
`distance_from_diagonal_scaling` and `squared_distance_from_diagonal_scaling` have complicated implementations with pow, sqrt, etc (all missing `std::`). As far as I can tell, up to a multiplicative constant (which we can probably omit), the first one should be just `std::abs(p.second-p.first)` and the second one the square of `p.second-p.first`.

yagu0:
Indeed. Moreover the comments are confusing: if the argument is the point coordinates, isn't p.second - p.first the lifespan of a point on the diagram ? So it wouldn't correspond to the "Euclidean distance of (b,d) from diagonal".

mglisse:
It actually computes the projection on the diagonal `(b+d)/2`, and then the squared distance to that. It happens to be equal to `|d-b|/sqrt(2)`. When computing the distance to the diagonal, Manhattan, Chebyshev and Euclidean are all proportional.

PR #1218, opened 2025-06-19
[Persistence_representations] Missing inline + other small fixes

hschreiber:
Should fix Issue #18 about the missing `inline`s and partially Issue #19 (missing `std::`).
I also replaced all `bool dbg = false; if (dbg)...` with `GUDHI_DEBUG` macros.
I also corrected the inconsistencies of the use `_` in variable names (sometimes used for arguments, sometimes for class attributes, but not all etc.)
I added a few `TODO` here and there. Please tell me which one should remain and which one I should remove (or solve?).

Once this PR is merged, I will take care of the rest of Issue #19 and #34 if it solvable quickly (putting all in the same PR looked a bit too messy).


src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
+      min -= std::fabs(max - min) / 100;
+      max += std::fabs(max - min) / 100;
+    }
 
-  if (dbg) {
-    std::cerr << "min_ : " << min_ << std::endl;
-    std::cerr << "max_ : " << max_ << std::endl;
+#ifdef GUDHI_DEBUG
+    std::cerr << "min : " << min << std::endl;
+    std::cerr << "max : " << max << std::endl;
```
VincentRouvreau: For debug traces, I would have used `std::clog` and not `std::cerr`

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
-                     std::vector<std::vector<double> > filter,
-                     size_t number_of_pixels,
-                     double min_,
-                     double max_)
-{
-  bool dbg = false;
-  if (dbg) {
+ private:
+  // if min == max, then the program is requested to set up the values itself based on persistence intervals
+  void _construct(const std::vector<std::pair<double, double> >& intervals,
```
VincentRouvreau, line 79: Why did you rename `construct` as  `_construct`. I think there is no such convention in C++, and moreover names beginning with an underscore or a double underscore are RESERVED for the C++ implementers.

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
+#endif
 
-  if (min_ == max_) {
-    // in this case, we want the program to set up the min_ and max_ values by itself.
-    min_ = std::numeric_limits<int>::max();
-    max_ = -std::numeric_limits<int>::max();
+    if (min == max) {
+      // in this case, we want the program to set up the min and max values by itself.
+      min = std::numeric_limits<int>::max();
+      max = -std::numeric_limits<int>::max();
```
VincentRouvreau:
```suggestion
      max = std::numeric_limits<int>::min();
```

As:
```cpp
    std::cout <<  std::numeric_limits<int>::max() << '\n';  //2147483647
    std::cout <<  std::numeric_limits<int>::min() << '\n';  // -2147483648
    std::cout << -std::numeric_limits<int>::max() << '\n';  // -2147483647
```

Moreover, we set a double value with some `std::numeric_limits<int>::max()`, strange...

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
     std::cerr << "Entering construct procedure \n";
-    getchar();
-  }
+#endif
 
-  if (min_ == max_) {
-    // in this case, we want the program to set up the min_ and max_ values by itself.
-    min_ = std::numeric_limits<int>::max();
-    max_ = -std::numeric_limits<int>::max();
+    if (min == max) {
```
VincentRouvreau, line 89: Oops... Here we test that 2 doubles are equal... Maybe not a good idea

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
  *    Copyright (C) 2016 Inria
  *
  *    Modification(s):
+ *      - 2025/06 Hannah Schreiber: Divers small bug fixes (missing `inline`s, `GUDHI_DEBUG`s etc.)
```
VincentRouvreau:
```suggestion
 *      - 2025/06 Hannah Schreiber: Various small bug fixes (missing `inline`s, `GUDHI_DEBUG`s etc.)
```
It seems to me that `divers` is not english, no ?

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
 
 #ifndef PSSK_H_
 #define PSSK_H_
 
 // standard include
-#include <limits>
-#include <utility>
+#ifdef GUDHI_DEBUG
+#include <iostream> // std::cerr, std::clog
+#endif
```
VincentRouvreau: Just a detail, but I think you can `#include <iostream>` without the header guard test.

src/Persistence_representations/include/gudhi/Persistence_heat_maps.h
```diff
@@ -6,37 +6,45 @@
...
 namespace Persistence_representations {
 
 /**
  * This is a simple procedure to create n by n (or 2*pixel_radius times 2*pixel_radius) cubical approximation of a
  * Gaussian kernel.
+ *
+ * @ingroup Persistence_representations
  **/
-std::vector<std::vector<double> > create_Gaussian_filter(size_t pixel_radius, double sigma)
+inline std::vector<std::vector<double> > create_Gaussian_filter(std::size_t pixel_radius, double sigma)
```
VincentRouvreau, line 46: This method could even be `static inline`

VincentRouvreau (commented):
1st round

src/Persistence_representations/include/gudhi/Persistence_heat_maps.h
```diff
@@ -189,85 +204,129 @@ class Persistence_heat_maps
...
+   * filter of a radius 5.
+   * @param erase_below_diagonal A boolean which determines if the area of image below the diagonal should be erased
+   * or not. Default value: false (it will not be erased).
+   * @param number_of_pixels The number of pixels in each direction. Default value: 1000.
+   * @param min The minimal x and y value of points that are to be taken into account. If set to the same value than
+   * `max_`, its value is re-inferred from the minimal point in the given interval. Assumed to be less or equal to
+   * `max_`. Default value: `std::numeric_limits<double>::max()`.
+   * @param max The maximal x and y value of points that are to be taken into account. If set to the same value than
+   * `min_`, its value is re-inferred from the maximal point in the given interval. Assumed to be greater or equal to
+   * `min_`. Default value: `std::numeric_limits<double>::max()`.
```
VincentRouvreau:
```suggestion
   * @param min The minimal x and y value of points that are to be taken into account. If set to the same value than
   * `max`, its value is re-inferred from the minimal point in the given interval. Assumed to be less or equal to
   * `max`. Default value: `std::numeric_limits<double>::max()`.
   * @param max The maximal x and y value of points that are to be taken into account. If set to the same value than
   * `min`, its value is re-inferred from the maximal point in the given interval. Assumed to be greater or equal to
   * `min`. Default value: `std::numeric_limits<double>::max()`.
```

We should use `min` and `max` in the documentation as `min_` and `max_` are the internal private attributes

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: Ah yes, it's "diverse" in english, I forgot the "e". But "Various" is also fine.

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: Yes, it was just to emphasize that `<iostream>` was not really a necessary include, but I agree that it does not make much of a difference. Should I remove the guards?

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: Separating public from private methods helped me to understand how the different classes in this modules are organized when I was going through everything without looking at the details. And if I remember well, `__` and `___` are problematic, not `_`. But there is no problem to revert everything. Just not for the attributes, the convention used here was too messy.

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: I agree, but that is the original code... We could make an issue out of this, but I will not correct it in this PR.

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: I am also not sure why they are used as integers here instead of double. In particular, `std::numeric_limits<int>::max()` should be much smaller than `std::numeric_limits<double>::max()`, so there is a possibility that the loop after does not initialize the values properly (even though, I guess, that in practice the intervals never have such big end values).

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: Yes, I didn't changed the original code here. But I will replace everything with `std::clog` as it is easy to do.

src/Persistence_representations/include/gudhi/Persistence_heat_maps.h
hschreiber: The method is relatively long. Even if we marked it as `inline`, it will probably not be inlined and marking it `static` would be a waste of space, as we cannot link against it anymore, no? I don't think, it is just a helper method for the module. I do think, it was intended to be called by a user too (even though probably no one does it...)

src/Persistence_representations/include/gudhi/Persistence_heat_maps.h
hschreiber: Indeed, I forgot them.

src/Persistence_representations/include/gudhi/PSSK.h
mglisse: I like the GUDHI_DEBUG check before including iostream...

src/Persistence_representations/include/gudhi/PSSK.h
mglisse: The use of `int` is probably a copy-pasto, you could fix it.

src/Persistence_representations/include/gudhi/PSSK.h
mglisse: Comparing the result of a computation is risky, but here, this is more used as an API, if the user calls f(42,42), interpret it as f_guess(), so I don't think the comparison is a big problem (although an overload without the last arguments would have been more natural IMO).

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
+      }
+      // now we have the structure filled in, and moreover we know min and max values of the interval, so we know the
+      // range.
 
-    for (size_t i = 0; i != intervals_.size(); ++i) {
-      if (intervals_[i].first < min_) min_ = intervals_[i].first;
-      if (intervals_[i].second > max_) max_ = intervals_[i].second;
+      // add some more space:
+      min -= std::fabs(max - min) / 100;
+      max += std::fabs(max - min) / 100;
```
mglisse: It would make more sense to compute `fabs(max-min)/100` once and then add/remove the same quantity from min/max. Adding a bit more space on the right is strange.

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,167 @@
...
-};
-
-// if min_ == max_, then the program is requested to set up the values itself based on persistence intervals
-void PSSK::construct(const std::vector<std::pair<double, double> >& intervals_,
-                     std::vector<std::vector<double> > filter,
-                     size_t number_of_pixels,
-                     double min_,
-                     double max_)
-{
-  bool dbg = false;
```
mglisse, line 80: IIRC Paweł did this on purpose, because he wanted to be able to turn these on individually (not all or nothing), at runtime, in the debugger. That doesn't mean we have to keep it, but I thought I'd post this for context.

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: Sure, but the user could have some weird way to use this, with computed inputs. An overload like you propose would have been much safer. Also, `min == max` could have been `min >= max` (not because it makes the comparison with double safer, but it is implicitly assumed that `min < max`, so a mistake less to happen).

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: I guessed so far. He even had some `getchar` methods in them to stop the computation temporarily at particular points, which is quite useful while debugging. But quite ugly inside release code, when there is just a low chance of being needed again.

Another way would have been to just add `constexpr` everywhere. But as there is a Gudhi macro for it, I thought, I should use it. But this does not allow individual enabling...

VincentRouvreau: Yes, Pawel did this on purpose, but I also prefer the debug traces mechanism that was put in places for all the other modules.

src/Persistence_representations/include/gudhi/PSSK.h
VincentRouvreau: Ah, you are right, cf. https://en.cppreference.com/w/cpp/language/identifiers.html
 > In declaration
 > Identifiers that appear as a token or preprocessing token of one of the following forms are reserved:
 > in the global namespace, identifiers that begin with an underscore

Here we are not in the global namespace. But I always find it strange to see identifiers starting with underscore, as this is not Python code.

hschreiber: I guess, that is a matter of taste. I just find it more visual to put the `_` in front than in the back, so I prefer indeed the Python convention for this. And as they are all private methods, I didn't though that someone would care. Should I change them back? Now that the most complicated part is done, I don't really care anymore.

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: I just found out that there is a macro command `#undef`. I guess, someone could play with it to stop other methods from printing if one only wants specific outputs. In the old version, he had to re-compile every time too. The only annoying part, is that we have to modify not only the method we want to print. Or we could add the possibility to compile gudhi in debug mode, but without defining `GUDHI_DEBUG`, so the user can define and undefine it by hand in the place he wants.

mglisse: > he had to re-compile every time too.

Why? If you compile without optimizations, you can change the value of each `dbg` at runtime in the debugger.

hschreiber: As the output where full of `getchar` (and another similar one with similar effect I don't remember the name), I do not think it was intended to be used with a debugger. That is why, I thought about the more classic version of debugging "by hand" with printings. Also, if you use a debugger, are the traces even useful? The debugger already gives you the values of most objects at break points. In that case, I would simply prefer to have `GUDHI_DEBUG` not defined.

src/Persistence_representations/include/gudhi/Persistence_heat_maps.h
```diff
@@ -552,298 +621,219 @@ class Persistence_heat_maps
...
       heat_maps[i][j] = to_compute_median[to_compute_median.size() / 2];
     }
   }
-  this->heat_map = heat_maps;
+  this->heat_map_ = heat_maps;
   this->min_ = maps[0]->min_;
   this->max_ = maps[0]->max_;
 }
 
 template <typename Scalling_of_kernels>
```
mglisse, line 807: What is the point of adding `inline` here (and in many other template functions)?

src/Persistence_representations/include/gudhi/PSSK.h
```diff
@@ -5,163 +5,168 @@
...
+                  double max = -1)
+  {
+#ifdef GUDHI_DEBUG
+    std::clog << "Entering construct procedure \n";
+#endif
+
+    if (min == max) {
+      // in this case, we want the program to set up the min and max values by itself.
+      min = std::numeric_limits<double>::max();
+      max = std::numeric_limits<double>::min();
```
mglisse: If we want to handle negative numbers, `numeric_limits<double>::min` is wrong, it should be `lowest` (or `-max`).

src/Persistence_representations/include/gudhi/Persistence_heat_maps.h
hschreiber: Oh, there isn't any. I guess I added it by reflex as I had to do it for a bunch of other methods. Good that you pointed that out, I will remove them. Even though, it don't think it makes much of a difference (?), we don't need to overload lines with useless keywords.

src/Persistence_representations/include/gudhi/PSSK.h
hschreiber: Ah, I never realized that `min` did not work the same for `int` and `double`/`float`. Weird choice, but I guess there must be a reason for it. I change it to `-max`.

src/Persistence_representations/include/gudhi/PSSK.h
VincentRouvreau: Let's forget this legacy point

PR #1221, opened 2025-06-26
Constify Rips and Cubical sklearn transform

### LINKED ISSUE #1206: Morally `transform` should be const for python sklearn interfaces
_Originally posted by @DavidLapous in https://github.com/GUDHI/gudhi-devel/pull/1202#discussion_r2095987040_

This comment also applies to RipsPersistence and CubicalPersistence.

cf. #1202 for its resolution

VincentRouvreau:
And also add Python types to CubicalPersistence
Fix #1206

### DIFF
--- a/src/python/gudhi/sklearn/rips_persistence.py
+++ b/src/python/gudhi/sklearn/rips_persistence.py
@@ -50,11 +50,11 @@ class RipsPersistence(BaseEstimator, TransformerMixin):
     def __init__(
         self,
         homology_dimensions: Union[int, Iterable[int]],
-        threshold: float = float('inf'),
+        threshold: float = float("inf"),
         input_type: Literal[
             "point cloud", "full distance matrix", "lower distance matrix", "distance coo_matrix"
-        ] = 'point cloud',
-        num_collapses: Union[int, Literal["auto"]] = 'auto',
+        ] = "point cloud",
+        num_collapses: Union[int, Literal["auto"]] = "auto",
         homology_coeff_field: int = 11,
         n_jobs: Optional[int] = None,
     ):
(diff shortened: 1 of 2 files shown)


src/python/gudhi/sklearn/rips_persistence.py
```diff
@@ -86,10 +86,22 @@ def __init__(
         self.homology_coeff_field = homology_coeff_field
         self.n_jobs = n_jobs
 
+        # Done twice (in __init__ and fit), but exception is better the sooner
+        dim_list = np.asarray(self.homology_dimensions, dtype=int)
```
mglisse, line 90: What happens if I pass as argument to `__init__` something like `homology_dimensions=(x for x in range(3) if x!=1)` (which is iterable)?

hschreiber: True, `ArrayLike` should be the right type. `np.asarray` will accept the iterable, but the returned object will remain a generator, so `ndim` will be 0. Or we need to change the tests using `ndim`. Completely missed that.

mglisse: My point is that reading the same generator a second time (in fit, after `__init__`) produces an empty sequence (the generator has already been consumed during the first call).
The fact that generators are not array-like is a separate problem, but a problem indeed. (I see a `numpy.fromiter` which could be relevant, or not)

DavidLapous: The `np.fromiter` is relevant, but slower on numpy arrays since it does a copy + iterator protocol, but that's completely fine in that case.
python iterators are too problematic so I guess it's better to ask for a `Sequence`. 
Scipy, sklearn & cie start their functions with `np.asarray` already so I think it's fine.

hschreiber: > My point is that reading the same generator a second time (in fit, after `__init__`) produces an empty sequence (the generator has already been consumed during the first call). The fact that generators are not array-like is a separate problem, but a problem indeed. (I see a `numpy.fromiter` which could be relevant, or not)

Yes, this is also a problem. I just wanted to point out that calling `np.asarray` with a generator does not unrolls the generator inside the array construction to explicitly copy its values into an array (would have been my first though). By replacing the typing `Iterable[int]` with `ArrayLike` we would just avoid having to deal with it. The old documentation just specified `list of int`, so it would not be a downgrade.

On the long run, if we want to start enabling more general iterables as input, we could do so, but this is a separate issue. Perhaps, we could create a github discussion about it?

mglisse: > The `np.fromiter` is relevant, but slower on numpy arrays since it does a copy + iterator protocol, but that's completely fine in that case.

I don't expect people to pass the dimensions as a numpy array, a tuple makes more sense to me in most cases. I am ok with Sequence, ArrayLike, or things like that, even if Iterable[int] does seem doable.

> Scipy, sklearn & cie start their functions with `np.asarray` already so I think it's fine.

They do that for data, not necessarily for everything (not for the shape of an array for instance).

PR #1222, opened 2025-06-27
[Persistence_matrix] Mixin reorganisation

hschreiber:
I re-organized the mixins of the Matrix classes, to make a few things easier. In particular, a better access to some maps I will need for the new version of representative cycle computation (whose PR will come after this one).


src/Persistence_matrix/include/gudhi/Fields/Zp_field_operators.h
```diff
@@ -67,15 +72,16 @@ class Zp_field_operators
...
 
   /**
    * @brief Sets the characteristic of the field.
    * 
    * @param characteristic Prime number corresponding to the desired characteristic of the field.
    */
   void set_characteristic(Characteristic characteristic) {
-    if (characteristic <= 1)
+    if (characteristic == nullCharacteristic)
       throw std::invalid_argument("Characteristic must be strictly positive and a prime number.");
```
VincentRouvreau: The exception message does not correspond to the test. Should be:
```c++
      throw std::invalid_argument("Characteristic cannot be null.");
```

or
```c++
      throw std::invalid_argument("Characteristic must be defined.");
```

mglisse: I like the fact that it gives all the information at once, not an exchange like
- let me try 0
- 0 not allowed
- let me try -2 or 4
- no, must be prime
(-2 was lucky, -5 would silently be interpreted as 2^32-5 which is prime...)

Not being 0 is actually a consequence of not being prime, so even "must be a positive prime number" would be ok.

hschreiber: After re-looking at the method, I realized that `<= 1` was the right test. Otherwise, nothing is thrown for the case the input argument is 1. So, I put it back.
The message is right again in that case (even though, I could have remove the "strictly" indeed as 0 is not prime).

hschreiber: > (-2 was lucky, -5 would silently be interpreted as 2^32-5 which is prime...)

Negative numbers are a bit tricky indeed as they are automatically converted to unsigned... My hope was, that because the type is renamed `Characteristic`, it would force people to look up the type in the documentation to avoid such type of mistake (there is also the template name which hints to it). But that is probably a bit naïve.

Perhaps the better solution is to type the input as `int` and let the method throw if it is negative as it does right now. It just feels contradictory...

VincentRouvreau:
I didn't review all the PR, it is still a WIP of a beta code. I merge it to help @hschreiber to go further.

Issue #530, opened 2021-09-13
[bug] tensorflow symbolicVector management with wasserstein_distance


tlacombe:
I tried to use a tensorflow model to generate persistence diagrams ; using keras backend, and `gudhi.wasserstein.wasserstein_distance` as a loss function. 

I got the error 
```
    TypeError: len is not well defined for symbolic Tensors. (IteratorGetNext:1) Please call `x.shape` rather than `len(x)` for shape information.
```
which arise when calling 
```
n = len(X)
```
in `gudhi.wasserstein.wasserstein_distance`. 

Below is a MWE to reproduce the error. 

I will investigate this to (hopefully) propose a fix.

```python
import numpy as np
import gudhi.wasserstein as gdw
import tensorflow as tf

def wloss(y_true, y_pred):
    loss = gdw.wasserstein_distance(y_true, y_pred, order=2, internal_p=2, enable_autodiff=True)**2
    return loss


if __name__=='__main__':
    X = np.random.rand(100, 50)  # 100 obs in dim 50
    Y = np.random.rand(100, 10, 2)  # 100 diagrams with 10 points each. 
    model = tf.keras.models.Sequential([
        tf.keras.layers.Dense(20),
        tf.keras.layers.Reshape((10,2))])
    model.compile(optimizer='SGD', loss=wloss)
    model.fit(X,Y, epochs=10)
```

JustNobody0204:
I think here you should convert the tensor object in the wloss function to the proper format that the function wasserstein_distance uses.

mglisse:
It helps if you pass `run_eagerly=True` to `model.compile()`.

PR #1228, opened 2025-08-14
Fix tensorflow len issues

RandomAnass:
## Summary
To fix  TypeError issues that occur when using TensorFlow symbolic tensors with `enable_autodiff=True` in GUDHI's wasserstein distance and KNN functions.

## Problem
The issue was that `len(X)` and `len(Y)` were called on TensorFlow symbalic tensors before checking if autodiff was enabled. TensorFlow symbolic tensors don't support the `len()` function, causing a TypeError.

This is related to [issue #530](https://github.com/GUDHI/gudhi-devel/issues/530).

## Changes
1. **wasserstein.py**: Added TensorFlow tensor detection before using `len()` in the `wasserstein_distance` function
2. **knn.py**: Added similar TensorFlow tensor detection in the `fit` method

## Details
- Use `tensor.shape[0]` for TensorFlow tensors instead of `len(tensor)`
- Tried to keep backward compatibility with numpy arrays 
- Added proper error handling for missing TensorFlow dependency
- Only apply the fix when `enable_autodiff=True`

## Testing
- Both functions now handle TensorFlow symbolic tensors without throwing TypeError
- Backward compatibility with numpy arrays is maintained (from the tests i had)
- Functions work correctly when TensorFlow is not available

## Files Changed
- `src/python/gudhi/wasserstein/wasserstein.py`
- `src/python/gudhi/point_cloud/knn.py`

### DIFF
--- a/src/python/gudhi/point_cloud/knn.py
+++ b/src/python/gudhi/point_cloud/knn.py
@@ -100,9 +100,22 @@ class KNearestNeighbors:
         Args:
             X (numpy.array): coordinates for reference points.
         """
-        if self.k > len(X):
+        # Handle TensorFlow tensors before using len()
+        if self.params.get("enable_autodiff", False):
+            try:
+                import tensorflow as tf
+                if hasattr(tf, 'is_tensor') and tf.is_tensor(X):
+                    n_samples = int(X.shape[0])
+                else:
+                    n_samples = len(X)
+            except (ImportError, AttributeError):
+                n_samples = len(X)
+        else:
+            n_samples = len(X)
+            
+        if self.k > n_samples:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


VincentRouvreau:
Thank you for the PR !
@RandomAnass I just had a look to the Wasserstein part, and I was wondering to know if you would have an example in mind that you want to solve ?

Another way to solve it would be to move the following code (l. 353-360):
```python
    if enable_autodiff:
        import eagerpy as ep

        X_orig = ep.astensor(X)
        Y_orig = ep.astensor(Y)
        X = X_orig.numpy()
        Y = Y_orig.numpy()
```
before the call to `len(X)`.

I tried and it seems to fix #530 if we call `model.compile(optimizer='SGD', loss=wloss, run_eagerly=True)` as @mglisse was suggesting.

PR #1231, opened 2025-08-20
Add Matomo web stats before </head> for c++ and python documentation

VincentRouvreau:


mglisse (approved):
I am not a big fan of these types of solutions. Either the information is already in the logs of the web server and I don't see why we need to add extra code in each webpage, or the information collected is excessive. But well, at least this one doesn't seem too horrible, so ok.

(I didn't look at the technical aspects, I'll trust you for that)

PR #1233, opened 2025-08-28
Fix some python warnings in tests

VincentRouvreau:
* Fix this remark https://github.com/GUDHI/gudhi-devel/pull/1164#issuecomment-3228641746 in diagram_vectorizations_distances_kernels.py
* Call `pot(...enable_autodiff=True)` with `keep_essential_parts=False` to avoid warnings and keep the behavior in test_wasserstein_with_tensors.py
* Fix numpy warning `Conversion of an array with ndim > 0 to a scalar` in cech_complex_sklearn_itf.py
* Catch warning and test it in test_wasserstein_barycenter.py

### DIFF
--- a/src/python/example/cech_complex_sklearn_itf.py
+++ b/src/python/example/cech_complex_sklearn_itf.py
@@ -31,7 +31,7 @@ nb_points = np.random.randint(100,high=300, size=dataset_size)
 X = []
 
 for idx in range(dataset_size):
-    radius = target[idx][0]
+    radius = target[idx, 0]
     pts = points.sphere(nb_points[idx], 2, radius=radius)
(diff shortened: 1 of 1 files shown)


src/python/example/cech_complex_sklearn_itf.py
```diff
@@ -31,8 +31,9 @@
 X = []
 
 for idx in range(dataset_size):
-    pts = points.sphere(nb_points[idx], 2, radius=target[idx])
-    ns = noise * target[idx] * np.random.rand(nb_points[idx], 2) - (noise / 2.)
+    radius = target[idx][0]
```
mglisse:
```suggestion
    radius = target[idx, 0]
```
may be more idiomatic in Python?

PR #1239, opened 2025-09-08
Len issue on tensors

### LINKED ISSUE #530: [bug] tensorflow symbolicVector management with wasserstein_distance

VincentRouvreau:
Fix #530 (AFAIK, we use EagerPy in the backend, and it works only for eager execution)

Fix also the same len issue on knn (no test here - if someone has an idea for a test, I take it)

### DIFF
--- a/src/python/gudhi/wasserstein/wasserstein.py
+++ b/src/python/gudhi/wasserstein/wasserstein.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2019 Inria
 #
 # Modification(s):
+#   - 2025/09 Vincent Rouvreau: Use `len` when input is a Sequence, else use `.shape[0]` (numpy or tensors)
 #   - YYYY/MM Author: Description of the modification
 
 __license__ = "MIT"
@@ -14,6 +15,7 @@ import numpy as np
 import scipy.spatial.distance as sc
 import warnings
 import ot
+from collections.abc import Sequence
 
 
 # Currently unused, but Théo says it is likely to be used again.
@@ -272,8 +274,14 @@ def wasserstein_distance(
     """
 
     # First step: handle empty diagrams
-    n = len(X)
-    m = len(Y)
+    if isinstance(X, Sequence):
+        n = len(X)
+    else:
+        n = X.shape[0]
+    if isinstance(Y, Sequence):
+        m = len(Y)
+    else:
+        m = Y.shape[0]
 
     if n == 0:
         if m == 0:
--- a/src/python/gudhi/point_cloud/knn.py
+++ b/src/python/gudhi/point_cloud/knn.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2020 Inria
 #
 # Modification(s):
+#   - 2025/09 Vincent Rouvreau: Move `len` test in fit as it may not be defined until eagerpy modify its format
 #   - YYYY/MM Author: Description of the modification
 
 __license__ = "MIT"
@@ -101,10 +102,6 @@ class KNearestNeighbors:
         Args:
             X (numpy.array): coordinates for reference points.
         """
-        if self.k > len(X):
-            raise ValueError(
-                f"Expected number of neighbors (aka. 'k') <= number of samples, but k={self.k} and number of samples={len(X)}"
-            )
         self.ref_points = X
         if self.params.get("enable_autodiff", False):
             import eagerpy as ep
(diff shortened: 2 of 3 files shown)


src/python/gudhi/wasserstein/wasserstein.py
```diff
@@ -272,8 +274,14 @@ def wasserstein_distance(
     """
 
     # First step: handle empty diagrams
-    n = len(X)
-    m = len(Y)
+    if isinstance(X, Sequence):
+        n = len(X)
+    else:
+        n = X.shape[0]
```
mglisse: I think try-catch (try to use `len`, if it fails it probably throws an exception, catch it and use the alternative) would be more traditional Python style, but maybe that has drawbacks? (speed, warning messages)

VincentRouvreau: I did so on [f68aae2](https://github.com/GUDHI/gudhi-devel/pull/1239/commits/f68aae23126ba326213a32ccca1fe5ea6506d16b)
I could also do:
```python
    try:
        n = X.shape[0]
    except AttributeError:
        n = len(X)
```
But `len` seems to be widely available:
```python
import numpy as np
import torch
import tensorflow as tf

len([1, 2, 3])                       # 3
len(np.array([1, 2, 3, 4]))          # 4
len(torch.tensor([[1, 2], [3, 4]]))  # 2
len(tf.constant([[1], [2], [3]]))    # 3
```

With my proposal, if the input type is not correct it will lead to the error:
```
AttributeError: 'X' object has no attribute 'shape'
```

VincentRouvreau: Argh... It works fine on my machine with `pip install tensorflow[and-cuda]` (in a conda env + some tricks) but fails on CI with the CPU version `pip install tensorflow` in a virtual environment.

VincentRouvreau: Ok, now I understand the issue.

On CPU, TensorFlow is stricter on out-of-bounds access than on GPU. Saying that, I added some traces after `if enable_autodiff:`, and on the second loop of the fit:
```
pairs_X_Y[:, 0] = [ 7 11 12 14 15 18 19 21 23 24 26 35 36 37 53 54 55 58 59 60 62]
pairs_X_Y[:, 1] = [60 36 42 38 24  8 32 62 18 46 40 30 35 20 14  0 34 12 28 27  2]
```
while `X_orig.shape[0]` and `Y_orig.shape[0]` are 32.

It fails because `len(X) = 32 - len(Y) = 32 - len(X_orig) = 32 - len(Y_orig) = 32` before the call to `_finite_part` on X and Y, and `len(X) = 64 - len(Y) = 64 - len(X_orig) = 32 - len(Y_orig) = 32` after the call to `_finite_part`.

`_finite_part` is awaiting a (n x 2) numpy array, but in autodiff case, input shape is, for example, `(32, 10, 2)` which outputs an array of shape `(64,2)`, but `X_orig` is still of shape `(32, 10, 2)`.

VincentRouvreau:
I turn this PR in draft back as it does not work.
The issue seems to come from `_finite_part` when `enable_autodiff=True`.

VincentRouvreau:
After more investigation, it seems to come from [keras loss function](https://keras.io/api/losses/) that provides a batch as an input.
I removed the UT and the fix mark and merge it like this.
We investigate in parallel with @tlacombe to see if the error comes from the loss function that should be:
```python
def wloss(y_true, y_pred):
    return [ gdw.wasserstein_distance(val, pred, order=2, internal_p=2, enable_autodiff=True, keep_essential_parts=False)**2
             for val, pred in zip(y_true, y_pred) ]

```

Issue #1158, opened 2024-12-02
Potentially erroneous assertion in Hera?


gspr:
This is really an [issue in Hera](https://github.com/anigmetov/hera/issues/14), but it seems to affect GUDHI very directly. I have higher hopes of seeing a response here.

Until some weeks ago, the GUDHI Debian packages accidentally built the (C++ part of the) Python bindings with `-DNDEBUG`, thus disabling all assertions. After that was changed, I've been seeing the test suite fail. That failure seems to stem from [this assertion in Hera](https://github.com/anigmetov/hera/blob/af67e4e8c263ee22569751cb86158ab099bee07e/include/hera/wasserstein/auction_oracle_kdtree_restricted.hpp#L100). I'm not well-versed in the Hera code, but the assertion seems wrong to me. Won't every point in the loop just above the assertion pass the `if_normal` test in case of entirely-off-diagonal persistence diagrams? Then the assertion trivially fails. 

This in turn causes problems for GUDHI – for example, [this part of a test](https://github.com/GUDHI/gudhi-devel/blob/7bd2eccb54886dd711cf9d98e5ce24a3299ba408/src/python/test/test_wasserstein_distance.py#L181) fails on the assertion failing. This might not easily get caught if one, like me, accidentally builds with `-DNDEBUG`.

Am I missing something?

mglisse:
Thanks for the report, we do indeed have something to fix in gudhi. It also reminded us that we wanted to run (part of) the CI with assertions enabled, Vincent is looking at it in #1163.

> Until some weeks ago, the GUDHI Debian packages accidentally built the (C++ part of the) Python bindings with `-DNDEBUG`, thus disabling all assertions.

What do you mean "accidentally"? Building the debian package with `-DNDEBUG` looks normal to me.

gspr:
Marc Glisse ***@***.***> writes:

> Thanks for the report, we do indeed have something to fix in gudhi.

Thanks for taking a look!

> It also reminded us that we wanted to run (part of) the CI with
> assertions enabled, Vincent is looking at it in #1163.

Sounds good.

>  Until some weeks ago, the GUDHI Debian packages accidentally built the (C++ part of the) Python
>  bindings with -DNDEBUG, thus disabling all assertions.
>
> What do you mean "accidentally"? Building the debian package with
> -DNDEBUG looks normal to me.

Debian desires to build our packages with assertions enabled, and so we
[...]

PR #1245, opened 2025-09-15
Hera matching debug mgmt v.2

VincentRouvreau:
I also precise it in hera wasserstein documentation.
The python test are launched this time (will assert if not correct) but without sanitizers.

This work shall be reworked when #1158 and https://github.com/anigmetov/hera/issues/14 will be fixed.


src/python/test/test_wasserstein_distance.py
```diff
@@ -303,5 +303,7 @@ def test_wasserstein_distance_pot():
 
 
 def test_wasserstein_distance_hera():
-    _basic_wasserstein(hera_wrap(delta=1e-12), 1e-12, test_matching=True)
-    _basic_wasserstein(hera_wrap(delta=0.1), 0.1, test_matching=True)
+    # Because of https://github.com/anigmetov/hera/issues/14 hera with test_matching is failing in debug mode
```
mglisse: To be fair, my workarounds in the python wrapper are bogus as well.

src/python/gudhi/hera.py
```diff
@@ -69,6 +69,10 @@ def wasserstein_distance(
     Returns:
         float|Tuple[float,numpy.array|None]: Approximate Wasserstein distance W_q(X,Y), and optionally the 
             corresponding matching
+    
+    .. warning::
+    
+        In debug mode, when matching is ``True``, there is a known bug and code will assert
```
mglisse: That's too specific. The code for `matching=True` is known to have bugs, we recommend using the POT version instead. It happened that we noticed it thanks to an assertion in debug mode, but IIRC you could also get wrong results (or crashes?) in release mode.

hschreiber:
> For dtm and knn Python test, I have an error in debug mode, that I do not have on my machine:
> 
> ```
> 50/56 Test #50: test_knn_py_test .....................................................................***Exception: Illegal 18.71 sec
> ============================= test session starts ==============================
> platform linux -- Python 3.12.3, pytest-8.3.5, pluggy-1.5.0
> rootdir: /__w/gudhi-devel/gudhi-devel
> configfile: pyproject.toml
> plugins: cov-6.1.1
> collected 4 items
> 
> ../../../src/python/test/test_knn.py .Fatal Python error: Illegal instruction
> 
> Current thread 0x00007f4e5bdd4080 (most recent call first):
>   File "<frozen importlib._bootstrap>", line 488 in _call_with_frames_removed

That is weird indeed, as it looks a lot like how usually a C-extension problem look like, but none of those two tests uses any "Gudhi C++". Could there be a version problem with any of the imports?

VincentRouvreau:
> That is weird indeed, as it looks a lot like how usually a C-extension problem look like, but none of those two tests uses any "Gudhi C++". Could there be a version problem with any of the imports?

I did some small steps on this issue.
On my machine with the docker image, compilation and tests are ok.

I have differences in numpy config. Locally, `python -c "import numpy; print(numpy.__config__.show())"` says:
```txt
SIMD Extensions:
  baseline:
  - SSE
  - SSE2
  - SSE3
  found:
  - SSSE3
  - SSE41
  - POPCNT
  - SSE42
  - AVX
  - F16C
  - FMA3
  - AVX2
  - AVX512F
  - AVX512CD
  - AVX512_SKX
  - AVX512_CLX
  - AVX512_CNL
  - AVX512_ICL
  not found:
  - AVX512_KNL
  - AVX512_KNM
```

While on the CI:
```
SIMD Extensions:
  baseline:
  - SSE
  - SSE2
  - SSE3
  found:
  - SSSE3
  - SSE41
  - POPCNT
  - SSE42
  - AVX
  - F16C
  - FMA3
  - AVX2
  not found:
  - AVX512F
  - AVX512CD
  - AVX512_KNL
  - AVX512_KNM
  - AVX512_SKX
  - AVX512_CLX
  - AVX512_CNL
  - AVX512_ICL
```

It could be some missing AVX instructions

.github/workflows/sanitize-unitary-tests-linux.yml
```diff
@@ -7,15 +7,20 @@ jobs:
     name: debug_tests
     runs-on: ubuntu-latest
     # cf. https://github.com/GUDHI/gudhi-deploy/blob/main/Dockerfile_for_circleci_image
-    container: gudhi/ci_for_gudhi:2025.04.01
+    container: gudhi/ci_for_gudhi:2025.04.02
     steps:
       - uses: actions/checkout@v4
         with:
           submodules: true
+      # Check python tests in debug mode and check adress sanitizer for the c++ unitary tests
```
mglisse:
```suggestion
      # Check python tests in debug mode and check address sanitizer only for the c++ unitary tests
```
Somehow I was reading the sentence too quickly and it took me a while to read it slowly enough to understand it. I am hoping that adding "only" makes the meaning slightly easier to grasp...

src/python/gudhi/hera.py
hschreiber: If it is always buggy for `matching=True`, even in release mode, we should perhaps add a warning in the method. Something like:
```Python
if matching is True:
    warnings.warn("The Hera background used by the method when `matching` is set to `True` is known to have bugs. We recommend using `gudhi.wasserstein` instead.", UserWarning)
```

src/python/test/test_wasserstein_distance.py
VincentRouvreau: Done on [ed68a20](https://github.com/GUDHI/gudhi-devel/pull/1245/commits/ed68a20389c3fe1a6c833dbbe2241a463bff0c73)

src/python/gudhi/hera.py
VincentRouvreau: I rewrote the doc warning on [ed68a20](https://github.com/GUDHI/gudhi-devel/pull/1245/commits/ed68a20389c3fe1a6c833dbbe2241a463bff0c73) and add a `UserWarning` on [ea73ad2](https://github.com/GUDHI/gudhi-devel/pull/1245/commits/ea73ad22a75d5654843e3e1a9d2c582dade305a9)

src/python/gudhi/hera.py
```diff
@@ -65,11 +66,24 @@ def wasserstein_distance(
...
 
     Returns:
         float|Tuple[float,numpy.array|None]: Approximate Wasserstein distance W_q(X,Y), and optionally the 
             corresponding matching
+    
     """
+    if matching:
+        warnings.warn(
+            """
+            The Hera background used by the method when `matching` is set to `True` is known to have bugs.
```
mglisse: s/background/backend/ ?
As you did for the doc, I think we could reformulate the warning so it doesn't put all the blame on Hera's maintainers.

VincentRouvreau: I reworked the warning on [587577f](https://github.com/GUDHI/gudhi-devel/pull/1245/commits/587577fbe66ff1f508e8562850ec7a155dbd388d)

src/python/gudhi/hera.py
```diff
@@ -65,11 +66,24 @@ def wasserstein_distance(
...
     Returns:
         float|Tuple[float,numpy.array|None]: Approximate Wasserstein distance W_q(X,Y), and optionally the 
             corresponding matching
+    
     """
+    if matching:
+        warnings.warn(
+            """
+            There is known bug (https://github.com/GUDHI/gudhi-devel/issues/881) when `matching` is set to `True` with
+            the Hera backend. We recommend using `gudhi.wasserstein.wasserstein_distance` instead.
```
mglisse: I'd be tempted to add "for now" or "for the moment" at the end, to clarify that this is temporary, but ok with or without that change.

mglisse (approved):

src/python/gudhi/hera.py
VincentRouvreau: Done on [414be1d](https://github.com/GUDHI/gudhi-devel/pull/1245/commits/414be1dac0f0ff87ecb68a089a8166e1be5eab51) + [e0cc1f5](https://github.com/GUDHI/gudhi-devel/pull/1245/commits/e0cc1f5ffbdf76080c1981ff55763b85fb4689b6)

Issue #768, opened 2022-12-12
Re-expand simplextrees after edge collapse


DavidLapous:
As far as I'm aware, edge collapses are only useful for graphs and clique complexes. 
Wouldn't that make sense to re-expand by default the simplextree to its original dimension after collapsing ? That will ensure that clique complexes simplextrees preserves its homology.
I'm thinking about something like this in the `simplextree.pyx` file
```python
def collapse_edges(self, max_dimension:int=None, num:int=1, ...)->SimplexTree:
    max_dimension = self.dimension() if max_dimension is None else max_dimension
    <old code for collapse>
    self.expansion(max_dimension) # Expands back the simplextree to the original dimension.
    return self
````

mglisse:
Thank you for this suggestion.
The main goal of edge collapses is to avoid computing the complex you would use as input in your code. That is, we only want to build triangles, tetrahedra, etc *after* doing the edge collapse. I am afraid that if we did what you suggest, some users would get confused into doing expansion *before* edge collapse, defeating the purpose.
There could be valid reasons to expand, then collapse and expand again, if only to see how many simplices you save with the collapses, but that looks like a rather niche use case, with an easy workaround (call expansion yourself afterwards), so for now I think I'd rather not do it. But I'll leave it this issue open for a while, to see if other opinions are coming.

Note that we will probably add an extra interface to edge collapses in Python, working on coo_matrix or similar.

DavidLapous:
> I am afraid that if we did what you suggest, some users would get confused into doing expansion before edge collapse, defeating the purpose.

That's a very good point. Do you think that adding a (python) warning in that case enough to prevent users to do that ?

mglisse:
> That's a very good point. Do you think that adding a (python) warning in that case would be enough to prevent users to do that ?

In collapse_edges, adding something like `if self.dimension() > 1: print("Warning: collapse_edges() ignores all the simplices of dimension 2 or more in this complex", file=sys.stderr)`? Yes, I think that would be ok.

mglisse:
> As the root of the problem is an ambiguity in the method name (see [discussion](https://github.com/GUDHI/gudhi-devel/pull/1196#discussion_r2131759283_) in PR [#1196](https://github.com/GUDHI/gudhi-devel/pull/1196)), my suggestion would be to solve the problem properly by renaming the method.

Note that renaming doesn't prevent from adding a warning.

[...]

PR #1246, opened 2025-09-16
[Python SimplexTree] renaming collapse_edges

hschreiber:
See Issue #768. If the new names are ok with everyone, this should solve the issue and the discussion about a possible generalization (which goes beyond this PR and the initial issue) should be transferred into a discussion topic.

### DIFF
--- a/src/python/gudhi/simplex_tree.py
+++ b/src/python/gudhi/simplex_tree.py
@@ -400,14 +400,17 @@ class SimplexTree(t._Simplex_tree_python_interface):
     def collapse_edges_of_flag_complex(
         self, nb_iterations: int = 1, max_expansion_dim: int = None, inplace: bool = True
     ) -> SimplexTree:
-        """Assuming the complex is a flag complex (e.g. a Rips complex), applies the edge collapses technique
-        (see :cite:`edgecollapsearxiv`) to the 1-skeleton of the complex and rebuilds it by extending the skeleton to a
-        new flag complex (with the given maximal dimension). The new complex will be smaller but with same persistent
-        homology as the original one (at least in the existing dimensions: all cycle classes with higher or equal
-        dimension than `max_expansion_dim` will be ignored even if they existed before).
+        """Assuming the complex is a flag complex (e.g. a Rips complex), possibly not expanded (in fact preferably, as
+        only the 1-skeleton is necessary and therefore there is no need to explicitly store higher dimensional simplices
+        just for this purpose), applies the edge collapses technique (see :cite:`edgecollapsearxiv`) to the 1-skeleton
+        of the complex and rebuilds it by extending the new skeleton to a new flag complex (with the given maximal
+        dimension). The new complex will be smaller but with same persistent homology as the original one (just note
+        that if :meth:`persistence()` is called directly on the new simplex tree, all cycle classes with
+        higher or equal dimension than `max_expansion_dim` will be ignored even if they existed before).
 
         If `max_expansion_dim` is set to `1` or less, the method is equivalent to the deprecated method
-        :meth:`collapse_edges()`.
+        :meth:`collapse_edges()`. As a flag complex can be completely deduced from its 1-skeleton, there are lot of
+        purposes where keeping the maximal dimension to 1 is much more efficient than explicitly expand it.
 
         :param nb_iterations: The number of edge collapse iterations to perform. If `nb_iterations` is strictly less
             than 1, the method does nothing and returns `self` (independently of the value of `inplace`). Default is 1.
@@ -418,15 +421,16 @@ class SimplexTree(t._Simplex_tree_python_interface):
             Default is `None`.
         :type max_expansion_dim: int
         :param inplace: If `True`, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
-            tree which is then returned. Default is `True`.
+            tree which is then returned. Note that this parameter is ignored if `nb_iterations` is strictly less than 1.
+            Default is `True`.
         :type inplace: bool
-        :returns: `self` (after modifications from collapses) if `inplace` is set to `True` and a new tree resulting
-            from the collapses otherwise.
+        :returns: `self` (after modifications from collapses) if `inplace` is set to `True` or `nb_iterations` is
+            strictly less than 1, or, a new tree resulting from the collapses otherwise.
         :rtype: SimplexTree
 
         .. warning::
-            The current simplex tree is assumed to be a flag complex. If it is not the case, some information may be
-            lost during the re-expansion, i.e., the persistent homology may differ from before.
+            The current simplex tree is assumed to be a flag complex (possibly not expanded). If it is not the case,
+            some information may be lost during the re-expansion, i.e., the persistent homology may differ from before.
         """
         if nb_iterations < 1:
             return self
@@ -446,7 +450,8 @@ class SimplexTree(t._Simplex_tree_python_interface):
     def collapse_edges(self, nb_iterations: int = 1) -> SimplexTree:
         """.. deprecated:: 3.12.0 
             The method was renamed :meth:`collapse_edges_of_flag_complex()` to clarify the expected
-            pre-conditions. Please use this with argument `max_expansion_dim` set to 1 instead.
+            pre-conditions (the simplex tree has to represent a flag complex. Note that this complex does not need to
+            be expanded.). Please use the new method with argument `max_expansion_dim` set to 1 instead.
 
         :param nb_iterations: The number of edge collapse iterations to perform. If `nb_iterations` is strictly less
             than 1, the method does nothing. Default is 1.
@@ -457,9 +462,9 @@ class SimplexTree(t._Simplex_tree_python_interface):
         warnings.warn(
             "Since Gudhi 3.12, `collapse_edges_of_flag_complex(nb_iterations, 1)` should be called instead of"
             + " `collapse_edges(nb_iterations)`. The method was renamed to clarify the pre-condition that the used"
-            + " simplex tree should be the 1-skeleton of a flag complex. Note that it is now also possible to return"
-            + " the flag complex expended to any dimension by specifying the `max_expansion_dim` parameter of the new"
-            + " method.",
+            + " simplex tree should be a flag complex or at least represent a flag complex (i.e., the 1-skeleton is"
+            + " sufficient, when not even preferable). Note that it is now also possible to return the flag complex"
+            + " expended to any dimension by specifying the `max_expansion_dim` parameter of the new method.",
             DeprecationWarning,
         )
         return self.collapse_edges_of_flag_complex(nb_iterations, max_expansion_dim=1, inplace=True)


src/python/include/python_interfaces/Simplex_tree_interface.h
```diff
@@ -195,32 +196,63 @@ class Simplex_tree_interface : public Simplex_tree<Simplex_tree_options_for_pyth
...
+      const Simplex_tree_interface& st,
+      int nb_collapse_iteration)
   {
     using Filtered_edge = std::tuple<Vertex_handle, Vertex_handle, Filtration_value>;
     std::vector<Filtered_edge> edges;
-    for (Simplex_handle sh : Base::skeleton_simplex_range(1)) {
-      if (Base::dimension(sh) == 1) {
-        typename Base::Simplex_vertex_range rg = Base::simplex_vertex_range(sh);
+    for (Simplex_handle sh : st.skeleton_simplex_range(1)) {
+      if (st.dimension(sh) == 1) {
```
VincentRouvreau: There is a new iterator `dimension_simplex_range` in the Simplex tree to do so:
```suggestion
    for (Simplex_handle sh : st.dimension_simplex_range(1)) {
```

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,76 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+        if inplace:
+            super()._collapse_edges_inplace(nb_iterations)
+            super().expansion(max_expansion_dim)
+            return self
+        collapsed_complex = super()._collapse_edges(nb_iterations)
+        collapsed_complex.expansion(max_expansion_dim)
+        return collapsed_complex
+
+    def collapse_edges(self, nb_iterations=1) -> SimplexTree:
+        """Deprecated since Gudhi 3.12.0. Please use :meth:`collapse_edges_as_graph()` instead.
```
VincentRouvreau: For a better documentation rendering, I would do:
```suggestion
        """.. deprecated:: 3.12.0 Please use :meth:`collapse_edges_as_graph()` instead.
```

And remove the `.. deprecated:: 3.12.0` at the end of the docstring

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,76 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+        :type nb_iterations: int
+
+        .. warning::
+            The current simplex tree is assumed to be a graph, that is of maximal dimension 1.
+            If it is not the case, all higher dimensional simplices will get lost during the
+            reduction process and not be reinserted. To regain them, call `expansion(max_dim)` afterwards.
+
+        .. deprecated:: 3.12.0
+        """
+        import warnings
```
VincentRouvreau: This import should be done at top of the file (with numpy)

src/python/test/test_simplex_tree.py
```diff
@@ -365,13 +365,49 @@ def test_collapse_edges():
...
+    assert st2.num_simplices() == 9
+    assert st2.find([0, 2]) == False  # [1, 3] would be fine as well
+    for simplex in st2.get_skeleton(0):
+        assert simplex[1] == 1.0
+
+    st.collapse_edges_as_graph()
     assert st.num_simplices() == 9
     assert st.find([0, 2]) == False  # [1, 3] would be fine as well
     for simplex in st.get_skeleton(0):
         assert simplex[1] == 1.0
```
VincentRouvreau:
```suggestion
   assert st == st2
```

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,76 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+        If `max_expansion_dim` is set to `1`, the method is equivalent to :meth:`collapse_edges_as_graph()`.
+
+        :param nb_iterations: The number of edge collapse iterations to perform. Default is 1.
+        :type nb_iterations: int
+        :param max_expansion_dim: The maximal dimension to which the new complex has to be expended to. If None, the
+            current dimension is chosen. Note that the final dimension of the new complex can be smaller if no
+            higher-dimensional simplex can exist. Default is None.
+        :type nb_iterations: int
+        :param inplace: If true, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
+            tree which is then returned. Default is True.
```
VincentRouvreau:
```suggestion
        :param inplace: If True, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
            tree which is then returned. Default is True.
```

Python value is True, but a boolean can be true. Maybe we should even be more strict and replace them with `True` to display it as a piece of code (between 2 back-ticks). We could do the same for None.

```suggestion
        :param inplace: If `True`, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
            tree which is then returned. Default is `True`.
```

src/python/gudhi/simplex_tree.py
hschreiber: I also moved up the `.. deprecated` of `initialize_filtration` with a message to homogenize.

src/python/gudhi/simplex_tree.py
```diff
@@ -76,10 +77,11 @@ def __deepcopy__(self) -> SimplexTree:
...
-        """This function initializes and sorts the simplicial complex
+        """.. deprecated:: 3.2.0 The initialization is done automatically if necessary; there is no need
+        to call this method since.
+        
+        This function initializes and sorts the simplicial complex
         filtration vector.
-
-        .. deprecated:: 3.2.0
         """
         import warnings
```
VincentRouvreau: This one can be removed too.
```suggestion
```

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +424,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+        than `max_expansion_dim` will be ignored even if they existed before).
+
+        If `max_expansion_dim` is set to `1` or less, the method is equivalent to :meth:`collapse_edges_as_graph()`.
+
+        :param nb_iterations: The number of edge collapse iterations to perform. Default is 1.
+        :type nb_iterations: int
+        :param max_expansion_dim: The maximal dimension to which the new complex has to be expended to. If `None`, the
+            current dimension is chosen. Note that the final dimension of the new complex can be smaller if no
+            higher-dimensional simplex can exist. Default is `None`.
+        :type nb_iterations: int
```
VincentRouvreau:
```suggestion
        :type max_expansion_dim: int
```

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +424,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+        :param inplace: If `True`, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
+            tree which is then returned. Default is `True`.
+        :type nb_iterations: bool
```
VincentRouvreau:
```suggestion
        :type inplace: bool
```

src/python/gudhi/simplex_tree.py
```diff
@@ -409,6 +413,9 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
 
         :param nb_iterations: The number of edge collapse iterations to perform. Default is 1.
         :type nb_iterations: int
+        :param inplace: If `True`, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
+            tree which is then returned. Default is `True`.
+        :type nb_iterations: bool
```
VincentRouvreau:
```suggestion
        :type inplace: bool
```

VincentRouvreau (approved):

src/python/gudhi/simplex_tree.py
```diff
@@ -17,6 +17,7 @@
 
 import numpy as np
 from numpy.typing import ArrayLike
+import warnings
```
mglisse, line 20: Only importing the dependency in the rare case where the user did something wrong (call a deprecated function) seemed good to me, but I know that's not the recommended Python way, and `warnings` is a standard package, so ok.

src/python/gudhi/simplex_tree.py
```diff
@@ -76,13 +77,12 @@ def __deepcopy__(self) -> SimplexTree:
         return self.copy()
 
     def initialize_filtration(self):
-        """This function initializes and sorts the simplicial complex
+        """.. deprecated:: 3.2.0 The initialization is done automatically if necessary; there is no need
+        to call this method since.
```
mglisse: image
I think if you want to put the message right after `::`, it needs to be on a single line. Otherwise the message should start on another line.

src/python/gudhi/simplex_tree.py
```diff
@@ -409,6 +411,9 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
 
         :param nb_iterations: The number of edge collapse iterations to perform. Default is 1.
         :type nb_iterations: int
+        :param inplace: If `True`, the collapse is done on this simplex tree. Otherwise, the collapse is done on a new
+            tree which is then returned. Default is `True`.
+        :type inplace: bool
 
         .. warning::
             The current simplex tree is assumed to be a graph, that is of maximal dimension 1.
```
mglisse: I thought an important part of the issue was adding a runtime warning to this function in case the input dimension is larger than 1, but I cannot find it?

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+            if max_expansion_dim > 1:
+                super().expansion(max_expansion_dim)
+            return self
+        collapsed_complex = super()._collapse_edges(nb_iterations)
+        if max_expansion_dim > 1:
+            collapsed_complex.expansion(max_expansion_dim)
+        return collapsed_complex
+
+    def collapse_edges(self, nb_iterations=1) -> SimplexTree:
+        """.. deprecated:: 3.12.0 Please use :meth:`collapse_edges_as_graph()` instead.
```
mglisse: Saying that it was renamed (or that it is equivalent to `self.collapse_edges_as_graph(nb_iterations, inplace=True)`) might give a bit more information than "please use xxx instead" (users can just replace the old name with the new name in their code without reading the doc) and allow us to remove the duplicated doc?

src/python/include/python_interfaces/Simplex_tree_interface.h
```diff
@@ -13,6 +13,7 @@
 #define INCLUDE_SIMPLEX_TREE_INTERFACE_H_
 
 #include <cstddef>
+#include <initializer_list>
```
mglisse, line 16: It is unnecessary (`<vector>` is guaranteed to include it), but ok if you want to be explicit.

src/python/include/python_interfaces/Simplex_tree_interface.h
```diff
@@ -195,32 +196,61 @@ class Simplex_tree_interface : public Simplex_tree<Simplex_tree_options_for_pyth
...
 
+  Simplex_tree_interface build_with_collapse_edges(int nb_collapse_iteration)
+  {
+    auto edges = get_collapse_edges(*this, nb_collapse_iteration);
+
+    // Keep only the vertices
+    Simplex_tree_interface st;
+    // vertices have possibility different filtration values, so insert_batch not possible
+    for (int v : Base::complex_vertex_range()) {
+      st.insert_simplex(std::initializer_list<int>{v}, Base::filtration(Base::find({v})));
```
mglisse: `Simplex_tree` has a member function `find_vertex`. It may be private right now, but that can be changed.

src/python/include/python_interfaces/Simplex_tree_interface.h
```diff
@@ -195,32 +196,61 @@ class Simplex_tree_interface : public Simplex_tree<Simplex_tree_options_for_pyth
...
+    }
```
mglisse, line 243: It feels like we should have a better/cheaper way to copy the vertices from an existing ST to an empty one. It is probably negligible here, so not necessarily this PR.

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
         """
         if nb_iterations < 1:
             return
-        super()._collapse_edges(nb_iterations)
-        return self
+        if inplace:
+            super()._collapse_edges_inplace(nb_iterations)
+            return self
+        return super()._collapse_edges(nb_iterations)
```
mglisse: That returns a `gudhi._simplex_tree_ext._Simplex_tree_python_interface`, not a `gudhi.simplex_tree.SimplexTree`.

src/python/gudhi/_simplex_tree.cc
```diff
@@ -285,9 +285,12 @@ computed with these values.
 This `notebook <https://github.com/GUDHI/TDA-tutorial/blob/master/Tuto-GUDHI-extended-persistence.ipynb>`_
 explains how to compute an extension of persistence called extended persistence.
            )doc")
-      .def("_collapse_edges",
+      .def("_collapse_edges_inplace",
            &gsti::collapse_edges,
            nb::call_guard<nb::gil_scoped_release>())
+      .def("_collapse_edges",
+           &gsti::build_with_collapse_edges,
```
mglisse: I think the code would be easier to follow if the name of the python wrapper and the name of the internal C++ function matched (there can be an underscore or even bigger differences, but the matching should be obvious without reading this file).

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
         """
         if nb_iterations < 1:
             return
```
mglisse:
```suggestion
            return self
```
?
If `inplace=False`, it isn't clear from the doc that the function is allowed to return the old ST, maybe it is supposed to `return new SimplexTree(self)` in that case?

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+            return self
+        return super()._collapse_edges(nb_iterations)
+
+    def collapse_edges_as_flag_complex(
+        self, nb_iterations: int = 1, max_expansion_dim: int = None, inplace: bool = True
+    ) -> SimplexTree:
+        """Assuming the complex is a flag complex (e.g. a Rips complex), applies the edge collapses technique
+        (see :cite:`edgecollapsearxiv`) to the 1-skeleton of the complex and rebuilds it by extending the skeleton to a
+        new flag complex (with the given maximal dimension). The new complex will be smaller but with same persistent
+        homology than the original one (at least in the existing dimensions: all cycle classes with higher dimension
```
mglisse:
```suggestion
        new flag complex (with the given maximal dimension). The new complex will be smaller but with same persistent
        homology as the original one (at least in the existing dimensions: all cycle classes with higher dimension
```
(larger than, same as)

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
+        than `max_expansion_dim` will be ignored even if they existed before).
```
mglisse: Homology in dimension `max_expansion_dim` won't be preserved either, since that could require simplices of dimension `max_expansion_dim+1`.

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
         if nb_iterations < 1:
             return
-        super()._collapse_edges(nb_iterations)
-        return self
+        if inplace:
+            super()._collapse_edges_inplace(nb_iterations)
+            return self
+        return super()._collapse_edges(nb_iterations)
+
+    def collapse_edges_as_flag_complex(
```
mglisse: Did anyone find a good use case for this function?
If it doesn't exist, it is just 3 lines of code, possibly 1 with chaining: `st.prune_above_dimension(1).collapse_edges_xxx(*).expansion(*)`. So in my opinion we should only provide it if we expect that this is something users will commonly want to use (and that they are right).
I am not 100% opposed to adding the function if others really want it, but at the very least the documentation should warn users that in most cases, before edge-collapse, they should only build the graph (**not** expand, which would be very costly), then call the other function, and only then expand.

src/python/gudhi/simplex_tree.py
```diff
@@ -397,7 +397,9 @@ def flag_persistence_generators(
             infinites = [np.array(d).reshape(-1, 2) for d in l]
         return (normal0, normals, infinite0, infinites)
 
-    def collapse_edges(self, nb_iterations=1) -> SimplexTree:
+    def collapse_edges_as_graph(
```
mglisse: If we are going to rename for clarity, I think I'd rather go all the way. `collapse_edges_as_graph` is still not super explicit, it doesn't say that some flag filtration is involved. `collapse_edges_of_graph_representing_a_flag_filtration` looks clearer. Maybe it is a bit exaggerated (`collapse_edges_of_graph_of_flag`?), it just feels strange to stop halfway with a name that is long but still not that clear.

src/python/gudhi/simplex_tree.py
```diff
@@ -417,5 +422,74 @@ def collapse_edges(self, nb_iterations=1) -> SimplexTree:
...
-        super()._collapse_edges(nb_iterations)
-        return self
+        if inplace:
+            super()._collapse_edges_inplace(nb_iterations)
+            return self
+        return super()._collapse_edges(nb_iterations)
+
+    def collapse_edges_as_flag_complex(
+        self, nb_iterations: int = 1, max_expansion_dim: int = None, inplace: bool = True
+    ) -> SimplexTree:
```
mglisse: Maybe `collapse_edges_of_flag_complex` (or filtration)? "as" makes it sound like we are reinterpreting something, when what we are actually doing is stating a precondition on the input.

src/python/gudhi/simplex_tree.py
hschreiber: Should be corrected now.

src/python/gudhi/simplex_tree.py
hschreiber: The purpose of renaming and adding a second method was actually to avoid having to add a warning. Pruning higher dimensions can totally be intended. Otherwise the C++ code should throw instead of calling `prune_above...`.

src/python/include/python_interfaces/Simplex_tree_interface.h
hschreiber: This one was added automatically by the IDE and as it is not wrong and more explicit, I kept it. There is no deeper thought behind it.

src/python/include/python_interfaces/Simplex_tree_interface.h
hschreiber: There definitely is a better way to do so. But nothing existing yet. We would need something like `insert_batch_vertices` but which allows individual filtration values or even better a "from_vertices(simplex_tree)" method (as the simplex tree does not provide a range over the filtration values). I don't know if calling `insert_batch_vertices` and then re-assign the filtration values would be more efficient than what I did.

src/python/gudhi/simplex_tree.py
hschreiber: Arf, you are right. This should be corrected now.

src/python/gudhi/_simplex_tree.cc
hschreiber: I renamed them the same.

src/python/gudhi/simplex_tree.py
hschreiber: As the simplex tree is not even pruned when `nb_iterations < 1`, it is difficult to be consistent. So, I just added a `self` for all cases and an explanation in the documentation.

src/python/gudhi/simplex_tree.py
hschreiber: Ah yes, I meant the "higher" as a "non-strict" one. But I agree that should be more explicit.

src/python/gudhi/simplex_tree.py
hschreiber: I will let you debate with @DavidLapous on this, I don't really care. Just be quick, such that this PR does not drag on. But a good point of this method is that its existence contrasts with `collapse_edges_as_graph`, making the purpose of the latter even clearer and less confusing.

src/python/gudhi/simplex_tree.py
hschreiber: Well the flag filtration is only implicitly involved in the collapse method. In the end, the python user does not really care which method was used, as long as it has the right proprieties on the barcode. Someone could give a graph as input (without having flag in mind) and it will keep its promises. We could make it more explicit in the documentation if you want, but I don't think it is necessary to overcharge the method name because of it.
I am also afraid that it will confuse the user because of the other `collapse_edges_of_flag_complex`  method which mentions the "flag" property.

I you really think that it is important to be more precise in the function name, I am also not completely against it. But we need to find a good name for both of the methods such that there is no confusion about their purposes (and no one can come whine because they refuse to read the documentation &mdash; "graph" alone is nicely radical).

src/python/include/python_interfaces/Simplex_tree_interface.h
hschreiber: If you are okay with making `find_vertex` public in its current state, I will do so and replace the `find` method used here with it. It should indeed be more efficient.

hschreiber:
> * in python, I don't know if the `inplace` is useful, it's shorter to do `simplextree.copy().collapse_edges()`

As Marc already said, there is a performance difference between "construct new empty simplex tree > add vertices from the other > add computed edges" and "copy entire simplex tree > prune simplices above dim 1 > add computed edges".
And, you were the one who asked for it in the issue... That is the only reason I added it. 

> * Is the default parameter 1 the best ? i.e., e.g., does `nb_iterations=100`  or `inf` take significantly more time (in practice) than using `1` ? (each of the edges adds a lot in an expansion)

I agree with Marc on this. That `1` is the sanest default option here. A lot of cases don't need to compute more and adding more would impact the computation time non-trivially. It should be the user to decide on the best trade-off for him.

src/python/include/python_interfaces/Simplex_tree_interface.h
mglisse: I think the current find_vertex is fine and can be made public, unless someone disagrees.

src/python/include/python_interfaces/Simplex_tree_interface.h
```diff
@@ -195,32 +196,67 @@ class Simplex_tree_interface : public Simplex_tree<Simplex_tree_options_for_pyth
...
       insert({std::get<0>(remaining_edge), std::get<1>(remaining_edge)}, std::get<2>(remaining_edge));
     }
   }
 
+  nanobind::object collapse_edges(int nb_collapse_iteration)
+  {
+    auto edges = get_collapse_edges(*this, nb_collapse_iteration);
+
+    nanobind::object inst = nanobind::module_::import_("gudhi.simplex_tree").attr("SimplexTree")();
+    Simplex_tree_interface &st = nanobind::cast<Simplex_tree_interface &>(inst);
```
mglisse: Wouldn't it be simpler to do that in the Python function? `new_st = new SimplexTree()` and then pass `new_st` as argument to `collapse_edges`? I am not saying doing it in C++ is wrong, but it seems more complicated.

src/python/gudhi/simplex_tree.py
mglisse: > Well the flag filtration is only implicitly involved in the collapse method. In the end, the python user does not really care which method was used, as long as it has the right proprieties on the barcode.

On the barcode **of the flag filtration** generated by this graph, not on the barcode of the graph.

> Someone could give a graph as input (without having flag in mind) and it will keep its promises.

Er, no edge can be collapsed in a graph. You can only collapse a simplex if it has a unique maximal strict coface (so dimension at least 2). edge_collapse certainly modifies the persistence diagram of the filtered graph, the user must have flag filtrations in mind when they use this function.

src/python/gudhi/simplex_tree.py
mglisse: > As the simplex tree is not even pruned when `nb_iterations < 1`,

I don't think that was deliberate (and it isn't documented). If it makes more sense to change the behavior in this corner case, we can change it.

src/python/gudhi/simplex_tree.py
mglisse: Trying to think of potential sensible behaviors for this particular function (not the one that re-expands) when called on a complex of dim >= 2
1. error (exception)
2. warning, then prune and continue
3. silently prune, continue

I think David was asking for 2. We currently have 3 and you want to keep 3 with the argument that the new name makes 1-2 less relevant, and that 1 is better than 2 anyway. Hmm, if David agrees with you (do you, David?), I don't mind...

src/python/gudhi/simplex_tree.py
hschreiber: The promise we do is that the barcode is the same in the flag complex, that is the property I was talking about. So, yes, the notion of flag is somewhere, but this seems to be more of common sense? Like you say, collapsing a graph makes no sense per se. I was more thinking of someone who just has some triangle meshing without a real sense of "flag filtration" as it is a technical term. He just wants the holes of its shape to appear without realizing that he was constructing a flag complex to do so. While he will certainly know that his meshing can be seen as a graph. I admit, that was not clear.

src/python/include/python_interfaces/Simplex_tree_interface.h
hschreiber: I don't feel it makes much of a difference? I would just have to change more files to add an argument. But if you think that it will have a better impact on the compilation or on the python<->.so exchange to do so, I can. I don't mind.

mglisse: @VincentRouvreau 3rd opinion?

src/python/gudhi/simplex_tree.py
hschreiber: > If it makes more sense to change the behavior in this corner case, we can change it.

You mean to change to old behavior or the behavior currently in the PR? 

The usual behavior of this method is (1) get a simplex tree with only vertices (by pruning when inplace, by copying the vertices otherwise), (2) computing new edges `nb_iterations` times and (3) adding those edges in the simplex tree. So, the consistent behavior of `nb_iterations < 1` would be to result with the same simplex tree than before but without any simplices above dimension 1. Ie. it has the same behavior as `prune_above_dimension(1)` just with the possibility of `inplace` being false.
As it is a very particular special case, I thought that returning `self` would be satisfying in all cases, as it is a more efficient "nothing was done" (also because it is what did the old version). Or do you think this could be an anoying behavior in the cases where this special case is needed?

src/python/gudhi/simplex_tree.py
DavidLapous: I agree that `collapse_edges_as_graph` may not be the best. 
`collapse_edges_of_graph_representing_a_flag_filtration` is definitely too long. 
Would `collapse_flag_complex` work ? 
Q : in  theory, what is collapsing is higher order simplices into edges, non ? in that case it shouldn't be collapse edge anyway

src/python/gudhi/simplex_tree.py
DavidLapous: I think a good idea would be `nb_iteration=0` -> `return self` and `nb_iteration < 0` collapse until not possible anymore. ofc, we copy if `inplace=False`

src/python/gudhi/simplex_tree.py
DavidLapous: I think I prefer to raise a warning, and re-expand (to preserve homology).

To avoid having too many warnings, a possibility is to add a flag in gudhi to enable or not the optimization warnings, e.g., additional copies / expansions here. Something like this, up to renamings :
```python
# True by default
gd.logs.copy_warning=False
gd.logs.unnecessary_computation=False
# etc.
```

hschreiber: > I think I prefer to raise a warning, and re-expand (to preserve homology).

We are talking about the method which does not re-expand. The one who does is the other new one `collapse_edges_of_flag_complex`. `dim >=2` is only an anomaly if the input is supposed to be a 1-skeleton.

src/python/gudhi/simplex_tree.py
hschreiber: > I think a good idea would be `nb_iteration=0` -> `return self` and `nb_iteration < 0` collapse until not possible anymore. ofc, we copy if `inplace=False`

We don't have a way to stop the iterations automatically for now, so this is not possible. You could add this suggestion to the issue (future discussion) #768.

src/python/gudhi/simplex_tree.py
hschreiber: > Would `collapse_flag_complex` work ? 

This is literally already the name for the other method. As I said, if you really want to rename one, we also need to rename the other. I need suggestions for both.

> Q : in theory, what is collapsing is higher order simplices into edges, non ? in that case it shouldn't be collapse edge anyway

Yes and no. We collapse indeed higher dimensional into edges, but also edges are pushed. Otherwise, the 1-skeleton would not change and nothing could collapse in the first place.

src/python/include/python_interfaces/Simplex_tree_interface.h
VincentRouvreau: I agree with you Marc and it involves quite few changes. In simplex_tree.py:
```python
    def collapse_edges_of_graph(
        self, nb_iterations: int = 1, inplace: bool = True
    ) -> SimplexTree:
...
        if inplace:
            super()._collapse_edges_inplace(nb_iterations)
            return self
        st = SimplexTree()
        super()._collapse_edges(nb_iterations, st)
        return st

    def collapse_edges_of_flag_complex(
        self, nb_iterations: int = 1, max_expansion_dim: int = None, inplace: bool = True
    ) -> SimplexTree:
...
        if inplace:
            super()._collapse_edges_inplace(nb_iterations)
            if max_expansion_dim > 1:
                super().expansion(max_expansion_dim)
            return self
        st = SimplexTree()
        super()._collapse_edges(nb_iterations, st)
        if max_expansion_dim > 1:
            st.expansion(max_expansion_dim)
        return st
```

Then in Simplex_tree_interface.h:
```cpp
  void collapse_edges(int nb_collapse_iteration, Simplex_tree_interface & st)
  {
    auto edges = get_collapse_edges(*this, nb_collapse_iteration);
    {
      nanobind::gil_scoped_release release;

      // keeping the vertices
      // vertices have possibility different filtration values, so insert_batch not possible
      for (int v : Base::complex_vertex_range()) {
        st.insert_simplex(std::initializer_list<int>{v}, Base::filtration(Base::find_vertex(v)));
      }

      // Insert remaining edges
      for (const auto& remaining_edge : edges) {
        st.insert({std::get<0>(remaining_edge), std::get<1>(remaining_edge)}, std::get<2>(remaining_edge));
      }
    }
  }
```

hschreiber: Done at https://github.com/GUDHI/gudhi-devel/pull/1246/commits/3366b2deab2ec6d02a921dc595dd126f0cc9156e.

src/python/test/test_collapse_edges.py
```diff
@@ -13,28 +13,28 @@
 from scipy.sparse import coo_matrix
 import pytest
 
-from gudhi.flag_filtration.edge_collapse import reduce_graph as collapse_edges_of_graph
+from gudhi.flag_filtration.edge_collapse import reduce_graph as collapse_edges
```
VincentRouvreau, line 16: IMO I think we shouldn't rename `reduce_graph` as `collapse_edges` in the tests, it could be misunderstood. Either use `collapse_edges_of_flag_complex` or keep its original name `reduce_graph`.

hschreiber: I was not sure why the method was renamed in the first place, so I just rolled back the changes made in this file, as it is not directly related to the PR. The more logical for me would to just keep `reduce_graph`, but I honestly don't care. 
@mglisse a particular opinion on this?

src/python/gudhi/simplex_tree.py
```diff
@@ -397,25 +397,69 @@ def flag_persistence_generators(
...
-            return
-        super()._collapse_edges(nb_iterations)
-        return self
+            return self
+        if max_expansion_dim is None:
+            max_expansion_dim = self.dimension()
+        if inplace:
+            super()._collapse_edges_inplace(nb_iterations)
+            if max_expansion_dim > 1:
+                super().expansion(max_expansion_dim)
```
mglisse, line 442: Do we need `super()` all over the place? I thought it was only needed when the current class has a method with the same name as a method from the parent, and you want to call the second one.

hschreiber: Because I find this more readable as a developer. If I want to look up something about a called method, I know that I have to search in the parent, not in the current class. The `super()` will have to be replaced by `self` anyway, so we don't even really gain shortness by not using it.

VincentRouvreau:
Replaced by #1270

PR #1250, opened 2025-09-22
[Persistence_matrix] Representative cycles fix

hschreiber:
Fixes the problems with representative cycles + adds the possibility of non-Z2.

Now, the only thing missing, is to agree on a C++ and Python interface.


src/Persistence_matrix/include/gudhi/persistence_matrix_options.h
```diff
@@ -44,6 +45,20 @@ enum class Column_types : std::uint8_t {
   INTRUSIVE_SET   /**< @ref Intrusive_set_column "": Underlying container is a boost::intrusive::set<@ref Entry>. */
 };
 
+// a column is said to behave well if its underlying content is alway ordered by increasing row index
+// and iterators cannot iterate over zero cells.
+template <Column_types column_type>
+struct is_well_behaved : std::true_type {};
```
mglisse: * Is the name that clear? Why not something like `has_increasing_iteration_over_nonzeros`? Or maybe split in 2, `iterates_by_increasing_row` and `iteration_guarantees_nonzero_coef`? (names used to clarify my comment, I don't recommend those exact names)
* A default of `true` seems a bit dangerous, if you add a new column and forget to specialize this class.
* Why not put this information inside the column type, so it is easier to find when creating a new column? Otherwise specializing this class is kind of part of the concept of a column but not documented there.

hschreiber: Honestly, I don't really like this struct (it does not really provide a good solution to the problem). I was hoping it being more of a temporary thing. The best would be to just implement a proper iterator for the "non well behaving"  columns. In the meantime, I could indeed just add a flag in the columns instead.

hschreiber:
> I still think it is bad that you end up writing `if constexpr(is_z2)` for every other operation. With a proper API, the code for non-z2 should also work for z2 and in most cases be optimized by the compiler to the same thing you currently specify in the z2 branch.

I don't use it that much. And mostly when there is a data-structure difference between the two versions. For example, `Cycle` is a vector of indices in Z2 and a vector of pairs in Zp. So I cannot construct them the same way. It feels kind of wasteful to use a vector of pairs for all cases (in particular when the knowledge is used, for example, to deduce numpy arrays from it etc.)
(By the way, now that I think about it, it could be more useful to use vectors of arrays of size 2 instead of vectors of pairs in some cases. Perhaps, I should change to that.)

The thing I could do, is isolating those cases better, such that we call some general mini-method which hides the `if constexpr` better.

src/Persistence_matrix/include/gudhi/persistence_matrix_options.h
hschreiber: Well, I ended up making the iterators and remove the struct.

PR #1251, opened 2025-09-23
Representation metrics hera bottleneck

### LINKED ISSUE #432: Hera's bottleneck in representations/metrics.py
Since we now have a second implementation of bottleneck_distance, should we provide a choice in BottleneckDistance, like we do for wasserstein_distance? We need to be careful that epsilon has a very different meaning in both cases...

VincentRouvreau:
Fix #432 
Fix partially #296 (only for metrics.py)

To be discussed : WassersteinDistance raises [NameError](https://docs.python.org/3/library/exceptions.html#NameError) exception when the argument mode is incorrect. IMHO, it should return [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError)

### DIFF
--- a/src/python/gudhi/representations/metrics.py
+++ b/src/python/gudhi/representations/metrics.py
@@ -204,10 +204,10 @@ def pairwise_persistence_diagram_distances(X, Y=None, metric="bottleneck", n_job
             If None, pairwise distances are computed from the first list only.
         metric: distance to use. It can be either a string ("sliced_wasserstein", "wasserstein", "hera_wasserstein"
             (Wasserstein distance computed with Hera - note that Hera is also used for the default option
-            "wasserstein"), "pot_wasserstein" (Wasserstein distance computed with POT), "bottleneck", "cgal_bottleneck"
-            (bottleneck distance computed with CGAL - note that CGAL is also used for the default option "bottleneck"),
-            "hera_bottleneck" (bottleneck distance computed with Hera), "persistence_fisher" or a function taking two
-            numpy arrays of shape (nx2) and (mx2) as inputs.
+            "wasserstein"), "pot_wasserstein" (Wasserstein distance computed with POT), "bottleneck",
+            "gudhi_bottleneck" (bottleneck distance computed with GUDHI - note that GUDHI is also used for the default
+            option "bottleneck"), "hera_bottleneck" (bottleneck distance computed with Hera), "persistence_fisher" or a
+            function taking two numpy arrays of shape (nx2) and (mx2) as inputs.
             If it is a function, make sure that it is symmetric and that it outputs 0 if called on the same two arrays.
         n_jobs (int): number of jobs to use for the computation. This uses joblib.Parallel(prefer="threads"), so
             metrics that do not release the GIL may not scale unless run inside a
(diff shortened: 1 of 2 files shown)


src/python/gudhi/representations/metrics.py
```diff
@@ -335,50 +328,72 @@ class BottleneckDistance(BaseEstimator, TransformerMixin):
...
+    :Requires: `CGAL <installation.html#cgal>`_ for ``mode='cgal'``.
     """
 
-    def __init__(self, e=None, n_jobs=None, epsilon=None):
+    def __init__(self, mode="cgal", e=None, delta=None, n_jobs=None, epsilon=None):
         """
         Constructor for the BottleneckDistance class.
 
         Parameters:
+            mode (str): method for computing bottleneck distance. Either ``'cgal'`` or ``'hera'``. Default set to
```
mglisse: I haven't read the PR yet, but I wanted to say that I don't think we should use "cgal" to designate the code. "hera" makes sense because the whole code comes from hera. But the other version was written by François for Gudhi, the code is in Gudhi, it just happens to use CGAL's kNN to look for nearest neighbors, but that's a detail, like the fact that we use a hashmap from boost for instance. The main code is not in CGAL but in Gudhi.

VincentRouvreau: You are right. I renamed the backend as 'gudhi' on [4762ff6](https://github.com/GUDHI/gudhi-devel/pull/1251/commits/4762ff6c44d5178ac05f5f9df587439c9a3c35c7)

src/python/gudhi/representations/metrics.py
```diff
@@ -188,60 +189,39 @@ def flat_metric(a, b):
...
-                metric=_sklearn_wrapper(pot_wasserstein_distance, X, Y, **kwargs),
-                n_jobs=n_jobs,
-            )
-        except ImportError:
-            print(
-                "POT (Python Optimal Transport) is not installed. Please install POT or use metric='wasserstein' or metric='hera_wasserstein'"
-            )
-            raise
-    elif metric == "sliced_wasserstein":
+    if metric == "sliced_wasserstein":
```
hschreiber, line 224: Why is this not part of the `if isinstance(metric, str):` condition below? `"sliced_wasserstein"` is also a string, no? I guess, because `_pairwise` is not called with the same arguments than the other, but it is a bit confusing when reading the code.

src/python/gudhi/representations/metrics.py
```diff
@@ -252,7 +232,20 @@ def pairwise_persistence_diagram_distances(
...
+            from .. import bottleneck_distance as gudhi_bottleneck_distance
+
+            PAIRWISE_DISTANCE_FUNCTIONS["bottleneck"] = gudhi_bottleneck_distance
+            PAIRWISE_DISTANCE_FUNCTIONS["gudhi_bottleneck"] = gudhi_bottleneck_distance
+
+        elif metric == "pot_wasserstein":
+            # Import here as it can fail if POT is not installed - Will throw ImportError
+            from gudhi.wasserstein import wasserstein_distance as pot_wasserstein_distance
+
+            PAIRWISE_DISTANCE_FUNCTIONS["pot_wasserstein"] = pot_wasserstein_distance
```
hschreiber, line 247: It looks like `PAIRWISE_DISTANCE_FUNCTIONS` is only used in this method and nowhere else. As we have to do those `if ... elif ...` clauses anyway because of the import, couldn't we directly make `if`s for every possibility instead of using the map? It is a bit weird to have a half filled map we update just for some cases on the run. Or we fill it from the beginning with every possibility ( `"sliced_wasserstein"` included) and we only keep the `if`s for the import and the modification of the arguments  `X` and `Y` of `_sklearn_wrapper` in the case of `"sliced_wasserstein"`?

src/python/gudhi/representations/metrics.py
```diff
@@ -335,50 +328,72 @@ class BottleneckDistance(BaseEstimator, TransformerMixin):
...
+                for details.
         """
         self.e = None
         if epsilon is not None:
             warnings.warn("epsilon is a deprecated argument, please consider using e", DeprecationWarning)
             self.e = epsilon
         if e is not None:
             self.e = e
+        self.mode = mode
+        self.delta = delta
```
hschreiber, line 356: As `e` and `delta` are both doubles and will never be used both in the same instance (as only one mode at a time is possible, right?), could we not use the same variable for both?

**Edit:** I just saw that you use them to raise `ValueError` depending on if they are set or not, but I guess you can use the argument variable instead? It would just be a bit annoying because of the possibility of `epsilon`...

src/python/gudhi/representations/metrics.py
```diff
@@ -335,50 +328,72 @@ class BottleneckDistance(BaseEstimator, TransformerMixin):
...
         Parameters:
             X (list of n x 2 numpy arrays): input persistence diagrams.
             y (n x 1 array): persistence diagram labels (unused).
         """
+        if self.mode not in ("gudhi", "hera"):
+            raise ValueError("Unknown mode. Current available values for mode are 'gudhi' and 'hera'")
+        if self.mode == "gudhi" and self.delta is not None:
+            raise ValueError("mode='gudhi' and set 'delta' is contradictory, as 'delta' only applies for mode='hera'")
+        if self.mode == "hera" and self.e is not None:
+            raise ValueError("mode='hera' and set 'e' is contradictory, as 'e' only applies for mode='gudhi'")
```
hschreiber, line 373: Would it not be better to test this in the `__init__` (same for the tests in `__call__`)? Or are `self.mode`, `self.e` and `self.delta` meant to be directly modified by the user? If yes, also forget about my comment above about `self.e` and `self.delta`.

PR #1252, opened 2025-09-25
Silence some warnings (unused variables)

DavidLapous:
Added a bunch of `[[maybe_unused]]` to silence warnings.
I don't know if these variable should be used, but they aren't in my code, and make compilation noise.


src/Bitmap_cubical_complex/include/gudhi/Bitmap_cubical_complex_base.h
```diff
@@ -810,7 +810,7 @@ void Bitmap_cubical_complex_base<T>::read_perseus_style_file(const char* perseus
 
 template <typename T>
 Bitmap_cubical_complex_base<T>::Bitmap_cubical_complex_base(const char* perseus_style_file,
-                                                            const std::vector<bool>& directions) {
+                                                            [[maybe unused]] const std::vector<bool>& directions) {
```
hschreiber: Several `_` are missing: here (l. 813), l. 822 and l. 832.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -586,10 +586,10 @@ class Simplex_tree {
     if constexpr (!Options::stable_simplex_handles) root_.members().reserve(root_source.size());
     for (auto& p : root_source.members()){
       if constexpr (Options::store_key && OtherSimplexTreeOptions::store_key) {
-        auto it = root_.members().try_emplace(
+        [[maybe_unused]] auto it = root_.members().try_emplace(
```
hschreiber: Here and just below, if `it` is really not used:

```suggestion
        root_.members().try_emplace(
```

mglisse:
Thanks for reporting those.
(you have a typo that breaks compilation)
I don't think we should blindly silence the warnings.
Unless it helps with the doxygen doc or something like that, I'd prefer to reserve `[[maybe_unused]]` for things that are really **maybe** unused, typically a variable only used in an assertion (so used in debug mode but not in release mode). But this is a style preference, if people love `[[maybe_unused]]` we can decide to use it more.
* `f(const extents &e)` could be just `f(const extent&)`, naming the variable `e` is pointless.
* `f(const std::vector<bool>& directions)` here the name of the ignored variable is relevant for documentation purposes, we could use the old style `f(const std::vector<bool>& /*directions*/)`. If it helped with doxygen, this could be a good reason to keep the name of the variable and have `[[maybe_unused]]`, but it looks like those functions are undocumented. We might also want to check if we can remove those functions, if they are undocumented and not used anywhere.
* `auto x=f();` could be just `f();` if we don't need `x`.

@hschreiber , most of those are in code you wrote, can you confirm that they are indeed supposed to be unused and not the sign of a bug?

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: You forgot the other one.

DavidLapous: Ah ye, I just used the auto-commit of gh. mb

DavidLapous:
I'm compiling it with DavidLapous/multipers#47. @hschreiber Can I just copy paste the files from this branch to there ?

hschreiber:
> I'm compiling it with [DavidLapous/multipers#47](https://github.com/DavidLapous/multipers/pull/47). @hschreiber Can I just copy paste the files from this branch to there ? (you have write access to both branch if needed)

You cannot copy/paste `Simplex_tree.h`, there are quite some significant differences. But it should be fine for the other files. Perhaps, just make a diff for the two column classes, to be sure.

VincentRouvreau:
The following code:
```cpp
  /** @brief maybe unused
   *
   * @param[in] value Probably unused.
   */
  void probably_unused([[maybe_unused]] const int value) {
      std::cout << "probably_unused\n";
  }
```
gives the following documentation:
image

While the following code:
```cpp
  /** @brief maybe unused
   *
   * @param[in] value Probably unused.
   */
  void mostly_unused(const int /*value*/) {
      std::cout << "mostly_unused\n";
  }
```
renders the following documentation:
image

I agree that if a parameter is not used, we should not document it, but as it appears in the method/function signature, I am more in favor of `[[maybe_unused]]`.

Moreover, in our Doxyfile, we have explicitly `WARN_IF_UNDOCUMENTED   = YES` which means we would also have a doxygen warning when building the documentation with the signature `void mostly_unused(const int /*value*/) `

VincentRouvreau (approved):

PR #1254, opened 2025-10-02
Parallelize compilation for CircleCI

VincentRouvreau:
Limitation is 1h and sometimes it goes further


VincentRouvreau:
Not better... Now ram consumption exceeds the limits

PR #1261, opened 2025-10-09
Try macos-14 to build universal pip packages

### LINKED ISSUE #1256: `macos-13` runner will be deprecated
Currently the GitHub Actions workflow (`pip-build-osx.yml` and `pip-packaging-osx.yml`) uses `macos-13` (x86_64-based macOS runner). According to [GitHub Changelog (Sep 19, 2025)](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/), this runner is scheduled for complete deprecation by 4 December 2025. Therefore, it must be replaced with either the Intel (x86_64) based `macos-15-intel` or the Apple Silicon (arm64) based `macos-15`.

### LINKED ISSUE #1260: CMake find_package Eigen 5.X
For CMake to find Eigen 5.0.0 (new version in August 2025), I had to replace:
https://github.com/GUDHI/gudhi-devel/blob/0cc53be098b46002bca6502f266b0ae48d27c14a/src/cmake/modules/GUDHI_third_party_libraries.cmake#L36

with:
```cmake
find_package(Eigen3 5.0 NO_MODULE)
```
GUDHI cmake shall be able to handle 3.X and 5.X versions of Eigen

VincentRouvreau:
Fix #1256 

... But we lose support for Python 3.9 and 3.10

Also fix https://github.com/GUDHI/gudhi-devel/issues/1260 (required by macos14 where brew install Eigen 5.0.0)

### DIFF
--- a/src/cmake/modules/GUDHI_third_party_libraries.cmake
+++ b/src/cmake/modules/GUDHI_third_party_libraries.cmake
@@ -41,8 +41,12 @@ endif()
 if(TARGET Eigen3::Eigen)
     # Not mandatory as it is set by Eigen3Config.cmake
     get_target_property(EIGEN3_INCLUDE_DIRS Eigen3::Eigen INTERFACE_INCLUDE_DIRECTORIES)
-    # EIGEN_VERSION_STRING for Eigen 5.X, and EIGEN3_VERSION_STRING otherwise
-    message("++ Eigen 3 version ${EIGEN3_VERSION_STRING}${EIGEN_VERSION_STRING}. Includes found in ${EIGEN3_INCLUDE_DIRS}")
+    # Eigen3_VERSION for Eigen 5.X, and EIGEN3_VERSION_STRING otherwise
+    # EIGEN3_VERSION_STRING also used by src/python/CMakeLists.txt
+    if(NOT EIGEN3_VERSION_STRING)
+      set(EIGEN3_VERSION_STRING ${Eigen3_VERSION})
+    endif()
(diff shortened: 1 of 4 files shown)


src/cmake/modules/GUDHI_third_party_libraries.cmake
```diff
@@ -34,10 +34,15 @@ if (TARGET CGAL::CGAL)
 endif ()
 
 find_package(Eigen3 3.3 NO_MODULE)
```
mglisse: Should we make this one `QUIET`, since there is a second call in case of failure? Should we look for the more recent version first?

mglisse (approved):

src/cmake/modules/GUDHI_third_party_libraries.cmake
VincentRouvreau: That's a good idea to look for the more recent version first in quiet mode.
Because with my proposal, with Eigen 5.0.0:
```
  CMake Warning at src/cmake/modules/GUDHI_third_party_libraries.cmake:36 (find_package):
    Could not find a configuration file for package "Eigen3" that is compatible
    with requested version "3.3".

    The following configuration files were considered but not accepted:

      /opt/homebrew/share/eigen3/cmake/Eigen3Config.cmake, version: 5.0.0

  Call Stack (most recent call first):
    CMakeLists.txt:21 (include)


  ++ Eigen 3 version 5.0.0. Includes found in /opt/homebrew/include/eigen3
```
And when no Eigen:
```
-- Could NOT find Eigen3 (missing: Eigen3_DIR)
-- Could NOT find Eigen3 (missing: Eigen3_DIR)
```

Which may be not understandable by the users

PR #1263, opened 2025-10-14
[Python] declare explicitly what is imported in `__init__.py` and declared in `__all__`

### LINKED ISSUE #412: gudhi.__all__ does not include classes
```python
from gudhi import *
SimplexTree
```
> NameError: name 'SimplexTree' is not defined

Whereas `from gudhi import SimplexTree` makes it available just fine.

Related, there seems to be a debate whether IDEs should use `__all__` or `__dir__` or something else for completion, but in particular some seem to be using `__all__` and thus fail to complete `gudhi.SimplexTr`.

Removing the line that defines `__all__` in `__init__.py` seems to help for the import. However, it doesn't seem to help for completion in PyCharm.

VincentRouvreau:
The idea here was to declare explicitly in `__init__.py` what we want to embed in the python module and also to declare it in `__all__`.

Tested with PyCharm-2025.2.3 and it fixes #412 

We can discuss in a further PR what should be private (for instance `persistence_graphical_tools.py` could be renamed `_persistence_graphical_tools.py` ...)

### DIFF
--- a/src/python/gudhi/persistence_graphical_tools.py
+++ b/src/python/gudhi/persistence_graphical_tools.py
@@ -199,9 +199,7 @@ def plot_persistence_barcode(
         if path.isfile(persistence_file):
             # Reset persistence
             persistence = []
-            diag = read_persistence_intervals_grouped_by_dimension(
-                persistence_file=persistence_file
-            )
+            diag = read_persistence_intervals_grouped_by_dimension(persistence_file=persistence_file)
             for key in diag.keys():
                 for persistence_interval in diag[key]:
                     persistence.append((key, persistence_interval))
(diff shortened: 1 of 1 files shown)


src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -23,8 +23,8 @@
 import os
 import shutil
 
-from gudhi.reader_utils import read_persistence_intervals_in_dimension
-from gudhi.reader_utils import read_persistence_intervals_grouped_by_dimension
+from gudhi import read_persistence_intervals_in_dimension
+from gudhi import read_persistence_intervals_grouped_by_dimension
```
hschreiber: I like the shortcuts for sklearn modules, but not really this one. Could you already do `from gudhi import read_persistence_intervals_in_dimension` before this PR and you just corrected this file or is it new ? If the first, I guess I have to live with it, but if not, I would prefer keeping `from gudhi.reader_utils import read_persistence_intervals_in_dimension`, as I don't think that the reader methods should be at the root.

VincentRouvreau: Both are working before and after my modifications. I can rollback this modification if you prefer.

hschreiber: I would be in favor of it, but if it already worked in this form before, it does not make that much of a difference (i.e. as you prefer).

hschreiber (approved):

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Change has been reverted on [fc1a79f](https://github.com/GUDHI/gudhi-devel/pull/1263/commits/fc1a79ff95e0cb477608a1c379dd2168932160b1)

DavidLapous (approved):

PR #1265, opened 2025-10-22
Use CMake version ranges to support Eigen versions from 3.3 up to 5.X.X

### LINKED ISSUE #1260: CMake find_package Eigen 5.X

VincentRouvreau:
CMake version ranges requires CMake >= 3.19

Fix #1260 

Tested with Eigen 3.3.1, 3.4.1 and 5.0.0

### DIFF
--- a/.github/next_release.md
+++ b/.github/next_release.md
@@ -27,6 +27,8 @@ Below is a list of changes:
 
 - Installation
      - Pip package is now available for OSx &ge; 13.0 (was &ge; 12.0).
+     - CMake minimal version is now 3.19
+     - Support Eigen 3.4.1 and 5.X.X
 
 - Miscellaneous
      - The [list of bugs that were solved](https://github.com/GUDHI/gudhi-devel/issues?q=label%3A3.12.0+is%3Aclosed) is available on GitHub.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.19)
 
 project(GUDHIdev)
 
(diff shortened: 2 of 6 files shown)


mglisse:
Eigen3ConfigVersion.cmake from 3.4.0 has
```cmake
  if(PACKAGE_FIND_VERSION_RANGE)
    # both endpoints of the range must have the expected major version
    math (EXPR CVF_VERSION_MAJOR_NEXT "${CVF_VERSION_MAJOR} + 1")
    if (NOT PACKAGE_FIND_VERSION_MIN_MAJOR STREQUAL CVF_VERSION_MAJOR
        OR ((PACKAGE_FIND_VERSION_RANGE_MAX STREQUAL "INCLUDE" AND NOT PACKAGE_FIND_VERSION_MAX_MAJOR STREQUAL CVF_VERSION_MAJOR)
          OR (PACKAGE_FIND_VERSION_RANGE_MAX STREQUAL "EXCLUDE" AND NOT PACKAGE_FIND_VERSION_MAX VERSION_LESS_EQUAL CVF_VERSION_MAJOR_NEXT)))
      set(PACKAGE_VERSION_COMPATIBLE FALSE)
```
(read the comment in particular)
i.e. it does not allow ranges that include both `3.*` and `4.*`. 
I don't think we can avoid having 2 separate calls to find_package to handle that...
It looks wrong that Eigen's doc suggests
```cmake
find_package(Eigen3 3.4...5 REQUIRED NO_MODULE)  # Any version >=3.4.0 but <6.0.0.
```
when that does not actually work with 3.4.0.

Issue #1264, opened 2025-10-21
[python] `plot_persistence_diagram` does not handle empty sub-diagram


hschreiber:
In python, if you give to `plot_persistence_diagram` a diagram of the form `[pairs_in_dim0_array, pairs_in_dim1_array, ...]` where one of the pair arrays is empty, we get the error `ValueError: too many values to unpack (expected 2)` at line 372 `x = [birth for (dim, (birth, death)) in persistence]`.

This case arises for examples when computing the diagram from `gudhi.sklearn.RipsPersistence`: the resulting diagram will always have the size of the max dimension given as parameter + 1. So, for example, the given maximal dimension was 2, but there is no pair in dimension 2, then the resulting diagram has size 3 and the last array is empty.
As this behavior makes sense to me, I think `plot_persistence_diagram` should be the one adapting.

Note:
I have to admit, I went over the problem very quickly, so the problem could be somewhere else. I will look at it more closely later (not before November), but I wanted to create the issue to not forget about it.

Issue #354, opened 2020-06-14
cmp_intervals_by_dim_then_length


mglisse:
The comparison `cmp_intervals_by_dim_then_length` is used in several places to sort persistence intervals. Having it in one example (custom_persistence_sort.cpp) to demonstrate how to do something unusual, sure. But it doesn't look like a good order to use everywhere (in particular in the python plugin). If we sort at all, I'd rather sort by length only, so that the most relevant intervals are grouped.

PR #1267, opened 2025-11-05
[python] corrects ploting of empty diagrams + plot instead of scatter

hschreiber:
Fixes Issue #1264

Example of code which was not working (because dim 2 is empty):

```python
import numpy as np
from gudhi.sklearn.rips_persistence import RipsPersistence
from gudhi import plot_persistence_diagram

rips_transformer = RipsPersistence(
        homology_dimensions=range(3),
        input_type="point cloud",
        homology_coeff_field=2
    )
pc = np.array([[0,2], [1,5], [4,0.5], [4,5]])
rips_transformer.fit(pc)
rips_target = rips_transformer.transform([pc])
plot_persistence_diagram(rips_target[0])
```

**Edit:** I also replaced the scatter plot with a plot plot to make the plotting faster (tested with a Rips filtration of 500 points). In a very naïve way (I didn't wanted to spend too much time on it and it works already faster this way).

### DIFF
--- a/src/python/gudhi/persistence_graphical_tools.py
+++ b/src/python/gudhi/persistence_graphical_tools.py
@@ -72,7 +72,7 @@ def _format_handler(a):
     # Array
     try:
         first_death_value = a[0][1]
-        if isinstance(first_death_value, (np.floating, float, np.integer, int)):
+        if np.issubdtype(type(first_death_value), np.number):
             pers = [[0, x] for x in a]
             return pers, 1
     except IndexError:
@@ -81,7 +81,7 @@ def _format_handler(a):
     try:
         for elt in a:  # check that death values have correct type
             if len(elt) != 0:
-                if not isinstance(elt[0][1], (np.floating, float, np.integer, int)):
+                if not np.issubdtype(type(elt[0][1]), np.number):
                     raise TypeError("Should be a list of (birth, death)")
         pers = [[fake_dim, x] for fake_dim, elt in enumerate(a) for x in elt]
         return pers, 2
@@ -97,14 +97,14 @@ def _get_number_of_pairs_by_dimension(barcode):
         return []
 
     try:
-        if isinstance(barcode[0][1], (np.floating, float, np.integer, int)):
+        if np.issubdtype(type(barcode[0][1]), np.number):
             # array of (b,d)
             return [len(barcode)]
     except IndexError:
         pass
 
     try:
-        if len(barcode[0]) == 0 or isinstance(barcode[0][0][1], (np.floating, float, np.integer, int)) :
+        if len(barcode[0]) == 0 or np.issubdtype(type(barcode[0][0][1]), np.number):
             # array of array of (b,d)
             return [len(barcode[d]) for d in range(len(barcode))]
     except TypeError:
@@ -225,7 +225,9 @@ def plot_persistence_barcode(
         if path.isfile(persistence_file):
             # Reset persistence
             persistence = []
-            diag = read_persistence_intervals_grouped_by_dimension(persistence_file=persistence_file)
+            diag = read_persistence_intervals_grouped_by_dimension(
+                persistence_file=persistence_file
+            )
             for key in diag.keys():
                 for persistence_interval in diag[key]:
                     persistence.append((key, persistence_interval))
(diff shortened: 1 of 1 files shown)


src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -79,8 +80,9 @@ def _format_handler(a):
     # Iterable of array
     try:
         for elt in a:  # check that death values have correct type
-            if not isinstance(elt[0][1], (np.floating, float, np.integer, int)):
-                raise TypeError("Should be a list of (birth, death)")
+            if len(elt) != 0:
+                if not isinstance(elt[0][1], (np.floating, float, np.integer, int)):
```
mglisse: I think numpy recommends using `np.issubdtype` instead of isinstance (and then as a bonus `np.floating` magically handles `float`).

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -89,6 +91,30 @@ def _format_handler(a):
     return a, 0
 
 
+# only necessary because _format_handler does not directly decompose everything into xd, yd arrays for plotting
```
mglisse, line 94: Maybe it should? I think @VincentRouvreau added a conversion from the new type to the old type because that was the easiest way to add support for the new type, but long term it should be the reverse.

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -369,9 +396,25 @@ def plot_persistence_diagram(
...
     y = [death if death != float("inf") else infinity for (dim, (birth, death)) in persistence]
-    c = [colormap[dim] for (dim, (birth, death)) in persistence]
 
-    axes.scatter(x, y, alpha=alpha, color=c)
+    i = 0
+    for d in range(len(sizes)):
+        if sizes[d] != 0:
+            j = i + sizes[d]
+            xd = x[i:j]
+            yd = y[i:j]
```
mglisse, line 416: Does something sort the points by dimension before this?

src/python/gudhi/persistence_graphical_tools.py
hschreiber: `format_handler_` is used by `plot_persistence_barcode`, `plot_persistence_diagram` and `plot_persistence_density` and they all use the output a little bit differently, so I see why this was the simplest way to adapt the possible barcode formats. But it feels like a lot of useless copying/moving around, i.e. not very optimal for all three. On the long run, it could be better to have a specialized version for all three, even if it is more annoying to maintain. And then, methods/loops like this one would become unnecessary.

src/python/gudhi/persistence_graphical_tools.py
hschreiber: Oh, good point, no, there isn't for the input format `(dim, (b, d))`. I have to correct this.

hschreiber: Done, but I would redo some benchmarking to be sure that I did not added to much overhead with the sort (and the sizes) compared to the old version without all this.

hschreiber: Ok, the times are still right:

Rips 1000 random points, 1598 bars

#### array or array of (b,d)

before: 28.1 ms ± 833 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
after: 19.3 ms ± 492 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)

#### array of (b,d)

before: 26.7 ms ± 457 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
after: 17.6 ms ± 35.8 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

#### (dim, (b,d))

before: 23 ms ± 557 μs per loop (mean ± std. dev. of 7 runs, 10 loops each)
after: 15.1 ms ± 182 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau:
```suggestion
                elt_type = type(elt[0][1])
                if not (np.issubdtype(elt_type, np.floating) or np.issubdtype(elt_type, np.integer)):
```
seems ok to me

src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Yes, that's not for the scope of this PR, but this is a reminder of #395 (and maybe #354) for the new persistence format

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -367,11 +394,29 @@ def plot_persistence_diagram(
     # line display of equation : birth = death
     axes.plot([axis_start, axis_end], [axis_start, axis_end], linewidth=1.0, color="k")
 
+    if input_type == 0:
+        sorted(persistence, key=lambda x: x[0])
```
mglisse: Did you test that?

hschreiber: Well, I tested it outside of the method but not within, which was an error. I really need to get used to Pythons way of dealing with inputs.
Anyway, it should be corrected now.

I also added tests for it, even though they would not detect the previous error. I just thought it could be useful if we do bigger changes at some point.

PR #1279, opened 2025-11-29
[Persistence_matrix] Functionality to insert a maximal cell using swaps

jlazovskis:
New function `insert_maximal_cell`, only for `RU_matrix`, which calls `insert_boundary` and then `vine_swap` to the requested position. Complementary to already existing `remove_maximal_cell`. Example file using this, removal, and swaps, along with cycle representative printing after each step. Dependent on pull request #1274 .

### DIFF
--- a/src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
+++ b/src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
@@ -24,8 +24,7 @@ struct RU_rep_cycles_options : Default_options<Column_types::INTRUSIVE_LIST, tru
 
 using RU_matrix = Gudhi::persistence_matrix::Matrix<RU_rep_cycles_options>;
 
-template <class Matrix>
-void print_current_representative_cycles(const Matrix& M) {
+void print_current_representative_cycles(const RU_matrix& M) {
   const auto& rc = M.get_representative_cycles();
   for (const auto& cycle : rc) {
     std::cout << M.get_column_dimension(cycle[0]);
@@ -42,27 +41,27 @@ void remove_insert_swap_with_cycles() {
   Matrix mp({{}, {}, {}, {}, {0, 1}, {0, 3}, {0, 2}, {1, 2}, {2, 3}});
 
   std::cout << "Representative cycles at input:\n";
-  print_current_representative_cycles<RU_matrix>(mp);
+  print_current_representative_cycles(mp);
 
   std::cout << "Representative cycles after swapping 6 and 7:\n";
   mp.vine_swap(6);
   mp.update_representative_cycles();
-  print_current_representative_cycles<RU_matrix>(mp);
+  print_current_representative_cycles(mp);
 
   std::cout << "Representative cycles after inserting 0-cell at position 4:\n";
   mp.insert_maximal_cell(4, {});
   mp.update_representative_cycles();
-  print_current_representative_cycles<RU_matrix>(mp);
+  print_current_representative_cycles(mp);
 
   std::cout << "Representative cycles after swapping 8 and 9:\n";
   mp.vine_swap(8);
   mp.update_representative_cycles();
-  print_current_representative_cycles<RU_matrix>(mp);
+  print_current_representative_cycles(mp);
 
   std::cout << "Representative cycles after removing 5:\n";
   mp.remove_maximal_cell(5);
   mp.update_representative_cycles();
-  print_current_representative_cycles<RU_matrix>(mp);
+  print_current_representative_cycles(mp);
 }
 
 int main() { remove_insert_swap_with_cycles<RU_matrix>(); }


src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
```diff
@@ -0,0 +1,99 @@
+/*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
+ *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
+ *    Author(s):       Hannah Schreiber
```
hschreiber: I think you can put your-self as Author here instead of the "modification section", even if you were inspired by my file.

src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
```diff
@@ -0,0 +1,99 @@
...
+            });
+
+  std::cout << "Representative cycles at input:\n";
+  auto rc = mp.get_representative_cycles();
+  for (auto cycle : rc) {
+    std::cout << mp.get_column_dimension(cycle[0]);
+    std::cout << "-cycle: ";
+    for (auto index : cycle) { std::cout << index << ", "; }
+    std::cout << "\n";
+  }
```
hschreiber: I would externalize this method in another method, as it is the same at each step. It makes it more readable and you could add a comment to explain what the loop does more easily.

src/Persistence_matrix/include/gudhi/Persistence_matrix/RU_matrix.h
```diff
@@ -166,6 +166,25 @@ class RU_matrix : public Master_matrix::RU_pairing_option,
...
+   * @tparam Boundary_range Range of @ref Matrix::Entry_representative. Assumed to have a begin(), end() and size()
+   * method.
+   * @param boundary Boundary generating the column to be inserted. The indices of the boundary have to correspond to the
+   * @p cellIndex values of precedent calls of the method for the corresponding cells and should be ordered in
+   * increasing order.
+   * @param dim Dimension of the cell whose boundary is given. If the complex is simplicial,
+   * this parameter can be omitted as it can be deduced from the size of the boundary.
+   */
+  template <class Boundary_range = Boundary>
+  void insert_maximal_cell(Index columnIndex, const Boundary_range& boundary, Dimension dim);
```
hschreiber: You need to add `Master_matrix::template get_null_value<Dimension>()` as the default value of `dim` here, otherwise the argument is not optional.

src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -854,6 +854,14 @@ class Matrix
                                    const Boundary_range& boundary,
                                    Dimension dim = Matrix::get_null_value<Dimension>());
 
+  /**
+   * @brief Only available for @ref boundarymatrix "RU matrices".
+   * 
+   * See also @ref remove_maximal_cell (for the complementary action) and @ref insert_boundary (for insertion at the highest index).
+   */
```
hschreiber, line 878: As this is the "front" matrix, the documentation is even more important, so this need more details. You can make a mix between the documentation of `remove_maximal_cell` and `insert_boundary` in this class if you want.

You also have to add that it is not available for RU matrices if `@ref PersistenceMatrixOptions::column_indexation_type` is set to `@ref Column_indexation_types::IDENTIFIER`. Or you have to add this method to `Id_to_index_overlay` (which is an overlay for RU making ID indices inputs possible instead of Position indices).

src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -854,6 +854,14 @@ class Matrix
...
+  template <class Boundary_range = Boundary>
+  void insert_maximal_cell(Index columnIndex, const Boundary_range& boundary, Dimension dim);
```
hschreiber:
```suggestion
  void insert_maximal_cell(Index columnIndex, const Boundary_range& boundary, Dimension dim = Matrix::get_null_value<Dimension>());
```

src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -1670,6 +1678,18 @@ Matrix<PersistenceMatrixOptions>::insert_boundary(ID_index cellIndex, const Boun
...
 }
 
+template <class PersistenceMatrixOptions>
+template <class Boundary_range>
+inline void Matrix<PersistenceMatrixOptions>::insert_maximal_cell(Index columnIndex, const Boundary_range& boundary, Dimension dim)
+{
+  static_assert(isNonBasic && PersistenceMatrixOptions::has_vine_update,
+                "'insert_maximal_cell' is not available for the chosen options.");
+  static_assert(PersistenceMatrixOptions::is_of_boundary_type || (PersistenceMatrixOptions::has_map_column_container &&
+                                                                  PersistenceMatrixOptions::has_column_pairings),
```
hschreiber: You can just keep `PersistenceMatrixOptions::is_of_boundary_type` here as the other condition was just for the chain matrix.

src/Persistence_matrix/include/gudhi/Persistence_matrix/RU_matrix.h
```diff
@@ -1,10 +1,11 @@
 /*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
  *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
- *    Author(s):       Hannah Schreiber
+ *    Author(s): Hannah Schreiber
```
hschreiber: There are spaces in the original copyright template here which were erased in all files you modified. I don't know how important it is, but it is probably better to remain consistent with the rest of gudhi where the spaces are existing.

src/Persistence_matrix/include/gudhi/Persistence_matrix/RU_matrix.h
```diff
@@ -538,6 +560,24 @@ inline void RU_matrix<Master_matrix>::insert_boundary(ID_index cellIndex, const
...
+  insert_boundary(boundary, dim);
+
+  // If started with 0 columns, no swaps are needed
+  if ( get_number_of_columns()==1 ) return;
+
+  for (Index curr = get_number_of_columns()-1; curr > columnIndex; --curr) {
+    Swap_opt::vine_swap(curr-1);
+  }
+
+}
```
hschreiber: Just a comment, as the matrix files do not respect the Gudhi format convention yet (by my fault), so you cannot really correct this automatically without making a mess (i.e., if you can correct it for this method only, you can do it, but otherwise, I will take care of it later): Gudhi follows the Google-style writting convention, which tells you how to do spacing, line breaks etc. Everything is encoded in the `src/.clang-format` file, so you just have to run it to automatically correct the style. Or, most IDE will do it for you and can also run it only on selected lines which is very handy.

For this portion of code, you will probably end up with:
```C++
{
  static_assert(Master_matrix::Option_list::has_vine_update,
                "'insert_maximal_cell' is not implemented for the chosen options.");

  insert_boundary(boundary, dim);

  // If started with 0 columns, no swaps are needed
  if (get_number_of_columns() == 1) return;

  for (Index curr = get_number_of_columns() - 1; curr > columnIndex; --curr) {
    Swap_opt::vine_swap(curr - 1);
  }
}
```

src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -854,6 +855,31 @@ class Matrix
...
+   * and moves the column using the @ref vine_swap "vine swaps" method to the requested position, which maintains a
+   * reduced decomposition. As a result, the input column (and other columns) may have a different boundary when it 
+   * is finally in the requested position `columnIndex`, compared with the boundary `boundary` upon input. No check 
+   * is made as to whether the order of cells, after the function has finished, will correspond to a valid filtration.
+   * 
+   * Only available for @ref boundarymatrix "RU matrices". Not available for RU matrices
+   * if @ref PersistenceMatrixOptions::column_indexation_type is set to @ref Column_indexation_types::IDENTIFIER.
+   * 
+   * See also @ref remove_maximal_cell (for the complementary action) and @ref insert_boundary (for inserting at the 
+   * highest index).
```
hschreiber: As most people don't like to read documentation, I think it is better to go straight to the point in the first sentence and than add the details afterwards. Something like this for example:

```suggestion
   * @brief Inserts the given cell boundary at the given position in the matrix/filtration such that the matrix remains
   * consistent.
   * Requires @ref PersistenceMatrixOptions::has_vine_update to be true. Only available for 
   * @ref boundarymatrix "RU matrices", but not if @ref PersistenceMatrixOptions::column_indexation_type is set to
   * @ref Column_indexation_types::IDENTIFIER.
   * 
   * Appends the input column to the end of the matrix using the @ref insert_boundary "insert boundary" method,
   * and moves the column using the @ref vine_swap "vine swaps" method to the requested position, which maintains a
   * reduced decomposition. As a result, the input column (and other columns) may have a different boundary when it 
   * is finally in the requested position `columnIndex`, compared with the boundary `boundary` upon input. No check 
   * is made as to whether the order of cells, after the function has finished, will correspond to a valid filtration.
   * 
   * See also @ref remove_maximal_cell (for the complementary action) and @ref insert_boundary (for inserting at the 
   * highest index).
```

src/Persistence_matrix/test/pm_matrix_tests.h
```diff
@@ -1,10 +1,11 @@
 /*    This file is part of the Gudhi Library - https://gudhi.inria.fr/ - which is released under MIT.
  *    See file LICENSE or go to https://gudhi.inria.fr/licensing/ for full license details.
- *    Author(s):       Hannah Schreiber
+ *    Author(s): Hannah Schreiber
  *
- *    Copyright (C) 2023 Inria
+ *    Copyright (C) 2025 Inria
```
hschreiber: @VincentRouvreau For the copyright, is it better to write `2023-25` than replacing with `2025` or does it not matter?

VincentRouvreau: The *normal* way to do it is to say `Copyright (C) 2023-2025 Inria`, but then every year you have to change **all** the copyrights with `Copyright (C) 2023-2026 Inria`.
And we do not want to do that, so we just provide the year the file was created with `Copyright (C) 2023 Inria`.
Some projects (like CGAL) have a script that generates header files with the proper date, but our license is very permissive (MIT), I think there is no real danger to do it this way.

VincentRouvreau: On scikit-learn they are even more concise:
```python
# Authors: The scikit-learn developers
# SPDX-License-Identifier: BSD-3-Clause
```

src/Persistence_matrix/include/gudhi/Persistence_matrix/RU_matrix.h
jlazovskis: Ok! This was just a compulsive formatting change, will change it back.

src/Persistence_matrix/include/gudhi/Persistence_matrix/RU_matrix.h
jlazovskis: Found the file, was able to apply the format with a sublimetext package. Did it for all the (parts of) files that I edited.

src/Persistence_matrix/include/gudhi/Matrix.h
jlazovskis: Noted, adjusted!

src/Persistence_matrix/test/pm_matrix_tests.h
jlazovskis: Ok! Reverted all the years back to how they were before I started, 2025 in the one file I created.

src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
```diff
@@ -0,0 +1,68 @@
...
+struct RU_rep_cycles_options : Default_options<Column_types::INTRUSIVE_LIST, true> {
+  static const bool can_retrieve_representative_cycles = true;
+  static const bool has_vine_update = true;        // for insert, remove, swap
+  static const bool has_removable_columns = true;  // for remove
+};
+
+using RU_matrix = Gudhi::persistence_matrix::Matrix<RU_rep_cycles_options>;
+
+template <class Matrix>
+void print_current_representative_cycles(Matrix M) {
```
hschreiber:
```suggestion
void print_current_representative_cycles(const Matrix& M) {
```
Or at least the `&`, otherwise you copy the matrix at each call.

Below, I would also use `const auto& rc` and `const auto& cycle` as those are non trivial objects.

For such a small matrix it won't make much of a difference, but as it is used as an example for users, it is better to be a bit rigorous.

hschreiber (approved):

src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
jlazovskis: Thanks, agreed, updated.

src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
```diff
@@ -0,0 +1,68 @@
...
+  std::cout << "Representative cycles at input:\n";
+  print_current_representative_cycles<RU_matrix>(mp);
+
+  std::cout << "Representative cycles after swapping 6 and 7:\n";
+  mp.vine_swap(6);
+  mp.update_representative_cycles();
+  print_current_representative_cycles<RU_matrix>(mp);
+
+  std::cout << "Representative cycles after inserting {1,3} at position 7:\n";
+  mp.insert_maximal_cell(4, {});
```
VincentRouvreau: It seems to me that there is a delta between the comment (`Representative cycles after inserting {1,3} at position 7`) and the actual code `mp.insert_maximal_cell(4, {});`. Maybe you have forgotten to update the comment after modifying the code, no ?

src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
```diff
@@ -0,0 +1,68 @@
...
+struct RU_rep_cycles_options : Default_options<Column_types::INTRUSIVE_LIST, true> {
+  static const bool can_retrieve_representative_cycles = true;
+  static const bool has_vine_update = true;        // for insert, remove, swap
+  static const bool has_removable_columns = true;  // for remove
+};
+
+using RU_matrix = Gudhi::persistence_matrix::Matrix<RU_rep_cycles_options>;
+
+template <class Matrix>
+void print_current_representative_cycles(const Matrix& M) {
```
VincentRouvreau: Here, I think we can skip the template as the type is always the same.

```suggestion
void print_current_representative_cycles(const RU_matrix& M) {
```

And you will also have to replace `print_current_representative_cycles<RU_matrix>(mp);` with `print_current_representative_cycles(mp);` in the `remove_insert_swap_with_cycles` function.

src/Persistence_matrix/example/example_remove_insert_swap_with_cycles.cpp
jlazovskis: Yes, that's right, I had modified the example steps (but not the comments) since the first draft of this file. Comment has been updated to align with what the code actually does.

VincentRouvreau:
Replaced by #1294 (there was a merge conflict), but it will be in next version

PR #1280, opened 2025-12-03
Gudhi random generator for reproducibility

VincentRouvreau:
Thanks @mglisse for the help on the `-fvisibility=hidden`.

## In C++

A `Random_generator` class that embeds `CGAL::Random` class:
* `Random_generator()` uses `std::random_device` instead of `std::time` (cf. `CGAL::Random()`)
* Sets visibility to `CGAL::Random::get_default_random()`

All functions from `src/common/include/gudhi/random_point_generators.h` have an extra `Gudhi::random::Random*` argument (sets by default).
To be discussed here if we should move these functions inside the `Gudhi::random` new namespace. I didn't do it, but I am in favor of that.

## In Python

because of https://github.com/CGAL/cgal/discussions/9135 there is a new `librandom.so` dynamic library. `_random_ext.*.so` and `datasets/generators/_points_ext.*.so` are linking with `librandom.so`.
For windows, I had to do some weird stuff in `datasets/generators/points.py` for the system to find `random.dll`:
```python
# Specific for Windows - To be done before import _ctorus and _sphere
if os.name == 'nt':
    # Add random.dll that is located in "../.."
    dll_dir = Path(__file__).parent / "../.."
    os.add_dll_directory(str(dll_dir.resolve()))
```

A new `gudhi.random.GudhiBitGenerator` python class, that wraps another new `GudhiRandomGenerator` (a nanobind capsule that embeds `Random_generator` C++ class), and works like:
```python
from gudhi.random import GudhiBitGenerator
import numpy as np
# Sets the numpy random generator with the gudhi one, with a seed
np.random.Generator(GudhiBitGenerator(42))
```

`gudhi.datasets.generators.points.sphere` (but also torus and ctorus) have a new `GudhiBitGenerator` argument:
```ipython
>>> from gudhi.datasets.generators import points
>>> from gudhi.random import GudhiBitGenerator
>>> points.sphere(n_samples = 5, ambient_dim = 2, radius = 1, sample = "random", rng=GudhiBitGenerator(42))
array([[ 0.60884028,  0.79329283],
       [-0.97854477, -0.20603432],
       [-0.82108121, -0.57081139],
       [ 0.02495144,  0.99968866],
       [-0.72194001, -0.69195565]])
```

## Results

All methods of `src/common/include/gudhi/random_point_generators.h` have reproducible results (one can test `src/common/utilities/off_file_from_shape_generator.cpp` for instance)
`gudhi.datasets.generators.points.*` have reproducible results.


src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+#ifdef DEBUG_TRACES
+      std::clog << "CGAL::get_default_random(seed=" << seed << "): " << &(CGAL::get_default_random()) << "\n";
+#endif  // DEBUG_TRACES    
+  }
+  Random_generator(Random_generator& other) = delete;
+  Random_generator(Random_generator&& other) = delete;
+  Random_generator& operator=(const Random_generator& other) = delete;
+  Random_generator& operator=(Random_generator&& other) = delete;
+  
+  Random* get_default_random() { return &(CGAL::get_default_random()); }
```
hschreiber: Is there a reason why the method returns a pointer and not directly a const reference?

src/common/include/gudhi/random_point_generators.h
```diff
@@ -186,62 +191,64 @@ std::vector<typename Kernel::Point_d> generate_points_on_torus_3D(std::size_t nu
...
-                                               double radius_noise_percentage = 0.,
-                                               std::vector<typename Kernel::FT> current_point =
-                                                       std::vector<typename Kernel::FT>()) {
+                                            OutputIterator out,
+                                            double radius_noise_percentage = 0.,
+                                            std::vector<typename Kernel::FT> current_point =
+                                              std::vector<typename Kernel::FT>(),
+                                            Gudhi::random::Random* rng =
+                                              Gudhi::random::Random_generator().get_default_random()) {
+  assert(rng);
```
hschreiber: There are several `assert`s like that. Does they have a different role than `GUDHI_CHECK`?

src/python/doc/reproducibility.rst
```diff
@@ -0,0 +1,49 @@
...
+    
+    rng = Generator(GudhiBitGenerator(42)) # Sets the seed
+    rng.random(5)
+    # Outputs: array([0.34270148, 0.11108528, 0.42233896, 0.08111117, 0.85644071])
+
+
+.. note::
+    
+    This is a work in progress, and for the moment, only `datasets generators <datasets.html#datasets-generators>`_
+    can accept a `GudhiBitGenerator` as an argument.
```
hschreiber: We could also mention here the problem with parallelism. And also in the class description?

src/python/gudhi/datasets/generators/points.py
```diff
@@ -42,13 +53,16 @@ def _generate_grid_points_on_torus(n_samples, dim):
     return array_points[array_points_idx].reshape(-1, 2 * dim)
 
 
-def torus(n_samples, dim, sample="random"):
+def torus(n_samples: int, dim: int, sample: str = "random", rng: GudhiBitGenerator = GudhiBitGenerator()):
```
hschreiber: Should we replace `str`, the type of `sample`, with `Literal["random", "grid"]`? (same with other functions)

src/python/gudhi/datasets/generators/points.py
```diff
@@ -42,13 +53,16 @@ def _generate_grid_points_on_torus(n_samples, dim):
...
     """
     Generate points on a flat dim-torus in R^2dim either randomly or on a grid
 
     :param n_samples: The number of points to be generated.
     :param dim: The dimension of the torus on which points would be generated in R^2*dim.
     :param sample: The sample type of the generated points. Can be 'random' or 'grid'.
+    :param rng: Default is `GudhiBitGenerator()`, but it can be set with a seed with `gudhi.random.GudhiBitGenerator(seed)`.
```
hschreiber: It is not so clear from the sentence if `gudhi.random.GudhiBitGenerator(seed)` should be passed as argument instead of default or if you can fix the seed of  `GudhiBitGenerator()` by calling `gudhi.random.GudhiBitGenerator(seed)` before the method.

src/python/include/python_interfaces/random_utils.h
```diff
@@ -0,0 +1,37 @@
...
+namespace random {
+
+void setup_bitgen(Random_generator* rng, nb::capsule capsule) {
+  bitgen_t* bg = static_cast<bitgen_t*>(capsule.data());
+  bg->state = rng;
+  bg->next_uint64 = next_uint64;
+  bg->next_uint32 = next_uint32;
+  bg->next_double = next_double;
+  bg->next_raw    = next_uint64;
+};
```
hschreiber: Do you intend to use this method outside of `_random.cc`?

hschreiber (commented):
For the problem with parallelism you pointed out, I think it is fine for now, as long as we mention this problem somewhere. As fixing this problem is not exactly trivial.

src/python/gudhi/datasets/generators/points.py
VincentRouvreau: It is a good practice. Done on [e2fa9c2](https://github.com/GUDHI/gudhi-devel/pull/1280/commits/e2fa9c2841dd94f6100aaf0f040ceb0a8921cc6d)

src/common/include/gudhi/Random_generator.h
VincentRouvreau: You are right, it is less bug prone when you write it this way.
I had a strange side effect on the python & c++ (cf. sphere and ctorus) side where I had to make rng Optional and set it each time it was called, otherwise it was staying in the same random state. Strange...
Done on [d16e66f](https://github.com/GUDHI/gudhi-devel/pull/1280/commits/d16e66fd5223fd755552fcf5cf850924b0bbc02f)

src/common/include/gudhi/random_point_generators.h
VincentRouvreau: side effect of [d16e66f](https://github.com/GUDHI/gudhi-devel/pull/1280/commits/d16e66fd5223fd755552fcf5cf850924b0bbc02f)

src/python/include/python_interfaces/random_utils.h
VincentRouvreau: Move setup_bitgen inside _random.cc and remove useless random_utils.h on [c9205c0](https://github.com/GUDHI/gudhi-devel/pull/1280/commits/c9205c0655b34631401b2dfea4420db8627b7c08)

src/python/gudhi/datasets/generators/points.py
VincentRouvreau: I tried to make it more explicit by modifying reproducibility text for points generator on [14825c3](https://github.com/GUDHI/gudhi-devel/pull/1280/commits/14825c3109ae899349e84d64098c559f3dc573ed)

src/python/doc/reproducibility.rst
VincentRouvreau: I added some precision about parallelism in the doc on [6221696](https://github.com/GUDHI/gudhi-devel/pull/1280/commits/6221696a56d8f847bb2a5c47b416e891fe503ec8)

hschreiber (approved):
There seems to be a problem with the `Optional[int|GudhiBitGenerator]` syntax in the CI, but otherwise, it looks good to me.

src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+ *      - YYYY/MM Author: Description of the modification
+ */
+
+#ifndef RANDOM_GENERATOR_H_
+#define RANDOM_GENERATOR_H_
+
+#include <cstdint>
+#include <random>
+
+// Must be done before #include <CGAL/Random.h> - Requires also to compile without -fvisibility=hidden
```
mglisse, line 17: `__attribute__((visibility("default")))` should override `-fvisibility=hidden` for this declaration, and if we don't have the flag, we shouldn't need the attribute. Do we really have problems if we don't have both?
"before" is because MSVC requires `dllimport` to be present in the first declaration?

src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+  Random_generator(unsigned int seed) {
+      CGAL::get_default_random() = Random(seed);
+#ifdef DEBUG_TRACES
+      std::clog << "CGAL::get_default_random(seed=" << seed << "): " << &(CGAL::get_default_random()) << "\n";
+#endif  // DEBUG_TRACES    
+  }
+  Random_generator(Random_generator& other) = delete;
+  Random_generator(Random_generator&& other) = delete;
+  Random_generator& operator=(const Random_generator& other) = delete;
+  Random_generator& operator=(Random_generator&& other) = delete;
```
mglisse, line 60: How did you choose the list of things to delete? There is `const` for `=` but not for the constructor. I think deleting one copy and one = should be sufficient to prohibit them all.

src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+  
+  const Random& get_default_random() { return CGAL::get_default_random(); }
+  uint32_t next_uint32() { return CGAL::get_default_random().uniform_int<uint32_t>(0, UINT32_MAX); }
+  uint64_t next_uint64() { return CGAL::get_default_random().uniform_int<uint64_t>(0, UINT64_MAX); }
+  double   next_double() { return CGAL::get_default_random().uniform_real<double>(); }
+
+ private:
+  std::random_device rand_dev;
```
mglisse, line 68: Why store it if it is only used once?

src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+
+#include <CGAL/Random.h>  // for CGAL::get_default_random()
+
+namespace Gudhi {
+namespace random {
+
+// Let's hide CGAL::Random
+using Random = CGAL::Random;
+
+class Random_generator {
```
mglisse, line 40: What is the goal of this class? It looks like it could be replaced by a function `set_seed`. Or is there some place where it is useful for it to be a class?
Ah, the way you use `Gudhi::random::Random_generator().get_default_random()` as a default argument means that you reseed CGAL's default random generator every time there is a call to a function using random numbers that does not explicitly pass a RNG as argument. That defeats most of the purpose of having a global object. IMO, in the case with no explicit argument, 2 better options would be
1. use the default global random generator **without reseeding**.
2. use a fresh separate instance of `Random`, seeded however you want.

The second one complicates the life of people who want reproducibility, now instead of setting a seed at the beginning of the program they have to modify all of their calls.

src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+  }
+  Random_generator(Random_generator& other) = delete;
+  Random_generator(Random_generator&& other) = delete;
+  Random_generator& operator=(const Random_generator& other) = delete;
+  Random_generator& operator=(Random_generator&& other) = delete;
+  
+  const Random& get_default_random() { return CGAL::get_default_random(); }
+  uint32_t next_uint32() { return CGAL::get_default_random().uniform_int<uint32_t>(0, UINT32_MAX); }
+  uint64_t next_uint64() { return CGAL::get_default_random().uniform_int<uint64_t>(0, UINT64_MAX); }
+  double   next_double() { return CGAL::get_default_random().uniform_real<double>(); }
```
mglisse, line 65: Should those (and their aliases below) move to the nanobind file, since they are only used there?

src/common/include/gudhi/random_point_generators.h
```diff
@@ -6,19 +6,22 @@
...
+#include <gudhi/Random_generator.h>
+
 #include <CGAL/number_utils.h>
-#include <CGAL/Random.h>
 #include <CGAL/point_generators_d.h>
 #include <CGAL/version.h>  // for CGAL_VERSION_NR
 
 #include <vector>  // for vector<>
 #include <boost/math/constants/constants.hpp> // for pi constant
+#include <thread>
```
mglisse, line 24: ?

src/common/include/gudhi/random_point_generators.h
```diff
@@ -122,13 +125,12 @@ std::vector<typename Kernel::Point_d> generate_points_on_plane(std::size_t num_p
 }
 
 template <typename Kernel>
-std::vector<typename Kernel::Point_d> generate_points_on_moment_curve(std::size_t num_points, int dim,
-                                                                      typename Kernel::FT min_x,
-                                                                      typename Kernel::FT max_x) {
+std::vector<typename Kernel::Point_d>
+generate_points_on_moment_curve(std::size_t num_points, int dim, typename Kernel::FT min_x, typename Kernel::FT max_x,
+                                Gudhi::random::Random rng = Gudhi::random::Random_generator().get_default_random()) {
```
mglisse, line 130: I looked at CGAL, and they seem to pass `Random& rnd = CGAL::get_default_random()` (i.e. a non-const reference, not a copy). Is the copy done on purpose?

src/common/include/gudhi/Random_generator.h
```diff
@@ -0,0 +1,86 @@
...
+#ifdef DEBUG_TRACES
+      std::clog << "CGAL::get_default_random(seed=" << seed << "): " << &(CGAL::get_default_random()) << "\n";
+#endif  // DEBUG_TRACES    
+  }
+  Random_generator(Random_generator& other) = delete;
+  Random_generator(Random_generator&& other) = delete;
+  Random_generator& operator=(const Random_generator& other) = delete;
+  Random_generator& operator=(Random_generator&& other) = delete;
+  
+  const Random& get_default_random() { return CGAL::get_default_random(); }
```
mglisse, line 62: Most methods of Random are non-const (otherwise they would have needed to mark all the members `mutable`).
If I make a copy of the default Random 3 times and generate one number with each, I'll get 3 times the same number, copying Random doesn't seem like the normal way to use it.

src/common/utilities/off_file_from_shape_generator.cpp
```diff
@@ -24,14 +26,14 @@ typedef CGAL::Epick_d< CGAL::Dynamic_dimension_tag > K;
 typedef K::Point_d Point;
 
 void usage(char * const progName) {
-  std::cerr << "Usage: " << progName << " in|on sphere|cube off_file_name points_number[integer > 0] " <<
-      "dimension[integer > 1] radius[double > 0.0 | default = 1.0]" << std::endl;
+  std::cerr << "Usage: " << progName << " in|on sphere|cube|curve|torus|klein off_file_name points_number(integer > 0) " <<
+      "dimension(integer > 1) [radius(double > 0.0, default = 1.0)] [seed(unsigned integer)]" << std::endl;
```
mglisse, line 30: I think I've seen [opt1 [opt2]] when both are optional but you cannot specify opt2 without specifying opt1.

src/python/doc/reproducibility.rst
```diff
@@ -0,0 +1,50 @@
+:orphan:
+
+.. To get rid of WARNING: document isn't included in any toctree
+
+===============
+Reproducibility
+===============
+
+Some of the GUDHI functionnalities are using randomness. In order to reproduce the results, a specific GUDHI random
+generator is available.
```
mglisse, line 10: As we discussed before, numpy was never the issue for reproducibility, users could simply call `numpy.random.seed`.

src/python/gudhi/datasets/generators/_points.cc
```diff
@@ -72,46 +77,18 @@ NB_MODULE(_points_ext, m)
 {
   m.attr("__license__") = "LGPL v3";
 
-  m.def("sphere",
```
mglisse, line 75: Can you comment why it is better to have 2 steps (`_sphere` and `sphere`) instead of one? Is it to make the doc prettier?

src/python/gudhi/datasets/generators/points.py
```diff
@@ -5,20 +5,32 @@
...
 
-from ._points_ext import ctorus, sphere
+from numpy.random import Generator
+from gudhi.random import GudhiBitGenerator
 
+# Specific for Windows - To be done before import _ctorus and _sphere
+if os.name == 'nt':
+    # Add random.dll that is located in "../.."
+    dll_dir = Path(__file__).parent / "../.."
+    os.add_dll_directory(str(dll_dir.resolve()))
```
mglisse, line 26: Man that's ugly... I guess loading the library from the toplevel `__init__.py` will be the natural evolution. That or turning the library into a module.
I guess this is ok for now.
What does delvewheel do with the library?

src/python/gudhi/datasets/generators/points.py
```diff
@@ -5,20 +5,32 @@
...
+    
+from ._points_ext import _ctorus, _sphere
 
-def _generate_random_points_on_torus(n_samples, dim):
+def _generate_random_points_on_torus(n_samples: int, dim: int, rng: GudhiBitGenerator = GudhiBitGenerator()):
```
mglisse, line 30: Looks like like one could accept any BitGenerator, it isn't restricted to the one from Gudhi?

src/python/gudhi/datasets/generators/points.py
```diff
@@ -41,14 +53,28 @@ def _generate_grid_points_on_torus(n_samples, dim):
         array_points_idx[..., i] = x
     return array_points[array_points_idx].reshape(-1, 2 * dim)
 
+def _get_bit_generator(rng: Optional[Union[int, GudhiBitGenerator]]):
```
mglisse, line 56: I thought you removed support for Python 3.9 in this PR? Ah, it was only on windows, but why then?

src/python/gudhi/datasets/generators/points.py
```diff
@@ -41,14 +53,28 @@ def _generate_grid_points_on_torus(n_samples, dim):
...
+def torus(n_samples: int, dim: int, sample: Literal["random", "grid"] = "random",
+          rng: Optional[Union[int, GudhiBitGenerator]] = None):
     """
     Generate points on a flat dim-torus in R^2dim either randomly or on a grid
 
     :param n_samples: The number of points to be generated.
     :param dim: The dimension of the torus on which points would be generated in R^2*dim.
     :param sample: The sample type of the generated points. Can be 'random' or 'grid'.
+    :param rng: For :doc:`reproducible results <reproducibility>`, pass a seed with `rng = seed` or
+        `rng = gudhi.random.GudhiBitGenerator(seed)`. Default is `None`, which means no reproducibility.
```
mglisse, line 75: In practice, the user wouldn't pass `gudhi.random.GudhiBitGenerator(seed)` for every call, they would create an object `rng` (probably with a seed) and pass the same object to several functions.

mglisse (commented):
I didn't finish reading the Python part, but let me post what I have.

The principle of replacing in C++ a local object `CGAL::Random rng;` with a function parameter that has a default value is good.

Please try to minimize unrelated formatting changes (next time, if you need to do such changes, do them first in a separate PR that you can merge in master without review), it is annoying to go through a diff checking if it only changed whitespace or more.

src/common/include/gudhi/Random_generator.h
VincentRouvreau: No need to store random_device on c09726984

src/common/include/gudhi/random_point_generators.h
VincentRouvreau: This is a forgetting from my side. I was using it when trying to see what happens in parallel. Removed on a728bfd16

src/common/include/gudhi/random_point_generators.h
```diff
@@ -208,35 +210,34 @@ static void generate_grid_points_on_torus_d(const Kernel &k, int dim, std::size_
...
   } else {
     for (std::size_t i = 0; i < num_points;) {
       double radius_noise_ratio = 1.;
       if (radius_noise_percentage > 0.) {
-        radius_noise_ratio = rng.get_double(
-                                            (100. - radius_noise_percentage) / 100.,
-                                            (100. + radius_noise_percentage) / 100.);
+        radius_noise_ratio = rng.get_double((100. - radius_noise_percentage) / 100.,
+                                             (100. + radius_noise_percentage) / 100.);
       }
```
mglisse, line 241: Is there a non-whitespace change here?

src/common/include/gudhi/random_point_generators.h
```diff
@@ -336,21 +338,16 @@ std::vector<typename Kernel::Point_d> generate_points_on_two_spheres_d(std::size
...
-          k.scaled_vector_d_object();
+      typename Kernel::Point_to_vector_d k_pt_to_vec = k.point_to_vector_d_object();
+      typename Kernel::Vector_to_point_d k_vec_to_pt = k.vector_to_point_d_object();
+      typename Kernel::Scaled_vector_d k_scaled_vec =  k.scaled_vector_d_object();
       p = k_vec_to_pt(k_scaled_vec(k_pt_to_vec(p), radius_noise_ratio));
     }
 
-    typename Kernel::Translated_point_d k_transl =
-        k.translated_point_d_object();
+    typename Kernel::Translated_point_d k_transl = k.translated_point_d_object();
```
mglisse, line 350: or here?

src/common/include/gudhi/random_point_generators.h
```diff
@@ -461,8 +461,8 @@ std::vector<typename Kernel::Point_d> generate_points_on_klein_bottle_4D(std::si
     Point p = construct_point(k,
                               (a + b * cos(v)) * cos(u) + (noise == 0. ? 0. : rng.get_double(0, noise)),
                               (a + b * cos(v)) * sin(u) + (noise == 0. ? 0. : rng.get_double(0, noise)),
-                              b * sin(v) * cos(u / 2) + (noise == 0. ? 0. : rng.get_double(0, noise)),
-                              b * sin(v) * sin(u / 2) + (noise == 0. ? 0. : rng.get_double(0, noise)));
+                              b * sin(v) * cos(u / 2) + (noise == 0. ? 0.   : rng.get_double(0, noise)),
+                              b * sin(v) * sin(u / 2) + (noise == 0. ? 0.   : rng.get_double(0, noise)));
```
mglisse, line 465: or there?

mglisse (commented):
In case I am not reading things properly.

src/python/gudhi/datasets/generators/points.py
VincentRouvreau: I agree, and now I think that maybe I explained badly this point.
A user could do:
```python
seed = 14
sph = points.sphere(n_samples = 50, ambient_dim = 2, radius = 1, sample = "random", rng=seed)
trs = points.ctorus(n_samples = 50, dim = 3, rng=seed)
```

Or:
```python
from gudhi.random import GudhiBitGenerator
rng = GudhiBitGenerator(14)
sph = points.sphere(n_samples = 50, ambient_dim = 2, radius = 1, sample = "random", rng=rng)
trs = points.ctorus(n_samples = 50, dim = 3, rng=rng)
```

Both will be reproducible results, even if both versions will not produce the same results (as `seed` is reset in first version when calling `trs = points.ctorus(n_samples = 50, dim = 3, rng=seed)`)

src/common/include/gudhi/random_point_generators.h
VincentRouvreau: Ok, sorry, now I got your point.

src/python/gudhi/datasets/generators/points.py
VincentRouvreau: Yes I know, and I am not proud of this move, but delvewheel was not fixing this one.
I will have a closer look to see if I can fix it.

src/common/include/gudhi/Random_generator.h
VincentRouvreau: Here, I wanted to forbid copy constructor, move constructor, copy assignment and move assignment.
I tested it because I didn't know that only those 2 lines where sufficient enough:
```cpp
  Random_generator(const Random_generator& other) = delete;
  Random_generator& operator=(const Random_generator& other) = delete;
```
Sometimes being more expressive doesn't hurt, no ?
I let the 4 of them but fixed `Random_generator(Random_generator& other) = delete;` to `Random_generator(const Random_generator& other) = delete;` on b5dc09009

VincentRouvreau:
Replaced by #1335

PR #1283, opened 2025-12-08
[Common] fixes forgotten static_extends method of Simple_mdspan

hschreiber:
`static_extent` was true like that before I added the possibility of static extends and I forgot to update it as I am not using this method anywhere (just part of the `std::md_span` interface). So I also added tests for it.

### DIFF
--- a/src/common/include/gudhi/simple_mdspan.h
+++ b/src/common/include/gudhi/simple_mdspan.h
@@ -177,6 +177,9 @@ class extents
   {
     stream << "[ " << sizeof...(Extents) << " ] ";
     ((stream << Extents << ' '), ...);
+    stream << " [";
+    for (rank_type r = 0; r < e.rank(); ++r) stream << e.extent(r) << " ";
+    stream << "]";
 
     return stream;
   }


src/common/test/test_simple_mdspan.cpp
VincentRouvreau, line 87: I have this warning here:
```log
In file included from /home/gailuron/workspace/gudhi/gudhi-devel/src/common/test/test_simple_mdspan.cpp:18:
In member function ‘T& Gudhi::Simple_mdspan<T, Extents, LayoutPolicy>::operator[](const IndexRange&) const [with IndexRange = std::initializer_list<long unsigned int>; T = int; Extents = const Gudhi::extents<long unsigned int, 18446744073709551615, 18446744073709551615>; LayoutPolicy = Gudhi::layout_right]’,
    inlined from ‘void test_simple_mdspan_access::test_method()’ at /home/gailuron/workspace/gudhi/gudhi-devel/src/common/test/test_simple_mdspan.cpp:87:17:
/home/gailuron/workspace/gudhi/gudhi-devel/src/common/include/gudhi/simple_mdspan.h:394:34: warning: iteration 4611686018427387903 invokes undefined behavior [-Waggressive-loop-optimizations]
  394 |     return *(ptr_ + map_(indices));
      |                                  ^
/home/gailuron/workspace/gudhi/gudhi-devel/src/common/test/test_simple_mdspan.cpp: In member function ‘void test_simple_mdspan_access::test_method()’:
/home/gailuron/workspace/gudhi/gudhi-devel/src/common/test/test_simple_mdspan.cpp:86:31: note: within this loop
   86 |     for (std::size_t j = 0; j != ms2.extent(1); j++) {
      |                             ~~^~~~~~~~~~~~~~~~
```

while `i` and `j` seem ok if I print their values:
```log
i=0 - j=0
i=0 - j=1
i=0 - j=2
i=0 - j=3
i=0 - j=4
i=0 - j=5
i=1 - j=0
i=1 - j=1
i=1 - j=2
i=1 - j=3
i=1 - j=4
i=1 - j=5
```

hschreiber: This warning was already there before and I suspect it is a false positive. When declaring this class, all dynamic extends are giving the value `std::numeric_limits<std::size_t>::max()` (which seems to be equal to 18446744073709551615) in compile time, but then are replaced by the given values at construction. The compiler is probably giving out the warning before making this change.

That is the only explanation I found, because the class always worked, when I used it...

VincentRouvreau: In `simple_mdspan.h` I added some traces in `operator[]` like:
```cpp
  template <class IndexRange = std::initializer_list<index_type> >
  reference operator[](const IndexRange& indices) const
  {
    for (auto idx : indices)
      std::cout << idx << ", ";
    std::cout << "\n";
    return *(ptr_ + map_(indices));
  }
```

... and the warning disappears...
I also do not have the warning with clang. Sorry for the noise

PR #1285, opened 2025-12-16
[Persistence_matrix] Update single cycle

hschreiber:
I added the possibility to update just a particular representative cycle in the matrix to avoid useless run times for uninteresting cycles. And the possibility to update only cycles of a particular dimension. Most people will only be interested by 1-cycles.

### DIFF
--- a/src/Persistence_matrix/include/gudhi/Persistence_matrix/Id_to_index_overlay.h
+++ b/src/Persistence_matrix/include/gudhi/Persistence_matrix/Id_to_index_overlay.h
@@ -625,7 +625,7 @@ class Id_to_index_overlay
    *
    * @return A const reference to the vector of representative cycles.
    */
-  const std::vector<Cycle>& get_all_representative_cycles();
+  const std::vector<Cycle>& get_all_representative_cycles() const;
   /**
    * @brief Only available if @ref PersistenceMatrixOptions::can_retrieve_representative_cycles is true.
    * Returns the cycle representing the given bar.
@@ -633,7 +633,7 @@ class Id_to_index_overlay
    * @param bar A bar from the current barcode.
    * @return A const reference to the cycle representing @p bar.
    */
-  const Cycle& get_representative_cycle(const Bar& bar);
+  const Cycle& get_representative_cycle(const Bar& bar) const;
 
  private:
   using Dictionary = typename Master_matrix::template Dictionary<Index>;
@@ -1043,14 +1043,14 @@ inline void Id_to_index_overlay<Underlying_matrix, Master_matrix>::update_repres
 
 template <class Underlying_matrix, class Master_matrix>
 inline const std::vector<typename Id_to_index_overlay<Underlying_matrix, Master_matrix>::Cycle>&
-Id_to_index_overlay<Underlying_matrix, Master_matrix>::get_all_representative_cycles()
+Id_to_index_overlay<Underlying_matrix, Master_matrix>::get_all_representative_cycles() const
 {
   return matrix_.get_all_representative_cycles();
 }
 
 template <class Underlying_matrix, class Master_matrix>
 inline const typename Id_to_index_overlay<Underlying_matrix, Master_matrix>::Cycle&
-Id_to_index_overlay<Underlying_matrix, Master_matrix>::get_representative_cycle(const Bar& bar)
+Id_to_index_overlay<Underlying_matrix, Master_matrix>::get_representative_cycle(const Bar& bar) const
 {
   return matrix_.get_representative_cycle(bar);
 }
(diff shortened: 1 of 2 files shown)


src/Persistence_matrix/include/gudhi/Persistence_matrix/chain_rep_cycles.h
```diff
@@ -63,20 +68,34 @@ class Chain_representative_cycles
...
    *
    * @return A const reference to a vector of @ref Matrix::Cycle containing all representative cycles.
    */
   const std::vector<Cycle>& get_representative_cycles();
   /**
    * @brief Returns the representative cycle corresponding to the given bar.
-   * If the matrix is modified later after the first call,
-   * @ref update_representative_cycles has to be called to update the returned cycles.
+   * If the matrix was modified since the last call, @ref update_representative_cycles or
+   * @ref update_representative_cycle has to be called to update the returned cycle.
```
VincentRouvreau: I am not a big fan with the naming of the methods get|update_representative_cycle versus get|update_representative_cycle**s**.
I think a more explicit name would help. Something like `get|update_representative_cycle_from_bar` and `get|update_list_of_representative_cycles` or some other proposal.

src/Persistence_matrix/include/gudhi/Persistence_matrix/chain_rep_cycles.h
```diff
@@ -122,33 +141,96 @@ inline void Chain_representative_cycles<Master_matrix>::update_representative_cy
...
+  if (birthToCycle_.size() <= bar.birth) {
+    birthToCycle_.resize(bar.birth + 1, nullValue);
+  }
+  if (birthToCycle_[bar.birth] == nullValue) {
+    birthToCycle_[bar.birth] = representativeCycles_.size();
+    representativeCycles_.resize(representativeCycles_.size() + 1);
+  }
+
+  if constexpr (Master_matrix::Option_list::has_vine_update) {
+    for (ID_index i = 0; i < nberColumns; i++) {
```
VincentRouvreau: Could this loop also be parallelized with TBB ?

src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -1409,25 +1409,38 @@ class Matrix
...
-   * @ref get_representative_cycles again if the matrix was modified in between. Otherwise the old cycles will be
-   * returned.
+   * the representative cycles of the current state of the filtration represented by the matrix. It needs to be called
+   * before calling @ref get_representative_cycles if the matrix was modified since last call. Otherwise the old cycles
+   * will be returned.
+   *
+   * @param dim If different from default value, only the cycles of the given dimension are updated.
+   * All others are erased.
+   */
+  void update_representative_cycles(Dimension dim = get_null_value<Dimension>());
```
VincentRouvreau: This is not the purpose of this PR, just for my curiosity, but should these methods get|update_representative_cycle(s) could be virtual, no ? Or maybe I missed something in the design of your code.

src/Persistence_matrix/include/gudhi/Persistence_matrix/chain_rep_cycles.h
hschreiber: As both don't have the same number of arguments, I thought it should be fine. But I renamed `get|update_representative_cycles` to `get|update_all_representative_cycles` to be more explicit without making the names too long.

src/Persistence_matrix/include/gudhi/Persistence_matrix/chain_rep_cycles.h
hschreiber: The `if` statement in the loop is only visited once, so I rewrote this part to make it clearer. Now the only loop left is `while (i < nberColumns && get_position(i) != bar.birth) ++i;` and I don't think using TBB would make it much faster?

src/Persistence_matrix/include/gudhi/Matrix.h
hschreiber: The class `Matrix` is never inherited, I use composition everywhere except for the mixins. So, `Chain|RU_representative_cycles` are inherited, not the other way around and so they cannot be virtual methods.

VincentRouvreau: Argh, yes sorry about the comment... It was hard for me to enter the design of the code. I asked copilot for a class diagram of the persistence matrix classes and it shows me some inheritance that does not exist in your code. I will be more careful next time

src/Persistence_matrix/include/gudhi/Persistence_matrix/chain_rep_cycles.h
VincentRouvreau: What disturbed me was that there was no `break` in the `for` loop. I am ok with a `break` or with the `while` proposal

hschreiber: Yes, I realized that I forgot it when I looked at it again. This is another reason why I rewrote it, it is easier to not make silly mistake that way...

Issue #1292, opened 2026-01-14
Rely more on mypy for the arguments type detection when sphinx generates the documentation


VincentRouvreau:
We started to use mypy for python, but IMHO, I think we should also rely on it for the documentation generation.

For example, in the `dealunay_complex.py`, we say:
```python
    def __init__(
        self,
        points: Sequence[Sequence[float]] = [],
        weights: Optional[Sequence[float]] = None,
        precision: Literal["fast", "safe", "exact"] = "safe",
    ):
        """Args:
              points (Sequence[Sequence[float]]): A list of points in d-Dimension.
              weights (Optional[Sequence[float]]): A list of weights. If set, the number of weights must correspond to
                  the number of points.
              precision (str): Complex precision can be `'fast'`, `'safe'` or `'exact'`. Default is `'safe'`.

        :raises ValueError: In case of inconsistency between the number of points and weights.
        """
```

which gives the following documentation:

<img width="1197" height="223" alt="Image" src="https://github.com/user-attachments/assets/fde5b7de-922b-4d52-9d51-b4ce8bdae4bf" />

But, if we rely on mypy for the arguments type detection when sphinx generates the documentation (here I remove the arguments types in between parenthesis):
```python
    def __init__(
        self,
        points: Sequence[Sequence[float]] = [],
        weights: Optional[Sequence[float]] = None,
        precision: Literal["fast", "safe", "exact"] = "safe",
    ):
        """Args:
              points: A list of points in d-Dimension.
              weights: A list of weights. If set, the number of weights must correspond to
                  the number of points.
              precision: Complex precision can be `'fast'`, `'safe'` or `'exact'`. Default is `'safe'`.

        :raises ValueError: In case of inconsistency between the number of points and weights.
        """
```

The documentation looks like:

<img width="1197" height="223" alt="Image" src="https://github.com/user-attachments/assets/7ee7ef93-ede5-4b3d-9493-4fbe832fccbe" />

PR #1297, opened 2026-02-19
Vineyard module

hschreiber:
C++ and Python module for Vineyard computations.

The PR is ready, but there are two points I would like to discuss (in addition to the review):
- what should be the import path for the Python interface? For now it is `from gudhi.vineyard import Vineyard`, which I don't really like. But the answer could depend on the change of interface we plan for persistence in general.
- In the plottings, how should we handle vines with coordinates at infinity? In the persistence diagram, we added an artificial "inf bar" to the diagram, but as the vineyard are 3D, it does not look as nice... For now, I just draw them a bit further away than the others, but that is not really satisfying either.

**PR dependecies:**
- #1298


src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -1711,13 +1744,35 @@ Matrix<PersistenceMatrixOptions>::insert_boundary(ID_index cellIndex, const Boun
...
+{
   static_assert(isNonBasic && PersistenceMatrixOptions::has_vine_update,
                 "'insert_maximal_cell' is not available for the chosen options.");
-  static_assert(PersistenceMatrixOptions::is_of_boundary_type,
+
+  if constexpr (isNonBasic && !PersistenceMatrixOptions::is_of_boundary_type &&
+                PersistenceMatrixOptions::column_indexation_type == Column_indexation_types::CONTAINER)
+    return matrix_.insert_maximal_cell(columnIndex, boundary, dim);
+  else
+    matrix_.insert_maximal_cell(columnIndex, boundary, dim);
```
VincentRouvreau, line 1757: We should return something in the else case
```suggestion
    return matrix_.insert_maximal_cell(columnIndex, boundary, dim);
```

src/Persistence_matrix/include/gudhi/Matrix.h
```diff
@@ -1711,13 +1744,35 @@ Matrix<PersistenceMatrixOptions>::insert_boundary(ID_index cellIndex, const Boun
...
+{
+  static_assert(isNonBasic && PersistenceMatrixOptions::has_vine_update,
                 "'insert_maximal_cell' is not available for the chosen options.");
-  matrix_.insert_maximal_cell(columnIndex, boundary, dim);
+
+  if constexpr (isNonBasic && !PersistenceMatrixOptions::is_of_boundary_type &&
+                PersistenceMatrixOptions::column_indexation_type == Column_indexation_types::CONTAINER)
+    return matrix_.insert_maximal_cell(columnIndex, cellIndex, boundary, dim);
+  else
+    matrix_.insert_maximal_cell(columnIndex, cellIndex, boundary, dim);
```
VincentRouvreau, line 1775: We should return something in the else case
```suggestion
    return matrix_.insert_maximal_cell(columnIndex, cellIndex, boundary, dim);
```

src/Vineyard/include/gudhi/vineyard_base.h
```diff
@@ -0,0 +1,392 @@
...
+      for (const auto& v : vyd.matrix_.get_column(i)) stream << v << ", ";
+      stream << "]\n";
+    }
+    stream << "]\n";
+    stream << "Permutation:\n";
+    for (auto v : vyd.order_) {
+      stream << v << " ";
+    }
+    stream << "\n";
+    if constexpr (Matrix_options::is_of_boundary_type) {
```
VincentRouvreau: From the code above, it seems `vyd.idToPos_` is only set when `!Matrix_options::is_of_boundary_type`. So I imagine you want to print it only when `!Matrix_options::is_of_boundary_type`
```suggestion
    if constexpr (!Matrix_options::is_of_boundary_type) {
```

src/Vineyard/include/gudhi/vineyard_base.h
```diff
@@ -0,0 +1,392 @@
...
+  void update(const Filtration_range& filtrationValues)
+  {
+    GUDHI_CHECK(filtrationValues.size() == order_.size(),
+                std::invalid_argument("Filtration value container size is not matching."));
+
+    for (Index i = 1; i < order_.size(); i++) {
+      int curr = i;
+      // speed up when ordered by dim, to avoid unnecessary swaps
+      while (curr > 0 && matrix_.get_column_dimension(curr) == matrix_.get_column_dimension(curr - 1) &&
+             filtrationValues[order_[curr]] < filtrationValues[order_[curr - 1]]) {
```
VincentRouvreau, line 287: It seems that this speed up requires some constraints that are not asserted, nor documented, no ?
Specifically when the user provides its own filtration.

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+        :param number_of_updates: Optional (for optimization purposes). Will allocate memory space for
+            `number_of_updates` additional steps in the vineyard after initialization. Defaults to 0.
+        :type number_of_updates: int, optional
+        :raises ValueError: If both `filtered_cpx` and (`boundaries` or `dimensions` or `filtration_values`) are
+            provided or if none of both are provided.
+        :return: Current state of the vineyard. See :meth:`get_current_vineyard_view`.
+        :rtype: list[read-only np.ndarray]
+        """
+        if filtered_cpx is not None:
+            if boundaries != None or dimensions != None or filtration_values != None:
```
VincentRouvreau: In Python we say is / is not with None
```suggestion
            if boundaries is not None or dimensions is not None or filtration_values is not None:
```

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+        :param filtration_values: Array of new filtration values such that `filtration_values[i]` corresponds to
+            the new value for `boundaries[i]` provided to :meth:`initialize`. If `boundaries` was not provided to
+            :meth:`initialize`, provide the argument `filtered_cpx` instead. Defaults to `None`.
+        :type filtration_values: np.ndarray or `None`
+        :raises ValueError: If both `filtered_cpx` and `filtration_values` are provided or none of both.
+        :return: Current state of the vineyard. See :meth:`get_current_vineyard_view`.
+        :rtype: list[read-only np.ndarray]
+        """
+        if filtered_cpx is not None:
+            if filtration_values != None:
```
VincentRouvreau: In Python we say is / is not with None
```suggestion
            if filtration_values is not None:
```

src/Vineyard/include/gudhi/vineyard_builder.h
```diff
@@ -0,0 +1,334 @@
...
+  std::vector<Index> numberOfBars_;       /**< Number fo vines by dimension. */
+  std::optional<Dimension> repCyclesDim_; /**< Dimension of stored cycles. */
+  std::optional<std::vector<std::tuple<Cycle, Dimension, Index> > > latest_representative_cycles_; /**< Cycles. */
+
+  template <class Filtration_range>
+  bool _store_cycle(const typename Base::Bar& bar, const Filtration_range& filtrationValues) const
+  {
+    if (!repCyclesDim_) return false;
+    if (*repCyclesDim_ != Base::nullDimension && bar.dim != *repCyclesDim_) return false;
+    if (filtrationValues[bar.birth] == Bar::inf) return false;
```
VincentRouvreau: Shouldn't we `GUDHI_CHECK` the cases where `birth == inf` ?
Here we just don't store it.

src/python/test/test_vineyard.py
```diff
@@ -0,0 +1,55 @@
...
+
+# not really safe, but as giving command line arguments to pytest while using ctest ask for some
+# more complicated gymnastic, I opted for this option for now...
+def get_file_path_prefix():
+    parent_dir_name = Path(__file__).resolve().parent.parent.parent.name
+    if parent_dir_name == "src":
+        # devel version
+        return "../../../data/points/unknotting_rings/rings"
+    # user version
+    return "../../data/points/unknotting_rings/rings"
```
VincentRouvreau: This is fragile...
I think we should think of fetching the dataset. Maybe a bit more complicated to do but we can think about it

data/points/unknotting_rings/README
```diff
@@ -0,0 +1,3 @@
+Data set of two linked rings unknotting step by step.
```
VincentRouvreau, line 1: This file should be named README of rings.COPYRIGHT.
We should also mention that the license of the dataset is MIT, or link to https://github.com/frabertoglio/TDAdynamicdata/blob/main/LICENSE

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+from gudhi.reader_utils import read_lower_triangular_matrix_from_csv_file
+
+
+# for debug only
+def _verify_validity(
+    vineyard: list[np.ndarray],
+    path_prefix: str,
+    path_suffix: str = ".txt",
+    first_index: int = 0,
+    delimiter: str = None,
```
VincentRouvreau:
```suggestion
    delimiter: Optional[str] = None,
```
requires
```python
from typing import Optional
```

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+    number_of_updates: int = None,
```
VincentRouvreau:
```suggestion
    number_of_updates: Optional[int] = None,
```

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+        :param cycles_dim: If `store_latest_cycles` is set to `True`: if set to a positive number, only the cycle of
+            dimension `cycles_dim` will be stored, otherwise stores all cycles. If `store_latest_cycles` is set to
+            `False`: ignored. Defaults to -1.
+        :type cycles_dim: int, optional
+        """
+        super().__init__(store_latest_cycles, cycles_dim)
+
+    def initialize(
+        self,
+        filtered_cpx: SimplexTree | CubicalComplex | PeriodicCubicalComplex = None,
```
VincentRouvreau:
```python
      variable: Optional[type] = None
```

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+            raise ValueError(
+                "Either a filtered complex or (boundaries/dimensions/filtration values) should be not `None`."
+            )
+        super()._initialize(boundaries, dimensions, filtration_values, number_of_updates)
+        return self.get_current_vineyard_view()
+
+    def update(
+        self,
+        filtered_cpx: SimplexTree | CubicalComplex | PeriodicCubicalComplex = None,
+        filtration_values: np.ndarray = None,
```
VincentRouvreau:
```python
      variable: Optional[type] = None
```

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1009 @@
...
+            return self.get_current_vineyard_view()
+
+        if filtration_values is None:
+            raise ValueError(
+                "Either a filtered complex or filtration values should be not `None`."
+            )
+        super()._update(filtration_values)
+        return self.get_current_vineyard_view()
+
+    def get_current_vineyard_view(self, dim: int = None) -> list[np.ndarray] | np.ndarray:
```
VincentRouvreau:
```python
      variable: Optional[type] = None
```

src/Persistence_matrix/include/gudhi/Matrix.h
hschreiber: The return type in the else case is `void` so there is no need to return.

It is, by the way, part of PR #1292, not really this one.

src/Vineyard/include/gudhi/vineyard_base.h
hschreiber: The comment above was a bit misleading here indeed, but there is no assert needed, as I know for sure that what I want is ordered by dimension (it is not the input `filtrationValues`, but private member `order_`). So there is also no need to document it, as it is a technical detail the user does not need to know.

src/Vineyard/include/gudhi/vineyard_builder.h
hschreiber: `birth == inf` is not wrong, so it should not throw, even in debug mode. It just means that `death` is also `inf` and therefore the bar is not only trivial, but only happens at infinity, i.e. never. Storing its representative cycle feels like a waste of space. But storing the bar it-self is important as we need to keep the matching from barcode to barcode in vineyard, so the number of bars have to be constant.

src/python/test/test_vineyard.py
hschreiber: I completely agree on this, but I am not sure how to do it. The best would be to have access to `${CMAKE_CURRENT_SOURCE_DIR}`, that is, to pass the path via command line to ctest and pytest, but this requires an additional `conftest.py` file. Should I do that?

src/Vineyard/include/gudhi/vineyard_base.h
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

data/points/unknotting_rings/README
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

Is it right that way?

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

~~But I used the `str | None` syntax instead as it is the recommended one: https://typing.python.org/en/latest/reference/best_practices.html#shorthand-syntax~~
Nevermind, that was the thing which does not work with python 3.9...

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

src/python/gudhi/vineyard.py
hschreiber: Done in 754874dce238b5e84aa4bdf88c7fa822a179f06d.

src/python/test/test_vineyard.py
hschreiber: Tested on 9b346aff9f794dcfef7a4cad9406619f3016b715 and 5e88fd854cc99ecd517dd3001641cf10c6059ba3 (forgot to commit the `conftest.py` file on the first one...). It was easier than I thought.

Works on my computer. Let's see how it passes on the CI.

hschreiber: Ah, the CI seems to call `pytest` by itself sometimes (without the cmake), e.g. `python -B -m pytest test/*.py --cov-report html --cov=gudhi`, and so it does not work as the path is not passed...

**Edit:** Added a skip if path is not given. See if it works.

VincentRouvreau:
> what should be the import path for the Python interface? For now it is from gudhi.vineyard import Vineyard, which I don't really like. But the answer could depend on the change of interface we plan for persistence in general.

And why not stick to the documentation categories that are already in place ? Something like:
```python
from gudhi.descriptors import Vineyard
```

Just add a `descriptors` folder with an `__init__.py` and move `vineyard.py` in it.
To be discussed.

mglisse:
> > what should be the import path for the Python interface? For now it is from gudhi.vineyard import Vineyard, which I don't really like.

I don't find that so ugly. I've seen elsewhere `from sklearn.pipeline import make_pipeline`, `from datetime import datetime`, etc.

> But the answer could depend on the change of interface we plan for persistence in general.

Of course.

> ```python
> from gudhi.descriptors import Vineyard
> ```

I don't find the word "descriptor" very clear...

src/common/doc/main_page.md
```diff
@@ -434,6 +434,29 @@
...
 
+### Vineyard
+
+<table>
+  <tr>
+    <td width="35%" rowspan=2>
+      \image html "vineyard_ex2.png"
+    </td>
+    <td width="50%">
+    Vineyards record changes in the barcode when cells change positions in a filtration.
```
VincentRouvreau, line 445: Here, we could be more verbose to advertise the feature, idk, something like:
```suggestion
    A vine is the trajectory traced by a single point in the persistence diagram as the filtration evolves continuously over time.
    If you stack all the persistence diagrams over time into a 3D space (x = birth, y = death, z = time), each topological feature traces a curve aka. a vine. A vineyard is the collection of all vines across all dimensions, forming a surface-like structure in this 3D space.
    The implementation is based on @cite vineyards and @cite zigzag.
```

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1061 @@
...
+                *[[np.min(scaling), np.max(scaling)]] * 2,
+                [getattr(ax, "get_{}lim".format("z"))()],
+            )
+        ax.set_aspect("equalxy", "box")
+
+        plt.show()
+        return
+
+
+class PointCloudRipsVineyard:
```
VincentRouvreau, line 465: An example/benchmark I used to test the feature:
```python
import numpy as np
from gudhi.vineyard import PointCloudRipsVineyard

radius = 1.
pts = points.sphere(n_samples = 70, ambient_dim = 2, radius = radius)

X = []
directions = pts/np.abs(pts)
for _ in range(10):
    pts = pts + (np.random.rand(pts.shape[0], pts.shape[1]) * radius / 10.) * directions
    X.append(pts)

v = PointCloudRipsVineyard.from_tensors(X, store_point_coordinates=True, store_cycles=True)
v.plot_vineyards(dim=1, noise_option="erase_diagonal")
```
It takes time because of a lot of swaps, but I am posting it here if you want to use it as an example/benchmark.

src/python/gudhi/vineyard.py
```diff
@@ -0,0 +1,1061 @@
...
+
+
+import os
+import glob
+import numpy as np
+from numpy.typing import ArrayLike
+from typing import Literal, Optional, Union
+import warnings
+import matplotlib as mpl
+import matplotlib.pyplot as plt
```
VincentRouvreau, line 21: This makes matplotlib is required by the feature. 2 options:
* Only import matplotlib in the functions/methods that are using matplotlib
* Specify it in the `requires:` field of documentation and `if(MATPLOTLIB_FOUND) add_gudhi_py_test(test_vineyard ...) endif()` in `src/python/CMakeLists.txt`

A third one would be to wait for PEP810 and `lazy import` 😉

src/python/CMakeLists.txt
```diff
@@ -69,6 +69,16 @@ function( add_gudhi_py_test THE_TEST )
   endif()
 endfunction( add_gudhi_py_test )
 
+function( add_gudhi_py_test_with_file_path THE_TEST THE_PATH)
```
VincentRouvreau, line 72: Argh... I am not a big fan of testing things with files, because in a near future, I would like devs and users to be able to do:
```bash
cd gudhi-devel-repo # or gudhi-devel-user-version
pip install --no-build-isolation -ve.[tests]
pytest # and here it is going to fail because of paths issue
```
I would prefer this to be an example (that we can also test) as it is more an integration test than a unitary test.
But keep it like this and we will think about it later.

src/python/test/test_vineyard.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+
+import pytest
+from pathlib import Path
+import glob
+import os
+import numpy as np
+
+from gudhi import RipsComplex
+from gudhi.vineyard import Vineyard, PointCloudRipsVineyard
```
VincentRouvreau, line 18: In [coverage report](https://output.circle-artifacts.com/output/job/f8fed905-6f2d-4e56-af4b-abe9bf79e49b/artifacts/0/htmlcov/z_8249f58cbc144a7c_vineyard_py.html), only 15% of the code is tested. We should do some effort here. I can help.

src/python/doc/vineyard_user.rst
```diff
@@ -0,0 +1,161 @@
...
+.. include:: vineyard_sum.inc
+
+Definition
+----------
+
+From a filtration of the form
+:math:`\mathcal{K}_0 \rightarrow \mathcal{K}_1 \rightarrow \cdots \rightarrow \mathcal{K}_n` we can compute the
+persistence diagram. For example, lets take two adjacent triangles and add they faces one after the other:
+
+.. code::
```
VincentRouvreau, line 16: This is a nice example where I think that [myst-nb](https://pypi.org/project/myst-nb/) sphinx plugin would help !

src/Vineyard/include/gudhi/vineyard_helper.h
```diff
@@ -0,0 +1,131 @@
...
+  }
+  return numberOfSimplices;
+}
+
+/**
+ * @ingroup vineyard
+ *
+ * @brief Builds a boundary matrix from the given matrix.
+ *
+ * @tparam FilteredComplex Complex type with following methods (see @ref FilteredComplex or @ref Gudhi::Simplex_tree
```
VincentRouvreau, line 55: You also added a `complex_simplex_range` method on Cubical_complex class (I am ok with it) to be able to perform `build_boundary_matrix_from_complex` on a cubical, but I am a bit lost with the naming of the template parameter here: Can we say that the cubical is a filtered complex ?
Because in the C++ documentation, we separated [cubical](https://gudhi.inria.fr/doc/latest/group__cubical__complex.html) from [Filtered Complexes](https://gudhi.inria.fr/doc/latest/group__simplex__tree.html)

src/python/gudhi/vineyard.py
hschreiber: Arf, I missed that.

I am thinking of a third option: I could externalize the plot functions like gudhi does for the diagrams and add the requirement only for this new file. It would be more work, but also the cleanest and most consistent (with other plot methods in Gudhi) solution. What do you think?

src/Vineyard/include/gudhi/vineyard_helper.h
hschreiber: The cubical complex stores filtration values and makes it possible to iterate over in that order, so it is a filtered complex. I don't know why the C++ documentation separates them, but if you look for example at the [FilteredComplex](https://gudhi.inria.fr/doc/latest/struct_filtered_complex.html) concept of the `Persistent_cohomology` class, it is also meant for the cubical complex. The doxygen group for filtered complexes was probably just forgotten?

src/common/doc/main_page.md
hschreiber: I added a more verbose description in 4b6c8dedad554141159ea005f875e895b4b91feb.
Does it look good? I tried to be a bit less formal than your version (also I was not sure about some points in your proposed description).

src/python/gudhi/vineyard.py
VincentRouvreau: Yes let's go this way !

PR #1322, opened 2026-04-01
gudhi.representations.BottleneckDistance shall respect BaseEstimator requirements

### LINKED ISSUE #1321: In representations, BottleneckDistance scikit-learn interfaces does not respect BaseEstimator requirements
#1164 (fix of #343 ) introduced a bug in `gudhi.representations.BottleneckDistance`

## Description

For GridSearchCV (cf. https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html ), BottleneckDistance scikit-learn interfaces should be able to perform `get_params`.

## Behaviour

```python
from gudhi.representations import BottleneckDistance
BottleneckDistance().get_params()
# E           AttributeError: 'BottleneckDistance' object has no attribute 'epsilon'
```

## Expected results

`BottleneckDistance().get_params()` shall work (even if epsilon is deprecated).

VincentRouvreau:
Fix #1321 
Fix also https://github.com/GUDHI/TDA-tutorial/blob/master/tutorials/Tuto-GUDHI-representations.ipynb

### DIFF
--- a/src/python/gudhi/representations/metrics.py
+++ b/src/python/gudhi/representations/metrics.py
@@ -383,14 +383,11 @@ class BottleneckDistance(BaseEstimator, TransformerMixin):
             numpy array of shape (number of diagrams in **diagrams**) x (number of diagrams in X): matrix of pairwise
                 bottleneck distances.
         """
-        if self.epsilon is not None:
-            return pairwise_persistence_diagram_distances(
-                X, self.diagrams_, metric="bottleneck", e=self.epsilon, n_jobs=self.n_jobs
-            )
-        else:
-            return pairwise_persistence_diagram_distances(
-                X, self.diagrams_, metric="bottleneck", e=self.e, n_jobs=self.n_jobs
-            )
+        e = self.epsilon if self.epsilon is not None else self.e
+        return pairwise_persistence_diagram_distances(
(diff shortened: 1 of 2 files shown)


src/python/gudhi/representations/metrics.py
```diff
@@ -368,18 +373,24 @@ def fit(self, X, y=None):
...
-        )
-        return Xfit
+        if self.epsilon is not None:
+            return pairwise_persistence_diagram_distances(
+                X, self.diagrams_, metric="bottleneck", e=self.epsilon, n_jobs=self.n_jobs
+            )
+        else:
+            return pairwise_persistence_diagram_distances(
+                X, self.diagrams_, metric="bottleneck", e=self.e, n_jobs=self.n_jobs
+            )
```
mglisse:
```suggestion
        e = self.epsilon if self.epsilon is not None else self.e
        return pairwise_persistence_diagram_distances(
            X, self.diagrams_, metric="bottleneck", e=e, n_jobs=self.n_jobs
        )

```
has a bit less repetition.
Another possibility (cleaner?) would be to set some variable (`self.e_`?) in `fit` and use that here.

mglisse (approved):

PR #1326, opened 2026-04-10
Random cpp for gudhi

VincentRouvreau:
A new C++11 Random module for GUDHI and its use in:
* src/Subsampling/include/gudhi/choose_n_farthest_points.h
* src/Subsampling/include/gudhi/pick_n_random_points.h 
* src/Persistence_representations/include/gudhi/Sliced_Wasserstein.h
* src/Nerve_GIC/include/gudhi/GIC.h

In src/Coxeter_triangulation/include/gudhi/Functions/random_orthogonal_matrix.h c++ random method was replaced by `CGAL::get_default_random().get_double()`, like that, when setting the CGAL seed, it sets it for the whole function.

~~Functions from gudhi/random_point_generators.h were moved in the Gudhi::random namespace.~~

User can set seed in off_file_from_shape_generator utility.

A new [topic 'Reproducibility'](https://output.circle-artifacts.com/output/job/5ac2181e-c029-4bb0-90d4-4ba84fa37711/artifacts/0/doxygen/group__reproducibility.html) in the c++ documentation.


src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,232 @@
...
+   * - @ref Gudhi::subsampling::pick_n_random_points
+   * - @ref Gudhi::Persistence_representations::Sliced_Wasserstein
+   * - @ref Gudhi::cover_complex::Cover_complex
+   * 
+   * In order to reproduce the results, a specific GUDHI random generator is available, and one can set its seed:
+   * 
+   * @code{.cpp}
+   * #include <gudhi/Random.h>
+   * 
+   * // Get the default static thread local GUDHI random generator - It is the one that is used internally
```
mglisse, line 42:
```suggestion
   * // Get the default thread local GUDHI random generator - It is the one that is used internally
```
`static` could be misinterpreted if we don't know it is inside a function.

src/common/include/gudhi/random_point_generators.h
```diff
@@ -27,6 +28,8 @@
 
 namespace Gudhi {
 
+namespace random {
```
mglisse, line 31: This breaks the API, unless there is a `using` line somewhere that I missed. Or are all the function in here undocumented?

src/Toplex_map/benchmark/benchmark_tm.cpp
```diff
@@ -58,20 +60,19 @@ int nb_insert_simplex2 = 3000;
 int nb_membership2 = 400000;
 
 Simplex random_simplex(int n, std::size_t d) {
-  std::random_device rd;
-  std::mt19937 gen(rd());
-  std::uniform_int_distribution<std::size_t> dis(1, n);
+  // Do not use Gudhi::random::get_default_random() here. Reset seeds randomly on each call - like it was previously
+  // Otherwise num_maximal_simplices is quite small
```
mglisse, line 64: Can you explain?

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,232 @@
...
+   private:
+    Engine gen_;
+  };
+  
+  /**
+   * @brief Returns a thread-local default random number generator instance.
+   * 
+   * @return A reference to the thread-local default @ref Random instance.
+   */
+  inline Random<>& get_default_random() {
```
mglisse, line 218: Personally, I would have gone with something much more minimalist
```c++
inline std::mt19937_64& get_default_random() {
    static thread_local std::mt19937_64 g{std::random_device{}()};
    return g;
}
```
(possibly with a typedef for `std::mt19937_64`)
without a class or wrappers for all the generation functions, remain as close as possible to C++11 random, just make the rng global.
But the functionality is there, so if you and Hannah like the wrapper, ok for this approach.

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,232 @@
...
+   * - @ref Gudhi::Persistence_representations::Sliced_Wasserstein
+   * - @ref Gudhi::cover_complex::Cover_complex
+   * 
+   * In order to reproduce the results, a specific GUDHI random generator is available, and one can set its seed:
+   * 
+   * @code{.cpp}
+   * #include <gudhi/Random.h>
+   * 
+   * // Get the default static thread local GUDHI random generator - It is the one that is used internally
+   * auto rng = Gudhi::random::get_default_random();
```
mglisse, line 43:
```suggestion
   * auto& rng = Gudhi::random::get_default_random();
```
?
This applies in many other places. Otherwise, the global generator remains stuck, and new copies will regenerate the same numbers.

src/common/include/gudhi/random_point_generators.h
VincentRouvreau: Yes it breaks the API, and yes these functions are undocumented. Mainly used internally and by the utility off_file_from_shape_generator

src/common/include/gudhi/Random.h
VincentRouvreau: I really liked your idea, so I rewrote it with a more minimalist approach !
Doc + tests + code on 8f70f4eb6

src/Toplex_map/benchmark/benchmark_tm.cpp
VincentRouvreau: This is no more the case with the rewrote on 8f70f4eb6

src/common/include/gudhi/Random.h
VincentRouvreau: This was true, but this is no more the case with the rewrote on 8f70f4eb6

src/common/include/gudhi/random_point_generators.h
VincentRouvreau: I documented it on 6f509bfd1

src/common/include/gudhi/Random.h
mglisse: I think there was a misunderstanding about what I meant: when I said "minimalist", I meant that get_default_random and set_seed would be the **only** functions, no get, no get_range, users keep constructing their own `std::uniform_real_distribution<double>`, they just call it on the default random engine instead of creating a new one.
That doesn't mean you have to remove the get/get_range wrappers, I just wanted to clarify my comment. Compared to the old code, it would just replace
```c++
    std::default_random_engine re;
```
or
```c++
    std::random_device rand_dev;
    std::mt19937 rand_engine(rand_dev());
```
with
```c++
    auto& re = random::get_default_random();
```

src/common/include/gudhi/random_point_generators.h
```diff
@@ -106,7 +109,7 @@ std::vector<typename Kernel::Point_d> generate_points_on_plane(std::size_t num_p
   typedef typename Kernel::Point_d Point;
   typedef typename Kernel::FT FT;
   Kernel k;
-  CGAL::Random rng;
+  CGAL::Random rng = CGAL::get_default_random();
```
mglisse:
```suggestion
  CGAL::Random& rng = CGAL::get_default_random();
```
everywhere?

VincentRouvreau: Thanks, I forgot this one. Done on 2ca10033c (also for coxeter)

src/common/include/gudhi/Random.h
VincentRouvreau: Oh ! Ok.
But if we do so, we would replace code like:
```cpp
#include <random>

//...
std::random_device rand_dev;
std::mt19937 rand_engine(rand_dev());

std::uniform_real_distribution<int> dis(min, max);
int rand_value = dis(rand_engine);
```
with:
```cpp
#include <random>
#include <gudhi/Random.h>

//...
std::uniform_real_distribution<int> dis(min, max);
int rand_value = dis(Gudhi::random::get_default_random());
```
While my proposal is to be able to do:
```cpp
#include <gudhi/Random.h>

//...
int rand_value = Gudhi::random::get<int>(min, max);
```
I think that `get` and `get_range` worth it (I could rename them `get_uniform` and `get_uniform_range`), but let's ask @hschreiber for a third point of view.

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,169 @@
...
+   * @tparam CustomRandomGenerator The type of the random generator. Default is @ref RandomGenerator.
+   * @param nbr The number of random numbers to generate.
+   * @param min The minimum value (inclusive).
+   * @param max The maximum value (inclusive).
+   * @param rng A random generator. Use the default one if not set.
+   * @return A vector of random numbers of type `Type` in the range [min, max].
+   */
+  template <typename Type, typename CustomRandomGenerator = RandomGenerator>
+  std::vector<Type> get_range(std::size_t nbr, const Type& min, const Type& max,
+                              CustomRandomGenerator& rng = get_default_random()) {
```
mglisse:
```suggestion
  template <typename Type, typename CustomRandomGenerator = RandomGenerator&>
  std::vector<Type> get_range(std::size_t nbr, const Type& min, const Type& max,
                              CustomRandomGenerator&& rng = get_default_random()) {
```
Out of curiosity, would something like this allow to remove the second overload?

VincentRouvreau: Yes it works ! I did so on for `get` and `get_range` on 835b173aa

src/common/include/gudhi/Random.h
hschreiber: I like the helper methods `get` and `get_range`. But I would indeed rename them `get_uniform*` or something similar to be more precise.

src/common/include/gudhi/Random.h
mglisse: Note that test_random.cpp also compiles if I just remove the second overload without changing the first one. The issues were probably with an earlier version.
(The current version if good, I was just trying to understand)

src/common/include/gudhi/Random.h
VincentRouvreau: > I like the helper methods `get` and `get_range`. But I would indeed rename them `get_uniform*` or something similar to be more precise.

I renamed them on 8094ee2a9

src/common/include/gudhi/random.h
```diff
@@ -0,0 +1,159 @@
...
+   * @tparam Type The type of the random numbers (must be integral or floating-point).
+   * @tparam CustomRandomGenerator The type of the random generator. Default is @ref RandomGenerator.
+   * @param nbr The number of random numbers to generate.
+   * @param min The minimum value (inclusive).
+   * @param max The maximum value (inclusive).
+   * @param rng A random generator. Use the default one if not set.
+   * @return A vector of random numbers of type `Type` in the range [min, max].
+   */
+  template <typename Type, typename CustomRandomGenerator = RandomGenerator&>
+  std::vector<Type> get_uniform_range(std::size_t nbr, const Type& min, const Type& max,
```
mglisse, line 109: Maybe we can mention `std::ranges::generate_random` (C++26) in a comment.

src/common/include/gudhi/random_point_generators.h
mglisse: Since they are not documented (I hadn't checked), you don't need to.

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,159 @@
...
+   * CGAL::get_default_random() = CGAL::Random(42);
+   * @endcode
+   *
+   * @{
+   */
+
+   /**
+    * The default random number generator type returned by @ref get_default_random
+    */
+  using RandomGenerator = std::mt19937_64;
```
mglisse: Are there other places where we use CamelCase for types in Gudhi?

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,159 @@
...
+   * 
+   * // Set the seed for the default generator that is used internally in GUDHI
+   * Gudhi::random::set_seed(42);
+   * @endcode
+   * 
+   * Other GUDHI functionnalities, that are based on CGAL, are using also randomness, but depends on
+   * <a href="https://doc.cgal.org/latest/Generator/classCGAL_1_1Random.html">CGAL::Random</a>.
+   * Here is the list:
+   * 
+   * - Gudhi::random::generate_points_on_plane
```
mglisse: It may be a bit strange to list undocumented functions in the doc.

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,159 @@
...
+    get_default_random().seed(seed);
+  }
+
+  /**
+   * @brief Generates a random number in the range [min, max].
+   *
+   * @tparam Type The type of the random number (must be integral or floating-point).
+   * @tparam CustomRandomGenerator The type of the random generator. Default is @ref RandomGenerator.
+   * @param min The minimum value (inclusive).
+   * @param max The maximum value (inclusive).
```
mglisse: The max is included for integers but not for reals, if I read the standard correctly.

src/common/include/gudhi/Random.h
VincentRouvreau: If I read well [code convention](https://github.com/GUDHI/gudhi-devel/blob/master/.github/code_conventions.md), it says:
 > * Concepts are named with camel case starting with uppercase. E.g. PersistentHomology for the concept of Persistence homology.
 > * Classes start with an uppercase letter and use underscore separators. E.g. Skeleton_blocker_contractor.

So I renamed `RandomGenerator` as `Random_generator` as it is a type.
But I didn't rename `CustomRandomGenerator` as it is a concept, cf.
```c++
   template <typename Type, typename CustomRandomGenerator = Random_generator&>
   Type get_uniform(const Type& min, const Type& max, CustomRandomGenerator&& rng = get_default_random()) { ... }
```

Am I right ?
Done on [bbb3217](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/bbb32176570911a82e2f4dac508cf06986be0844)

src/common/include/gudhi/random.h
VincentRouvreau: I added a comment on [2a5ecd6](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/2a5ecd61b25bbe0d6ff7942682f92c61f4ded083) to explain this function could be replaced with `std::ranges::generate_random` when gudhi will go with c++26 standard

src/common/include/gudhi/Random.h
VincentRouvreau: Ok, but it is also strange to say `you can set the seed for the CGAL random values`, but with no reason.
And the reason is that some methods are binded in Python (and also the utility). I agree with your comment but I don't know how to deal with it...
Maybe I could open an issue to explain that we should document these random functions, no ?

src/common/include/gudhi/Random.h
VincentRouvreau: I precised it in the documentation on [c477d71](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/c477d71b3e5212ea0906b4d9a11878b497c30633)
But maybe also the syntax `in the range [min, max]` should be reworked. I am not sure

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,160 @@
...
+
+namespace Gudhi {
+
+namespace random {
+  
+  /**
+   * @defgroup reproducibility Reproducibility
+   * 
+   * Some of the GUDHI functionnalities are using randomness. Here is a list of classes and methods that are using
+   * internally @ref Gudhi::random::Random
```
mglisse: I think the more important property of those functions is that they use `get_default_random` (and are thus affected by `set_seed`).

src/common/include/gudhi/Random.h
```diff
@@ -0,0 +1,160 @@
...
+   * - Gudhi::random::generate_points_in_ball_d
+   * - Gudhi::random::generate_points_in_cube_d
+   * - Gudhi::random::generate_points_on_two_spheres_d
+   * - Gudhi::random::generate_points_on_3sphere_and_circle
+   * - Gudhi::random::generate_points_on_klein_bottle_3D
+   * - Gudhi::random::generate_points_on_klein_bottle_4D
+   * - Gudhi::random::generate_points_on_klein_bottle_variant_5D
+   * - Gudhi::coxeter_triangulation::random_orthogonal_matrix
+   * 
+   * In order to reproduce the results, one can set the seed of the default CGAL random generator, that is used
```
mglisse: Add a link to CGAL's doc?

src/common/include/gudhi/Random.h
mglisse: If no documented C++ function uses it, then maybe we don't need to document it in C++?
Documenting the other functions is also a possibility (if we do that, I don't know if `random` is such a good namespace for them, some of the generated point sets are not random), but preferably not in this PR, to avoid delays.

src/common/include/gudhi/Random.h
VincentRouvreau: I added a link to CGAL::Random on [850120d](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/850120dd4d5e660e65f08d0f6ce6f76fa9289c96). Link to `CGAL::Random::getdefault_random` does not seem permanent (`https://doc.cgal.org/latest/Generator/group__PkgGeneratorsRef.html#gaaf28146567a47f590c4ee036527f1706`)

src/common/include/gudhi/Random.h
VincentRouvreau: Sorry about that, I forget to reformulate this sentence. Done on [850120d](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/850120dd4d5e660e65f08d0f6ce6f76fa9289c96)

src/Persistence_representations/include/gudhi/Sliced_Wasserstein.h
```diff
@@ -251,11 +252,8 @@ class Sliced_Wasserstein
       double epsilon = 0.0001;
       double thresh_y = (max_ordinate - min_ordinate) * epsilon;
       double thresh_x = (max_abscissa - min_abscissa) * epsilon;
-      std::random_device rd;
-      std::default_random_engine re(rd());
-      std::uniform_real_distribution<double> uni(-1, 1);
       for (int i = 0; i < num_pts_dgm; i++) {
-        double u = uni(re);
+        double u = Gudhi::random::get_uniform<double>(-1., 1.);
```
mglisse, line 256: Note that there is a small overhead compared to the previous code: this calls get_default_random once per loop, and gcc is bad at detecting at compile-time that the thread_local has already been initialized. Probably negligible here though.

src/common/include/gudhi/random_point_generators.h
VincentRouvreau: Ok, I removed the namespace and API break on [3b9e325](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/3b9e325fdfa37e3c956df29131d73dae9cfd78be)

src/common/include/gudhi/Random.h
VincentRouvreau: Ok, I removed the namespace and API break on [3b9e325](https://github.com/GUDHI/gudhi-devel/pull/1326/commits/3b9e325fdfa37e3c956df29131d73dae9cfd78be). I also removed the CGAL part in the reproducibility documentation, but I will have to talk about it in the python part.

mglisse (approved):

PR #1329, opened 2026-04-15
[Simplex_tree] Euler characteristic

hschreiber:
More "naive" and specialized version of PR #606 to use until someone wants to take care of a more generalized (and optimized?) version of it.

As it is a functionality which was asked several times.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -906,6 +906,21 @@ class Simplex_tree {
     return dimension_;
   }
 
+  /**
+   * @brief Computes and returns the euler characteristic of the non-filtered underlying complex represented
+   * by the simplex tree.
+   */
+  int compute_euler_characteristic() const {
+    auto dimension_count = num_simplices_by_dimension();
+    int euler = 0;
+    int sign = 1;
+    for (const int d : dimension_count) {
+      euler += sign * d;
+      sign = -sign;
+    }
+    return euler;
+  }
+
   /** \brief Returns true if the node in the simplex tree pointed by
    * the given simplex handle has children.*/
   template<class SimplexHandle>
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -907,7 +907,7 @@ class Simplex_tree {
   }
 
   /**
-   * @brief Computes and returns the euler characteristic of the non-filtered underlying complex represented
+   * @brief Computes and returns the Euler characteristic of the non-filtered underlying complex represented
    * by the simplex tree.
    */
   auto euler_characteristic() const {
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -860,7 +860,7 @@ class Simplex_tree {
   }
 
  public:
-  /** \brief Returns the number of simplices of each dimension in the complex. */
+  /** \brief Computes and returns the number of simplices of each dimension in the complex. */
   std::vector<size_t> num_simplices_by_dimension() const {
     if (is_empty()) return {};
     // std::min in case the upper bound got crazy
@@ -910,7 +910,7 @@ class Simplex_tree {
    * @brief Computes and returns the euler characteristic of the non-filtered underlying complex represented
    * by the simplex tree.
    */
-  int compute_euler_characteristic() const {
+  int euler_characteristic() const {
     auto dimension_count = num_simplices_by_dimension();
     int euler = 0;
     int sign = 1;
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -860,7 +860,7 @@ class Simplex_tree {
   }
 
  public:
-  /** \brief Returns the number of simplices of each dimension in the simplex tree. */
+  /** \brief Returns the number of simplices of each dimension in the complex. */
   std::vector<size_t> num_simplices_by_dimension() const {
     if (is_empty()) return {};
     // std::min in case the upper bound got crazy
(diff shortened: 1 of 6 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -906,6 +906,21 @@ class Simplex_tree {
...
+  /**
+   * @brief Computes and returns the euler characteristic of the non-filtered underlying complex represented
+   * by the simplex tree.
+   */
+  int compute_euler_characteristic() const {
+    auto dimension_count = num_simplices_by_dimension();
+    int euler = 0;
+    int sign = 1;
+    for (const int d : dimension_count) {
+      euler += sign * d;
```
mglisse: I would use a different letter than `d`, which looks too much like a dimension whereas it represents a number.

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -906,6 +906,21 @@ class Simplex_tree {
     return dimension_;
   }
 
+  /**
+   * @brief Computes and returns the euler characteristic of the non-filtered underlying complex represented
+   * by the simplex tree.
+   */
+  int compute_euler_characteristic() const {
```
mglisse, line 913: I am not a big fan of randomly adding `compute_`, `get_`, etc in front of functions, I find `cplx.euler_characteristic()` perfectly readable. But if you and @VincentRouvreau prefer with `compute_`, ok.

src/python/gudhi/_simplex_tree.cc
```diff
@@ -139,6 +139,23 @@ This function returns the dimension of the simplicial complex.
 This function returns a valid dimension upper bound of the simplicial complex.
 
 :returns:  an upper bound on the dimension of the simplicial complex.
+:rtype:  int
+           )doc")
+      .def("num_simplices_by_dimension",
+           &gsti::num_simplices_by_dimension,
+           R"doc(
+Returns the number of simplices of each dimension in the simplex tree.
```
mglisse: There are several places in the doc where I would rather talk about a "complex" instead of a "simplex tree", so the doc of `num_simplices_by_dimension` would remain the same in a different implementation of a complex. But again if you and Vincent prefer this, ok.

mglisse (commented):
Thanks. Just a few details.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: I added the `compute_` just to emphasize that something is computed and that is not just a constant time look-up of a pre-computed value (e.g. avoid using it in a loop of a cont Simplex tree). But I have nothing against removing it, if both of you prefer.

src/python/gudhi/_simplex_tree.cc
hschreiber: Should the method then be renamed `num_cells_by_dimension` to have something more general? Well, I guess, that would not be very consistent with all other function names which are shared for the FilteredComplex concept...

I am fine with using "complex" in the documentation as it is clearly not wrong.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: Most of the computation is done by `num_simplices_by_dimension`, which does not have `compute_`, although that's probably because it was already there and you would have added it otherwise. Personally, I find it ok that `insert` has a completely different complexity for unordered_map, map, and flat_map, I don't feel the need to give them 3 different names. But let's wait for the 3rd vote.

VincentRouvreau: If you have a look to [eulearning](https://github.com/vadimlebovici/eulearning), they named everything `compute_*`, but I am not a big fan, and I would also prefer naming the method `euler_characteristic()`. Even in Tangential complex, there is a hidden method (no doc) in another Simplicial complex data structure that is called `euler_characteristic()`.
I vote for `euler_characteristic()` naming.

hschreiber: Ok, changed it to `euler_characteristic()`!

VincentRouvreau:
I put it here just to keep in mind (and thanks @alex-tanaka for the benchmark):
> a small (non-rigorous) runtime test for computing the Euler characteristic of a 3-dimensional complex (27441 simplices) with various GUDHI methods:
>  * compute_persistence, betti_numbers and external Python functions: 9.4E-03 s
>  * get_simplices and external Python functions: 6.1E-03 s
>  * num_vertices, prune_above_dimension,  num_simplices and external Python functions: 8.8E-04 s
>  * num_simplices_by_dimension and external Python functions: 2.1E-04 s
>  * euler_characteristic: 1.9E-04 s

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -906,6 +906,21 @@ class Simplex_tree {
     return dimension_;
   }
 
+  /**
+   * @brief Computes and returns the euler characteristic of the non-filtered underlying complex represented
+   * by the simplex tree.
+   */
+  int euler_characteristic() const {
```
mglisse: I wonder if `int` is the right type. In practice, it should be fine, but for a huge complex, it might be safer to use something larger.

hschreiber: Replaced it with `std::make_signed_t<std::size_t>` as Vincent told me you suggested.

src/python/gudhi/_simplex_tree.cc
```diff
@@ -139,6 +139,23 @@ This function returns the dimension of the simplicial complex.
...
+           R"doc(
+Computes and returns the number of simplices of each dimension in the complex.
+
+:returns:  Array containing at index `d` the number of simplices of dimension `d` in the complex.
+:rtype:  1-dimensional numpy.array of length `maximal dimension + 1`.
+           )doc")
+      .def("euler_characteristic",
+           &gsti::euler_characteristic,
+           R"doc(
+This function computes and returns the euler characteristic of the non-filtered underlying complex represented
```
mglisse:
```suggestion
This function computes and returns the Euler characteristic of the non-filtered underlying complex represented
```
(nom propre)

mglisse (approved):

PR #1330, opened 2026-04-16
[python] plot fix

### LINKED ISSUE #1328: Wrong color in plot_persistence_diagram when max_intervals is exceeded
With gudhi 3.12.0 from conda-forge
```python
import gudhi
import numpy as np
import matplotlib.pyplot as plt
gudhi.plot_persistence_diagram([np.array([[0,1],[1,2]]),np.array([[0,2]])],max_intervals=4)
plt.show()
```
has 2 red points and 1 blue point, normal.
```python
gudhi.plot_persistence_diagram([np.array([[0,1],[1,2]]),np.array([[0,2]])],max_intervals=2)
plt.show()
```
1 red point disappears (normal), but the blue point has turned red ?!
This may be related to the recent issue where the code assumed that all points of dim 0 came first.

hschreiber:
Fixes Issue #1328.


src/python/gudhi/persistence_graphical_tools.py
VincentRouvreau: Return type has been modified, the documentation of this function should be changed accordingly

src/python/gudhi/persistence_graphical_tools.py
```diff
@@ -364,16 +364,15 @@ def plot_persistence_diagram(
...
-        pass
+    persistence, input_type = _format_handler(persistence)
+    persistence, changed = _limit_to_max_intervals(
+        persistence, max_intervals, key=lambda life_time: life_time[1][1] - life_time[1][0]
+    )
+    if changed or input_type == 0:
+        persistence = sorted(persistence, key=lambda x: x[0])
+    min_birth, max_death = _min_birth_max_death(persistence, band)
+
+    sizes, max_dim = _get_number_of_pairs_by_dimension(persistence)
```
VincentRouvreau: I would not call the second return value of `_get_number_of_pairs_by_dimension` as `max_dim`, because there is the case where it `return res, max(res) + 1` (so `max_dim + 1`). I understand the purpose of this, and how we want to process it with a loop with `range`. I propose to rename it `n_dims`,  `num_dims`,or `dims_count`, ...

```suggestion
    sizes, n_dims = _get_number_of_pairs_by_dimension(persistence)
```

src/python/test/test_persistence_graphical_tools.py
```diff
@@ -92,15 +92,15 @@ def test_limit_to_max_intervals():
     # check no warnings if max_intervals equals to the diagrams number
     with warnings.catch_warnings():
         warnings.simplefilter("error")
-        truncated_diags = gd.persistence_graphical_tools._limit_to_max_intervals(
+        truncated_diags, _ = gd.persistence_graphical_tools._limit_to_max_intervals(
             diags, 10, key=lambda life_time: life_time[1][1] - life_time[1][0]
         )
```
VincentRouvreau: Why not testing it ?
```suggestion
        truncated_diags, sorted = gd.persistence_graphical_tools._limit_to_max_intervals(
            diags, 10, key=lambda life_time: life_time[1][1] - life_time[1][0]
        )
        assert sorted == False
```

src/python/test/test_persistence_graphical_tools.py
```diff
@@ -92,15 +92,15 @@ def test_limit_to_max_intervals():
...
         # check diagrams are not sorted
         assert truncated_diags == diags
 
     # check warning if max_intervals lower than the diagrams number
     with pytest.warns(UserWarning) as record:
-        truncated_diags = gd.persistence_graphical_tools._limit_to_max_intervals(
+        truncated_diags, _ = gd.persistence_graphical_tools._limit_to_max_intervals(
             diags, 5, key=lambda life_time: life_time[1][1] - life_time[1][0]
         )
```
VincentRouvreau: Why not testing it ?
```suggestion
        truncated_diags, sorted = gd.persistence_graphical_tools._limit_to_max_intervals(
            diags, 5, key=lambda life_time: life_time[1][1] - life_time[1][0]
        )
        assert sorted == True
```

src/python/test/test_persistence_graphical_tools.py
VincentRouvreau, line 118: I had to do some unitary tests to understand the code. If you want to add them, it would improve the test coverage:
```python

def test_get_number_of_pairs_by_dimension():
    nb_pairs_by_dim, n_dims = gd.persistence_graphical_tools._get_number_of_pairs_by_dimension([])
    assert n_dims == 0
    assert nb_pairs_by_dim == []
    
    diags = np.array([[1, 2], [3, 4], [5, 6]], float)
    nb_pairs_by_dim, n_dims = gd.persistence_graphical_tools._get_number_of_pairs_by_dimension(diags)
    assert n_dims == 1
    assert nb_pairs_by_dim == [len(diags)]

    diags = [(1.0, 2.0), (3.0, 4.0), (5.0, 6.0)]
    nb_pairs_by_dim, n_dims = gd.persistence_graphical_tools._get_number_of_pairs_by_dimension(diags)
    assert n_dims == 1
    assert nb_pairs_by_dim == [len(diags)]

    diags = [(0, (1.0, 2.0)), (0, (3.0, 4.0)), (0, (5.0, 6.0))]
    nb_pairs_by_dim, n_dims = gd.persistence_graphical_tools._get_number_of_pairs_by_dimension(diags)
    assert n_dims == 1
    for val in nb_pairs_by_dim.values():
        assert val == len(diags)
    for k in nb_pairs_by_dim.keys():
        assert k == 0

    diags = [(2, (1.0, 2.0)), (2, (3.0, 4.0))]
    nb_pairs_by_dim, n_dims = gd.persistence_graphical_tools._get_number_of_pairs_by_dimension(diags)
    assert n_dims == 3
    for val in nb_pairs_by_dim.values():
        assert val == len(diags)
    for k in nb_pairs_by_dim.keys():
        assert val == 2
```

src/python/gudhi/persistence_graphical_tools.py
hschreiber: Ah, indeed, I forgot it. Done in 16f83c36ce78601be0dafd81d65548a301c34d95.

src/python/gudhi/persistence_graphical_tools.py
hschreiber: Ok, changed it to `num_dims` in 16f83c36ce78601be0dafd81d65548a301c34d95.

src/python/test/test_persistence_graphical_tools.py
hschreiber: Added the test for both calls in 16f83c36ce78601be0dafd81d65548a301c34d95.

src/python/test/test_persistence_graphical_tools.py
hschreiber: Good idea, I added the test in 16f83c36ce78601be0dafd81d65548a301c34d95.
I am used to only test public methods in C++, so I always forget them in python...

VincentRouvreau (approved):

PR #1332, opened 2026-04-24
Add Steenrod_persistence module (Sq^k barcodes over GF(2))

ammedmar:
## Summary

This PR adds a new gudhi module, **`Steenrod_persistence`**, which computes the
ordinary persistence barcode together with the Sq^k Steenrod barcode of a
filtered simplicial complex over GF(2).

Steenrod squares are cohomology operations

    Sq^k : H^n(X; F_2) → H^{n+k}(X; F_2)

that provide finer discriminatory information than Betti numbers alone.
The Steenrod barcode (Lupo–Medina-Mardones–Tauzin, 2022) generalises the
ordinary mod-2 cohomology barcode: for k = 0 one recovers it.

### What's included

| Component | Description |
|---|---|
| `src/Steenrod_persistence/` | Header-only C++ library (7 headers) |
| `src/Steenrod_persistence/test/` | Boost.Test C++ test (`ctest -R Steenrod`) |
| `src/Steenrod_persistence/example/` | RP² worked example |
| `src/Steenrod_persistence/doc/` | Doxygen intro, COPYRIGHT |
| `src/python/include/python_interfaces/Steenrod_barcode_interface.h` | nanobind bindings |
| `src/python/gudhi/simplex_tree.py` | `SimplexTree.compute_steenrod_barcodes(k, absolute=True)` |
| `src/python/test/test_simplex_tree_steenrod.py` | 11 pytest tests |
| `src/python/doc/steenrod_persistence_user.rst` | Sphinx user manual |
| `src/python/doc/steenrod_persistence_sum.inc` | Index entry |
| `biblio/bibliography.bib` | Three new bib entries |

### Algorithm

Four stages following [Lupo–Medina-Mardones–Tauzin 2022](https://doi.org/10.1007/s41468-022-00093-7)
with the cup-i formulas of [Medina-Mardones 2023](https://doi.org/10.1016/j.comgeo.2023.101999):

1. Filtration sort (dimension-within-value order)
2. Cohomology twist-reduction with clearing → R = DV decomposition
3. Sq^k applied to each cocycle representative (OpenMP-parallelised)
4. Augmented column reduction → Steenrod barcode (OpenMP-parallelised)

### Python API

```python
from gudhi import SimplexTree

st = SimplexTree()
# ... insert simplices ...

ordinary, steenrod = st.compute_steenrod_barcodes(k=2)
# ordinary[d] — list of (birth, death) pairs for H^d
# steenrod[d] — list of (birth, death) pairs for the Sq^k barcode in degree d
```

No prior call to `compute_persistence` is required.

### References

- U. Lupo, A. M. Medina-Mardones, G. Tauzin, *Persistence Steenrod modules*, J. Appl. Comput. Topol. (2022). https://doi.org/10.1007/s41468-022-00093-7
- A. M. Medina-Mardones, *New formulas for cup-i products and fast computation of Steenrod squares*, Comput. Geom. (2023). https://doi.org/10.1016/j.comgeo.2023.101999
- A. M. Medina-Mardones & Z. Zhou, *Persistent cohomology operations*, arXiv:2503.17130 (2025).

### Test plan

- [x] `ctest -R Steenrod` passes (2/2 Boost.Test)
- [x] `pytest test_simplex_tree_steenrod.py` passes (11/11)
- [x] Full `ctest` regression: 207/212 pass (5 pre-existing failures unrelated to this PR)
- [ ] CI green on upstream runners
- [ ] Sphinx doc build clean (no warnings on `steenrod_persistence_user.rst`)
- [ ] Gudhi maintainer review of CMake integration and module conventions


biblio/bibliography.bib
```diff
@@ -454,6 +454,18 @@ @inproceedings{10.5555/3327546.3327645
...
+  title    = {Persistence {S}teenrod modules},
+  journal  = {Journal of Applied and Computational Topology},
+  volume   = {6},
+  number   = {4},
+  pages    = {507--537},
+  year     = {2022},
+  doi      = {10.1007/s41468-022-00093-7},
+  url      = {https://doi.org/10.1007/s41468-022-00093-7}
+}
+
```
hschreiber: The citations are ordered by title, not by authors, so `Persistence Steenrod modules` should go to P and `New formulas for cup-i products and fast computation of Steenrod squares` to N.

src/python/gudhi/simplex_tree.py
```diff
@@ -18,10 +18,82 @@
 import numpy as np
 from numpy.typing import ArrayLike
 import warnings
-
 from gudhi import _simplex_tree_ext as t
```
hschreiber, line 22: Just a detail, but I would prefer keeping the blank line here to keep a visual separation between "internal" imports and "external" imports.

src/python/gudhi/simplex_tree.py
```diff
@@ -18,10 +18,82 @@
...
+    n = len(bars_by_dim)
+    result: list = [[] for _ in range(n)]
+    for d, bars in enumerate(bars_by_dim):
+        for pair in bars:
+            if _is_essential_relative(pair):
+                _, birth = pair
+                result[d].append((birth, _POS_INF))
+            elif d > 0:
+                result[d - 1].append(tuple(pair))
+    return result
```
hschreiber: Those two methods (`_to_absolute` and `_to_absolute_steenrod`) would probably be faster in C++ in the Steenrod interface (in particular seeing all those for-loops). You can just add a `_compute_absolute` method additionally to the already existing `_compute`.

src/python/gudhi/simplex_tree.py
```diff
@@ -249,6 +321,147 @@ def compute_persistence(
         self._pers._compute_persistence(homology_coeff_field, min_persistence)
         return self
 
+    def compute_steenrod_barcodes(self, k=1, absolute=False, max_dim=None, n_jobs=-1):
```
hschreiber:
```suggestion
    def compute_steenrod_barcodes(self, k: int = 1, absolute: bool = False, max_dim: int | None = None, n_jobs: int = -1):
```
It is still not done in a lot of our old functions, so you could not guess, but we try to type all new methods (and hopefully do the same with the old ones one day too...).

src/python/gudhi/simplex_tree.py
```diff
@@ -249,6 +321,147 @@ def compute_persistence(
...
+            # absolute Steenrod bar and so must not trip the warning.
+            check_max = (len(ordinary) - 1) if max_dim is None else int(max_dim)
+            check_max = min(check_max, len(ordinary) - 1)
+            # Essential relative bars carry the sentinel ``death = -inf``;
+            # see _to_absolute.
+            problematic_dims = [
+                d for d in range(1, check_max + 1)
+                if any(_is_essential_relative(pair) for pair in ordinary[d])
+            ]
+            if problematic_dims:
```
hschreiber, line 354: My suggestion is probably just a useless micro-optimization which will not bring much compared to the rest of the computation, but I was wondering if we could avoid constructing this list (the nested inner loop could be not that fast if there are really a lot of bars in the barcode) by computing the absolute versions first and while doing so, raising a flag when the problematic case appears. So the list gets only constructed when the flag is raised.

src/python/include/python_interfaces/Steenrod_barcode_interface.h
```diff
@@ -0,0 +1,125 @@
...
+ * ``death = +infinity`` marks essential bars. Coefficients are GF(2); no field
+ * parameter is exposed.
+ */
+template <class Simplex_tree_t>
+class Steenrod_barcode_interface {
+ public:
+  using Filtration_value = typename Simplex_tree_t::Filtration_value;
+  using Birth_death = std::pair<double, double>;
+  using Bars_one_dim = std::vector<Birth_death>;
+  using Bars_by_dim = std::vector<Bars_one_dim>;
```
hschreiber: In the future, we would like to homogenize the barcode format through all Gudhi python modules in a more useful format then we have now. It would really be nice, if this code could already adopt the new format such that we do not have to deprecate it too. I can help with that if you lack time to rewrite `to_filt_pairs`, the tests etc. for this purpose.

The new types are, using your naming:
```C++
using value_type = double;
using Birth_death = std::array<value_type, 2>;
using Bars_one_dim_finite = std::vector<Birth_death>;
using Bars_one_dim_infinite = std::vector<value_type>;
using Bars_by_dim_finite = std::vector<Bars_one_dim_finite>;
using Bars_by_dim_infinite = std::vector<Bars_one_dim_infinite>;
// this last type is not necessary, you can just keep two separate objects
// of type Bars_by_dim_finite and Bars_by_dim_infinite instead of a pair
using Bars = std::pair<Bars_by_dim_finite, Bars_by_dim_infinite>;
```
After constructing `Bars intervals`, you can then convert it into python types while minimizing new allocations by doing:
```C++
nanobind::list finites;
nanobind::list infinites;

// _wrap_as_numpy_array makes no copies but transfers the ownership
// of the underlying data to a numpy object. Note that this only works because
// we use std::array instead of std::pair for the type Birth_death
for (int d = 0; d < dim_max; ++d) {
  finites.append(_wrap_as_numpy_array(std::move(intervals.first[d])));
  infinites.append(_wrap_as_numpy_array(std::move(intervals.second[d]), intervals.second[d].size()));
}

return nanobind::make_tuple(finites, infinites);
```

In python, this will give you a tuple with two numpy arrays: the first with shape `(num_bars, 2)` (i.e. all finite bars) and the second with shape `(num_bars)` (i.e. all infinite bars with only the birth values stored).

Our plan is then to store this in a object with the methods `get_finite_intervals`, `get_infinite_intervals` and `get_intervals` (where both arrays are unified in a single numpy array of  shape `(num_bars, 2)`). But you do not have to care about this last step.

src/Steenrod_persistence/include/gudhi/Steenrod_reduction.h
```diff
@@ -0,0 +1,281 @@
...
+
+/** \brief Map from a simplex (sorted vertex list) to its relative index within a dimension.
+ * \ingroup steenrod_persistence
+ */
+using Spx_to_idx = std::unordered_map<Simplex, Index, Simplex_hash>;
+
+/** \brief All data for one dimension of the filtration. \ingroup steenrod_persistence */
+struct Dim_filtration {
+  std::vector<Index> idxs;     /**< Absolute filtration positions of simplices. */
+  std::vector<Simplex> tups;   /**< Sorted vertex lists, one per simplex. */
```
hschreiber, line 56: I didn't looked too much into the algorithm yet, but are the absolute positions used for something else than access to the filtration values? As the persistence computation usually only depends on the relative positions within a same dimension.
If it is only used to do `filtration_values[idx]`, then it would be reasonable to ask the user to give us the filtration in the "by_dim" order too (as it is already the case for the simplices), such that we can get rid of the extra array `idxs`.

src/python/include/python_interfaces/Steenrod_barcode_interface.h
```diff
@@ -0,0 +1,125 @@
...
+
+  /** Returns ``(ordinary, steenrod)`` barcodes.
+   *
+   * @param[in] n_jobs  Number of OpenMP threads for the parallelised stages.
+   *                    ``-1`` (default) uses all available cores.
+   */
+  std::pair<Bars_by_dim, Bars_by_dim> compute(int n_jobs = -1) {
+    using steenrod_persistence::Index;
+    using steenrod_persistence::Filtration_by_dim;
+    using steenrod_persistence::Dim_filtration;
```
hschreiber: This type (Dim_filtration) does not seem to be used.

src/Steenrod_persistence/include/gudhi/Steenrod_barcode.h
```diff
@@ -0,0 +1,104 @@
...
+ * @return ``Barcodes_result`` with the ordinary and Steenrod barcodes.
+ */
+inline Barcodes_result barcodes(int k,
+                                const Filtration_by_dim& fbd,
+                                const std::vector<double>* filtration_values = nullptr,
+                                int n_jobs = -1,
+                                int maxdim = -1) {
+  if (k < 0) {
+    throw std::invalid_argument("Sq^k exponent k must be non-negative");
+  }
```
hschreiber: As we are not in the Python interface, you could just type `k` with `unsigned int` to avoid negative values. If you prefer keeping `int` to avoid quiet conversion, use the macro `GUDHI_CHECK(k >= 0, std::invalid_argument("Sq^k exponent k must be non-negative"))` instead. This will only activate in Debug mode (a more fine grained macro is planned, but for now, that is what we have to use).

src/Steenrod_persistence/include/gudhi/Steenrod_barcode.h
```diff
@@ -0,0 +1,104 @@
...
+  const int n_dims_in = static_cast<int>(fbd.size());
+  const bool truncate = (maxdim >= 0) && (maxdim + 1 < n_dims_in);
+
+  Filtration_by_dim fbd_capped;
+  const Filtration_by_dim* fbd_used = &fbd;
+  if (truncate) {
+    fbd_capped.reserve(maxdim + 1);
+    for (int d = 0; d <= maxdim; ++d) fbd_capped.push_back(fbd[d]);
+    fbd_used = &fbd_capped;
+  }
```
hschreiber: This generates a non trivial (and potentially quite heavy) copy of your filtration. I think it would be better to do something like:
```C++
maxdim = maxdim < 0 || maxdim + 1 >= n_dims_in ? n_dims_in - 1 : maxdim;
auto rt = compute_reduced_triangular(fbd, maxdim);
```
And then remove the line where `maxdim` is defined inside `compute_reduced_triangular` to use the argument value instead. As `maxdim` is the ending condition of the loop, all higher dimensions will be automatically ignored.

hschreiber (commented):
A first batch of comments.

ammedmar:
Thanks for the careful review @hschreiber. I've pushed two follow-up
commits with the small / mechanical items and one perf fix.  Three of
your bigger comments interact so I'd like to land them as one revision
once we agree on the shape — questions below.

Item labels below (R1 … R10) link to the original threads in the order
they were posted.

### Pushed: [266f17aa5](https://github.com/Steenroder/gudhi-devel/commit/266f17aa5) (cleanups)

- **[R1](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3303993396)** Bib entries: `lupo2022persistencesteenrod` → P section,
  `medina2023cupifast` → N section.
- **[R2](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304047127)** Restored the blank line between external and internal imports.
- **[R4](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304165875)** Type-annotated the signature:
  `k: int = 1, absolute: bool = False, max_dim: int | None = None, n_jobs: int = -1`.
  (`pyproject.toml` already requires Python ≥ 3.10, so PEP-604 syntax is fine.)
- **[R8](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304872274)** Removed the unused `using steenrod_persistence::Dim_filtration;`.
- **[R9](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3305000168)** Kept `int k` + the active `throw std::invalid_argument(...)`,
  with a short comment explaining the choice.  Reasoning: `unsigned int`
  silently turns a user-passed `-1` into a huge positive `k` via
  integral conversion (so a Python user passing `-1` through nanobind
  hits the original underflow); `GUDHI_CHECK` is Debug-only, so direct
  C++ callers in Release builds also still hit the underflow.  Active
  `throw` covers both.  Happy to revisit if you'd rather match the
  house style strictly.

### Pushed: [458540449](https://github.com/Steenroder/gudhi-devel/commit/458540449) ([R10](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3305078146), perf)

Dropped the `fbd_capped` deep copy.  `compute_reduced_triangular` now
takes `int maxdim = -1` and uses it as the loop bound (negative or
out-of-range values are clamped to `fbd.size() - 1` inside, so the
default keeps the existing "process all dimensions" semantics).
Downstream functions (`compute_barcode_and_coho_reps`,
`compute_steenrod_matrix`, `compute_steenrod_barcode`) derive their
own `n_dims` from `rt.idxs.size()` and descendants, so the truncation
propagates without further plumbing.  No behaviour change; tests
unchanged (13 Python + 1 C++ unit).

### `idxs` audit ([R7](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304850073))

You asked whether `rt.idxs` is used for anything beyond
`filtration_values[idx]` lookup.  Auditing every read across
`Steenrod_reduction.h`, `Steenrod_cocycles.h`, `Steenrod_barcode.h`,
`Steenrod_barcode_impl.h`, `Steenrod_sq.h`, the answer splits into
three use cases:

1. **Filtration-value lookup (the case you described).**
   `Steenrod_cocycles.h:106-107, 128` stores absolute indices as bar
   endpoints; these are later resolved to filtration values in
   `Steenrod_barcode_interface.h::to_filt_pairs`.  This pass could use
   `(dim, rel_idx)` pairs and resolve to values at the very end —
   trivial refactor.

2. **Essential-class detection.**  `Steenrod_cocycles.h:98-127`
   populates a per-dim `std::unordered_set<Index> all_birth_idxs` with
   absolute indices of dim-simplices that get "claimed" as births by
   (dim-1)-columns, then declares any dim-simplex *not* in the set as
   essential.  The set is per-dim, but the convenience of one identity
   per simplex motivates absolute indices.  Could be re-expressed in
   per-dim relative indices with a small restructure.

3. **Cross-dim ordering in the Steenrod barcode reduction.**
   `Steenrod_barcode_impl.h:71, 81-91` compares
   `idxs_prev_dim[i] >= births_dim_minus_k[ii]` — the Algorithm 2 step
   that "unlocks" a Steenrod column once the (dim-1) reduction has
   advanced past its birth.  This **fundamentally requires the global
   filtration ordering across dimensions**.  Either we keep `idxs` to
   carry that ordering, or we rebuild a global rank from
   `filtration_values` on the fly (same memory).

So the premise *"persistence usually only depends on relative
positions within a same dim"* holds for the ordinary cohomology
reduction (which only walks dim-by-dim), but the Sq^k barcode step
from Lupo-Medina-Mardones-Tauzin Algorithm 2 inherently uses global
indices.  I could simplify (1) and (2) — and they'd carry their own
small memory wins — but `idxs` stays in the picture for (3) regardless.

If you'd still like the (1) + (2) cleanup as a clarity win, I can do
it.  But the "drop the extra array `idxs` entirely" direction looks
infeasible without also rewriting the Steenrod barcode reduction.

### Substantive ([R3](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304117087), [R6](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304586720)) — one open question

[R3](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304117087)
(`_to_absolute*` in C++) and
[R6](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304586720)
(unified `(finite, infinite)` numpy format) are interconnected: in the
new format finite vs essential are *structurally* separate, so the
absolute conversion can live entirely on the C++ side — define
`_compute_absolute` that returns the new format with the
relative→absolute shift already applied, and the Python helpers go
away.  This also makes
[R5](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304263731)
dissolve: the duality flag can be set during the conversion loop
instead of by a separate post-pass.

One question before I touch code:

**Could you point me at the reference branch / draft PR / header where
the unified `(finite, infinite)` format is being designed?**  Want to
match whatever lands upstream so the Steenrod module doesn't drift
again.  Happy to take the first cut at writing it once I can read the
design.

### Summary

| Item | Status |
|---|---|
| [R1](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3303993396), [R2](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304047127), [R4](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304165875), [R8](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304872274) | done in 266f17aa5 |
| [R9](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3305000168) | done in 266f17aa5 with note; happy to switch if you prefer |
| [R10](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3305078146) | done in 458540449 |
| [R7](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304850073) | audit above; (1)+(2) optional, (3) needs `idxs` regardless |
| [R3](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304117087), [R5](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304263731), [R6](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304586720) | waiting on design pointer for the unified format |

src/Steenrod_persistence/include/gudhi/Steenrod_barcode.h
hschreiber: > Kept `int k` + the active `throw std::invalid_argument(...)`,
>       with a short comment explaining the choice.  Reasoning: `unsigned int`
>       silently turns a user-passed `-1` into a huge positive `k` via
>       integral conversion (so a Python user passing `-1` through nanobind
>       hits the original underflow); `GUDHI_CHECK` is Debug-only, so direct
>       C++ callers in Release builds also still hit the underflow.  Active
>       `throw` covers both.  Happy to revisit if you'd rather match the
>       house style strictly.

Your python interface already has its own `throw` in case `k` is negative (and we should keep it to avoid python to crash). So this one is only for C++ use and not having active throws in a release build is not that unusual; that is what the debug build is for. Of course this is debatable (that is why I said that `GUDHI_CHECK` could be more fine-grained), but that is the subject for another pull request, not this one. So, it is better to follow the house style for now.

src/Steenrod_persistence/include/gudhi/Steenrod_reduction.h
hschreiber: > You asked whether `rt.idxs` is used for anything beyond `filtration_values[idx]` lookup. Auditing every read across `Steenrod_reduction.h`, `Steenrod_cocycles.h`, `Steenrod_barcode.h`, `Steenrod_barcode_impl.h`, `Steenrod_sq.h`, the answer splits into three use cases:
> 
> 1. **Filtration-value lookup (the case you described).**
>        `Steenrod_cocycles.h:106-107, 128` stores absolute indices as bar
>        endpoints; these are later resolved to filtration values in
>        `Steenrod_barcode_interface.h::to_filt_pairs`.  This pass could use
>        `(dim, rel_idx)` pairs and resolve to values at the very end —
>        trivial refactor.
> 
> 2. **Essential-class detection.**  `Steenrod_cocycles.h:98-127`
>        populates a per-dim `std::unordered_set<Index> all_birth_idxs` with
>        absolute indices of dim-simplices that get "claimed" as births by
>        (dim-1)-columns, then declares any dim-simplex _not_ in the set as
>        essential.  The set is per-dim, but the convenience of one identity
>        per simplex motivates absolute indices.  Could be re-expressed in
>        per-dim relative indices with a small restructure.
> 
> 3. **Cross-dim ordering in the Steenrod barcode reduction.**
>        `Steenrod_barcode_impl.h:71, 81-91` compares
>        `idxs_prev_dim[i] >= births_dim_minus_k[ii]` — the Algorithm 2 step
>        that "unlocks" a Steenrod column once the (dim-1) reduction has
>        advanced past its birth.  This **fundamentally requires the global
>        filtration ordering across dimensions**.  Either we keep `idxs` to
>        carry that ordering, or we rebuild a global rank from
>        `filtration_values` on the fly (same memory).
> 
> 
> So the premise _"persistence usually only depends on relative positions within a same dim"_ holds for the ordinary cohomology reduction (which only walks dim-by-dim), but the Sq^k barcode step from Lupo-Medina-Mardones-Tauzin Algorithm 2 inherently uses global indices. I could simplify (1) and (2) — and they'd carry their own small memory wins — but `idxs` stays in the picture for (3) regardless.
> 
> If you'd still like the (1) + (2) cleanup as a clarity win, I can do it. But the "drop the extra array `idxs` entirely" direction looks infeasible without also rewriting the Steenrod barcode reduction.

As the absolute position seems necessary for `steenrod_barcode_single_dim` (3), just keeping it as it is, is totally fine, in particular if it simplifies also (1) and (2). It was really just in case it was just used as a look-up for filtration values.

src/python/include/python_interfaces/Steenrod_barcode_interface.h
hschreiber: > **Could you point me at the reference branch / draft PR / header where the unified `(finite, infinite)` format is being designed?** Want to
> match whatever lands upstream so the Steenrod module doesn't drift
> again. Happy to take the first cut at writing it once I can read the
> design.

It is far from finished, but here the branch with a beginning of the new interface: https://github.com/hschreiber/gudhi-devel/tree/python_interface
In particular, look in the new folder `invariants` in the `python/gudhi` directory.
The idea is to replace `simplex_tree.persistence()` or `cubical_complex.persisence()` with a common method `compute_persistence(complex)` which returns an object `res` which then can do things like `res.get_intervals()`, `res.get_finite_intervals()`, `res.get_betti()` etc.
For steenrod, we would then have a method `compute_steenrod_persistence(complex)` (where `complex` could be a simplex tree or directly the boundary matrix), which returns the same kind of object with methods `get_intervals`, `get_finite_intervals` and `get_infinite_intervals` returning standard persistence, but additionally also with methods `get_steenrod_intervals`, `get_steenrod_finite_intervals` and `get_steenrod_infinite_intervals` for steenrod persistence.

The new format was chosen to fit well with all machine learning pipelines we have and because numpy arrays are much more efficient when manipulated than native tuples or lists.

But as all of that (except the barcode format) is still unstable in its end design, you don't really have to care for it. If the barcode is in the format I explained above, that is already a big help.

hschreiber:
A general comment about the C++ structure (a rather subjective one, so you can ignore it if you don't want to modify everything, but I thought it could be useful if you want to continue doing C++ projects in the future):

You use the `steenrod_persistence` namespace more like a class than just a namespace. That is "grammatically" not wrong, but you code reads more like C than C++ because of that. I think the code would gain in readability if you used a more object oriented organization (for example, you could keep the method `barcodes` in `Steenrod_barcode.h`, but hide the actual computation behind builder classes or any other design patterns you prefer). This way, you could "hide" within a class private methods and types which are not meant for the public user and share more easily some arguments. It also helps to make the relations between types/structures and methods more clear.

src/Steenrod_persistence/example/rp2_steenrod.cpp
```diff
@@ -0,0 +1,66 @@
...
+      std::sort(s.begin(), s.end());
+      all_simplices.insert(s);
+    }
+  }
+
+  std::vector<Simplex> sorted(all_simplices.begin(), all_simplices.end());
+  std::sort(sorted.begin(), sorted.end(), [](const Simplex& a, const Simplex& b) {
+    if (a.size() != b.size()) return a.size() < b.size();
+    return a < b;
+  });
```
hschreiber: `all_simplices` is a `std::set` which will automatically sort its entries (using the default `operator<` if you don't specify anything). So you could define it with this comparison method instead to avoid re-sorting a second time here. Then you can iterate over it in the loop after to avoid copying it in a vector.

src/Steenrod_persistence/example/rp2_steenrod.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+using namespace Gudhi::steenrod_persistence;
+
+// Close the complex spanned by ``top`` and return a Filtration_by_dim whose
+// filtration order is (dim, lexicographic).
+static Filtration_by_dim make_filtration_from_top(const std::vector<Simplex>& top) {
+  std::set<Simplex> all_simplices;
+  for (const Simplex& t : top) {
+    const int n = static_cast<int>(t.size());
+    for (int mask = 1; mask < (1 << n); ++mask) {
```
hschreiber, line 38: As this is an example file, meant to help user, it would be good to add a small comment here to explain what `(1 << n)` (and the `mask & (1 << i)` below) makes, as bit operation tricks are more advance notions.

src/Steenrod_persistence/example/rp2_steenrod.cpp
```diff
@@ -0,0 +1,66 @@
...
+  std::vector<Simplex> sorted(all_simplices.begin(), all_simplices.end());
+  std::sort(sorted.begin(), sorted.end(), [](const Simplex& a, const Simplex& b) {
+    if (a.size() != b.size()) return a.size() < b.size();
+    return a < b;
+  });
+
+  Filtration_by_dim fbd;
+  for (std::size_t i = 0; i < sorted.size(); ++i) {
+    const int dim = static_cast<int>(sorted[i].size()) - 1;
+    while (static_cast<int>(fbd.size()) <= dim) fbd.emplace_back();
```
hschreiber: As `sorted[i]` should never be empty, `dim` will always be positive. So you can just do:
```suggestion
    const auto dim = sorted[i].size() - 1;
    if (fbd.size() <= dim) fbd.resize(dim + 1);
```
to make it more readable (as `dim` is not used elsewhere and therefore does need to be `int`) and more efficient (a `resize` will be faster then a series of `emplace_back`s).

src/Steenrod_persistence/example/rp2_steenrod.cpp
```diff
@@ -0,0 +1,66 @@
...
+  const std::vector<Simplex> top = {
+      {1, 2, 4}, {2, 3, 4}, {1, 3, 5}, {2, 3, 5}, {1, 4, 5},
+      {1, 2, 6}, {1, 3, 6}, {3, 4, 6}, {2, 5, 6}, {4, 5, 6}};
+  const auto fbd = make_filtration_from_top(top);
+  const auto result = barcodes(/*k=*/1, fbd);
+
+  std::cout << "RP^2 — Sq^1 Steenrod barcode, bars per dimension:\n";
+  for (std::size_t d = 0; d < result.steenrod.size(); ++d) {
+    std::cout << "  dim " << d << " : " << result.steenrod[d].size() << " bar(s)\n";
+  }
```
hschreiber: A **very optional** suggestion: as `result` is a self designed structure `Barcodes_result`, you could add to this structure a `operator<<` friend method such that you could do `std::cout << result << "\n";` instead of the whole loop.

src/Steenrod_persistence/include/gudhi/Steenrod_barcode_impl.h
```diff
@@ -0,0 +1,221 @@
...
+ * @return Barcode of finite + essential bars. ``death == -1`` marks essentials.
+ */
+inline Barcode steenrod_barcode_single_dim(const Matrix& steenrod_matrix_dim,
+                                           int n_idxs_dim,
+                                           const std::vector<Index>& idxs_prev_dim,
+                                           const Matrix& reduced_prev_dim,
+                                           const std::vector<Index>& births_dim_minus_k) {
+  const int n_cols_red = static_cast<int>(reduced_prev_dim.size());
+  const int n_cols_st = static_cast<int>(steenrod_matrix_dim.size());
+  const int n_births = static_cast<int>(births_dim_minus_k.size());
```
hschreiber: All the `static_cast<int>` are useless here, as the type of the variables are already explicitly `int` (so the conversion is already forced).

src/Steenrod_persistence/include/gudhi/Steenrod_barcode_impl.h
```diff
@@ -0,0 +1,221 @@
...
+
+  // Augmented = [reduced_prev_dim | steenrod_matrix_dim]
+  std::vector<Column> augmented;
+  augmented.reserve(n_cols_red + n_cols_st);
+  for (const Column& c : reduced_prev_dim) augmented.push_back(c);
+  for (const Column& c : steenrod_matrix_dim) augmented.push_back(c);
```
hschreiber: If I didn't miss anything, `rt.reduced` (which corresponds to `reduced_prev_dim`) is only used here and in `compute_barcode_and_coho_reps` which is called before already. Same for `steenrod_matrix_dim` which is last used here. Can you confirm that? If that is the case, we can avoid copying both matrices here and use an index shift (if necessary) instead for the columns in `steenrod_matrix_dim`, as it is not a problem for later if they are modified. This should be much more time efficient.

src/Steenrod_persistence/include/gudhi/Steenrod_barcode_impl.h
```diff
@@ -0,0 +1,221 @@
...
+    int n_cols_st_curr = 0;
+    {
+      const Index max_prev = idxs_prev_dim.back();
+      for (int ii = 0; ii < n_births; ++ii) {
+        n_cols_st_curr = ii;
+        if (max_prev >= births_dim_minus_k[ii]) break;
+      }
+      // If the loop ran to completion, n_cols_st_curr stays at n_births - 1
+      // (mirrors Python ``enumerate`` semantics).
+    }
```
hschreiber, line 92:
```suggestion
    int n_cols_st_curr = 0;
    while (n_cols_st_curr < n_births - 1 && idxs_prev_dim.back() < births_dim_minus_k[n_cols_st_curr]) {
        ++n_cols_st_curr;
    }
```
Or if `births_dim_minus_k` is sorted as it looks like:
```C++
auto it = std::upper_bound(births_dim_minus_k.begin(), births_dim_minus_k.end(), idxs_prev_dim.back());
int n_cols_st_curr = it == births_dim_minus_k.end() ? n_births - 1 : std::distance(births_dim_minus_k.begin(), it);
```
Looks uglier, but is potentially faster (depending on how long `births_dim_minus_k` is, as `std::upper_bound` is supposed to be O(long n) with std::vector while the previous loop is O(n)).

src/Steenrod_persistence/include/gudhi/Steenrod_barcode_impl.h
```diff
@@ -0,0 +1,221 @@
...
+      // (exclusive) and idx (inclusive).
+      if (n_cols_st_curr < n_cols_st) {
+        const Index next_birth = births_dim_minus_k[n_cols_st_curr];
+        if (idx >= next_birth) {
+          if (i > 0) {
+            const Index next_idx = idxs_prev_dim[i - 1];
+            for (int jj = n_cols_st_curr; jj < n_cols_st; ++jj) {
+              if (next_idx >= births_dim_minus_k[jj]) break;
+              ++n_cols_st_curr;
+            }
```
hschreiber:
```suggestion
            for (int jj = n_cols_st_curr; jj < n_cols_st && next_idx < births_dim_minus_k[jj]; ++jj) {
              ++n_cols_st_curr;
            }
```

src/Steenrod_persistence/include/gudhi/Steenrod_barcode_impl.h
```diff
@@ -0,0 +1,221 @@
...
+
+#ifndef STEENROD_BARCODE_IMPL_H_
+#define STEENROD_BARCODE_IMPL_H_
+
+#include <utility>
+#include <vector>
+
+#ifdef _OPENMP
+#include <omp.h>
+#endif
```
hschreiber: @VincentRouvreau Do we compile Gudhi with the `openmp` link?

If I remember well, we only use the TBB libary (https://www.intel.com/content/www/us/en/developer/tools/oneapi/onetbb.html) for parallelization.

src/Steenrod_persistence/include/gudhi/Steenrod_sq.h
```diff
@@ -0,0 +1,193 @@
...
+                                                  const std::vector<Simplex>& tups_dim,
+                                                  const Spx_to_idx& spx2idx_dpk,
+                                                  int length,
+                                                  int n_jobs = -1) {
+  const int n_reps = static_cast<int>(coho_reps_dim.size());
+#ifdef _OPENMP
+  const int n_threads = (n_jobs <= 0) ? omp_get_max_threads() : n_jobs;
+#else
+  (void)n_jobs;
+#endif
```
hschreiber: Just put a `[[maybe_unused]]` in front of `int n_jobs = -1` in the arguments to avoid this branch of the `#ifdef` (which is only there to avoid a "unused variable warning" while compiling). It makes it clearer that `n_jobs` is ignored when you cannot use it and suppresses the warning already.

hschreiber (commented):
Second batch of comments.

ammedmar:
Update for @hschreiber on the second batch and the leftover first-batch items. All ten line-level comments from the second batch are now addressed. R3, R5 and R6 from the first batch are also done in a single follow-up commit.

### Trivial / mechanical (commit `d01e4a18b`)

Five items applied exactly as suggested:

- **[B2](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335414278)**: short comment explaining the `(1<<n)` / `mask & (1<<i)` subset-enumeration bit trick.
- **[B3](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335452945)**: `auto dim + resize` in place of `int dim` + `static_cast` + `emplace_back` loop.
- **[B5](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335542596)**: dropped the three useless `static_cast<int>` on `size_t → int` assignments.
- **[B8](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335791075)**: inner `if (...) break;` merged into the for-loop condition.
- **[B10](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335860733)**: `[[maybe_unused]] int n_jobs` in place of the `#else (void)n_jobs;` branch. Applied in both `Steenrod_sq.h` and the matching pattern in `compute_steenrod_barcode` in `Steenrod_barcode_impl.h`.

### Example refactors (commit `30fa06cbe`)

- **[B1](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335388700)**: `std::set` now takes a custom `(size, lex)` comparator; the explicit copy-into-vector + `std::sort` is gone; the loop iterates the set directly into the `Filtration_by_dim`.
- **[B4](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335485636)** (explicitly marked optional, done anyway): added a `friend operator<<` on `Barcodes_result` that prints bars-per-dimension for both ordinary and Steenrod barcodes; the example uses it instead of the hand-rolled print loop.

### Perf (commit `9b50f963d`)

- **[B6](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335636947)**: dropped the explicit `augmented = [R | M]` deep copy in `steenrod_barcode_single_dim`. The helper now takes `reduced_prev_dim` and `steenrod_matrix_dim` by non-const reference and accesses the logical augmented column `i` via a small inline accessor. Audit shows `rt.reduced` is last consumed by `compute_barcode_and_coho_reps` (before the Steenrod step) and by this helper; `steenrod_matrix` is consumed only here; nothing reads either after `compute_steenrod_barcode` returns. Per-dim slot disjointness keeps the parallelised outer loop safe.
- **[B7](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335777803)**: replaced the linear-scan `n_cols_st_curr` initialisation with a binary search (O(log n)). Note: the snippet in the review used `std::upper_bound`, which skips equality and disagrees with the original loop when `idxs_prev_dim.back()` equals one of the `births_dim_minus_k` entries (e.g. for `{10, 8, 6}` and value 8, the loop returns 1 but `upper_bound` returns 2). We use `std::lower_bound` with `std::greater<Index>{}` instead — same O(log n), behaviour-preserving on the equality case.

### Unified `(finite, infinite)` numpy format (commit `f85739055`)

This combines **[R3](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304117087)** (move `_to_absolute*` to C++), **[R5](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304263731)** (let the duality flag fall out of the conversion loop), and **[R6](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304586720)** (adopt the unified numpy format) into one revision.

The nanobind interface now returns each barcode as a pair `(finites_per_dim, infinites_per_dim)`. `finites[d]` is a `(n_bars, 2)` numpy array of `(death, birth)` rows with `death < birth` — same numerical contents and same slot order as before; `infinites[d]` is a `(n_bars,)` numpy array of birth values (the `-inf`-in-slot-0 sentinel is no longer needed). Arrays are zero-copy via `_wrap_as_numpy_array`. Heavy work still runs without the GIL; the GIL is re-acquired only to wrap the buffers.

Two methods on the interface: `_compute(n_jobs)` for the relative convention; `_compute_absolute(n_jobs, max_dim)` for the absolute convention, which does the relative → absolute dim shift in C++ and returns the duality flag as a third tuple element listing the problematic dims. The Python `_to_absolute*` helpers and the `_is_essential_relative` sentinel detector are gone.

13 / 13 Python tests pass; 2 / 2 C++ tests (unit + `rp2_steenrod` example) pass; user manual `testcode/testoutput` regenerated.

### B9 — OpenMP → TBB (in commit `be3eec70f`)

**[B9](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3335838337).** Audit confirmed gudhi has no CMake wiring for OpenMP at all, so `_OPENMP` was undefined on every build and the parallel loops in `Steenrod_sq.h` and `Steenrod_barcode_impl.h` ran serially with `n_jobs` silently ignored. Every other gudhi module that parallelises uses TBB (`Simplex_tree`, `Persistence_matrix`, `Tangential_complex`, `Persistent_cohomology`, …) — Steenrod was the only OpenMP outlier and the only one whose parallelism didn't actually run. Converted to `tbb::parallel_for` under the standard `#ifdef GUDHI_USE_TBB` guard with a serial fallback; `n_jobs` is now honoured by wrapping the body of `barcodes()` in a `tbb::task_arena`. All docstrings, doc comments, and the user manual that mentioned OpenMP are updated.

### Other post-review fixes (in commit `be3eec70f`)

Bundled with the B9 conversion above for landing speed — ideally these would have been a separate commit; flagging in case it matters for review. Two issues caught during a self-review pass:

- **Empty `SimplexTree` segfault.** `compute_steenrod_barcodes` on an empty simplex tree dereferenced `rt.idxs[0]` in `compute_barcode_and_coho_reps` and crashed (the `assert(n_dims_in >= 1)` in `compute_reduced_triangular` disappears in release builds). Added an empty-input early return at the top of `barcodes()` in `Steenrod_barcode.h`, plus a regression test covering both relative and absolute modes.
- **Negative `max_dim` was silently inconsistent across modes.** The relative path's Python wrapper sliced with `cutoff = max_dim + 1`, so `max_dim = -1` returned empty lists. The absolute path's C++ interface treated negative values as "all dimensions". `None` already means "all dimensions" in the public API, so the Python wrapper now rejects `max_dim < 0` outright with `ValueError`. Regression test added.

### R7 (idxs), R9 (k<0 guard)

Already addressed in previous round: **[R7](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3304850073)** accepted as-is per your reply ([here](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3334634420)); **[R9](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3305000168)** switched to `GUDHI_CHECK` per your reply ([here](https://github.com/GUDHI/gudhi-devel/pull/1332#discussion_r3334525202)) in commit `e30900363`.

### Still open

- **[C++ structure general comment](https://github.com/GUDHI/gudhi-devel/pull/1332#issuecomment-4594075006)** — agreed in spirit (the current namespace-as-module style does read more C than C++), but the format/conversion churn in `f85739055` already touches every Steenrod file; stacking a builder-class refactor on top of that doubles the blast radius. Proposing to defer to a separate follow-up PR — does that work for you?

ammedmar:
@hschreiber — one self-contained follow-up while the review is ongoing / and a question on the merge state.

max_dim now caps the reduction, not just the output. Previously the argument ran the full reduction and dropped the higher-dimensional bars at the end; it now stops the reduction at max_dim + 1 — the point at which H^max_dim classes can still be killed by (max_dim + 1)-simplices, so nothing below the cap changes. The returned barcodes are identical to before; the difference is purely the work skipped above the cap (≈1.5–2× on dense Vietoris–Rips complexes, scaling with the simplex mass above max_dim).
I checked the capped output against a full reduction in both the relative and absolute conventions, across RP², CP², CCP² and simplices, over k = 1..3 and every cutoff, comparing the finite and essential bars separately.

On the merge conflict: the branch now conflicts with master in a single file, I didn't want to slot the Steenrod entry in unilaterally: should it go in 3.13.0, or the next cycle? Happy to resolve it whichever way you prefer, or leave the changelog placement to you at integration. I'll also correct the entry's wording — it still says OpenMP, but the module now uses TBB.

Happy to hold all of this as a separate pass if you'd rather review the current state first. Thanks!

PR #1334, opened 2026-05-24
Avoid crash in `flag_persistence_generators` on non-flag filtrations

### LINKED ISSUE #1331: Crash when running flag_persistence_generators on non-flag complex
The simplex tree function `flag_persistence_generators` is supposed to be used on a flag complex (mainly Rips complex). When ran on something that is not a flag, like alpha complex, we should get an error. Instead, gudhi seems to just crash.

MWE:
```
import gudhi
import numpy as np

pts = np.random.rand(100,2)
alpha = gudhi.AlphaComplex(points=pts).create_simplex_tree()
alpha.compute_persistence()
alpha.flag_persistence_generators()
```

ryan-charette:
Resolves #1331 

Checks whether `edge_with_same_filtration()` actually finds an edge before `flag_persistence_generators()` tries to read its vertices.

For non-flag filtrations, the function can hit a simplex whose filtration was not inherited from any edge. Previously that could lead to calling `simplex_vertex_range()` on `null_simplex()`, which could crash the Python process. With this change, it raises a `ValueError` instead.

I also added a regression test with a small non-flag simplex tree to cover the error path.

### DIFF
--- a/src/python/include/python_interfaces/Persistent_cohomology_interface.h
+++ b/src/python/include/python_interfaces/Persistent_cohomology_interface.h
@@ -15,6 +15,7 @@
 #include <vector>
 #include <utility>    // for std::pair
 #include <algorithm>  // for sort
+#include <stdexcept>
 #include <unordered_map>
 
 #include <nanobind/nanobind.h>
@@ -228,6 +229,8 @@ class Persistent_cohomology_interface
     auto& diags = out.first;
     // diagsinf[0] should be interpreted as vector<int> and other diagsinf[i] as vector<array<int,2>>
     auto& diagsinf = out.second;
+    constexpr const char* non_flag_complex_error =
+        "flag_persistence_generators() requires a flag complex: a simplex has no edge with the same filtration";
     for (auto pair : Base::get_persistent_pairs()) {
       auto s = std::get<0>(pair);
       auto t = std::get<1>(pair);
@@ -240,6 +243,9 @@ class Persistent_cohomology_interface
           diagsinf[0].push_back(v);
         } else {
           auto e = stptr_->edge_with_same_filtration(s);
+          if (e == stptr_->null_simplex()) {
+            throw std::invalid_argument(non_flag_complex_error);
+          }
           auto&& e_vertices = stptr_->simplex_vertex_range(e);
           auto i = std::begin(e_vertices);
           auto v1 = *i;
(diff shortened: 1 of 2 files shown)


mglisse (approved):
A quick benchmark to check that it doesn't have a measurable impact in the normal, flag case would have been nice, but I guess this is ok.
I don't have a strong opinion on the exact type of the C++ exception.
IIUC for other xyz_generators functions (like lower_star), we return nonsense (vertex -1) but don't crash on invalid input, which sounds good enough, it was only this function that crashed?

ryan-charette:
Thanks!

Yeah, a benchmark would have been nice. I assumed the extra null simplex checks after the existing lookups wouldn’t be measurable in the normal flag case, but I can add one if you think it’s worth it.

And yes, that’s my understanding too: the issue here was specifically that `flag_persistence_generators()` could hit `simplex_vertex_range(null_simplex())` and crash. The other invalid-input cases I found return nonsense, but didn’t seem to crash.

VincentRouvreau (commented):
Just a question:
Would it help the users if we were documenting the exception in [flag_persistence_generators](https://gudhi.inria.fr/python/latest/simplex_tree_ref.html#gudhi.SimplexTree.flag_persistence_generators) ?
In [simplex_tree.py docstring](https://github.com/GUDHI/gudhi-devel/blob/11adba038499e4b59a9d46fe069d480c23915e77/src/python/gudhi/simplex_tree.py#L367-L380), we could add something like:
```python
        :raises ValueError: In case this is not a flag complex.
``` 

(I know that we didn't document all exceptions, but maybe for this one...)

mglisse:
> Just a question: Would it help the users if we were documenting the exception

I thought about it, but documenting it feels like encouraging the users to use it badly. Being flag is a precondition, it is up to the user to ensure it, we just happen not to crash the interpreter (currently through an exception, but it could be by returning nonsense instead in a future version, like we do for lower-star) if it isn't satisfied. I could be wrong about common practices in the Python community though, and I don't feel very strongly about it.

VincentRouvreau:
> I thought about it, but documenting it feels like encouraging the users to use it badly. Being flag is a precondition, it is up to the user to ensure it, we just happen not to crash the interpreter (currently through an exception, but it could be by returning nonsense instead in a future version, like we do for lower-star) if it isn't satisfied. I could be wrong about common practices in the Python community though, and I don't feel very strongly about it.

ok, so let’s go like that.

VincentRouvreau (approved):

PR #1336, opened 2026-05-31
Fix TangentialComplex intrinsic_dim spelling

### LINKED ISSUE #1195: [python TangentialComplex] argument name misspelling
https://github.com/GUDHI/gudhi-devel/blob/fb808b3cb34db7eb164665e241448a2a10da1f1c/src/python/gudhi/tangential_complex.pyx#L50

It should be `intrinsic_dim` instead of `instrisic_dim`. But as it is a python argument name, we cannot correct it without any consequences as someone could have used it with label.

ryan-charette:
Renames `intrisic_dim` to `intrinsic_dim` in the TangentialComplex Python interface, tests, example, and documentation.

Resolves #1195

### DIFF
--- a/src/python/gudhi/tangential_complex.py
+++ b/src/python/gudhi/tangential_complex.py
@@ -24,11 +24,11 @@ class TangentialComplex(t._Tangential_complex_interface):
     can be run to attempt to remove inconsistencies.
     """
 
-    def __init__(self, intrisic_dim, points=None, off_file=""):
+    def __init__(self, intrinsic_dim, points=None, off_file=""):
         """TangentialComplex constructor.
 
-        :param intrisic_dim: Intrinsic dimension of the manifold.
-        :type intrisic_dim: integer
+        :param intrinsic_dim: Intrinsic dimension of the manifold.
+        :type intrinsic_dim: integer
 
         :param points: A list of points in d-Dimension.
         :type points (Sequence[Sequence[float]]): list of list of double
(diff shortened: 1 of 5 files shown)


src/python/gudhi/tangential_complex.py
```diff
@@ -40,14 +40,14 @@ def __init__(self, intrisic_dim, points=None, off_file=""):
         """
```
VincentRouvreau, line 43: At least, for python users, we could do something like:

```suggestion
# ^^ import warnings

    def __init__(self, intrinsic_dim, intrisic_dim=None, points=None, off_file=""):
        """TangentialComplex constructor.

        :param intrinsic_dim: Intrinsic dimension of the manifold.
        :type intrinsic_dim: integer

        :param intrisic_dim: **[deprecated]** consider using `intrinsic_dim` instead.

        :param points: A list of points in d-Dimension.
        :type points (Sequence[Sequence[float]]): list of list of double

        Or

        :param off_file: An OFF file style name.
        :type off_file: string
        """
        if intrisic_dim is not None:
            warnings.warn(
                "The 'intrisic_dim' argument is deprecated and will be removed in a future version. Use 'intrinsic_dim' instead.",
                DeprecationWarning
            )
            intrinsic_dim = intrisic_dim
```

ryan-charette: I'm not sure how necessary it is (most users use this as a positional argument), but I don't see any reason to not add the deprecation warning. I'll go ahead and apply the suggestion.

src/python/gudhi/tangential_complex.py
VincentRouvreau, line 16:
```suggestion
import warnings

```

src/python/gudhi/tangential_complex.py
```diff
@@ -24,11 +24,16 @@ class TangentialComplex(t._Tangential_complex_interface):
     can be run to attempt to remove inconsistencies.
     """
 
-    def __init__(self, intrisic_dim, points=None, off_file=""):
+    def __init__(self, intrinsic_dim, points=None, off_file=""):
+# ^^ import warnings
+
```
VincentRouvreau:
```suggestion

```

VincentRouvreau (commented):
Sorry about that, my suggestion was not perfect.
I tried to fix it with these 2 suggestions.
It should be something like:
```python
import os
import warnings

from gudhi import _tangential_complex_ext as t
from gudhi.simplex_tree import SimplexTree


# TangentialComplex python interface
class TangentialComplex(t._Tangential_complex_interface):
    """The class Tangential_complex represents a tangential complex. After the
    computation of the complex, an optional post-processing called perturbation
    can be run to attempt to remove inconsistencies.
    """

    def __init__(self, intrinsic_dim, intrisic_dim=None, points=None, off_file=""):
```

VincentRouvreau (approved):

PR #1338, opened 2026-06-09
[Simplex_tree] Versioning for serialization

hschreiber:
Added a version number in front of the serialization buffer. The number will be incremented to 1 in PR #1191.

I named the version number with capital letters like a Macro, as it is a bit how it is used, but perhaps I should match it to the other variable names instead?

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -2898,12 +2898,13 @@ class Simplex_tree {
    *   architecture.
    */
   std::size_t get_serialization_size() const {
+    const std::size_t version_byte_size = sizeof(int);
     const std::size_t vh_byte_size = sizeof(Vertex_handle);
     std::size_t fv_byte_size = 0;
     const std::size_t tree_size = num_simplices_and_filtration_serialization_size(&root_, fv_byte_size);
-    const std::size_t buffer_byte_size = vh_byte_size + fv_byte_size + tree_size * 2 * vh_byte_size;
+    const std::size_t buffer_byte_size = version_byte_size + vh_byte_size + fv_byte_size + tree_size * 2 * vh_byte_size;
 #ifdef DEBUG_TRACES
-      std::clog << "Gudhi::simplex_tree::get_serialization_size - buffer size = " << buffer_byte_size << std::endl;
+    std::clog << "Gudhi::simplex_tree::get_serialization_size - buffer size = " << buffer_byte_size << std::endl;
 #endif  // DEBUG_TRACES
     return buffer_byte_size;
   }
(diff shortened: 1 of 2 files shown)


VincentRouvreau:
> I named the version number with capital letters like a Macro, as it is a bit how it is used, but perhaps I should match it to the other variable names instead?

No, that's correct, this is what is written in code conventions
https://github.com/GUDHI/gudhi-devel/blob/ac7ab417459c7177bd06da347e75dfd6a327f902/.github/code_conventions.md?plain=1#L11

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -2898,12 +2898,13 @@ class Simplex_tree {
    *   architecture.
    */
   std::size_t get_serialization_size() const {
+    const std::size_t version_byte_size = sizeof(int);
```
VincentRouvreau: `int` is assumed to match the size of the processor registers and at least on 16 bits. Should we be more strict and use something like `int16_t` or some equivalent ?

VincentRouvreau (approved):
Bad version number is handled and tested correctly to me.
Just a small comment on the version type

src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -3090,6 +3101,11 @@ class Simplex_tree {
   /** \brief Upper bound on the dimension of the simplicial complex.*/
   mutable int dimension_;
   mutable bool dimension_to_be_lowered_;
+
+  /**
+   * @brief Serialization version number. Should be incremented for each change in the serialization strategy.
+   */
+  static constexpr int SER_VERSION = 0;
```
mglisse: I think we could spell out the full word, it only appears 3 times in the code and users don't use it.

src/Simplex_tree/include/gudhi/Simplex_tree.h
mglisse: I don't think portability matters (we use non-portable things everywhere in the dump), but it is true that 16 bits is more than enough, so why not int16_t indeed.

mglisse (commented):
> The number will be incremented to 1 in PR https://github.com/GUDHI/gudhi-devel/pull/1191.

Do we want to update the number in every PR that modifies the format, or at most once per release? Even with 16 bits we have more than enough space for the former, I just wanted to ask.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Ok, changed it to `std::int16_t`.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Renamed it to `SERIALIZATION_VERSION`.

hschreiber:
> Do we want to update the number in every PR that modifies the format, or at most once per release? Even with 16 bits we have more than enough space for the former, I just wanted to ask.

My first thought was to change it every time the serialization method changes incompatibly to also ensure coherence between the branches/forks. But changing it only if it is for a different release seems also reasonable. I don't have a strong opinion about that.

VincentRouvreau:
> Do we want to update the number in every PR that modifies the format, or at most once per release? Even with 16 bits we have more than enough space for the former, I just wanted to ask.

I would update it only when a PR modifies the format. It will complicate, and one may forget to update it if we modify this number on each version (major and minor). I agree to increase the number on major version (if we go for gudhi 4.X.X), but not for minor version IMHO.

mglisse:
> > Do we want to update the number in every PR that modifies the format, or at most once per release? Even with 16 bits we have more than enough space for the former, I just wanted to ask.
> 
> I would update it only when a PR modifies the format.

Note that the alternative I was mentioning was to update it less, only if both criteria are satisfied:
1. the PR modifies the format
2. no other merged PR has increased the number since the last release

But let's go with the "every PR" version (except possibly on a case by case basis).

mglisse (approved):

PR #1340, opened 2026-06-15
Fix and test f-contiguous numpy arrays

### LINKED ISSUE #1339: RipsComplex produces wrong persistence diagrams for F-contiguous numpy arrays (pd.DataFrame.values)
## Environment
- GUDHI version: 3.12.0
- Python version: 3.13.13
- numpy version: 2.4.4
- OS: macOS 26.4.1 (arm64)

## Summary
`RipsComplex(points=arr)` produces incorrect persistence diagrams when `arr` is
a Fortran-contiguous (F-contiguous) numpy array. This is a regression from 3.5.0,
where the Cython-based binding correctly handled F-contiguous arrays via Python's
iteration protocol.

`pd.DataFrame.values` always returns an F-contiguous array for homogeneous
DataFrames, making this a common silent error for users who pass DataFrame
windows as point clouds.

## Minimal Reproducible Example

import numpy as np
import gudhi as gd

rng = np.random.default_rng(42)
base = rng.standard_normal((10, 2))

pts_c = np.ascontiguousarray(base)   # C-contiguous
pts_f = np.asfortranarray(base)      # F-contiguous (same data)

assert np.array_equal(pts_c, pts_f)  # identical values

def get_h1(pts):
    st = gd.RipsComplex(points=pts).create_simplex_tree(max_dimension=2)
    st.persistence()
    return st.persistence_intervals_in_dimension(1)

print(get_h1(pts_c))  # correct result
print(get_h1(pts_f))  # different (wrong) result

## Expected Behavior
Both calls should produce identical H1 persistence intervals,
since the underlying point coordinates are the same.

## Actual Behavior
C-contiguous input: shape (0, 2) — []
F-contiguous input: shape (3, 2) — [[0.53, 0.54], [0.97, 1.33], [1.47, 1.63]]

## Root Cause Analysis
The nanobind binding signature:

  __init__(self, arg0: ndarray[dtype=float64, shape=(*, *), writable=False], ...)

accepts any memory layout but the underlying C++ code appears to use raw pointer
arithmetic (ptr[i * ncols + j]) without accounting for strides, assuming
C-contiguous layout.

The fix on GUDHI's side would be to add a c_contig constraint to the nanobind
template, which causes numpy to automatically produce a C-contiguous copy when
needed:

  ndarray[dtype=float64, shape=(*, *), c_contig, writable=False]

## Workaround
Users can work around this by calling np.ascontiguousarray() before passing
to RipsComplex:

  pts = np.ascontiguousarray(df.values)
  gd.RipsComplex(points=pts)

## Verification
The list-of-lists path (__init__ overload 1) correctly handles the same data,
confirming the issue is specific to the numpy array path (overload 2).

VincentRouvreau:
I added the test proposed by @jnahm0
And the fix proposed by @mglisse 

Fix #1339

### DIFF
--- a/src/python/include/python_interfaces/numpy_utils.h
+++ b/src/python/include/python_interfaces/numpy_utils.h
@@ -185,7 +185,7 @@ class Numpy_2d_span
 
   const_pointer get_start_ptr() const { return &array_view_(0, 0); }
 
-  const_pointer get_end_ptr() const { return get_start_ptr() + array_view_.shape(0) * array_view_.shape(1); }
+  const_pointer get_end_ptr() const { return get_start_ptr() + array_view_.shape(0) * array_view_.stride(0); }
 };
 
 #endif  // INCLUDE_NUMPY_UTILS_PYTHON_H_
--- a/src/python/test/test_rips_complex.py
+++ b/src/python/test/test_rips_complex.py
@@ -6,6 +6,7 @@
 
     Modification(s):
       - 2025/04 Hannah Schreiber: Add tests to verify possibility of tensor input
+      - 2026/06 Joo-Heung Nahm and Vincent Rouvreau: Add tests for F-contiguous numpy arrays
       - YYYY/MM Author: Description of the modification
 """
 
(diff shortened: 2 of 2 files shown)


mglisse (approved):
We can start with that so the bug is fixed quickly.
Afterwards, I think a less minimal test would be better, it is quite possible with 10 points to have an empty H1 (and not crash or raise an exception) but still do nonsense. Minimality was great for a bug report, but in the tests I'd rather compare diagrams that have actual points in them.
Plus we should also test non-contiguous arrays, not just C vs F contiguous.

VincentRouvreau:
**[Edit]**
Tests are now on:
* C-contiguous | F-contiguous | not contiguous
* points | distance_matrix
* sparse | not sparse.

I removed H1 computation, SimplexTree equality test is enough

mglisse (approved):

PR #1342, opened 2026-06-19
[Simplex_tree] add number of parameters

hschreiber:
Split off PR #1191 to isolate the changes made to the simplex tree.

I increased `SERIALIZATION_VERSION` to 1, as I don't know if this will be merged before the next release and I think it is less problematic to forget to put it back to 0 (as real changes where made to the serialization methods), than to forget to put it to 1.

### DIFF
--- a/src/Simplex_tree/include/gudhi/Simplex_tree.h
+++ b/src/Simplex_tree/include/gudhi/Simplex_tree.h
@@ -377,7 +377,7 @@ class Simplex_tree {
   }
 
   /** \brief Returns a range over the simplices of the simplicial complex, in the order of the filtration.
-   * If the `operator<` method of the filtration value type does not induce a total ordering of the simplices,
+   * If the `operator<` method of the filtration value type does not induce a total ordering of the filtration values,
    * be sure to call @ref initialize_filtration(Comparator&&, Ignorer&&) const "initialize_filtration" with a proper
    * comparison operator at some point before. Otherwise the behaviour is undefined.
    *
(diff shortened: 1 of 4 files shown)


src/Simplex_tree/include/gudhi/Simplex_tree.h
```diff
@@ -392,6 +395,13 @@ class Simplex_tree {
    * was initialized, please call `clear_filtration()` or `initialize_filtration()` to recompute it.
    *
    * @note Not thread safe
+   * @par
+   * @note If @ref filtration_simplex_range or
+   * @ref initialize_filtration(Comparator&&, Ignorer&&) const "initialize_filtration" was never called before, this
+   * method will try to initialize the filtration order, assuming that `operator<` for @ref Filtration_value induces a
+   * total order of the simplices. If the assumption is wrong,
```
mglisse: That looks misleading. It assumes a total order on the filtration values, but accepts multiple simplices sharing the same filtration value.

mglisse (approved):
ok if you make the total order comment clearer.

src/Simplex_tree/include/gudhi/Simplex_tree.h
hschreiber: Is it better now? Or should I be more precise?

mglisse: ok for me.

src/Simplex_tree/test/simplex_tree_serialization_unit_test.cpp
```diff
@@ -12,6 +12,7 @@
 #include <iostream>
 #include <cstring>  // for std::size_t and strncmp
 #include <stdexcept>
+#include <random>
```
VincentRouvreau: You shall remove this include that is no more used here. `#include <gudhi/random.h>` is doing the job now.

hschreiber: Done.

VincentRouvreau (approved):

PR #1351, opened 2026-06-28
Add some test of f-contiguity and non contiguity for cubical

VincentRouvreau:


src/python/test/test_cubical_complex.py
```diff
@@ -14,7 +15,7 @@
 import pytest
 
 from gudhi import CubicalComplex, PeriodicCubicalComplex
-
+from numpy.testing import assert_almost_equal
```
mglisse, line 18: I am not convinced we need the "almost" since there is no computation involved, but whatever, it isn't important.

mglisse (approved):

src/python/test/test_cubical_persistence_low_dim.py
```diff
@@ -55,3 +56,29 @@ def test_basic_persistence_on_rectangle_from_top_cells():
     # Check persistence are equal, but maybe not sorted, maybe some duplicates
     assert_almost_equal(np.unique(pers0, axis=0), np.unique(persrec[0], axis=0))
     assert_almost_equal(np.unique(pers1, axis=0), np.unique(persrec[1], axis=0))
+
+def test_contiguity_for_persistence_on_a_line():
+    rng = np.random.default_rng(42)
+    # Get a 10x10 random values
```
mglisse: Oups, just 10, not 10x10.

src/python/test/test_cubical_persistence_low_dim.py
```diff
@@ -55,3 +56,29 @@ def test_basic_persistence_on_rectangle_from_top_cells():
...
+    base = rng.standard_normal((10))
+    persline_c = _persistence_on_a_line(np.ascontiguousarray(base))
+    persline_f = _persistence_on_a_line(np.asfortranarray(base))
+    assert_almost_equal(persline_c, persline_f)
+
+    base_repeat = np.repeat(base, 2, axis=0)
```
mglisse: Actually, repeating the values does not change persistence, so persistence(base_repeat) is the same as persistence(base), maybe not the best test.

It might be better to start from a random array of size 20, extract ::2 from it, compute persistence, use ascontiguousarray, recompute persistence. This way, the skipped values are meaningful.
Same for the rectangle case.

src/python/test/test_cubical_persistence_low_dim.py
VincentRouvreau: I fixed this comment on [8f16394](https://github.com/GUDHI/gudhi-devel/pull/1351/commits/8f16394a49f014760735631d50c316e27b5a69b5)

src/python/test/test_cubical_persistence_low_dim.py
VincentRouvreau: I took your proposal as I agree that it improves the test. I did so on the 4 tests I added on [8f16394](https://github.com/GUDHI/gudhi-devel/pull/1351/commits/8f16394a49f014760735631d50c316e27b5a69b5)

src/python/test/test_cubical_complex.py
VincentRouvreau: Testing float can lead to issues. You are right that it is not mandatory here, but I prefer to use `assert_almost_equal` as a good practice.

PR #1361, opened 2026-07-03
[Python, SimplexTree, *CubicalComplex] write_persistence_file

VincentRouvreau:
* Fix write_persistence_file for SimplexTree and test it.
* Add write_persistence_file for *cubical and test it

### DIFF
--- a/src/python/gudhi/cubical_complex.py
+++ b/src/python/gudhi/cubical_complex.py
@@ -356,7 +356,7 @@ class CubicalComplex(_Bitmap_cubical_complex_interface):
     def write_persistence_diagram(self, persistence_file):
         """This function writes the persistence intervals of the cubical complex in a user given file name.
 
-        :param persistence_file: Name of the file.
+        :param persistence_file: Name of the `persistence diagram <fileformats.html#persistence-diagram>`_ file.
         :type persistence_file: string
 
         :note: write_persistence_diagram method requires :func:`compute_persistence` method to be launched first.
--- a/src/python/gudhi/cubical_complex.py
+++ b/src/python/gudhi/cubical_complex.py
@@ -6,6 +6,7 @@
 #
 # Modification(s):
 #   - 2025/03 Hannah Schreiber: Use nanobind instead of Cython for python bindings.
+#   - 2026/07 Vincent Rouvreau: Add write_persistence_file
 #   - YYYY/MM Author: Description of the modification
 
 __license__ = "MIT"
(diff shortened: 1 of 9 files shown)


src/python/gudhi/cubical_complex.py
```diff
@@ -351,3 +352,17 @@ def persistence_intervals_in_dimension(self, dimension):
...
             return np.empty(shape=[0, 2])
         return piid
+
+    def write_persistence_diagram(self, persistence_file):
+        """This function writes the persistence intervals of the cubical complex in a user given file name.
+
+        :param persistence_file: Name of the file.
+        :type persistence_file: string
+
+        :note: intervals_in_dim function requires :func:`compute_persistence` function to be launched first.
```
mglisse: This comment about a different function looks strange.

mglisse (commented):
Can you clarify why we want this?
I think the direction was to avoid having file interfaces in every class (in particular as input), instead taking and outputting python objects, that can be read/written in files by separate functions.

VincentRouvreau:
> Can you clarify why we want this?

Here, I would to remove `__root_source_dir__` calls from [Persistence graphical tools user manual](https://gudhi.inria.fr/python/latest/persistence_graphical_tools_user.html).

To do so, I would like on the first example (aka. `plot_persistence_barcode`) to generate random points on a torus (with a seed), compute persistence and save the persistence in a file. Then in the next examples (aka. `plot_persistence_diagram` and `plot_persistence_density`), I would like to read persistence files, instead of computing them again.

src/python/gudhi/cubical_complex.py
VincentRouvreau: Sorry, it was a bad copy paste on my side. Fixed on 0a6271bd4

mglisse:
> To do so, I would like on the first example (aka. `plot_persistence_barcode`) to generate random points on a torus (with a seed), compute persistence and save the persistence in a file. Then in the next examples (aka. `plot_persistence_diagram` and `plot_persistence_density`), I would like to read persistence files, instead of computing them again.

Does that require those new functions? Couldn't it be done in 2 steps: the example computes the diagram (as a Python object) then saves it to a file (with pickle?), and the other examples read that back?

VincentRouvreau:
> Does that require those new functions? Couldn't it be done in 2 steps: the example computes the diagram (as a Python object) then saves it to a file (with pickle?), and the other examples read that back?

The fix for [SimplexTree.write_persistence_diagram](https://gudhi.inria.fr/python/latest/simplex_tree_ref.html#gudhi.SimplexTree.write_persistence_diagram) is required to write [persistence file](https://gudhi.inria.fr/python/latest/fileformats.html#persistence-diagram) (was buggy for the SimplexTree). Then, I noticed that we forgot to bind these methods for cubical and periodic_cubical, so I added and tested them.

I use this method, because we do have a section about [persistence file format](https://gudhi.inria.fr/python/latest/fileformats.html#persistence-diagram), but you are right that pickling the diagrams would be more pythonic.

Do you recommend to mark as deprecated `write_persistence_diagram` and persistence file format instead ?

VincentRouvreau:
In case we decide to keep the feature, I added some cross references between [file formats doc](https://gudhi.inria.fr/python/latest/fileformats.html#persistence-diagram) and the methods that are using this format.

PR #1362, opened 2026-07-09
Fix MapperComplex get_params

### LINKED ISSUE #1320: MapperComplex scikit-learn interfaces does not respect BaseEstimator requirements
## Description

For GridSearchCV (cf. https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html ), MapperComplex scikit-learn interfaces should be able to perform `get_params`.

## Behaviour

```python
from gudhi.cover_complex import MapperComplex
MapperComplex().get_params()
# E           AttributeError: 'MapperComplex' object has no attribute 'colors'
```

## Expected results

`MapperComplex().get_params()` shall work.

ryan-charette:
## Summary

Resolves #1320.

`MapperComplex` already accepts `colors` in its constructor, but it was not saving that value on the instance. That breaks sklearn’s `BaseEstimator.get_params()`, which expects every constructor argument to exist as an attribute.

This PR stores `colors` during initialization and adds a small regression test for `MapperComplex().get_params()`.

## Validation

- Targeted check that `MapperComplex().get_params()["colors"] is None`

### DIFF
--- a/src/python/gudhi/cover_complex.py
+++ b/src/python/gudhi/cover_complex.py
@@ -335,7 +335,6 @@ class MapperComplex(CoverComplexPy):
         self,
         *,
         input_type="point cloud",
-        colors=None,
         min_points_per_node=0,
         filter_bnds=None,
         resolutions=None,
@@ -379,7 +378,6 @@ class MapperComplex(CoverComplexPy):
             gains,
             clustering,
         )
-        self.colors = colors
         self.input_type, self.min_points_per_node, self.N, self.beta, self.C = (
             input_type,
             min_points_per_node,
--- a/src/python/gudhi/cover_complex.py
+++ b/src/python/gudhi/cover_complex.py
@@ -379,6 +379,7 @@ class MapperComplex(CoverComplexPy):
             gains,
             clustering,
         )
+        self.colors = colors
         self.input_type, self.min_points_per_node, self.N, self.beta, self.C = (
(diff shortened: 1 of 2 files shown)


src/python/gudhi/cover_complex.py
```diff
@@ -379,6 +379,7 @@ def __init__(
             gains,
             clustering,
         )
+        self.colors = colors
```
mglisse: It looks like the argument `colors` of `__init__` is undocumented and unused, maybe we should remove it instead?

ryan-charette: I agree, removing it is probably the better fix. I'll update the PR shortly

src/python/test/test_cover_complex.py
```diff
@@ -21,6 +21,10 @@ def test_empty_constructor():
     cover = CoverComplex()
 
 
+def test_mapper_complex_get_params():
+    assert "colors" not in MapperComplex().get_params()
```
VincentRouvreau, line 25: From my point of view, I think a better test would be to test `get_params` of all estimators. Something like:
```suggestion
def test_base_estimators_get_params():
    for complex in [MapperComplex, GraphInducedComplex, NerveComplex]:
        assert complex().get_params()["input_type"] == "point cloud"
```

or:
```python
        assert "input_type" in complex().get_params()
```

PR #1363, opened 2026-07-22
Make reader utilities fail explicitly on invalid file paths

### LINKED ISSUE #1225: [Python] No need of an empty default file name for file read functions in reader_utils and off_utils
And verify that if the file does not exist an exception is thrown.
List of modifications:
* [ ] `read_lower_triangular_matrix_from_csv_file(csv_file='', separator=';')`
* [ ] `read_persistence_intervals_grouped_by_dimension(persistence_file='')`
* [ ] `read_persistence_intervals_in_dimension(persistence_file='', only_this_dim=-1)`
* [ ] `read_points_from_off_file(off_file='')`

### LINKED ISSUE #1226: Missing error management in reader method
The method `read_lower_triangular_matrix_from_csv_file` in `reader_utils.h` does not throw anything when the given file is not found and just returns an empty vector. 

```C++
std::vector<std::vector<Filtration_value>> result;
std::ifstream in;
in.open(filename.c_str());
if (!in.is_open()) {
  return result;
}
```

> The behavior of the C++ functions doesn't look very good then, the calling function won't notice if there is an error (unless it somehow redirects cerr and monitors it). If it threw an exception, then the python wrapper would have nothing to do.

_Originally posted by @mglisse in https://github.com/GUDHI/gudhi-devel/pull/1196#discussion_r2189833486_

As `reader_utils.py` is only there to verify the existence of the input file, it could be completely removed if the C++ method throws an appropriate error (`std::invalid_argument` + message giving the file name looks sufficient?).

proffitteoy:
## Description

This PR updates the error-handling behavior of the reader utilities when a file path is missing or invalid, ensuring that the relevant APIs raise explicit exceptions instead of silently returning empty results.

Main changes:

- The C++ CSV reader now throws `std::invalid_argument` containing the provided file path when the file cannot be opened.
- Removed the empty-string default path from four Python reader APIs, so callers must now provide the path explicitly.
- Python readers now propagate `ValueError` or `FileNotFoundError` instead of printing an error and returning an empty list.
- Added C++ and Python regression tests covering missing arguments, invalid paths, and successful reads.

## Validation

- C++ reader test passed: `1/1`.
- `_reader_utils_ext` built successfully.
- Isolated tests using the actual Python extension passed, covering:
  - file path arguments are required;
  - invalid paths raise the expected exceptions;
  - valid files are read successfully.

Closes #1225
Closes #1226

### DIFF
--- a/src/common/include/gudhi/reader_utils.h
+++ b/src/common/include/gudhi/reader_utils.h
@@ -21,6 +21,7 @@
 #include <fstream>
 #include <map>
 #include <limits>  // for numeric_limits
+#include <stdexcept>
 #include <string>
 #include <vector>
 #include <utility>  // for pair
@@ -215,6 +216,8 @@ bool read_hasse_simplex(std::istream& in_, std::vector<Simplex_key>& boundary, F
  * ...<br>
  * Dj1;Dj2;...;Dj(j-1);<br>
  *
+ * @throws std::invalid_argument If `filename` cannot be opened.
+ *
  **/
 template <typename Filtration_value>
 std::vector<std::vector<Filtration_value>> read_lower_triangular_matrix_from_csv_file(const std::string& filename,
@@ -226,7 +229,10 @@ std::vector<std::vector<Filtration_value>> read_lower_triangular_matrix_from_csv
   std::ifstream in;
   in.open(filename.c_str());
   if (!in.is_open()) {
-    return result;
+    std::string error_str("read_lower_triangular_matrix_from_csv_file - Unable to open file ");
+    error_str.append(filename);
+    std::cerr << error_str << std::endl;
+    throw std::invalid_argument(error_str);
   }
 
   std::string line;
--- a/src/common/include/gudhi/reader_utils.h
+++ b/src/common/include/gudhi/reader_utils.h
@@ -231,7 +231,6 @@ std::vector<std::vector<Filtration_value>> read_lower_triangular_matrix_from_csv
   if (!in.is_open()) {
     std::string error_str("read_lower_triangular_matrix_from_csv_file - Unable to open file ");
     error_str.append(filename);
-    std::cerr << error_str << std::endl;
     throw std::invalid_argument(error_str);
   }
 
--- a/src/common/test/test_distance_matrix_reader.cpp
+++ b/src/common/test/test_distance_matrix_reader.cpp
@@ -23,10 +23,7 @@ using Distance_matrix = std::vector<std::vector<double>>;
 BOOST_AUTO_TEST_CASE( invalid_distance_matrix_file )
 {
   const std::string filename = "missing_distance_matrix.csv";
-  BOOST_CHECK_EXCEPTION(Gudhi::read_lower_triangular_matrix_from_csv_file<double>(filename), std::invalid_argument,
-                        [&filename](const std::invalid_argument& exception) {
-                          return std::string(exception.what()).find(filename) != std::string::npos;
-                        });
+  BOOST_CHECK_THROW(Gudhi::read_lower_triangular_matrix_from_csv_file<double>(filename), std::invalid_argument);
 }
 
 BOOST_AUTO_TEST_CASE( lower_triangular_distance_matrix )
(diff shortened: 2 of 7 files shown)


src/common/include/gudhi/reader_utils.h
```diff
@@ -226,7 +229,10 @@ std::vector<std::vector<Filtration_value>> read_lower_triangular_matrix_from_csv
   std::ifstream in;
   in.open(filename.c_str());
   if (!in.is_open()) {
-    return result;
+    std::string error_str("read_lower_triangular_matrix_from_csv_file - Unable to open file ");
+    error_str.append(filename);
+    std::cerr << error_str << std::endl;
```
VincentRouvreau: I would remove this error printing. The exception is sufficient in itself.

src/common/test/test_distance_matrix_reader.cpp
```diff
@@ -20,6 +20,15 @@
 
 using Distance_matrix = std::vector<std::vector<double>>;
 
+BOOST_AUTO_TEST_CASE( invalid_distance_matrix_file )
+{
+  const std::string filename = "missing_distance_matrix.csv";
+  BOOST_CHECK_EXCEPTION(Gudhi::read_lower_triangular_matrix_from_csv_file<double>(filename), std::invalid_argument,
+                        [&filename](const std::invalid_argument& exception) {
+                          return std::string(exception.what()).find(filename) != std::string::npos;
+                        });
```
VincentRouvreau: I understand that you wanted to test that the file name is in the exception, but this test would be sufficient enough from my point of view.
```suggestion
  BOOST_CHECK_THROW(Gudhi::read_lower_triangular_matrix_from_csv_file<double>(filename), std::invalid_argument);
```

(not mandatory)

src/common/include/gudhi/reader_utils.h
proffitteoy: done

src/python/gudhi/reader_utils.py
```diff
@@ -27,15 +26,13 @@ def read_lower_triangular_matrix_from_csv_file(csv_file="", separator=";"):
...
     :rtype: List[List[float]]
+
+    :raises ValueError: If the file cannot be opened.
     """
-    if csv_file:
-        if path.isfile(csv_file):
-            return t._read_matrix_from_csv_file(csv_file, separator[0])
-    print("file " + csv_file + " not set or not found.")
-    return []
+    return t._read_matrix_from_csv_file(csv_file, separator[0])
```
VincentRouvreau: I am not sure if this is the pythonic way of doing so, but we could also throw a `FileNotFoundError` instead (and keep the context), by doing (for the 3 methods you modified):
```suggestion
    try:
        return t._read_matrix_from_csv_file(csv_file, separator[0])
    except ValueError as original_exception:
        raise FileNotFoundError(f"{csv_file} not found") from original_exception
```

Then the error is:
```
>>> import gudhi as gd
>>> gd.read_lower_triangular_matrix_from_csv_file(csv_file="pouetpouettralala.toubiloubabdou")
read_lower_triangular_matrix_from_csv_file - Unable to open file pouetpouettralala.toubiloubabdou
  File "/gudhi-devel/build/src/python/gudhi/reader_utils.py", line 33, in read_lower_triangular_matrix_from_csv_file
    ret = t._read_matrix_from_csv_file(csv_file, separator[0])
ValueError: read_lower_triangular_matrix_from_csv_file - Unable to open file pouetpouettralala.toubiloubabdou

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/gudhi-devel/build/src/python/gudhi/reader_utils.py", line 36, in read_lower_triangular_matrix_from_csv_file
    raise FileNotFoundError(f"{csv_file} not found")
FileNotFoundError: pouetpouettralala.toubiloubabdou not found
```

proffitteoy: Addressed,I think I've done.

VincentRouvreau (approved):

PR #1364, opened 2026-07-24
[python] inverse Sequence and Tensor order in bindings

hschreiber:
Just realized that nanobind is not differentiating method overloads by closest type, but by "first argument type which fits" in order of ".def" definition. Which means that if you do:
```C++
.def(nb::init<const Sequence1D&>())
.def(nb::init<const Tensor1D&>())
```
the second version is never reached as any numpy array etc. can also be converted into a std::vector<...>. And a copy is made.
So I changed everything to:
```C++
.def(nb::init<const Tensor1D&>())
.def(nb::init<const Sequence1D&>())
```


VincentRouvreau (commented):
Nice catch ! I tested it by adding some traces. I don't know how we could do it without traces...
Just a small remark, we need to add a comment in the code somewhere for not to lose this important information.

mglisse (approved):

PR #1367, opened 2026-08-12
Fix bottleneck distance asymmetry

proffitteoy:
## Root Cause

The issue is located in the early-exit condition of `Graph_matching::multi_augment()`:

```cpp
unmatched_in_u.size() > sqrt(gp->size()) &&
max_depth >= sqrt(gp->size())
```

The exact Bottleneck search reuses a partial matching obtained at a smaller radius.

At the correct threshold for this counterexample, we have $n=6$, $k=3$, and $\ell=3$.The original condition triggers because $3>\sqrt 6$ and $3\geq\sqrt 6$, so `multi_augment()` returns `false` early. However, an augmenting path of depth $3$ still exists, and continuing the augmentation leads to a perfect matching.The problem is not caused by candidate-distance generation, but by the interaction between matching-state reuse and the early-exit shortcut.

## Mathematical Justification

Suppose the current matching leaves $k$ vertices in $U$ unmatched, and the shortest augmenting path has depth $\ell$. To obtain a perfect matching, the remaining $k$ unmatched vertices must eventually be matched through $k$ vertex-disjoint augmenting paths.

Each such augmenting path uses at least $(\ell+1)/2$ vertices from $U$. Since there are only $n$ vertices in $U$, a necessary condition is

$$
k\frac{\ell+1}{2}\leq n,
\qquad\text{hence}\qquad
k(\ell+1)\leq 2n.
$$

In particular, if $k\ell>2n$, then the current matching cannot be extended to a perfect matching.

Now choose the shortcut threshold $r=\sqrt{2n}$. If both $k>r$ and $\ell\geq r$, then $k\ell>r^2=2n$, so the necessary condition above is violated and returning `false` is safe.

The original threshold $r=\sqrt n$ only guarantees $k\ell>n$, which is not sufficient to rule out a perfect matching.

The counterexample demonstrates exactly this gap: for $n=6$, $k=3$, and $\ell=3$, we have $k\ell=9$. Thus $k\ell>n=6$, so the original $\sqrt n$ shortcut triggers, but $k\ell<2n=12$, so a perfect matching is still possible.

Therefore, replacing $\sqrt n$ with $\sqrt{2n}$ gives a sufficient condition for the early exit while retaining the intended optimization.

## Validation

After the change:

- Both `A -> B` and `B -> A` return `0.28348368108915978` on the minimal $3\times 3$ counterexample.
- A regression test has been added to check both exact Bottleneck symmetry and the expected numerical value.

### DIFF
--- a/src/Bottleneck_distance/include/gudhi/Graph_matching.h
+++ b/src/Bottleneck_distance/include/gudhi/Graph_matching.h
@@ -67,8 +67,9 @@ inline bool Graph_matching::multi_augment() {
     return false;
   Layered_neighbors_finder layered_nf(layering());
   int max_depth = layered_nf.vlayers_number()*2 - 1;
+  double rn = sqrt(2. * gp->size());
   // verification of a necessary criterion in order to shortcut if possible
-  if (max_depth < 0)
+  if (max_depth < 0 || (unmatched_in_u.size() > rn && max_depth >= rn))
     return false;
   bool successful = false;
   std::vector<int> tries(unmatched_in_u.cbegin(), unmatched_in_u.cend());
--- a/src/Bottleneck_distance/include/gudhi/Graph_matching.h
+++ b/src/Bottleneck_distance/include/gudhi/Graph_matching.h
@@ -67,9 +67,8 @@ inline bool Graph_matching::multi_augment() {
     return false;
   Layered_neighbors_finder layered_nf(layering());
   int max_depth = layered_nf.vlayers_number()*2 - 1;
-  double rn = sqrt(gp->size());
   // verification of a necessary criterion in order to shortcut if possible
-  if (max_depth < 0 || (unmatched_in_u.size() > rn && max_depth >= rn))
+  if (max_depth < 0)
     return false;
   bool successful = false;
   std::vector<int> tries(unmatched_in_u.cbegin(), unmatched_in_u.cend());
--- a/src/Bottleneck_distance/test/bottleneck_unit_test.cpp
+++ b/src/Bottleneck_distance/test/bottleneck_unit_test.cpp
@@ -132,7 +132,23 @@ BOOST_AUTO_TEST_CASE(graph_matching) {
   BOOST_CHECK(!m1.perfect());
 }
 
-BOOST_AUTO_TEST_CASE(exact_bottleneck_is_symmetric_with_reused_matching) {
+void check_bottleneck_counterexample(const std::vector<std::pair<double, double>>& a,
+                                     const std::vector<std::pair<double, double>>& b,
+                                     double expected) {
+  double exact_ab = bottleneck_distance(a, b, 0.);
+  double exact_ba = bottleneck_distance(b, a, 0.);
+  // Python's e=None maps to this overload's default value, the smallest positive double.
+  double default_ab = bottleneck_distance(a, b);
+  double default_ba = bottleneck_distance(b, a);
+
+  BOOST_CHECK_EQUAL(exact_ab, exact_ba);
+  BOOST_CHECK_CLOSE_FRACTION(exact_ab, expected, 1e-14);
+  BOOST_CHECK_EQUAL(default_ab, default_ba);
+  BOOST_CHECK_CLOSE_FRACTION(default_ab, expected, 1e-14);
+}
+
+// With the old shortcut, MSVC reaches n=6, k=3, l=3: k>sqrt(n), l>=sqrt(n), and k(l+1)=2n.
+BOOST_AUTO_TEST_CASE(bottleneck_windows_counterexample) {
   std::vector<std::pair<double, double>> a = {
       {0.8657832199712898, 1.1256902536912314},
       {0.16921569257803723, 0.6031897594038342},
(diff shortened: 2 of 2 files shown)


VincentRouvreau:
I am not a mathematical expert enough to understand your mathematical justification, sorry about that.

But I tested your test proposal on master branch and it works fine. I also tested in Python, and with hera and the test is ok in both case, so maybe the test is not robust enough.

If you have an example where asymmetry fails, that would be great ! (but maybe complicated...).
I will let math experts review your justification.

proffitteoy:
Thank you for testing this. I want to clarify what I mean by “asymmetry,” because I may not have explained it clearly enough before.

The bottleneck distance is a metric, so it must satisfy symmetry:

$$
d_B(A,B)=d_B(B,A).
$$

However, for this specific $3\times 3$ persistence diagram example, the current exact GUDHI implementation gives me:

```text
A -> B: 0.31024791223701775
B -> A: 0.2834836810891598
```

Therefore,

$$
d_B(A,B)\neq d_B(B,A),
$$

which is exactly the asymmetry I am referring to.

At the same time, Hera returns:

```text
Hera: 0.2834836810891598
```

So the issue is not merely that the bottleneck distance is slightly inaccurate. The same pair of persistence diagrams produces two different exact bottleneck distances when the input order is reversed.

The reproduction code I am using is:

```python
print("A -> B:", gudhi.bottleneck_distance(A, B, e=0))
print("B -> A:", gudhi.bottleneck_distance(B, A, e=0))
print("Hera:", gudhi.hera.bottleneck_distance(A, B, delta=0))
```

So I would like to confirm one point: when you tested this on the master branch, did you explicitly compare both

```python
gudhi.bottleneck_distance(A, B, e=0)
```

and

```python
gudhi.bottleneck_distance(B, A, e=0)
```

?

If both directions return the same value in your environment, then the bug may also depend on platform or iteration order. But in my environment, the input above reproduces the asymmetry consistently.

VincentRouvreau:
> So I would like to confirm one point: when you tested this on the master branch, did you explicitly compare both
> 
> ```python
> gudhi.bottleneck_distance(A, B, e=0)
> ```
> 
> and
> 
> ```python
> gudhi.bottleneck_distance(B, A, e=0)
> ```
> 
> ?
> 
> If both directions return the same value in your environment, then the bug may also depend on platform or iteration order. But in my environment, the input above reproduces the asymmetry consistently.

Thanks for the clarification. My first test was not with `e=0`. By the way, I do have the same result with `e=0`.

```python
from gudhi import bottleneck_distance
import numpy as np
diag1 = np.array([[0.8657832199712898, 1.1256902536912314], [0.16921569257803723, 0.6031897594038342], [-0.04675641745672246, 0.5202109447215970]])
diag2 = np.array([[0.15043295756523575, 0.18414854990571342], [0.15043295756523575, 0.18414854990571342], [0.6796253715781101, 1.3001211960521455]])
bottleneck_distance(diag1, diag2, e=0)
# 0.2834836810891598
bottleneck_distance(diag2, diag1, e=0)
# 0.2834836810891598
```

So I think you are right when you say
 >  the bug may also depend on platform or iteration order.

I will give a try on the CI with a new branch, just by curiosity.

proffitteoy:
We initially discovered this issue on **Windows**. On Windows, the original pair of persistence diagrams causes GUDHI's exact bottleneck distance to become asymmetric, i.e.

$$
d_B(A,B)\neq d_B(B,A).
$$

I then tested exactly the same input under WSL2/Linux and found that the original counterexample does not trigger the issue there: both directions return the same value. Therefore, the original example is not a cross-platform stable counterexample.

From the current source inspection, this platform-dependent behavior appears to be related to the **different traversal paths** taken by the matching algorithm. Unmatched vertices are stored in a `std::unordered_set`, while neighbor selection relies on CGAL KD-tree `search_any_point()`. When several equivalent neighbors or augmenting paths are available, the actual path selected can depend on the container iteration order and KD-tree search order, which are not guaranteed to be identical on Windows and Linux.

In other words, the original 3×3 input happens to follow a matching path on Windows that exposes the bug, while on Linux it follows a different path and therefore does not produce the asymmetry.

**It took me some time to construct a counterexample that also fails on Linux, since the traversal order there is different from Windows.**

The Linux counterexample is:

```python
A = np.array([
    [-0.7734086850435968, -0.7734065850126042],
    [ 0.4057382451550824,  0.40574118001625126],
    [-0.8218366177063472, -0.8136106266959778],
    [ 0.4057382451550824,  0.40574118001625126],
], dtype=np.float64)


B = np.array([
    [-0.6759744467664093,  0.17199054436205374],
    [-0.4787571504589536, -0.47692522493813944],
    [-0.6759744467664093,  0.17199054436205374],
    [-0.6759744467664093,  0.17199054436205374],
], dtype=np.float64)
```

On Linux, GUDHI's exact implementation returns:

```text
A -> B: 0.9453971293746579
B -> A: 0.9856011710580316
```

More importantly, both values are incorrect. Hera returns:

```text
A -> B: 0.4239824955642315
B -> A: 0.4239824955642315
```

and the exact GUDHI implementation on Windows also returns:

```text
A -> B: 0.4239824955642315
B -> A: 0.4239824955642315
```

So the issue does not appear to be Windows-specific. More precisely, different platforms require different inputs to trigger the bug because the traversal/matching paths differ. The original 3×3 counterexample triggers it on Windows, while this new 4×4 counterexample triggers it on Linux.

This is also consistent with the mathematical issue: the original early-exit shortcut uses $\sqrt n$ as the threshold, which only implies $k\ell>n$, and that is not sufficient to conclude that the current matching cannot be extended to a perfect matching.

proffitteoy:
As for the test, since different counterexamples are needed to trigger the issue on Windows and Linux, I think it may be better for you to decide how to add the regression test. It may also be fine not to add one; so far, after removing this problematic mathematical restriction, all of these counterexamples return the correct result.

mglisse:
Thanks for the report. I haven't checked in details yet, but I notice that in the explanation, you say that we are missing a factor 2, while in the patch, you are removing the shortcut completely. Could you comment on that?
This may have an impact on the speed of the program. It would be interesting to know how high this impact is, to know how much effort we should spend trying to retain part of this shortcut (of course correctness has priority).

proffitteoy:
Thanks for the comment.When I first submitted this PR, I was not completely confident that the mathematical argument behind the shortcut was fully correct. Since an incorrect early-exit condition can affect correctness, I chose the conservative approach and removed the shortcut entirely rather than keeping a condition that I could not formally justify.

Regarding the performance impact, based on my experiments, this shortcut was triggered only once in several thousand test cases, so removing it had a negligible impact in my benchmarks.

However, after the discussions over the past few days, I now tend to think that the proof is indeed correct.I will take another look at the implementation and verify whether we can safely restore this optimization.

proffitteoy:
This update covers three things:

1. Confirmed that the issue can also affect `e=None`, so the bug is not limited to `e=0`.
2. Reworked the regression tests with two counterexamples for Windows and two for Linux. Each case lies in the old shortcut’s “danger zone” .
3. Restored the shortcut using @mglisse ’s proposed fix.

VincentRouvreau:
From #1369 CI, I confirm that:
- `bottleneck_linux_[pertubed_]counterexample` fail on Linux.
- `bottleneck_linux_[pertubed_]counterexample` and `bottleneck_windows_[pertubed_]counterexample` fail on macos.
- Windows CI fails because of vcpkg that makes requests on github (that does not respond well these times)

VincentRouvreau:
I also have launched the benchmark in `src/Bottleneck_distance/benchmark` (I just added `Gudhi::random::set_seed(42);` before the for loop in bottleneck_chrono.cpp), and there is a (really) small gap in the timings:

|nb_pairs| time (ms) - master | bottleneck distance - master|  time (ms) - PR1367 | bottleneck distance - PR1367|
|---|---|---|---|---|
| 1000 | 30 | 3.99697   | 31 | 3.99697 |
| 2000 | 117 | 3.9976   | 120 | 3.9976 |
| 3000 | 272 | 3.99747  | 281 | 3.99747 |
| 4000 | 448 | 3.99781  | 458 | 3.99781 |
| 5000 | 844 | 3.99981  | 864 | 3.99981 |
| 6000 | 1066 | 3.99967 | 1091 | 3.99967 |
| 7000 | 499 | 3.99941  | 512 | 3.99941 |
| 8000 | 1040 | 3.99887 | 1065 | 3.99887 |
| 9000 | 1473 | 3.99934 | 1507 | 3.99934 |
| 10000 | 1583 | 3.9999 | 1619 | 3.9999 |

I think it is negligible

VincentRouvreau (approved):

mglisse (approved):
I haven't looked at the theory and won't have time in the near future. Let's merge this:
* there is an input where we currently get the wrong result
* the patch makes us give the right output on that input
* the patch should be safer than the current code (apply a shortcut in fewer cases)
* the performance impact doesn't seem too bad.

However, I'd like to keep something open (the issue? another one?) to remember to have another look at it. If the right condition is an inequality on some product, then we should test that instead of comparing both operands with the square root. But we need some reviewer to check the theory and confirm this (I don't even remember what the layers are supposed to represent in this algorithm...).

proffitteoy:
> I'd like to keep something open (the issue? another one?) to remember to have another look at it

I opened #1371 to further derive and discuss the product criterion behind this shortcut.

PR #1370, opened 2026-08-21
Different filtration values support in python

VincentRouvreau:
The aim of this PR is to discuss on the manner to manage float32/float64/... types for GUDHI's datastructures.
I started with the Cubical.

* src/python/include/python_interfaces/filtration_value_types_support.h
    - struct Supported { using List = type_list<float, double>; };
    - (equivalent) struct Supported_for_data_structure { using List = type_list<Data_structure<float>, Data_structure<double>>; };
    - for_each_filtration_value_type to loop on Supported or Supported_for_data_structure

* New classes generated in loop thanks to for_each_filtration_value_type in src/python/gudhi/_cubical_complex.cc
    - _Bitmap_cubical_complex_interface[_float32|_float64]
    - _Cubical_complex_persistence_interface[_float32|_float64]
    - _Periodic_cubical_complex_interface[_float32|_float64]
    - _Periodic_cubical_complex_persistence_interface[_float32|_float64]

* functions generated in loop thanks to for_each_filtration_value_type in src/python/gudhi/_pers_cub_low_dim.cc
    - _persistence_on_a_line nb::arg("data").noconvert() -> wrap_persistence_1d<[double|float]>(nb::ndarray<nb::c_contig>) 
    - _persistence_on_rectangle_from_top_cells nb::arg("data").noconvert() -> wrap_persistence_2d<[double|float]>(nb::ndarray<nb::c_contig>)
    - API change: _persistence_on_a_line and _persistence_on_rectangle_from_top_cells now accept only c_contig arrays

* src/python/gudhi/sklearn/cubical_persistence.py
    - def __transform(self, cells):
          cells = np.asarray(cells, order="C") # Conversion is now done here
          # This one is just to satisfy a test with integers. Should we also need int64 ?
          if not np.issubdtype(type(cells.flat[0]), np.floating):
              cells = cells.astype("float64")

I also needed to fix some unitary tests.
I didn't fix CubicalComplex and PeriodicCubicalComplex as conditional inheritance is not easy here.
I tried to add int64 type and only *few changes* were required (I can show on a gist).

### DIFF
--- a/src/python/gudhi/sklearn/cubical_persistence.py
+++ b/src/python/gudhi/sklearn/cubical_persistence.py
@@ -153,7 +153,9 @@ class CubicalPersistence(BaseEstimator, TransformerMixin):
                 diags = _persistence_on_rectangle_from_top_cells(cells, self.min_persistence)
             return [diags[i] if i in (0, 1) else np.empty((0, 2)) for i in self._dim_list]
 
-        cells = np.asarray(cells, order="F")
+        cells = np.asarray(cells)
+        if not cells.flags.f_contiguous:
+            cells = cells.T
         dimensions = cells.shape
         cells = cells.ravel(order="F")
         CubicalComplexItf = self._DTYPE_CUBICAL_COMPLEX_MAP[cells.dtype]
--- a/src/python/gudhi/sklearn/cubical_persistence.py
+++ b/src/python/gudhi/sklearn/cubical_persistence.py
@@ -90,6 +90,8 @@ class CubicalPersistence(BaseEstimator, TransformerMixin):
 
     def __transform(self, cells):
         cells = np.asarray(cells)
+        if not np.issubdtype(type(cells.flat[0]), np.floating):
+            cells = cells.astype("float64")
         if len(cells.shape) == 1 and self.min_persistence >= 0:
             res = _persistence_on_a_line(cells)
             if self.min_persistence > 0:
--- a/src/python/gudhi/sklearn/cubical_persistence.py
+++ b/src/python/gudhi/sklearn/cubical_persistence.py
@@ -16,7 +16,12 @@ from typing import Literal, Optional
 from sklearn.base import BaseEstimator, TransformerMixin
 from joblib import Parallel, delayed
 
-from .. import CubicalComplex
+from .._cubical_complex_ext import (
+    _Bitmap_cubical_complex_interface_float64,
+    _Cubical_complex_persistence_interface_float64,
+    _Bitmap_cubical_complex_interface_float32,
+    _Cubical_complex_persistence_interface_float32,
+)
 from .._pers_cub_low_dim_ext import (
     _persistence_on_a_line,
     _persistence_on_rectangle_from_top_cells,
@@ -41,6 +46,16 @@ class CubicalPersistence(BaseEstimator, TransformerMixin):
     """
     This is a class for computing the persistence diagrams from a cubical complex.
     """
+
+    _DTYPE_CUBICAL_COMPLEX_MAP = {
+        np.dtype(np.float32): _Bitmap_cubical_complex_interface_float32,
+        np.dtype(np.float64): _Bitmap_cubical_complex_interface_float64,
+    }
+
+    _DTYPE_CUBICAL_COMPLEX_PERSISTENCE_MAP = {
+        np.dtype(np.float32): _Cubical_complex_persistence_interface_float32,
+        np.dtype(np.float64): _Cubical_complex_persistence_interface_float64,
+    }
 
     def __init__(
         self,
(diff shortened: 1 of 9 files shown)


src/python/gudhi/sklearn/cubical_persistence.py
```diff
@@ -88,8 +109,29 @@ def fit(self, X, Y=None):
...
+            
     def __transform(self, cells):
-        cells = np.asarray(cells)
+        cells = np.asarray(cells, order="C")
+        # cf. src/python/test/test_sklearn_cubical_persistence.py - test_1d
+        # a = np.array([2, 4, 3, 5])
+        # ri = CubicalPersistence(0).fit_transform([a])[0]
+        # assert ri.dtype == np.dtype("float64") # whereas a.dtype == np.int64
+        if not np.issubdtype(cells.dtype, np.floating):
+            cells = cells.astype("float64")
```
mglisse, line 134: What happens if the input dtype is np.float16 or np.float128? (what do we want to happen?)

src/python/gudhi/sklearn/cubical_persistence.py
```diff
@@ -111,17 +153,16 @@ def __transform(self, cells):
...
-        elif self.input_type == "vertices":
-            cubical_complex = CubicalComplex(vertices=cells)
-        else:
-            raise ValueError("input_type can only be 'top_dimensional_cells' or 'vertices'")
-        cubical_complex.compute_persistence(
-            homology_coeff_field=self.homology_coeff_field,
-            min_persistence=self.min_persistence,
-        )
-        return [cubical_complex.persistence_intervals_in_dimension(dim) for dim in self._dim_list]
+        cells = np.asarray(cells, order="F")
```
mglisse: So if the user kindly provides an F-contiguous array, we first copy it to a C-contiguous array, then copy it again to an F-contiguous one?
A and A.T (the transpose) should give the same diagram (only the names of the generators differ), so switching between C- and F- contiguity should not require a copy.

mglisse (commented):
Couple quick comments (I haven't really read the diff yet).
* Does this increase the size of the binary? (probably a bit but not too much)
* Does this increase the overhead when computing the persistence of small arrays (say an array of shape (3,), (3,3) or (3,3,3). I think for MNIST-sized arrays we already spend more time in the python wrapper than in the C++ code, and I wonder at what point it becomes a problem.

VincentRouvreau:
>     * Does this increase the size of the binary? (probably a bit but not too much)

On master:
286K _cubical_complex_ext.cpython-310-x86_64-linux-gnu.so*
133K _pers_cub_low_dim_ext.cpython-310-x86_64-linux-gnu.so*

On branch different_filtration_values_support_in_python:
406K _cubical_complex_ext.cpython-310-x86_64-linux-gnu.so*
153K _pers_cub_low_dim_ext.cpython-310-x86_64-linux-gnu.so*

VincentRouvreau:
>     * Does this increase the overhead when computing the persistence of small arrays (say an array of shape (3,), (3,3) or (3,3,3). I think for MNIST-sized arrays we already spend more time in the python wrapper than in the C++ code, and I wonder at what point it becomes a problem.

I did some quick benchmark using this code:
```python
import numpy as np
from gudhi.sklearn import CubicalPersistence

X = np.random.random([1000000, 3])
cub = CubicalPersistence(homology_dimensions=0, n_jobs=1)
start = timer(); diags = cub.fit_transform(X); print(f"{timedelta(seconds=timer()-start)} sec for {X.shape=}")

X = np.random.random([1000000, 3, 3])
cub = CubicalPersistence(homology_dimensions=[0, 1], n_jobs=1)
start = timer(); diags = cub.fit_transform(X); print(f"{timedelta(seconds=timer()-start)} sec for {X.shape=}")

X = np.random.random([100000, 3, 3, 3])
cub = CubicalPersistence(homology_dimensions=[0, 1, 2], n_jobs=1)
start = timer(); diags = cub.fit_transform(X); print(f"{timedelta(seconds=timer()-start)} sec for {X.shape=}")
```

The results:
| dataset | master in sec. | different_filtration_values in sec. |
| --- | --- | --- |
| (3) | 5.8 | 6.5 |
| (3,3) | 8.97 | 9.66 |
| (3,3,3) | 8.79 | 8.64 |

It seems to have more impact on `persistence_on_line` and `persistence_on_rectangle`, maybe because these are using function bindings, while in the (3,3,3) case a class binding is used. I am not sure here

src/python/gudhi/sklearn/cubical_persistence.py
VincentRouvreau: Yes, that is a good question, maybe we could say, for instance, that float64 is the default when dtype in not one of the supported types.

But, currently, even on master version all of these cases are not handled properly:
```python
import numpy as np
from gudhi.sklearn import CubicalPersistence

cub = CubicalPersistence(homology_dimensions=[0, 1, 2], n_jobs=1)
X = np.random.random([1, 3, 3, 3])
X = np.asarray(X, dtype=np.float16)
diags = cub.fit_transform(X) # diags dtype=float64
X = np.asarray(X, dtype=np.float128)
diags = cub.fit_transform(X) # diags dtype=float64

cub = CubicalPersistence(homology_dimensions=[0, 1], n_jobs=1)
X = np.random.random([1, 3, 3])
X = np.asarray(X, dtype=np.float16)
diags = cub.fit_transform(X) # diags dtype=float64
X = np.asarray(X, dtype=np.float128)
diags = cub.fit_transform(X) # TypeError: _persistence_on_rectangle_from_top_cells(): incompatible function arguments

cub = CubicalPersistence(homology_dimensions=0, n_jobs=1)
X = np.random.random([1, 3])
X = np.asarray(X, dtype=np.float16)
diags = cub.fit_transform(X) # diags dtype=float64
X = np.asarray(X, dtype=np.float128)
diags = cub.fit_transform(X) # TypeError: _persistence_on_a_line(): incompatible function arguments
```

I have an idea on how to do loop on supported types with a default case (that would be the last of the list to be detected for instance)

src/python/gudhi/sklearn/cubical_persistence.py
VincentRouvreau: It seems to work as suggested:
```python
a=np.array([[1.,2.,3.], [4.,5.,6.]])
a.flags.f_contiguous
# False
a.flags.c_contiguous
# True
a.T
# array([[1., 4.],
#        [2., 5.],
#        [3., 6.]])
a.T.flags.f_contiguous
# True
a.T.flags.c_contiguous
# False
```

I modified code on 46cbd8ef9

PR #1381, opened 2026-09-11
Better pandas management

### LINKED ISSUE #1015: sklearn set_output
Since version 1.2, scikit-learn has a set_output API that can make transformers output pandas dataframes.

- [ ] Investigate how that affects us.

VincentRouvreau:
* Better pandas management by cmake
* pytest should skip when no pandas
* Add set_output for CubicalPersistence (and its test)           

Fix #1015  (was already fixed but not for all sklearn interfaces)


src/python/CMakeLists.txt
```diff
@@ -273,7 +273,7 @@ add_gudhi_py_test(test_cubical_complex)
 add_gudhi_py_test(test_cubical_persistence_low_dim)
 
 # Datasets are fetched for these tests
-if(SCIKIT-LEARN_FOUND AND WITH_GUDHI_REMOTE_TEST)
+if(SCIKIT-LEARN_FOUND AND PANDAS_FOUND AND WITH_GUDHI_REMOTE_TEST)
```
mglisse: Since you already disable part of the test from the inside when pandas is missing, it looks strange to skip the test completely without pandas.

VincentRouvreau: I removed this test. No need to test if pandas is found at cmake level, test is just skipped by pytest if not found.
Done on 82414ee80

mglisse (approved):

